PR #7259, opened 2018-04-21
Fix i386 build at OpenBSD

mazocomp:


mazocomp:
Oups, I wanted to write 32 bit first and forgot to remove word "bit".

PR #7262, opened 2018-04-21
Sync with FreeBSD's and OpenBSD's patches

mazocomp:
https://svnweb.freebsd.org/ports/head/games/minetest/files/patch-src_CMakeLists.txt?revision=442786&view=co
https://cvsweb.openbsd.org/cgi-bin/cvsweb/~checkout~/ports/games/minetest/patches/patch-src_CMakeLists_txt?rev=1.12&content-type=text/plain


mazocomp:
Oh dear compatibility, you're so tough.

src/CMakeLists.txt
```diff
@@ -323,12 +325,12 @@ else()
 		endif(HAVE_LIBRT)
 	endif(APPLE)
 
-	if(NOT HAIKU AND NOT APPLE)
+	if(NOT HAIKU AND NOT APPLE AND NOT (CMAKE_SYSTEM_NAME MATCHES "kFreeBSD.*|FreeBSD"))
```
SmallJoker: Can't we just use `if (DEFINED ENV{DISPLAY})` to detect whether X11 is in use or not?

src/CMakeLists.txt
```diff
@@ -62,6 +62,8 @@ set(USE_GETTEXT FALSE)
 
 if(ENABLE_GETTEXT)
 	find_package(GettextLib)
+	find_package(Iconv)
+	find_package(Intl)
```
SmallJoker: These are not required for Linux.

src/CMakeLists.txt
```diff
@@ -740,14 +744,12 @@ else()
 		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DWIN32_LEAN_AND_MEAN")
 	endif()
 
-	set(CMAKE_CXX_FLAGS_RELEASE "-DNDEBUG ${RELEASE_WARNING_FLAGS} ${WARNING_FLAGS} ${OTHER_FLAGS} -Wall -pipe -funroll-loops")
```
SmallJoker: Please keep these flags for Linux if *BSD doesn't know them.

src/CMakeLists.txt
```diff
@@ -740,14 +744,12 @@ else()
...
-	if(CMAKE_SYSTEM_NAME MATCHES "(Darwin|FreeBSD)")
+	set(CMAKE_CXX_FLAGS_RELEASE "-DNDEBUG ${RELEASE_WARNING_FLAGS} ${WARNING_FLAGS} ${OTHER_FLAGS} -Wall")
+	if(APPLE)
 		set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -Os")
-	else()
-		set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -O3 -ffast-math -fomit-frame-pointer")
-	endif(CMAKE_SYSTEM_NAME MATCHES "(Darwin|FreeBSD)")
+	endif(APPLE)
 	set(CMAKE_CXX_FLAGS_SEMIDEBUG "-g -O1 -Wall -Wabi ${WARNING_FLAGS} ${OTHER_FLAGS}")
-	set(CMAKE_CXX_FLAGS_DEBUG "-g -O0 -Wall -Wabi ${WARNING_FLAGS} ${OTHER_FLAGS}")
```
SmallJoker: Why do you remove `-O0` here whereas it's fine for semi-debug?

src/CMakeLists.txt
```diff
@@ -740,14 +744,12 @@ else()
 		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DWIN32_LEAN_AND_MEAN")
 	endif()
 
-	set(CMAKE_CXX_FLAGS_RELEASE "-DNDEBUG ${RELEASE_WARNING_FLAGS} ${WARNING_FLAGS} ${OTHER_FLAGS} -Wall -pipe -funroll-loops")
-	if(CMAKE_SYSTEM_NAME MATCHES "(Darwin|FreeBSD)")
+	set(CMAKE_CXX_FLAGS_RELEASE "-DNDEBUG ${RELEASE_WARNING_FLAGS} ${WARNING_FLAGS} ${OTHER_FLAGS} -Wall")
+	if(APPLE)
 		set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -Os")
-	else()
-		set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -O3 -ffast-math -fomit-frame-pointer")
```
SmallJoker: Also try to keep these optimization flags for Linux.

src/CMakeLists.txt
mazocomp: Sorry, I didn't test that.

src/CMakeLists.txt
mazocomp: It actually knows them, but they are broken.

src/CMakeLists.txt
mazocomp: Ok.

src/CMakeLists.txt
mazocomp: I didn't test semi-debug. I take all what is present in patches because I am annoyed to get compiletime/runtime errors.

src/CMakeLists.txt
mazocomp: @SmallJoker Won't that cause trouble to people which compile, for example, via SHH or virtual terminal?

SmallJoker: Right, I see. There's still `FindX11` in CMake - but a reliable detection doesn't seem to be possible with that either.

src/CMakeLists.txt
mazocomp: @SmallJoker Is it worth adding if(CMAKE_SYSTEM_NAME MATCHES "kOpenBSD.*|OpenBSD.*")?

src/CMakeLists.txt
mazocomp: @SmallJoker I've asked OpenBSD's ports team, it appears that these optimizer flags are set by ports or the user, not the ported software. Semi-debug is not reached by any usual port builds.
https://marc.info/?l=openbsd-ports&m=152430251624488&w=2

src/CMakeLists.txt
mazocomp: @SmallJoker argh, they were required in the past, so ignore this patch.

src/CMakeLists.txt
mazocomp: @SmallJoker I'm sorry for my ignorance, they aren't broken, they are just set by user or ports.

PR #7265, opened 2018-04-21
Optimizer flags are set by ports/user, ok?

mazocomp:
https://marc.info/?l=openbsd-ports&m=152430251624488&w=2


sfan5:
Unless there is some significant difference between patches to master vs. backports, there is no need to create separate pull requests as I will just cherry-pick them when the backports branch gets updated.

sfan5:
You should've kept the PR to merge into `master` and closed this one.

mazocomp:
@sfan5 Oh, what an implicit language, should I close these two and reopen those two?

sfan5:
Yes please

PR #7308, opened 2018-05-08
WIP: Fix confirm registration: snprintf in windows

Dumbeldor:
I fix crash in windows when you first join a server :

image

**But now the text displayed is as follows:**

image

**And when I click a button**

image

I continue to study this problem tomorrow. But if you have any ideas do not hesitate. Maybe I misunderstood. 
**This page works on linux?**

### DIFF
--- a/src/gui/guiConfirmRegistration.cpp
+++ b/src/gui/guiConfirmRegistration.cpp
@@ -99,18 +99,21 @@ void GUIConfirmRegistration::regenerateGui(v2u32 screensize)
 
 #ifdef _MSC_VER
 		int sizeBuffer = sizeof(info_text_buf) -
-				  (info_text_template.length() - 8 + address.length() +
-						  m_playername.length());
+				 (info_text_template.length() - 8 + address.length() +
+						 m_playername.length());
 
 		if (sizeBuffer >= 0) {
-			sprintf(info_text_buf, info_text_template.c_str(), address.c_str(), 
-					m_playername.c_str());
+			sprintf(info_text_buf, info_text_template.c_str(),
+					address.c_str(), m_playername.c_str());
 		} else {
-			warningstream << "the info_text_buf buffer is too small" << std::endl;
+			warningstream << "Failed to generate info_text. Buffer is too "
+					 "small"
+				      << std::endl;
 			int new_size_text_template =
 					(int)info_text_template.size() + (sizeBuffer - 4);
 			if (new_size_text_template > 0) {
-				std::string info_text_template_trunq = info_text_template.substr(0,
+				std::string info_text_template_trunq =
+						info_text_template.substr(0,
 								new_size_text_template);
 				info_text_template_trunq.append("...");
 				sprintf(info_text_buf, info_text_template_trunq.c_str(),
--- a/src/gui/guiConfirmRegistration.cpp
+++ b/src/gui/guiConfirmRegistration.cpp
@@ -112,11 +112,11 @@ void GUIConfirmRegistration::regenerateGui(v2u32 screensize)
 			int new_size_text_template =
 					(int)info_text_template.size() + (sizeBuffer - 4);
 			if (new_size_text_template > 0) {
-				std::string info_text_template_trunq =
+				std::string info_text_template_trunc =
 						info_text_template.substr(0,
 								new_size_text_template);
-				info_text_template_trunq.append("...");
-				sprintf(info_text_buf, info_text_template_trunq.c_str(),
+				info_text_template_trunc.append("...");
+				sprintf(info_text_buf, info_text_template_trunc.c_str(),
 						address.c_str(), m_playername.c_str());
 			}
 		}
--- a/src/gui/guiConfirmRegistration.cpp
+++ b/src/gui/guiConfirmRegistration.cpp
@@ -98,8 +98,25 @@ void GUIConfirmRegistration::regenerateGui(v2u32 screensize)
 		char info_text_buf[1024];
 
 #ifdef _MSC_VER
-		sprintf(info_text_buf, info_text_template.c_str(), address.c_str(),
-				m_playername.c_str());
+		int sizeBuffer = sizeof(info_text_buf) -
+				  (info_text_template.length() - 8 + address.length() +
+						  m_playername.length());
+
+		if (sizeBuffer >= 0) {
+			sprintf(info_text_buf, info_text_template.c_str(), address.c_str(), 
+					m_playername.c_str());
+		} else {
+			warningstream << "the info_text_buf buffer is too small" << std::endl;
+			int new_size_text_template =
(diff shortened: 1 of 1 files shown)


Dumbeldor:
For the problem of the text that is displayed badly I can do this kind of thing that works. But it may not be great.

```
const std::string search_server = "%1$s";
info_text_template.replace(info_text_template.find(search_server), search_server.size(), address);
```

srifqi:
What Windows version are you using? I'm using Windows 10 and Minetest (compiled using gcc) and it just works. I'll try to find the cause.

Dumbeldor:
@srifqi i compile with VS 2017

Dumbeldor:
"Beginning with the UCRT in Visual Studio 2015 and Windows 10, snprintf is no longer identical to _snprintf. The snprintf function behavior is now C99 standard compliant."

https://msdn.microsoft.com/en-US/library/2ts7cx93.aspx

nerzhul:
waiting for LINT fix to merge

Dumbeldor:
I have a little fix to do again and we can merge this first correction.

src/gui/guiConfirmRegistration.cpp
```diff
@@ -96,8 +96,14 @@ void GUIConfirmRegistration::regenerateGui(v2u32 screensize)
 				"Join to confirm account creation or click Cancel to "
 				"abort.");
 		char info_text_buf[1024];
+
+#ifdef _WIN32
```
sfan5: since this doesn't happen on mingw, it should be `#ifdef _MSC_VER` instead

Dumbeldor:
@sfan5 thanks, I continue to study this tonight. In any case, there is a problem with snprintf, but I may have misidentified the problem.

src/gui/guiConfirmRegistration.cpp
Dumbeldor: Yes you are right, thanks

Dumbeldor:
snprintf works when in the template i have %s and it replaces the %s with the variable data. But when the template has %1$s it crashes. 
So with sprintf i have no problem. What do you think ?

sfan5:
Well, `sprintf` is inherently unsafe, not a good idea to use it.
Have you tried `_snprintf_s`?

Dumbeldor:
I tried and I have the same problem :(

Dumbeldor:
It works with this template : 
``
static const std::string info_text_template = strgettext(
				"You are about to join the server at %s with the "
				"name \"%s\" for the first time. If you proceed, a "
				"new account using your credentials will be created "
				"on this server.\n"
				"Please retype your password and click Register and "
				"Join to confirm account creation or click Cancel to "
				"abort.");
``

srifqi:
How about we just change the template text? That way, we still can use the safe `snprintf()`.

Dumbeldor:
no because the order of the variables in template can change according to the language.

Dumbeldor:
@srifqi I made the same mistake as you at the beginning :D

Dumbeldor:
Currently there is no error handling. If snprintf returns -1, so that the text is truncated, nothing special is done ?

sfan5:
I'm not asking for error handling, if you absolutely do need to call `sprintf` (no <b>n</b>) check the size first to avoid a possible buffer overflow.

src/gui/guiConfirmRegistration.cpp
```diff
@@ -96,8 +96,34 @@ void GUIConfirmRegistration::regenerateGui(v2u32 screensize)
...
+			sprintf(info_text_buf, info_text_template.c_str(),
+					address.c_str(), m_playername.c_str());
+		} else {
+			warningstream << "Failed to generate info_text. Buffer is too "
+					 "small"
+				      << std::endl;
+			int new_size_text_template =
+					(int)info_text_template.size() + (sizeBuffer - 4);
+			if (new_size_text_template > 0) {
+				std::string info_text_template_trunq =
```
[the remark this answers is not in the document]
Dumbeldor: Exactly, thank you

src/gui/guiConfirmRegistration.cpp
```diff
@@ -96,8 +96,34 @@ void GUIConfirmRegistration::regenerateGui(v2u32 screensize)
 				"Join to confirm account creation or click Cancel to "
 				"abort.");
 		char info_text_buf[1024];
+
+#ifdef _MSC_VER
+		int sizeBuffer = sizeof(info_text_buf) -
```
nerzhul: bufferSize

please also add a comment to define what is this calculation

src/gui/guiConfirmRegistration.cpp
```diff
@@ -96,8 +96,34 @@ void GUIConfirmRegistration::regenerateGui(v2u32 screensize)
...
+			warningstream << "Failed to generate info_text. Buffer is too "
+					 "small"
+				      << std::endl;
+			int new_size_text_template =
+					(int)info_text_template.size() + (sizeBuffer - 4);
+			if (new_size_text_template > 0) {
+				std::string info_text_template_trunc =
+						info_text_template.substr(0,
+								new_size_text_template);
+				info_text_template_trunc.append("...");
```
nerzhul: instead of copying the whole data, just replace the last 3 chars directly (info_text_template[info_text_template.size() - 4] = '.')

src/gui/guiConfirmRegistration.cpp
```diff
@@ -96,8 +96,34 @@ void GUIConfirmRegistration::regenerateGui(v2u32 screensize)
...
+				 (info_text_template.length() - 8 + address.length() +
+						 m_playername.length());
+
+		if (sizeBuffer >= 0) {
+			sprintf(info_text_buf, info_text_template.c_str(),
+					address.c_str(), m_playername.c_str());
+		} else {
+			warningstream << "Failed to generate info_text. Buffer is too "
+					 "small"
+				      << std::endl;
```
sfan5: Since this should never happen in practice, IMO it's okay (and much simpler) to abort the dialog or not show any text at all

Dumbeldor: at least the player can still play

SmallJoker:
I'm not quite sure what I should think about this fix - or rather about MS' decision on changing the function behaviour. However, could you please move this code into a new function for `util/string.*` so it can be used for all situations with two translation insertion strings?

src/gui/guiConfirmRegistration.cpp
sfan5: Realistically, when is length(server name) + length(player name) + some fixes amount around 300 ever going to exceed 1024?

Instead of trying to do the right thing in weird cases, it's better to fail.

Dumbeldor: ok i do that

SmallJoker:
Superseded by #7353

PR #7315, opened 2018-05-10
Update mipmapping and textarea documentation

HybridDog:
textarea with non-empty name:
screenshot_20180510_152939
textarea with empty name:
screenshot_20180510_152929


builtin/settingtypes.txt
```diff
@@ -418,7 +418,10 @@ enable_particles (Digging particles) bool true
 
 [**Filtering]
 
-#    Use mip mapping to scale textures. May slightly increase performance.
+#    Use mip mapping to scale textures. May slightly increase performance,
+#    especially when using a high resolution texture pack.
+#    Note that it currently lacks gamma correct downscaling generating the mip
```
paramat: I don't think failings of current mipmap belongs in docs.

doc/lua_api.txt
```diff
@@ -2201,13 +2201,15 @@ appearing when you don't expect them to. See `no_prepend[]`
...
-* if the name is empty the textarea is readonly, the label is not displayed.
+* If the text overflows, a vertical scrollbar is added.
+* If the name is empty, the textarea is readonly and
+  the background is not shown.
 
 #### `label[<X>,<Y>;<label>]`
 * `x` and `y` work as per field
 * `label` is the text on the label
 * Position and size units are inventory slots
+* If you want to show text with multiple lines, use textarea with empty name.
```
Ezhh: More concise option: "To show text with ... "
But it's better to list things under what you are talking about. If you need to explain usage it should be "Use for single line text labels" and then move this to textarea section.

builtin/settingtypes.txt
HybridDog: Downscaling with missing gamma correction happens at many places, e.g. in firefox. I think this weakness should be documented.

SmallJoker: It's documented in form of an issue in this repository - it's a bug.

4w: Documentation is not advertising. If something lacks a feature it should be documented and properly linked to the issue until the issue was fixed.

doc/lua_api.txt
HybridDog: I've just noticed that it's not said what label actually does: It displays text.

builtin/settingtypes.txt
paramat: It doesn't help in any way to mention this in setting docs, as the user will be aware of the problem when they use it, and there's nothing they can do to avoid the problem. Many aspects of MT are imperfect but unless there is a useful reason to do so we don't write about this in docs, otherwise docs would double in size.

paramat:
:-1: As explained.

doc/lua_api.txt
Ezhh: You've basically ignored my comment. Don't talk about textarea in label section. Talk about textarea in textarea section.

Ezhh:
:-1: for me too in current state.

doc/lua_api.txt
```diff
@@ -2201,13 +2201,17 @@ appearing when you don't expect them to. See `no_prepend[]`
...
+* If the name is empty, the textarea is readonly and
+  the background is not shown, which corresponds to a multi-line label.
 
 #### `label[<X>,<Y>;<label>]`
 * `x` and `y` work as per field
-* `label` is the text on the label
-* Position and size units are inventory slots
+* The label formspec element displays the text set in `label`
+  at the specified position.
+* For text with multiple lines, refer to textarea.
```
Ezhh: Don't talk about textarea in label section. Talk about textarea in textarea section.

builtin/settingtypes.txt
```diff
@@ -418,7 +418,10 @@ enable_particles (Digging particles) bool true
...
+#    maps, thus the textures further away look a bit darker.
```
numberZero: *darker*? How can you be sure they can’t look *lighter* with some monitor settings?
Seriously, that doesn't belong to docs. These tiny imperfections don’t worth it. (BTW, remember that in-game “gamma” is applied to light levels and not to resulting pixels, that may be much more noticeable).

doc/lua_api.txt
```diff
@@ -2201,13 +2201,17 @@ appearing when you don't expect them to. See `no_prepend[]`
...
+* To display text which can contain formspec code, e.g. text set by a player,
+  use `minetest.formspec_escape`.
```
numberZero: That’s not label-specific. As well as note on units you removed.

builtin/settingtypes.txt
paramat: 2 core devs support this being removed.

doc/lua_api.txt
```diff
@@ -2201,13 +2201,17 @@ appearing when you don't expect them to. See `no_prepend[]`
 
 #### `textarea[<X>,<Y>;<W>,<H>;<name>;<label>;<default>]`
 * Same as fields above, but with multi-line input
-* if the text overflows a vertical scrollbar is added
-* if the name is empty the textarea is readonly, the label is not displayed.
+* If the text overflows, a vertical scrollbar is added.
+* If the name is empty, the textarea is readonly and
```
paramat: 'read only'?

paramat:
Has some issues and no response or action for a month. Reopen on progress.

doc/lua_api.txt
HybridDog: Often people forget to use minetest.formspec_escape (see travelnet), so I mentioned it here.

When I removed the note on units I thought that somewhere it is mentioned that the note pertains on all size units.
This note does not make sense to me because it appears for some formspec elements and for others no information about the size units is given. Why is this?

doc/lua_api.txt
HybridDog: Do you think that modders can easily find out about how to put multi-line text?

doc/lua_api.txt
HybridDog: I agree, it should be changed to "read-only".

builtin/settingtypes.txt
HybridDog: I still think missing gamma correction should be mentioned, but using only very few words.

> These tiny imperfections don’t worth it.

The brick texture mipmaps look quite different with gamma correction. (images from #6917)
default:
d

gamma-correct:
g
(you can ignore the bilinear or so filter)

builtin/settingtypes.txt
HybridDog: > as the user will be aware of the problem when they use it

Did anyone notice missing gamma correction in mip mapping before I did?

builtin/settingtypes.txt
numberZero: @HybridDog What you post as “gamma-corret,” in fact, looks *worse* on my display.

doc/lua_api.txt
numberZero: > it appears for some formspec elements and for others no information about the size units is given. Why is this?

[Historically](https://github.com/minetest/minetest/commit/3899f83da879e11cf531c88514772b890d060721), it was repeated in each element description. But later people probably considered that too obvious to repeat.

I suggest to move both size note and escaping explanation somewhere in the Formspec section itself, and keep short reminders like “Remember to escape text with formspec_escape” in elements documentation.

doc/lua_api.txt
Ezhh: You need to make it explicit that label is for single line instead. It's even fine to state it's not for multi-line use. Then if we make sure the textarea section mentions usage includes multi-line, the problem is gone. This way each section is explaining its own use and limits, and not explaining other things' uses and limits.

builtin/settingtypes.txt
```diff
@@ -420,7 +420,9 @@ enable_particles (Digging particles) bool true
 
 [**Filtering]
 
-#    Use mip mapping to scale textures. May slightly increase performance.
+#    Use mip mapping to scale textures. May slightly increase performance,
+#    especially when using a high resolution texture pack.
+#    Gamma correct downscaling is not supported.
```
paramat: This is ok for me.

doc/lua_api.txt
```diff
@@ -1777,6 +1777,13 @@ supported. It is a string, with a somewhat strange format.
 Spaces and newlines can be inserted between the blocks, as is used in the
 examples.
 
+Position and size units are inventory slots, `X` and `Y` position the formspec
```
HybridDog: I'm not sure if it is clear what X refers to because X appears later in the formspec element descriptions.

SmallJoker: In some way it's a prelude to the documentation which follows afterwards. I think that's fine.

doc/lua_api.txt
```diff
@@ -1777,6 +1777,13 @@ supported. It is a string, with a somewhat strange format.
...
+element relative to the top left of the menu and `W` and `H` are its width and
```
SmallJoker: `of the menu or container. ``W`` and ``H`` are its width and height values.`

HybridDog: https://github.com/minetest/minetest/pull/7315/commits/550bb6348a557986f997e5be782f0128f965a56d
I didn't put ` `` ` instead of `` ` `` around W and H.

paramat:
@Ezhh is it ok for you now?

Ezhh:
@paramat Fine for me now.

PR #7320, opened 2018-05-12
Add the override keyword to some inherited classes.

opk4406opk:
If it is an inherited class, it would be safer to use the override keyword.

### DIFF
--- a/src/shader.cpp
+++ b/src/shader.cpp
@@ -269,8 +269,8 @@ public:
 
 		The id 0 points to a null shader. Its material is EMT_SOLID.
 	*/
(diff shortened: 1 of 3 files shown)


src/shader.cpp
```diff
@@ -281,25 +281,25 @@ class ShaderSource : public IWritableShaderSource
 		for processing.
 	*/
 
-	u32 getShader(const std::string &name,
-		const u8 material_type, const u8 drawtype);
+	virtual u32 getShader(const std::string &name,
```
nerzhul: it's every virtual or override, not both.

PR #7331, opened 2018-05-13
Added pitch and roll for entities, saving rotations should also work

CoderForTheBetter:
This should add pitch and roll for entities.

See these related items around Github:
[My issue (closed) (#7323)](https://github.com/minetest/minetest/issues/7323)
[Another issue requesting pitch and roll (#2251)](https://github.com/minetest/minetest/issues/2251)

Here's this pull request in action:
39964985-52756f48-565e-11e8-89e8-256ab236f611
Here's the Lua code that rotates the helicopter in the above gif.
```
function Helicopter:on_step(_, dtime)
  
  if timer > -1 and timer < 90 then
    y=y+0.1
    p=0
    r=0
    minetest.chat_send_all("YAW: " .. tostring(y))
  elseif timer > 90 and timer < 180 then
    y=0
    p=p+0.1
    r=0
    minetest.chat_send_all("PITCH: " .. tostring(p))
  elseif timer > 180 and timer < 270 then
    y=0
    p=0
    r=r+0.1
    minetest.chat_send_all("ROLL: " .. tostring(r))
  end
  
  self.object:set_yaw(y)
  self.object:set_pitch(p)
  self.object:set_roll(r)

  timer = timer + 1
  if timer > 270 then
    timer = 0
  end
end
```

One thing I'm concerned with and didn't know how to handle (style issue) is the overriding of get/set pitch of the player's view and the methods I made. This can be found in the content_sao.h file. Also, I'm not sure if I should add the ability to automatically rotate entities on the pitch and roll axes like we have for yaw (automatic_rotate in Lua entity definition).

If these changes are good then I will add documentation to the Lua API in this pull request or another.

Please mark this WIP since it is still being tested and added to, thank you.


paramat:
Good feature.

CoderForTheBetter:
Honestly, I think this pull request is ready to be reviewed. If an automatic rotate should be added for pitch and roll then it probably should be added in a different pull request, in my opinion.

I'm not entirely sure this is a problem, and I'll bring the code to you guys, but in the file [content_sao.h](https://github.com/CoderForTheBetter/minetest/blob/38a099683170d243073b3d0dc430edf88492515d/src/content_sao.h), in the UNITSAO class I added definitions for getting pitch in rad and deg, but the PlayerSAO class inherits from the UNITSAO class ([right here](https://github.com/CoderForTheBetter/minetest/blob/38a099683170d243073b3d0dc430edf88492515d/src/content_sao.h#L219)). However, the PlayerSAO class already has a function for getting the pitch called the same name "getPitch" and "getRadPitch" ( [right here: PlayerSAO](https://github.com/CoderForTheBetter/minetest/blob/38a099683170d243073b3d0dc430edf88492515d/src/content_sao.h#L250) ) just like the methods I added in the UNITSAO class ([right here: UNITSAO](https://github.com/CoderForTheBetter/minetest/blob/38a099683170d243073b3d0dc430edf88492515d/src/content_sao.h#L43))

I'm wondering if I should just get rid of the function definitions/declarations in the PlayerSAO class since it will inherit it from the UNITSAO class.

**In summary:**
We have two classes: UNITSAO and PlayerSAO with the following functions in them:
```
class UnitSAO: public ServerActiveObject
{
public:
        virtual void setPitch(const float pitch) { m_pitch = pitch; }
        float getPitch() const { return m_pitch; };
	f32 getRadPitch() const { return m_pitch * core::DEGTORAD; }
};

class PlayerSAO : public UnitSAO
{
public:
	void setPitch(const float pitch);
	// Data should not be sent at player initialization
	void setPitchAndSend(const float pitch);
	f32 getPitch() const { return m_pitch; }
	f32 getRadPitch() const { return m_pitch * core::DEGTORAD; 
};
```
And as can be seen, in PlayerSAO getPitch() and getRadPitch will be overridden by what it inherits from UNITSAO, and my question is should I just get rid of them?

Thank you for your time! Hopefully it is clear what I am asking.

CoderForTheBetter:
Thanks for the response and review @rubenwardy. I would hope that a network break would acceptable here since this is 0.5.0 and it is a breaking a lot of things so that 0.4.16 can't work with it. I don't don't see why it would matter if it breaks network communication with other non-current/updated 0.5.0-dev clients and servers. This is a feature has needed to be added for years and now seems the best time.

If there is a way to go about this where it doesn't break the network as much, then please clarify and I will try my best.

I'll take care of your requested changes, thanks again!

src/script/lua_api/l_object.cpp
```diff
@@ -1856,6 +1908,10 @@ const luaL_Reg ObjectRef::methods[] = {
 	luamethod_aliased(ObjectRef, get_acceleration, getacceleration),
 	luamethod_aliased(ObjectRef, set_yaw, setyaw),
 	luamethod_aliased(ObjectRef, get_yaw, getyaw),
+        luamethod_aliased(ObjectRef, set_pitch, setpitch),
```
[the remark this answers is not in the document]
CoderForTheBetter: Sorry about the tabs and spaces. My IDE was set to use spaces when pressing TAB and I did not realize it.

src/script/lua_api/l_object.cpp
```diff
@@ -914,6 +914,58 @@ int ObjectRef::l_get_yaw(lua_State *L)
 	return 1;
 }
 
+// set_pitch(self, radians) (X-AXIS)
+int ObjectRef::l_set_pitch(lua_State *L)
+{
+	NO_MAP_LOCK_REQUIRED;
+	ObjectRef *ref = checkobject(L, 1);
+	LuaEntitySAO *co = getluaobject(ref);
+	if (co == NULL) return 0;
```
SmallJoker: The old code is already formatted wrong, please use either:
```C++
if (!co)
	return 0;
```
or
```C++
if (co == nullptr)
	return 0;
```
Same for your functions below.

src/script/lua_api/l_object.cpp
```diff
@@ -914,6 +914,58 @@ int ObjectRef::l_get_yaw(lua_State *L)
...
+	float pitch = luaL_checknumber(L, 2) * core::RADTODEG;
+	// Do it
```
SmallJoker: Very helpful. Could you please remove this one in all your functions?

src/content_sao.cpp
```diff
@@ -477,7 +483,9 @@ void LuaEntitySAO::step(float dtime, bool send_recommended)
 		move_d += m_last_sent_move_precision;
 		float vel_d = m_velocity.getDistanceFrom(m_last_sent_velocity);
 		if (move_d > minchange || vel_d > minchange ||
-				std::fabs(m_yaw - m_last_sent_yaw) > 1.0) {
+				std::fabs(m_yaw - m_last_sent_yaw) ||
+				std::fabs(m_pitch - m_last_sent_pitch) ||
```
SmallJoker: `> 1.0`

src/content_sao.cpp
```diff
@@ -477,7 +483,9 @@ void LuaEntitySAO::step(float dtime, bool send_recommended)
 		move_d += m_last_sent_move_precision;
 		float vel_d = m_velocity.getDistanceFrom(m_last_sent_velocity);
 		if (move_d > minchange || vel_d > minchange ||
-				std::fabs(m_yaw - m_last_sent_yaw) > 1.0) {
+				std::fabs(m_yaw - m_last_sent_yaw) ||
```
SmallJoker: `> 1.0`

src/genericobject.cpp
```diff
@@ -56,6 +58,10 @@ std::string gob_cmd_update_position(
 	writeV3F1000(os, acceleration);
 	// yaw
 	writeF1000(os, yaw);
+	// pitch
+	writeF1000(os, pitch);
+	// roll
+	writeF1000(os, roll);
```
SmallJoker: Only send pitch and roll here if automatic_rotate is above `0.001f` or `GenericCAO::processMessage` will be offset at reading the packet when the automatic rotation is disabled.

paramat:
Automatic rotate is not much needed for pitch and roll, it exists for yaw mostly because item entities rotate. It could be a later PR.

SmallJoker:
@CoderForTheBetter 
> I would hope that a network break would acceptable here

Yes, currently is 0.5.0-dev meant to be unstable regarding network communication. We already had a few commits which broke the compatibility in some way, so it's fine to break it again.

> I feel a name change to "setLookPitch()", "setLookPitchAndSend()", "getLookPitch()", and "getRadLookPitch()" would be better 

Yes, that would help to distinguish both values.

src/content_sao.cpp
CoderForTheBetter: Not sure what you want here. Should it be `> 1.0f`?

CoderForTheBetter: Thanks, it's checking if it is true, not greater than. Can't believe I missed that.

src/script/lua_api/l_object.cpp
CoderForTheBetter: I'll need some more clarification here. "this one" what? Remove that line? Then the pitch and roll will never be assigned. It is exactly the same way that yaw is set (already implemented before I did anything). I'm sorry you have to deal with my inexperience, but I'm trying my best to get this implemented since it will break things and 0.5.0 is the best time to do that. Thanks again.

src/genericobject.cpp
CoderForTheBetter: Hmm. This kind of a hard one. Right now, on the client side in [content_cao.cpp](https://github.com/CoderForTheBetter/minetest/blob/31f38a23f7143a38738d8c0c35d8ef5b472acd52/src/content_cao.cpp#L1307) it is read like this due to automatic_rotate:
```
if (std::fabs(m_prop.automatic_rotate) < 0.001f) {
	m_yaw = readF1000(is);
	m_pitch = readF1000(is);
	m_roll = readF1000(is);
} else {
	readF1000(is);
}
```
I'll just point that out for now, so to accomplish what you're asking could I introduce a new parameter into gob_cmd_update_position() ([defined here, where we are talking about it](https://github.com/CoderForTheBetter/minetest/blob/31f38a23f7143a38738d8c0c35d8ef5b472acd52/src/genericobject.cpp#L39)) for automatic_rotate?

**To summerize:**
I want to add a parameter to gob_cmd_update_position() [here](https://github.com/CoderForTheBetter/minetest/blob/31f38a23f7143a38738d8c0c35d8ef5b472acd52/src/genericobject.cpp#L39) so that it takes on the form of:
```
std::string gob_cmd_update_position(
	v3f position,
	v3f velocity,
	v3f acceleration,
	f32 yaw,
	f32 pitch,
	f32 roll,
        f32 automatic_rotate
	bool do_interpolate,
	bool is_movement_end,
	f32 update_interval
)
(Note the new parameter "automatic_rotate." Hopfully f32 type is fine/good)
```
With the new parameter I can do this inside of gob_cmd_update_position():
```
...
// yaw
writeF1000(os, yaw);
if (automatic_rotate < 0.001f) {
	// pitch
	writeF1000(os, pitch);
	// roll
	writeF1000(os, roll);
}
...
```

The only places now that gob_cmd_update_position() is called from (has always been like this) is in the file content_sao.cpp in the function [LuaEntitySAO::sendPosition](https://github.com/CoderForTheBetter/minetest/blob/31f38a23f7143a38738d8c0c35d8ef5b472acd52/src/content_sao.cpp#L786) and in the function [PlayerSAO::step](https://github.com/CoderForTheBetter/minetest/blob/31f38a23f7143a38738d8c0c35d8ef5b472acd52/src/content_sao.cpp#L1117).

**The LuaEntitySAO::sendPosition call of gob_cmd_update_position will look like so:**
```
std::string str = gob_cmd_update_position(
	m_base_position,
	m_velocity,
	m_acceleration,
	m_yaw,
	m_pitch,
	m_roll,
	m_prop.automatic_rotate
	do_interpolate,
	is_movement_end,
	update_interval
);
```

**The PlayerSAO::step call of gob_cmd_update_position will look like so:**
```
std::string str = gob_cmd_update_position(
	pos,
	v3f(0.0f, 0.0f, 0.0f),
	v3f(0.0f, 0.0f, 0.0f),
	m_yaw,
	m_pitch,
	m_roll,
	m_prop.automatic_rotate
	true,
	false,
	update_interval
);
```
Should I just make commit so that you can review it and understand it easier, @SmallJoker?

src/script/lua_api/l_object.cpp
paramat: I think this is sarcasm referring to the 'Do it' comment, which is unnecessary.

src/content_sao.cpp
paramat: `1.0f` is better to define a 'float' instead of a 'double'

src/content_sao.cpp
paramat: 1.0f

src/script/lua_api/l_object.cpp
CoderForTheBetter: Ah, I see. I was just keeping the comments the same since it was already in the yaw function. Will remove it. Wasn't my doing, blame whoever wrote the first function (and me, I should have thought the same).

paramat: Ah ok, yes current code sometimes has stuff we now prefer to change.

src/content_sao.cpp
CoderForTheBetter: I thought so! Floats are being compared not doubles.

src/script/lua_api/l_object.cpp
CoderForTheBetter: There are a ton of other functions in the same area that are doing the same thing I did. They will need to be fixed but I will not touch them since I do not want my pull request to spread too far.

src/content_sao.h
```diff
@@ -33,10 +33,18 @@ class UnitSAO: public ServerActiveObject
 	virtual ~UnitSAO() = default;
 
 	virtual void setYaw(const float yaw) { m_yaw = yaw; }
+        virtual void setPitch(const float pitch) { m_pitch = pitch; }
+        virtual void setRoll(const float roll) { m_roll = roll; }
```
paramat: These 2 lnes seem to have an extra space of indent.

src/script/lua_api/l_object.cpp
paramat: Yes it's fine to leave the unrelated code issues out of this PR.

src/script/lua_api/l_object.cpp
paramat: Again no need to remove these comments from unrelated functions.

src/content_sao.h
paramat: No problem :) it's not that messy actually, and pretty good for a first contribution.

src/script/lua_api/l_object.cpp
```diff
@@ -907,13 +907,65 @@ int ObjectRef::l_get_yaw(lua_State *L)
...
 	return 1;
 }
 
+// set_pitch(self, radians) (X-AXIS)
+int ObjectRef::l_set_pitch(lua_State *L)
+{
+	NO_MAP_LOCK_REQUIRED;
+	ObjectRef *ref = checkobject(L, 1);
+	LuaEntitySAO *co = getluaobject(ref);
+	if (!co) return 0;
```
paramat:
```
if (!co)
    return;
```
The action is always on a new line

CoderForTheBetter: Again, I'm learning bad habits from old code, will fix it.

src/script/lua_api/l_object.cpp
```diff
@@ -907,13 +907,65 @@ int ObjectRef::l_get_yaw(lua_State *L)
...
+	float pitch = luaL_checknumber(L, 2) * core::RADTODEG;
+
```
paramat: Best have the empty line immediately after the `return;`
Below too.

src/script/lua_api/l_object.cpp
paramat: See https://dev.minetest.net/Code_style_guidelines if you haven't already.

src/script/lua_api/l_object.cpp
CoderForTheBetter: Would this suffice? Any improvements? Looks like too many empty lines:
```
int ObjectRef::l_set_yaw(lua_State *L)
{
	NO_MAP_LOCK_REQUIRED;
	if (!co) 
	    return 0;
	
	float yaw = luaL_checknumber(L, 2) * core::RADTODEG;

	co->setYaw(yaw);
	return 0;
}
```

src/script/lua_api/l_object.cpp
CoderForTheBetter: Yeah, I looked at it and I've started following as well as I can.

src/server.cpp
```diff
@@ -1811,13 +1811,13 @@ void Server::SendMovePlayer(session_t peer_id)
 	assert(sao);
 
 	NetworkPacket pkt(TOCLIENT_MOVE_PLAYER, sizeof(v3f) + sizeof(f32) * 2, peer_id);
-	pkt << sao->getBasePosition() << sao->getPitch() << sao->getYaw();
+	pkt << sao->getBasePosition() << sao->getLookPitch() << sao->getYaw();
```
CoderForTheBetter: I was confused by sao.getYaw() here but remember that PlayerSAO inherits from UNITSAO. UNITSAO has contains the function for setYaw(), not in PlayerSAO. It is getting the player model yaw.

CoderForTheBetter:
As far as I can tell, all that is left is for @SmallJoker to answer my question [here.](https://github.com/minetest/minetest/pull/7331#discussion_r188103961) @rubenwardy, if you could give it another review sometime I would be very thankful, although paramat has pointed out a lot of errors in my style/formatting.

Thanks again, everyone! I really appreciate your time.

src/script/lua_api/l_object.cpp
paramat: Agreed, just one empty line is enough, often we add an empty line after a 'return'.

src/genericobject.cpp
CoderForTheBetter: @SmallJoker Great! That is much more simple. This is great discussion if automatic_rotate should ever be added for pitch and roll. I may open an issue after this that refers to this thread. I will fix this and review all of my code once again for style issues, then I'll push everything.

CoderForTheBetter:
A lot of stuff has gone down in this pull request. Here's a list as of commit [30763cb](https://github.com/minetest/minetest/pull/7331/commits/30763cbabb12e22d37e3cbed10a207f65a8e75ee):

- Added Lua methods/API for pitch and roll
- Save/serialize the rotations
- Made pitch & roll independent of automatic_rotate in ProcessMessage (maybe add later)
- Fixed indent spaces by making them tab indents
- Fixed if statement style -> egyptian style
- Got rid of non-empty empty lines
- Renamed/Refactored C++ function names for player yaw and look pitch across the engine
- Added documentation to lua_api.txt for pitch and roll

Hopefully that is all

paramat:
Code style seems fine now.

SmallJoker:
There's a little problem with the player pitch: https://www.youtube.com/watch?v=-zsb09gTwtQ

CoderForTheBetter:
@SmallJoker That's really messed up. This is most likely because of the functions I renamed/refactored.

paramat:
Just wanted to check something.
What are pitch and roll relative to? First yaw is applied as rotation around Y axis, then pitch should always be rotation around a yawed left-right axis through the model, not around the X or Z axis. Roll should be rotation around the yawed and pitched front-back axis through the model, not around the X or Z axis.
Hopefully this is already taken care of in the engine.

CoderForTheBetter:
@paramat I think Irrlicht has got our back if I understand what you're saying. Here's a [gif](https://thumbs.gfycat.com/GiganticTautAnemoneshrimp-size_restricted.gif) of pitch and roll. 
axis-size_restricted
It rolls (increments up) first and then its pitch increments upwards. I did not test the yaw but I can if you want me to. After I fix the player pitch bug I'm going to test everything I can think of: yaw, pitch, roll, and with players/objects attached/detached and attached again.

Here's Lua code for this gif:
```
local p = 0
local r = 0

function Helicopter:on_step(_, dtime)  
  if r < 2 then
    r=r+0.01
    self.object:set_roll(r)
    
  elseif p < 2 then
    p=p+0.01
    self.object:set_pitch(p)
    
  else
    p = 0
    r = 0
    self.object:set_roll(r)
    self.object:set_pitch(p)
    
  end
  minetest.chat_send_all("Roll: " .. tostring(r))
  minetest.chat_send_all("Pitch: " .. tostring(p))
  minetest.chat_send_all(" ")
end
```

paramat:
Ok that look correct.
Green is roll axis, model front-back.
Red is pitch axis, model left-right.
Blue is yaw axis, model up-down.

CoderForTheBetter:
@SmallJoker latest commit seems to have fixed the player pitch issue. It may be kind of a hacky way of doing things (maybe not though, up to you). I'm not sure what will happen if I just don't send pitch and roll when the player yaw is sent, so I just sent a pitch and roll of zero.

src/database/database-postgresql.cpp
```diff
@@ -544,8 +544,8 @@ bool PlayerDatabasePostgreSQL::loadPlayer(RemotePlayer *player, PlayerSAO *sao)
 		PQclear(results);
 		return false;
 	}
-
-	sao->setPitch(pg_to_float(results, 0, 0));
+
```
SmallJoker: Tabulator

src/content_sao.cpp
```diff
@@ -930,9 +948,13 @@ std::string PlayerSAO::getClientInitializationData(u16 protocol_version)
 	writeU8(os, 1); // version
 	os << serializeString(m_player->getName()); // name
 	writeU8(os, 1); // is_player
-	writeS16(os, getId()); //id
+	writeS16(os, getId()); // id
 	writeV3F1000(os, m_base_position);
 	writeF1000(os, m_yaw);
+
+	// Send 0 for player pitch and roll (pitch 1st roll 2nd) since we only want the player model to turn/yaw
```
SmallJoker: Comment too long. Please split.

CoderForTheBetter: Fixed.

CoderForTheBetter:
Going back to paramat's previous comment: I don't think the rotation is correct: [GIF](https://giant.gfycat.com/ThriftyAfraidEasteuropeanshepherd.webm)
thriftyafraideasteuropeanshepherd-size_restricted
The rotations are as follows: Pitch -> Roll -> Yaw. I think I know how to do this with Quaternions.

stujones11:
From what I can tell this does not work the way @paramat suggested it ought to. I also believe that does need to be the case for this to be easily understood and utilized. This can indeed be done using quaternions, however, iirc Irrlicht may already provide some helper methods for this.

@CoderForTheBetter Do you happen to have the test media/mod files available somewhere so I can try this out? I may be able to provide a fix.

CoderForTheBetter:
@stujones11, Here's the mod that I have been using to test this PR: [prytest](https://github.com/CoderForTheBetter/prytest). Make sure to rename it to 'prytest' and use the command '/spawn axis' to spawn an axis entity.

Your code looks pretty short and clean. I will have to look at it more.

paramat:
Looking at that animation, pitch looks correct, then roll seems to be around a world axis instead of the green model axis, then yaw is very strange.

paramat:
I think the order that the 3 are applied in Irrlicht is important here. I would consider the logical order of importance: yaw, pitch, roll. With yaw being applied first which rotates the pitch axis (model left-right axis, red), pitch being applied next, then lastly roll is applied around the model front-back axis (green). maybe try different orders of application to see if it makes sense.

CoderForTheBetter:
@paramat, great idea! I have noticed that these problems are not consistent when using different orders of applying rotations. I hope that is the problem. Great observation!

paramat:
> Looking at that animation, pitch looks correct, then roll seems to be around a world axis instead of the green model axis, then yaw is very strange and is around an axis perpendicular to the red model axis.

Just realised roll could be around the 'yaw vector' instead of a world axis. What i mean by 'yaw vector' is a vector confined to the XZ plane that rotates around Y as yaw is applied. At the start of the animation the yaw vector is pointing along the green axis and the yaw axis is vertical.

Now i see the animation as correct:
* Pitch is applied as a rotation around the red model left-right axis. The yaw vector and yaw axis do not change (the yaw vector and yaw axis are always at 90 degrees to each other).
* Roll is applied as a rotation around the yaw vector. The yaw vector does not change but the yaw axis tilts away from vertical.
* Yaw is applied as a rotation around the tilted yaw axis.

You can probably make it clearer by not rotating more than 90 degrees in any one step, around 30-60 degrees is clearest.

Since there are multiple correct ways to apply these 3 rotations Irrlicht had to choose one, but that one will be unintuitive for some uses, like aircraft.

paramat:
Thinking about yaw and pitch only, this system is suited to a player head or a ship's gun: yaw is always a rotation around the Y axis (because only roll tilts the yaw axis, pitch does not).
In games i can imagine that yaw is the most useful rotation, pitch is second, roll is least useful, and the most useful implementation is to always keep the yaw axis vertical. So it makes sense that Irrlicht chose this system.

CoderForTheBetter:
@paramat. This is going to take me a little bit to understand I think. For now, here are some animations.

1. Yaw -> Pitch -> Roll in increments of 0.1 radians (5.7 degrees) and stop once above 2.25 radians (129 degrees): [GIF #1](https://giant.gfycat.com/GeneralCompleteCleanerwrasse.webm)

2. Yaw -> Pitch -> Roll in increments of 0.1 radians (5.7 degrees) and stop once above 0.79 radians (45 degrees): [GIF #2](https://giant.gfycat.com/TintedSoftCricket.webm)

stujones11:
@paramat I think you had the right idea to begin with. Correct yaw, pitch and roll should always be around the local model axes. Irrlicht does not have any concept of these methods, only `setRotation` and it is not enough to simply modify the respective axis. If you want to just expose Irrlicht's

stujones11:
@paramat I think you had the right idea to begin with. Correct yaw, pitch and roll should always be around the local model axes. Irrlicht does not have any concept of these methods, only `setRotation` and it is not enough to simply modify the respective axis. If you want to just do as Irrlicht does then we need to stop calling it pitch yaw and roll and instead have a `set_rotation` method of our own.

However, for this to be any real use to modders, minetest engine should take care of things so that pitch yaw and roll always behave as expected (which it is perfectly doable). Otherwise anyone thinking of using this for say a flight sim mod will be forced to take care of all the heavy-duty quaternion/matrix maths in Lua to avoid such problems as [gimbal-lock](https://en.wikipedia.org/wiki/Gimbal_lock)

CoderForTheBetter:
@stujones11, I got sort of good results with your code.

In the file content_cao.h in the class class GenericCAO : public ClientActiveObject I made these variables:
```
float m_last_yaw = 0.0f;
float m_last_pitch = 0.0f;
float m_last_roll = 0.0f;
```
Which are then used as so in content_cao.cpp (along with your function):
```
void rotateAroundLocalAxis(scene::ISceneNode *node, f32 delta, v3f axis)
{
	core::matrix4 m, n;
	m.setRotationDegrees(node->getRotation());
	n.setRotationDegrees(axis * delta);
	m *= n;
	node->setRotation(m.getRotationDegrees());
}

void GenericCAO::updateNodePos()
{
	if (getParent() != NULL)
		return;

	scene::ISceneNode *node = getSceneNode();

	if (node) {
		v3s16 camera_offset = m_env->getCameraOffset();
		node->setPosition(pos_translator.vect_show - intToFloat(camera_offset, BS));
		if (node != m_spritenode) { // rotate if not a sprite
//			v3f rot = node->getRotation();
//			rot.Y = -m_yaw;
//			rot.X = -m_pitch;
//			rot.Z = -m_roll;
//			node->setRotation(rot);
		    
			// Only one axis should be modified at a time due to
			// how the server and client send/process modifications.
			// Check if yaw changed first since used most for player
			if (m_yaw != m_last_yaw) {
			    rotateAroundLocalAxis(node, irr::core::abs_(m_yaw - m_last_yaw), v3f(0, 1, 0));
			} else if (m_pitch != m_last_pitch) {
			    rotateAroundLocalAxis(node, irr::core::abs_(m_pitch - m_last_pitch), v3f(1, 0, 0));
			} else if (m_roll != m_last_roll){
			    rotateAroundLocalAxis(node, irr::core::abs_(m_roll - m_last_roll), v3f(0, 0, 1));
			}
		}
	}
}
```

Here's a gif of the result: [](GIF). Notice what happens when all the axes go back to zero. The error could be becuase of using abs value, and getting rid of that did not help and changing the order of what is subtratcted from eachother did not help. Not sure what is happening but I will continue to work on it.

Also, it might be nessacry to save m_last_yaw, m_last_pitch, and m_last_roll when the world shuts down but let's get this working first.

stujones11:
I think perhaps the problem is calculating the correct delta when the angle goes over 360 degrees back to zero and vice-versa. Maybe converting to radians would help, although there is likely a better method.

SmallJoker:
If moving over 360° towards 0° and back is an issue: I've already worked on a function which [handles this transition](https://github.com/minetest/minetest/pull/6825/files#diff-a408137ced7567bc503aa6f6386af956R375) while staying inside the min (0) and max (definable) value range.

paramat:
stujones11 i agree, the way Irrlicht works will be a problem for some uses, like aircraft and spacecraft, the Lua code required to convert would be a nightmare for modders. so options and engine code seems needed to get different behaviours.

The yaw axis always being vertical when only using yaw and pitch is useful for some uses, and matches how our current player view direction works.
Maybe there can be a choice of 2 systems:
* Yaw plus pitch only. Works as Irrlicht does it, like a player head, yaw axis always vertical.
* Yaw, pitch, roll. Works as rotations around model axes like an aircraft, extra code does conversion.

stujones11:
Having given this some more consideration, maybe adding the complexity of compound pitch, roll & yaw rotations to MTE is simply not worth the hassle.

Instead, we could simply expose `set_rotation` & `get_rotation` methods which would also be consistent with attachments. An additional method such as `object:rotate(axis, degrees)` that would rotate a model's local axis by a specified amount could be very helpful for vehicle mods in general.

IMO these ideas combined would offer maximum flexibility with minimum overhead and complexity.

CoderForTheBetter:
I still feel like we should make the engine figure this out. All I need to do with this now is figure out how to calculate the delta for what you provided @stujones11, as I am almost positive that is the problem. I tried what SmallJoker linked to no success.

orwell96:
Paramat's idea seems good. I would prefer a simple yaw-always-global approach too, because else I would need to
-set pitch to 0
-set new yaw
-set pitch to original value
everytime I need to rotate my entity around the global y axis.

SmallJoker:
@CoderForTheBetter I haven't tested this yet - but why can't you wrap the angle delta and update all axis after eachother when they were changed? 
```C++
if (m_yaw != m_last_yaw)
    rotateAroundLocalAxis(node, wrapDegrees_0_360(m_yaw - m_last_yaw), v3f(0, 1, 0));
if (m_pitch != m_last_pitch)
    rotateAroundLocalAxis(node, wrapDegrees_0_360(m_pitch - m_last_pitch), v3f(1, 0, 0));
if (m_roll != m_last_roll)
     rotateAroundLocalAxis(node, wrapDegrees_0_360(m_roll - m_last_roll), v3f(0, 0, 1));

m_last_yaw = m_yaw
m_last_pitch = m_pitch
m_last_roll = m_roll
```

stujones11:
@SmallJoker That would probably better but a potential problem is that the resulting rotation is order dependent. Unfortunately, I don't understand the network side of things very well but what I suspect is that it may be possible that multiple rotations could accumulate in a singe server step yet the order in which they were applied is lost.

For this method to work each successive call to pitch, roll or yaw must be processed in the correct order, otherwise the result may not be as intended.

@CoderForTheBetter I would also much prefer to have working pitch, yaw and roll methods, however, if that is not going to be feasible ATM then I much prefer the idea of direct access to the rotation vector over only allowing pitch.

I'm not really sure what could be up with your test code, the matrix rotation method appears to work flawlessly in my app. However, it is using a fixed 'delta' which makes me wonder if it's some sort accumulated rounding or conversion error somewhere. Also, what is not clear from your code is where `m_last_yaw` and friends are updated.

I have not yet had a chance to try this out but should be able to spend some time on this at the weekend. However, if my suspicions over the update are correct then there may not be an easy fix :(

CoderForTheBetter:
@stujones11 I've looked at forums posts for Irrlicht, Unity3D, and Urho3D but they never seem to apply correctly or Unity3D handles the problem itself with closed-source code/functions. I have also seen posts about parenting to an empty node but it seems dirty to me but if it works and no one minds it then it could definitely be implemented.

stujones11:
@CoderForTheBetter This is your PR but implementing true YPR in minetest is going to be somewhat non-trivial. If you want this merged anytime soon then you might be well advised to simply drop the 'roll'. Unless you are some sort of merge conflict ninja :)

However, I do think that adding direct access to the rotation vector should also be done at some point, at least. The incremental rotate method I suggested could also save a lot of `get_yaw` & `set_yaw` calls in vehicle mods.

CoderForTheBetter:
@stujones11 I agree with everything you are saying. Direct access should be implemented and incremental rotation should be implemented. To avoid wrapping angles and unintuitive rotations, like I have been testing, do you agree that we should _only_ allow incrementing pitch, roll, and yaw with a function like `incYaw(angleAmount)` etc... This would mean when an entity is spawned into the world its pitch, roll, and yaw are all zero, but then something like `incPitch(5)` is called and the new pitch will be lastPitch+pitch or 0+5 (upwards pitch) and then to pitch down `incPitch(-5)` would be 5+(-5) = 0. I would only implement the last part since I do not want this to get too big.

Also, I have not tested only pitch and yaw by themselves. For all I know at this point there could still be the same problem but with one less axis of rotation but maybe you already have that figured out and I just missed it.

By the way, I really appreciate your help with this PR, thank you.

paramat:
If roll is a big problem we could reserve a position for it in network protocol but just not use it immediately and implement yaw and pitch at first? If so, would this mean yaw is always around the Y axis? I guess it has to be to avoid order issues.

stujones11:
@CoderForTheBetter You can have working pitch and yaw by simply removing all mention of roll from your current implementation. Two axes play well together, three is a crowd.

The incremental rotate method could simply expose the same matrix rotation method I've shown where the axis is specified as unit vector.

PS. I am glad to help with this as it's something I have wanted to see added for some time. It's nice to see someone taking the initiative :)

CoderForTheBetter:
@paramat that seems like a good compromise. Although, if we use the pitch and yaw that would be implemented in the way you are talking about, my idea of only allowing incrementing of pitch, roll, and yaw would have to be implemented with more variables in the network (very not idea) or an extra bit for using those variables in different ways. For example, if bit = 0 then directly use those variables (p,r,y) to set rotation; if bit = 1 use those variables to increment pitch, roll, and yaw. I hope that makes sense.

CoderForTheBetter:
@stujones11 Great, I think for this pull request I will I will do exactly what paramat said above: I will keep pitch, **roll**, and yaw in the network protocol, but I will only use pitch and yaw since they work together, like you said, without any extra work. This means we will only have Lua methods called setYaw() and setPitch() but no setRoll().

Once this PR is approved and pushed to master (after making the above changes) I will open another PR with the incremented approach in my last comments which will, of course, use your matrix code (sorry that I did not make it clear in my last comment that I was planning on using it) and an extra bit in the network protocol for using the pitch, roll, and yaw for direct use or incremented use.

Hopefully that all makes sense.

CoderForTheBetter:
@paramat About the yaw axis, I think it needs to be since that is already what the player rotates on (the y-axis) and for order issues. I have been using this [picture](https://pterneas.com/wp-content/uploads/2017/05/kinect-joint-rotation.png) as a reference because of the player yaw being on the y-axis.

paramat:
@CoderForTheBetter your plan seems fine to me.
Remember to have these 3 stated in code and ordered in network protocol in the order of decreasing importance YP(R).

CoderForTheBetter:
Thinking about it some more, the method we just came up with has downsides and some of the other ideas seem better.

**Method 1**

Implement set_rotation(), get_rotation(), incPitch()/Roll()/Yaw() and getIncPitch()/Roll()/Yaw()

**What will be added in the network:**
Variables for pitch, roll, and yaw, and to differenticaite between directly setting rotation (set_rotation()) and incremeting by the varaibles a bit (bool) will be added to the network.

**PROS**
- A lot of flexiblity since players can come up with other method themselves when directly exposing set_rotation() in Lua.
- No axes are wasted.
- If a mod wants to only use pitch or yaw they hae two methods to do so.
- Easy to make vehicles that pitch, roll, and yaw.

**CONS**

- Another piece of data has to added to the network protocal (only a single bit).
- A check has to be made (using extra bit) when setting rotation. May not seem like a big problem but right now yaw is applied with little question (only checks if it is a sprite) and an extra if-statement might impact performance.


**Method 2**
Only implement set_rotation() and get_rotation()

**What will be added in the network:**
variables for pitch, roll, and yaw.

**PROS**
- Again, a lot of flexiblity since players can come up with other method themselves when directly exposing set_rotation() in Lua.
- Again, no axes are wasted.
- No extra if-statement when applying the rotation.
- No extra bit in the network.

**CONS**

- Harder to make vehicles that pitch, roll, and yaw.
- Yawing and pitching not easy or intuitve.

CoderForTheBetter:
Sorry about accidental close (canceling a comment).

stujones11:
> there will be easy methods for pitch, roll, and yaw

Rather than bloat the api with three separate methods, I would simply add a `rotate(vector)` method that multiplies the entity matrix with the given vector matrix, so to rotate 5 degrees around the Y axis you would simply write `obj:rotate({x=0, y=5, z=0})` This is actually more powerful as it would also allow for compound rotations where I believe the order is Y (yaw), X (pitch), Z (roll) [needs verification]

More importantly, however, if such a method is added then it is imperative (IMO) that we also have `get_rotation` & `set_rotation` methods otherwise it will be practically impossible to know the exact heading or restore the orientation.

I agree this should be a separate PR, I just though I would mention these ideas in advance.

CoderForTheBetter:
See #7395 

By the way, this PR can be closed since I will not be adding to it anymore. The rest will take place in the PR I linked and another later on.

PR #7351, opened 2018-05-19
Allow enter to select items from combobox's list.

basicer:
Rebased and addressed concerns in #6570.


SmallJoker:
It does what it should. Would it probably be possible for you to change it so that only the opened combobox (dropdown) is closed on escape? A 2nd escape press should then close the form.

@paramat 
> Only RETURN should select, Escape / Cancel / Inventory keys should continue to close the form.

I'm a bit late but would it be okay for you if we could make it so that escape doesn't always close the form directly?

basicer:
@SmallJoker I had hoped to do that, however the simple approach of rejecting the escape key similar to the return key and letting it bubble to the combo box doesn't work.  Nothing happens in that case.

PR #7339, opened 2018-05-15
Add discord rich presence necessary merge pliz

Dumbeldor:
I re-open my PR #7303 . I invite you to read the comments of the previous one to understand the problem. The idea is to add the rich presence of discord on Minetest. But it seems that some core developer Minetest refuse. **I invite you to express your point of view.**

But please, can we evolve and stop falling in the "Discord is owner, it's evil"

PR #7353, opened 2018-05-19
Add a MSVC / Windows compatible snprintf function

nOOb3167:
This almost makes me want to cry but basically Microsoft did not consider printf-style positional arguments (format specifier such as "%1$s") important enough for inclusion in standard library. Positional arguments are instead supported by separate functions (standard printf without positional support versus _printf_p with positional support).
To make things better, these separate positional functions sometimes behave differently from the standard ones.

In this PR a function named mt_snprintf is added to util/string.* (see @SmallJoker comment on #7308), implementing snprintf. This is done using _vsprintf_p, in conjunction with _vscprintf_p correcting the return value.

For other platforms, mt_snprintf can just call vsnprintf.

A use case in minetest for a snprintf equivalent supporting positional arguments is translation support sometimes having to change the argument order (example: "%1$d tomatoes bought by player %2$s" versus "igralec %2$s kupi %1$d paradiznikov").
This PR improves upon #7308 by removing the need for ifdef WIN32 codepath with manual 'bufferSize' calculation. Instead snprintf calls can just be changed to mt_snprintf.

### However

Author of #7308 is also developing #7339 and requested that:

> Dumbeldor commented
>> PS : Merge PR : #7316 and #7308 before this one

The #7308 snprintf commit is in the middle of about a dozen #7339 commits, and would thus be more complicated for Dumbeldor to rebase.
I recommend possibly coordinating with #7339 to be resolved before applying this.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -53,6 +53,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include "nodedef.h"         // Needed for determining pointing to nodes
 #include "nodemetadata.h"
 #include "particles.h"
+#include "porting.h"
 #include "profiler.h"
 #include "quicktune_shortcutter.h"
 #include "raycast.h"
@@ -64,7 +65,6 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include "util/basic_macros.h"
 #include "util/directiontables.h"
 #include "util/pointedthing.h"
-#include "util/string.h"
 #include "irrlicht_changes/static_text.h"
 #include "version.h"
 #include "script/scripting_client.h"
@@ -4024,7 +4024,7 @@ void Game::showPauseMenu()
 
 	 char control_text_buf[600];
 
-	 mt_snprintf(control_text_buf, ARRLEN(control_text_buf), control_text_template.c_str(),
+	 porting::mt_snprintf(control_text_buf, ARRLEN(control_text_buf), control_text_template.c_str(),
 			GET_KEY_NAME(keymap_forward),
 			GET_KEY_NAME(keymap_backward),
 			GET_KEY_NAME(keymap_left),
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -64,6 +64,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include "util/basic_macros.h"
 #include "util/directiontables.h"
 #include "util/pointedthing.h"
+#include "util/string.h"
 #include "irrlicht_changes/static_text.h"
 #include "version.h"
 #include "script/scripting_client.h"
@@ -4023,7 +4024,7 @@ void Game::showPauseMenu()
 
 	 char control_text_buf[600];
 
-	 snprintf(control_text_buf, ARRLEN(control_text_buf), control_text_template.c_str(),
+	 mt_snprintf(control_text_buf, ARRLEN(control_text_buf), control_text_template.c_str(),
 			GET_KEY_NAME(keymap_forward),
 			GET_KEY_NAME(keymap_backward),
 			GET_KEY_NAME(keymap_left),
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -4024,7 +4024,7 @@ void Game::showPauseMenu()
 
 	 char control_text_buf[600];
 
-	 porting::mt_snprintf(control_text_buf, ARRLEN(control_text_buf), control_text_template.c_str(),
+	 porting::mt_snprintf(control_text_buf, sizeof(control_text_buf), control_text_template.c_str(),
 			GET_KEY_NAME(keymap_forward),
 			GET_KEY_NAME(keymap_backward),
 			GET_KEY_NAME(keymap_left),
--- a/src/gui/profilergraph.cpp
+++ b/src/gui/profilergraph.cpp
@@ -18,6 +18,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
 
+#include "porting.h"
 #include "profilergraph.h"
 #include "util/string.h"
 
@@ -95,12 +96,12 @@ void ProfilerGraph::draw(s32 x_left, s32 y_bottom, video::IVideoDriver *driver,
 
 		s32 texth = 15;
 		char buf[10];
-		mt_snprintf(buf, 10, "%.3g", show_max);
+		porting::mt_snprintf(buf, 10, "%.3g", show_max);
 		font->draw(utf8_to_wide(buf).c_str(),
 				core::rect<s32>(textx, y - graphh, textx2,
 						y - graphh + texth),
 				meta.color);
-		mt_snprintf(buf, 10, "%.3g", show_min);
+		porting::mt_snprintf(buf, 10, "%.3g", show_min);
 		font->draw(utf8_to_wide(buf).c_str(),
 				core::rect<s32>(textx, y - texth, textx2, y), meta.color);
 		font->draw(utf8_to_wide(id).c_str(),
--- a/src/gui/profilergraph.cpp
+++ b/src/gui/profilergraph.cpp
@@ -95,12 +95,12 @@ void ProfilerGraph::draw(s32 x_left, s32 y_bottom, video::IVideoDriver *driver,
 
 		s32 texth = 15;
 		char buf[10];
-		snprintf(buf, 10, "%.3g", show_max);
+		mt_snprintf(buf, 10, "%.3g", show_max);
 		font->draw(utf8_to_wide(buf).c_str(),
 				core::rect<s32>(textx, y - graphh, textx2,
 						y - graphh + texth),
 				meta.color);
-		snprintf(buf, 10, "%.3g", show_min);
+		mt_snprintf(buf, 10, "%.3g", show_min);
 		font->draw(utf8_to_wide(buf).c_str(),
 				core::rect<s32>(textx, y - texth, textx2, y), meta.color);
 		font->draw(utf8_to_wide(id).c_str(),
--- a/src/gui/profilergraph.cpp
+++ b/src/gui/profilergraph.cpp
@@ -96,12 +96,12 @@ void ProfilerGraph::draw(s32 x_left, s32 y_bottom, video::IVideoDriver *driver,
 
 		s32 texth = 15;
 		char buf[10];
-		porting::mt_snprintf(buf, 10, "%.3g", show_max);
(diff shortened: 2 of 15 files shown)


src/util/string.cpp
```diff
@@ -945,3 +947,22 @@ std::wstring translate_string(const std::wstring &s) {
...
+
+int mt_snprintf(char * const buf, const size_t buf_size, const char *fmt, ...) {
+	// https://msdn.microsoft.com/en-us/library/bt7tawza.aspx
+	//  Many of the MSVC / Windows printf-style functions do not support positional arguments (eg. "%1$s").
+	//  We just forward the call to vsnprintf for sane platforms, but defer to _vsprintf_p on MSVC / Windows.
+	// https://github.com/FFmpeg/FFmpeg/blob/5ae9fa13f5ac640bec113120d540f70971aa635d/compat/msvcrt/snprintf.c#L46
+	//  _vsprintf_p has to be shimmed with _vscprintf_p on -1 (for an example see above FFmpeg link).
+	va_list args;
+	va_start(args, fmt);
+#ifndef _WIN32
```
SmallJoker: Should be `_MSC_VER`
https://github.com/minetest/minetest/pull/7308#discussion_r186995693

sfan5:
To avoid having to change all code, maybe util/string.h could redefine `snprintf` to point to `mt_snprintf`.

> The #7308 snprintf commit is in the middle of about a dozen #7339 commits, and would thus be more complicated for Dumbeldor to rebase.

This is not our problem

nOOb3167:
Yes i would rather chew on styrofoam than coordinate too :+1:

sfan5:
?

nOOb3167:
> ?

I am saying that I agree with you. Tracing back the post chain:

nOOb3167:
> The #7308 snprintf commit is in the middle of about a dozen #7339 commits, and would thus be more complicated for Dumbeldor to rebase.
> I recommend possibly coordinating with #7339 to be resolved before applying this.

sfan5:
> This is not our problem

nOOb3167:
> Yes i would rather chew on styrofoam than coordinate too +1

It is better this way. Working to get this fix in not waiting for #7339.


### Anyway more importantly:

> To avoid having to change all code, maybe util/string.h could redefine snprintf to point to mt_snprintf.

This is infeasible: irrTypes.h defines snprintf to sprintf_s already. It is an Irrlicht header not in the minetest source tree so it can't easily be fixed.
If minetest defines snprintf, for example in util/string.h, as suggested, the redefinition would have to come after irrTypes.h is read by the compiler, overriding the Irrlicht definiton.

Consider the following:
```
#include "minetestheader1"
#include "irrlichtheader1"
snprintf(...)
```
Here imagine than irrlichtheader1 leads to irrTypes.h eventually being included.

We can't just place an include directive for util/string.h after all other includes:
```
#include "minetestheader1"
#include "irrlichtheader1"
#include "util/string.h"
snprintf(...)
```
If minetestheader1 includes some other header, which includes some other header, etc, which then includes util/string.h anywhere on the way, then util/string.h will be included before irrTypes.h.

Since order of inclusion of headers cannot be reasonably guaranteed, especially considering such multiple nested levels of inclusion, we cannot reliably override the snprintf definition.

This means the snprintf calls should be changed.
The change footprint should not be much bigger than redefining snprintf:
Essentially even if redefining snprintf, the include "util/string.h" statement would still be necessary
in every code file using snprintf.

The way I've pushed just now, (with changing snprintf calls to mt_snprintf), we get include "util/string.h" and about one or two renames per every code file using snprintf.

nerzhul:
related to #7308

nOOb3167:
I'd prefer using porting::snprintf, but since the 'snprintf' part of the construct is still a separate preprocessor token, it is subject to macro substitution.
It is now porting::mt_snprintf.

src/gui/profilergraph.cpp
```diff
@@ -95,12 +96,12 @@ void ProfilerGraph::draw(s32 x_left, s32 y_bottom, video::IVideoDriver *driver,
 
 		s32 texth = 15;
 		char buf[10];
-		snprintf(buf, 10, "%.3g", show_max);
+		porting::mt_snprintf(buf, 10, "%.3g", show_max);
```
SmallJoker: Looks like a superfluous change. The size of the resulting string is known and there are no `%1$s` replacements.

sfan5: I think the point here is that you would always use `mt_snprintf` so you can be sure everything works as you expect in all possible cases.

src/game.cpp
```diff
@@ -4023,7 +4024,7 @@ void Game::showPauseMenu()
 
 	 char control_text_buf[600];
 
-	 snprintf(control_text_buf, ARRLEN(control_text_buf), control_text_template.c_str(),
+	 porting::mt_snprintf(control_text_buf, ARRLEN(control_text_buf), control_text_template.c_str(),
```
sfan5: while you're at it: `ARRLEN` can be replaced with `sizeof`

src/gui/profilergraph.cpp
sfan5: constant can be replaced with `sizeof`

src/gui/profilergraph.cpp
```diff
@@ -95,12 +96,12 @@ void ProfilerGraph::draw(s32 x_left, s32 y_bottom, video::IVideoDriver *driver,
...
 		font->draw(utf8_to_wide(buf).c_str(),
 				core::rect<s32>(textx, y - graphh, textx2,
 						y - graphh + texth),
 				meta.color);
-		snprintf(buf, 10, "%.3g", show_min);
+		porting::mt_snprintf(buf, 10, "%.3g", show_min);
```
sfan5: constant can be replaced with `sizeof`

src/map.cpp
```diff
@@ -1727,13 +1727,13 @@ std::string ServerMap::getSectorDir(v2s16 pos, int layout)
 	switch(layout)
 	{
 		case 1:
-			snprintf(cc, 9, "%.4x%.4x",
+			porting::mt_snprintf(cc, 9, "%.4x%.4x",
```
sfan5: constant can be replaced with `sizeof`

src/map.cpp
```diff
@@ -1727,13 +1727,13 @@ std::string ServerMap::getSectorDir(v2s16 pos, int layout)
...
 				(unsigned int) pos.X & 0xffff,
 				(unsigned int) pos.Y & 0xffff);
 
 			return m_savedir + DIR_DELIM + "sectors" + DIR_DELIM + cc;
 		case 2:
-			snprintf(cc, 9, (std::string("%.3x") + DIR_DELIM + "%.3x").c_str(),
+			porting::mt_snprintf(cc, 9, (std::string("%.3x") + DIR_DELIM + "%.3x").c_str(),
```
sfan5: constant can be replaced with `sizeof`

src/map.cpp
```diff
@@ -1791,7 +1791,7 @@ v3s16 ServerMap::getBlockPos(const std::string &sectordir, const std::string &bl
 std::string ServerMap::getBlockFilename(v3s16 p)
 {
 	char cc[5];
-	snprintf(cc, 5, "%.4x", (unsigned int)p.Y&0xffff);
+	porting::mt_snprintf(cc, 5, "%.4x", (unsigned int)p.Y&0xffff);
```
sfan5: constant can be replaced with `sizeof`

src/porting.h
```diff
@@ -327,6 +327,8 @@ bool secure_rand_fill_buf(void *buf, size_t len);
 
 // This attaches to the parents process console, or creates a new one if it doesnt exist.
 void attachOrCreateConsole();
+
+int mt_snprintf(char *const buf, const size_t buf_size, const char *fmt, ...);
```
sfan5: useless `const`

src/script/cpp_api/s_base.cpp
```diff
@@ -293,7 +293,7 @@ void ScriptApiBase::stackDump(std::ostream &o)
 				break;
 			case LUA_TNUMBER:  /* numbers */ {
 				char buf[10];
-				snprintf(buf, 10, "%lf", lua_tonumber(m_luastack, i));
+				porting::mt_snprintf(buf, 10, "%lf", lua_tonumber(m_luastack, i));
```
sfan5: constant can be replaced with `sizeof`

src/serverenvironment.cpp
```diff
@@ -1786,9 +1787,9 @@ static void print_hexdump(std::ostream &o, const std::string &data)
 			int i = i0 + di;
 			char buf[4];
 			if(di<thislinelength)
-				snprintf(buf, 4, "%.2x ", data[i]);
+				porting::mt_snprintf(buf, 4, "%.2x ", data[i]);
```
sfan5: constant can be replaced with `sizeof`

src/serverenvironment.cpp
```diff
@@ -1786,9 +1787,9 @@ static void print_hexdump(std::ostream &o, const std::string &data)
...
 			else
-				snprintf(buf, 4, "   ");
+				porting::mt_snprintf(buf, 4, "   ");
```
sfan5: constant can be replaced with `sizeof`

src/unittest/test_connection.cpp
```diff
@@ -280,7 +281,7 @@ void TestConnection::testConnectSendReceive()
 			if (i % 2 == 0)
 				infostream << " ";
 			char buf[10];
-			snprintf(buf, 10, "%.2X",
+			porting::mt_snprintf(buf, 10, "%.2X",
```
sfan5: constant can be replaced with `sizeof`

src/unittest/test_connection.cpp
```diff
@@ -323,7 +324,7 @@ void TestConnection::testConnectSendReceive()
 			if (i % 2 == 0)
 				infostream << " ";
 			char buf[10];
-			snprintf(buf, 10, "%.2X", ((int)(recvdata[i])) & 0xff);
+			porting::mt_snprintf(buf, 10, "%.2X", ((int)(recvdata[i])) & 0xff);
```
sfan5: constant can be replaced with `sizeof`

sfan5:
Should be fine since you're using `#ifndef _MSC_VER` not `#ifndef _WIN32`

nOOb3167:
The positional specifiers fail on outdated MinGW (tested with version released in 2015, program outputs "hello $s $s", ignoring the specifiers), but otherwise should be fine.

mingwinstall


```
#include <stdio.h>
int main(int argc, char **argv) { printf("hello %2$s %1$s", "abc", "def"); return 0; }
```

sfan5:
MinGW that old will probably not support C++11, so should be fine.

Issue #7354, opened 2018-05-19
Allow Overriding of minetest.get_server_status


Lejo1:
##### Issue type
 - Feature request
 
##### Minetest version
0.4.16
##### Summary
I still tried this it works for the command but not for the message, which you get on join.

Lejo1:
I can also just disable it with show_statusline_on_connect = false
And then I have to do this just on_join, because in lua the overriding works.

Lejo1:
Done ^

PR #7357, opened 2018-05-20
Make the server status message customizable

SmallJoker:
Implements feature request in #7354  

Remove now redundant setting show_statusline_on_connect
Improve documentation of `minetest.get_server_status`

Difference to before: The current player name is now shown in the player list too on join.
```
# Server: version=0.5.0-dev-debug, uptime=0.871, max_lag=0.5459, clients={Foobar}
*** Foobar joined the game.
```

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -1032,11 +1032,6 @@ PlayerSAO* Server::StageTwoClientInit(session_t peer_id)
 	// Send Breath
 	SendPlayerBreath(playersao);
 
-	// Note things in chat if not in simple singleplayer mode
-	if (!m_simple_singleplayer_mode && g_settings->getBool("show_statusline_on_connect")) {
-		// Send information about server to player in chat
-		SendChatMessage(peer_id, ChatMessage(CHATMESSAGE_TYPE_SYSTEM, getStatusString()));
-	}
 	Address addr = getPeerAddress(player->getPeerId());
 	std::string ip_str = addr.serializeString();
 	actionstream<<player->getName() <<" [" << ip_str << "] joins game. " << std::endl;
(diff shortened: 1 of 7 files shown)


src/server.cpp
```diff
@@ -1032,11 +1032,6 @@ PlayerSAO* Server::StageTwoClientInit(session_t peer_id)
 	// Send Breath
 	SendPlayerBreath(playersao);
 
-	// Note things in chat if not in simple singleplayer mode
-	if (!m_simple_singleplayer_mode && g_settings->getBool("show_statusline_on_connect")) {
-		// Send information about server to player in chat
-		SendChatMessage(peer_id, ChatMessage(CHATMESSAGE_TYPE_SYSTEM, getStatusString()));
```
sfan5: does this make `getStatusString()` dead code?

SmallJoker: No, it's still used (in one line) for `core.get_server_status()`.

SmallJoker: I know but `CHATMESSAGE_TYPE_RAW` (from `core.chat_send_player`) and `CHATMESSAGE_TYPE_SYSTEM` are handled exactly the same way in `Client::getChatMessage`.
These different types don't seem to be used anywhere outside the previously mentioned function. Is the purpose for them still in progress or is it legacy left-over?

nerzhul: it was introduced with the chat rework, intended to CSM to know which type is it and handle it as it wants, like adding colors for each type, for example

SmallJoker: Sounds good, but this will definitely need undoing the 3rd argument for `chat_send_player`, so the chat message type can be specified by Lua (prepend removed in 9a3b771).
After all also the player join message is [sent with that function](https://github.com/minetest/minetest/blob/master/builtin/game/misc.lua#L48), which should actually be marked as a server message too.

nerzhul: yes it's exact

PR #7361, opened 2018-05-21
Added crossview support

otdav33:
This is the type of 3d that you get when you cross your eyes. Here's a tutorial I found on the internet for those not familiar: http://www.tutorial-page.de/EN/3d-tutorial-1-an-introduction/

### DIFF
--- a/src/client/render/sidebyside.cpp
+++ b/src/client/render/sidebyside.cpp
@@ -23,8 +23,8 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include "client/hud.h"
 
 RenderingCoreSideBySide::RenderingCoreSideBySide(
-	IrrlichtDevice *_device, Client *_client, Hud *_hud, bool _horizontal)
-	: RenderingCoreStereo(_device, _client, _hud), horizontal(_horizontal)
+	IrrlichtDevice *_device, Client *_client, Hud *_hud, bool _horizontal, bool _flipped)
+	: RenderingCoreStereo(_device, _client, _hud), horizontal(_horizontal), flipped(_flipped)
 {
 }
 
@@ -62,7 +62,7 @@ void RenderingCoreSideBySide::drawAll()
 void RenderingCoreSideBySide::useEye(bool _right)
 {
 	driver->setRenderTarget(_right ? right : left, true, true, skycolor);
-	RenderingCoreStereo::useEye(_right);
+        RenderingCoreStereo::useEye(flipped != _right);
 }
 
 void RenderingCoreSideBySide::resetEye()
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -621,9 +621,10 @@ fall_bobbing_amount (Fall bobbing factor) float 0.0
 #    -    interlaced: odd/even line based polarisation screen support.
 #    -    topbottom: split screen top/bottom.
 #    -    sidebyside: split screen side by side.
+#    -    crossview: Cross-eyed 3d
 #    -    pageflip: quadbuffer based 3d.
 #    Note that the interlaced mode requires shaders to be enabled.
-3d_mode (3D mode) enum none none,anaglyph,interlaced,topbottom,sidebyside,pageflip
+3d_mode (3D mode) enum none none,anaglyph,interlaced,topbottom,sidebyside,crossview,pageflip
 
 #    In-game chat console height, between 0.1 (10%) and 1.0 (100%).
 console_height (Console height) float 0.6 0.1 1.0
--- a/src/client/render/factory.cpp
+++ b/src/client/render/factory.cpp
@@ -40,6 +40,10 @@ RenderingCore *createRenderingCore(const std::string &stereo_mode, IrrlichtDevic
 	if (stereo_mode == "sidebyside")
 		return new RenderingCoreSideBySide(device, client, hud);
 	if (stereo_mode == "topbottom")
-		return new RenderingCoreSideBySide(device, client, hud, true);
+		return new RenderingCoreSideBySide(device, client, hud,
+                        true);
+	if (stereo_mode == "crossview")
+		return new RenderingCoreSideBySide(device, client, hud,
+                        false, true);
 	throw std::invalid_argument("Invalid rendering mode: " + stereo_mode);
 }
--- a/src/client/render/sidebyside.h
+++ b/src/client/render/sidebyside.h
@@ -27,6 +27,7 @@ protected:
 	video::ITexture *left = nullptr;
 	video::ITexture *right = nullptr;
 	bool horizontal = false;
+	bool flipped = false;
 	core::dimension2du image_size;
 	v2s32 rpos;
 
@@ -36,7 +37,7 @@ protected:
 	void resetEye() override;
 
 public:
-	RenderingCoreSideBySide(IrrlichtDevice *_device, Client *_client, Hud *_hud,
-			bool _horizontal = false);
+	RenderingCoreSideBySide(IrrlichtDevice *_device, Client *_client,
+                Hud *_hud, bool _horizontal = false, bool _flipped = false);
 	void drawAll() override;
 };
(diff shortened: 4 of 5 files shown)


src/client/render/sidebyside.cpp
```diff
@@ -62,7 +62,7 @@ void RenderingCoreSideBySide::drawAll()
 void RenderingCoreSideBySide::useEye(bool _right)
 {
 	driver->setRenderTarget(_right ? right : left, true, true, skycolor);
-	RenderingCoreStereo::useEye(_right);
+        RenderingCoreStereo::useEye(flipped != _right);
```
sfan5: `_right ^ flipped` might be clearer

paramat:
I can get crosseyed 3D to work by looking at that example, but are you suggesting we add a rendering that requires players to go crosseyed? That would be unnacceptable obviously, it's uncomfortable even to do this for a few seconds and is unhealthy.

otdav33:
@paramat I can do it for long stretches of time and it doesn't affect my vision badly or give me headaches. If it gives you a headache, just don't use that rendering option. It's not required or even default.

paramat:
How you conciously react to it is unexpected, however this is obviously unhealthy for eyesight (this should be obvious to anyone), going crosseyed with your eyes angled towards a point a few inches in front of your face, while your eyes are actually depth-focussed on the screen.
I recommend you don't use this method, you may not notice issues immediately but you run the risk of harming your eyesight or causing other eyestrain problems.
If something is harmful, making it optional doesn't make it ok for inclusion.
Sorry, i do appreciate you contributing :)

sfan5:
> you run the risk of harming your eyesight

do you have a source for that?
or rather: do you have something that refutes [this](https://health.stackexchange.com/questions/3696/#answer-3926)?

paramat:
Thats easy to refute:

"Why doesn't crossing your eyes cause any damage?

Now, that I've gone over the what, I'll go over the why. Many times throughout the day, you will cross your eyes on purpose, but won't actually think of it as crossing your eyes. Crossing your eyes is the natural reaction to looking at something very close to your face. We need to angle our eyes so we can see objects clearly at a close range. This is essentially the same exact movement you would make when crossing your eyes. Our eyes are designed to be able to move up, down, left, and right, and, while it may not seem like the most natural eye movement, crossing your eyes voluntarily is completely normal."

This is just an opinion by a random person. It is an invalid argument is because it refers to everyday normal viewing of close objects, which is of course natural and not harmful. That is completly different to simulated crossview 3D where your eye orientation and focus are highly unnatural.

Less harmful is the scientifically proven strain of the difference between the screen and the 3D objects that appear in front or behind https://techcrunch.com/2011/07/22/who-could-have-guessed-3d-hurts-your-eyes. This applies to all 3D methods.

There are many opinions on the internet that 3D is not harmless, science says otherwise. Also, everyone knows 3D is hard on the eyes, many peole get nauseous and some people have extreme reactions to it, and this is from the less harsh glasses-based methods used in cinema.

I found this opinion by a doctor (not necessarily correct but makes a good point):
"So 3D viewing with cross eye technique has side effects as your brain has to put extra efforts and your eyes have to maintain unusual orientation for fusion of images , which leads to headache , retro-orbital pain, congestion of eyes etc."
Which makes the important point that crossview 3D is far more hard on eyesight than other methods.

The methods that use coloured or polarised glasses make sure each eye only sees what it should see. Also the eyes are in a fairly normal viewing state.
('Cardboard VR' for smartphones adds a headset that also makes sure that each eye only sees the image it should see, and has lenses too.)

The above methods are far less hard on eyesight because of the hardware that does most of the work for you. However crossview has no hardware and relies on a highly unnatural eye state. It requires the brain to work extra hard to make it work, as well as each eye not seeing just the image it is meant to see, but seeing 2 images side by side and going crosseyed to make them overlap.

We already have support for multiple 3D methods, the less harmful ones, adding a far more harmful one is not a good idea and is not needed.

sfan5:
> Less harmful is the scientifically proven strain of the difference between the screen and the 3D objects that appear in front or behind https://techcrunch.com/2011/07/22/who-could-have-guessed-3d-hurts-your-eyes. This applies to all 3D methods.

Wouldn't it be logical to remove all 3D methods then?

> I found this opinion by a doctor (not necessarily correct but makes a good point): 
> [...]

This describes non-permanent symptoms, which is not equivalent to "harming your eyesight".
Besides, everyone is unique, while it may be uncomfortable for you to use this even for a few seconds, others may be able to comfortably use it for longer times.

> We already have support for multiple 3D methods, the less harmful ones, adding the most harmful one is not a good idea and is not needed.

Aside from the discussion about harmfulness(?), it isn't Minetest's job to prevent users from doing/experiencing harmful things. If someone wants to use cross-eyed 3D, let them.

Semi-related but I'll just say it here:
It's honestly puzzling why you seek to reject pull requests that come with absolutely no downside for the project or its development.

paramat:
Most 3D methods are low-harm because they use viewing hardware and keep the eyes in a more natural viewing state, as explained.
Non-permanent is still harm, pain is harm.
>  it isn't Minetest's job to prevent users from doing/experiencing harmful things.

It should be obviously, you shouldn't encourage people to harm themselves. In extreme cases the law would come into effect.

sfan5, it's very silly how i am having to defend myself here. We should obviously reject useless and ridiculous gimmicks to avoid bloating the code unnecessarily, simplicity is what the roadmap is about. Adding useless stuff has a downside in maintenance.

From the roadmap and celeron55's posts:
"Keep the common base small enough to be manageable, as it is bound to be a monolithic chunk. This is what is called the Minetest Engine."
"Keep the common base lightweight, simple and well-written enough for it to not become some monster that everyone fears, because everyone is dependent on it."
"One question regarding to a mission statement for the engine that really drives reluctance to miscellaneous new features is simplicity. Minetest is meant to be simple."

Crossview is a cheap gimmick for viewing images without hardware for a few seconds, it is unsuitable for video or games, it is also a poor way to do 3D compared to the hardware implementations, and relies on a ridiculous crosseyed technique. This should be blindingly obvious to anyone.

There are some actually useful stereo issues and PRs that are going forgotten, please can we put effort into these instead?
One is Google Cardboard support #4026
Another is an improved implementation of stereo using off-axis #3620 and #5073

We already have a 'side by side' stereo implementation, why is this not usable for crossview?

nerzhul:
okay for me too. It's a user choose, user health is user responsibility, not us, and it's not enabled by default

paramat:
@otdav33 please could you try the existing side-by-side stereo to see if crossview is possible with that?

Also, is the crossview implementation useful as an alternative side-by-side for use with 3D headsets?

otdav33:
>sfan5, it's very silly how i am having to defend myself here. We should obviously reject useless and ridiculous gimmicks to avoid bloating the code unnecessarily. Adding useless stuff has a downside in maintenance.

It will hardly bloat the code seeing that it only adds 4 lines of code and changes 6. I wanted to use the feature, so I implemented it.

>We already have a 'side by side' stereo implementation, is this not usable for crossview?

No. The frames are reversed, which makes it incompatible. For that you need a separate crossview 3d mode.
>Is crossview implementation usable for Google cardboard or other 3D headsets?

No. The frames are reversed, which makes it incompatible. I suspect that you could just use the side-by-side mode for google cardboard, but I haven't tried it.
Another option would be to use crossview instead of google cardboard on your phone, though I don't suspect that you in particular would like it.

paramat:
Ok thanks for the information about the implementation, that should have been included in a description in the first post, to help reviewers but also to help non core-devs understand what is being done.

otdav33:
@paramat I thought someone would look at my code like @sfan5 or @SmallJoker did, so they would have an idea of what was going on. The actual line count of changed code is in the upper-right hand corner of the web view, shown here:
 here.

I suppose that I should've been a bit more descriptive, but much more and I'd be writing more documentation than code.

paramat:
The size of the changes is irrelevant to me, i would still disapprove, sorry.

A description really helps making understanding what the PR does faster and easier, even for core devs who are good with code and even for fairly simple PRs like this. Otherwise to get the basic understanding you have to study the code. It's good that devs and non-devs can have a basic understanding without the effort of studying the code.

A description is required, see https://github.com/minetest/minetest/blob/master/CONTRIBUTING.md

No bad feelings toward you at all despite my opinion of the feature, i appreciate that you contributed to MT. Sorry for the fuss.

PR #7367, opened 2018-05-23
mouse sensitivity

sseemmiin:
To Linux users, the default settings for mouse sensitivity is little bit high.
So, we thought that mouse sensitivity is fixed into "0.005" is better for Ubuntu.

### DIFF
--- a/src/defaultsettings.cpp
+++ b/src/defaultsettings.cpp
@@ -240,7 +240,7 @@ void set_default_settings(Settings *settings)
 
 	// Input
 	settings->setDefault("invert_mouse", "false");
-	settings->setDefault("mouse_sensitivity", "0.2");
+	settings->setDefault("mouse_sensitivity", "0.005");
 	settings->setDefault("repeat_rightclick_time", "0.25");
 	settings->setDefault("safe_dig_and_place", "false");
 	settings->setDefault("random_input", "false");


sfan5:
Th default is fairly high too here on Arch Linux / XFCE.
I have `mouse_sensitivity = 0.075` in my config.

paramat:
The problem with changing the default is that the majority of MT users who are used to it and are happy with it will now have to use a setting to get back what they're used to. The varying values people use shows that it's impossible to judge what is optimum. So it seems best to leave it as it is :-1:

sseemmiin:
Oh, i got it. But we can change the mouse sensitivity in game, so 
It would be better to set similar values ​​rather than too high?

jastevenson303:
Different mice use different DPIs.

nerzhul:
for me default DPI is just good on archlinux with wayland with default options. I think Ubuntu has missed something in their X11 replacement shit

CloudyProton:
I had to bump mine down to 0.1. It is not a matter of DPI as this same mouse if fine in other game's defaults. At the very least, MT mouse sensitivity should reflect other games.

ClobberXD:
Mine's perfect at 3.0; yes it's quite high :)

Mouse sensitivity preferences vary from mouse to mouse _and_ from player to player.

Calinou:
Note that Minetest does not use raw mouse input on any OS, so it will take OS-defined mouse acceleration into account (which is often considered undesirable for first-person games).

I don't know if Irrlicht exposes a way to get raw input though.

paramat:
Current default is ok for me.

SmallJoker:
Closing this PR because it does not solve the actual problem with the varying mouse sensitivity on different systems.

Issue #6869, opened 2018-01-01
Support running server as homeless user


dpflug:
I'm attempting to set up a server and ran into a snag. Minetest assumes the user it's running as will have a home folder. The FreeBSD port sets up a user that doesn't, by default.

path_user and path_cache seemingly can't be overridden at runtime.  They're defined in porting.cpp.

Not having a home directory helps with security, so I suggest adding a couple --flags to define path_user and path_cache. `--data` and `--cache`, perhaps. Thoughts?

4w:
That is not BSD specific. Homeless users are quite common on Linux-based operating systems, too.

Wuzzy2:
Note that “homeless user” here refers to an user without the `home` directory, not an user who is literally homeless. xD

sofar:
nothing wrong with making e.g. `/var/lib/minetest` the homedir of the user.

sofar:
Note that in most unixes, having a `home` folder is required. It may not have to exist, but every account needs to have one defined.

dpflug:
Those are both very good points and are good enough for me. I hadn't thought about making `/var/lib/minetest` the homedir. Unless someone has a compelling reason not to, we should probably just close this.

M1dgard:
On Linux, if `HOME` is not set, the server will crash with
```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
```
due to https://github.com/minetest/minetest/blob/c830347a57c63698f14803233dfc43adeb3fd51f/src/porting.cpp#L469-L470

---

If `HOME` is set to a location not writeable by the minetest user, the server will fail with
```
Cannot create user data directory
```

---

I hope this info might help people to find this issue.

Background info: on Linux I'm using [`runit`](http://smarden.org/runit/) as system service supervisor, and I use its [`chpst`](http://smarden.org/runit/chpst.8.html) command to change the process's permissions, such as the user. It doesn't alter environment variables like HOME.

nerzhul:
now we properly trigger a useful end user error, thanks @M1dgard

dch:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=235355 with a fix for this issue upstream.

PR #7376, opened 2018-05-26
Print error when HOME is not set

M1dgard:
Related to #6869. When no HOME is set, the server crashes with
```
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_M_construct null not valid
```
which leaves the user in the dark. With this commit, the server will crash with
```
minetest/src/porting.cpp:1b3: bool porting::setSystemPaths(): A fatal error occured: Required environment variable HOME is not set
```
instead which is arguably nicer. Feel free to make this prettier if you like.

This PR modifies code for macOS and other Unices too, but tested only on Linux.

### DIFF
--- a/src/porting.cpp
+++ b/src/porting.cpp
@@ -351,6 +351,21 @@ bool getCurrentExecPath(char *buf, size_t len)
 #endif
 
 
+//// Non-Windows
+#if !defined(_WIN32)
+
+const char *getHomeOrFail()
+{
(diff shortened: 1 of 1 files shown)


nerzhul:
nice job, but please factorize the check in one point

M1dgard:
@nerzhul I rewrote the commit.

nerzhul:
Okay for me

PR #7377, opened 2018-05-27
minetest.log: Add mod origin

HybridDog:
See #6205
The log functions can be used for messages other than those generated by minetest.log, too.

Lots of mods prepend `[<modname>] ` to log messages because minetest didn't do this.

No level provided:
```
2018-05-27 12:39:24: [Server]: [luacmd] test
```

Action:
```
2018-05-27 12:39:31: ACTION[Server]: [luacmd] test
```

Warning:
```
2018-05-27 12:39:48: WARNING[Server]: test (at /home/ulrich/.minetest/mods/luacmd/init.lua:26)
```

Deprecated (same as deprecated was before):
```
2018-05-27 12:40:12: WARNING[Server]: test (at /home/ulrich/.minetest/mods/luacmd/init.lua:26)
```

Error:
```
2018-05-27 12:45:34: ERROR[Server]: test
2018-05-27 12:45:34: ERROR[Server]: stack traceback:
2018-05-27 12:45:34: ERROR[Server]: 	[C]: in function 'log'
2018-05-27 12:45:34: ERROR[Server]: 	[string "/lua command"]:1: in function 'cmdFunc'
2018-05-27 12:45:34: ERROR[Server]: 	/home/ulrich/.minetest/mods/luacmd/init.lua:26: in function </home/ulrich/.minetest/mods/luacmd/init.lua:12>
[…]
```
Is there a way to remove the log call from the stack traceback?

### DIFF
--- a/src/script/common/c_internal.cpp
+++ b/src/script/common/c_internal.cpp
@@ -155,6 +155,31 @@ static void script_log(lua_State *L, const std::string &message,
 		infostream << script_get_backtrace(L) << std::endl;
 }
 
+std::string get_optional_origin(lua_State *L, const std::string &modname)
+{
+	// L can be NULL if we get called from scripting_game.cpp
+	if (!L)
+		return "";
+
+	if (modname != "")
+		return "[" + modname + "] ";
+
+	lua_Debug ar;
+
+	if (!lua_getstack(L, 2, &ar))
+		FATAL_ERROR_IF(!lua_getstack(L, 1, &ar), "lua_getstack() failed");
+	FATAL_ERROR_IF(!lua_getinfo(L, "Sl", &ar), "lua_getinfo() failed");
+
+	std::ostringstream os(std::ios::binary);
+	os << " (at " << ar.short_src << ":" << ar.currentline << ")";
+	return os.str();
+}
+
+void log_warning(lua_State *L, const std::string &message)
+{
+	warningstream << message << get_optional_origin(L, "") << std::endl;
+}
+
 void log_deprecated(lua_State *L, const std::string &message, int stack_depth)
 {
 	static thread_local bool configured = false;
@@ -175,4 +200,9 @@ void log_deprecated(lua_State *L, const std::string &message, int stack_depth)
 
 	if (do_log)
 		script_log(L, message, warningstream, do_error, stack_depth);
+}
+
+void log_error(lua_State *L, const std::string &message)
+{
+	errorstream << message << std::endl << script_get_backtrace(L) << std::endl;
 }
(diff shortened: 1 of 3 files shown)


src/script/lua_api/l_util.cpp
```diff
@@ -69,7 +69,16 @@ int ModApiUtil::l_log(lua_State *L)
 			level = LL_NONE;
 		}
 	}
-	g_logger.log(level, text);
+	if (level == LL_ERROR) {
+		log_error(L, text);
+		return 0;
+	}
+	if (level == LL_WARNING) {
```
nerzhul: switch case + return 0 outside of the conditions

src/script/common/c_internal.cpp
```diff
@@ -132,34 +132,51 @@ void script_run_callbacks_f(lua_State *L, int nargs,
...
+	// L can be NULL if we get called from scripting_game.cpp
+	if (!L)
+		return "";
+
+	if (modname != "")
+		return "[" + modname + "] ";
+
+	lua_Debug ar;
+
+	if (!lua_getstack(L, 2, &ar))
```
SmallJoker: Argument 2 is specifically used for the builtin functions, which throw warnings about function calls being deprecated. With argument 1 only, it would log the builtin function instead of the actual calling mod line.
Please provide either a full backtrace (which I should've done when touching this recently) or try to locate the first best backtrace line which actually corresponds to a mod or game source line.

SmallJoker: `minetest.log` should not need any manual adjustments to work correctly. Providing stack 1 and 2 (if available) backtrace ought to be enough for the most common cases. This includes warnings from builtin and those which are called directly in mods.

HybridDog: 1 seems to give the path for mods and builtin and 
2 gives the same path for builtin and is absent for mods.

HybridDog:
Printing the origin of the warning message is not helpful when a mod implements a `toint(num)` function which gives a warning if num is nan, another mod uses the toint function in another function and a third mod actually gets the number from player input and uses it to call the function of the second mod.
In this case a full backtrace can be used to find where nan comes from, 
printing only the place where the message is emitted does not show the problem's origin but it's a lot more helpful than printing no path and
printing the last three places can be used as compromise between usefulness and spam.

nerzhul:
i don't know why it doesn't have been finished and merged, but it's very useful feature, can you rebase it ?

HybridDog:
I've extracted the first to, unrelated changes: #9697

nerzhul:
thanks, you can rebase now :)

MoNTE48:
@nerzhul, it rebased a few months ago. Will you do a review?

appgurueu:
> Is there a way to remove the log call from the stack traceback?

Yes, but I don't think it should be removed; if it were, you'd have to grep the code for the log message to find the code path it originated from.

Zughy:
@HybridDog the PR is now "Supported by core dev". I understand the frustration, but I must ask: are you going to rebase this or should I close it (as the rebase request is from 13 March 2021)?

PR #7387, opened 2018-05-29
Update embedded Lua to 5.1.5

sfan5:
We were still using a version from 2008, Lua 5.1.5 is from 17 Feb 2012.


nerzhul:
wtf i thought i updated it at a point

nerzhul:
@SmallJoker then i missed to push the PR heh :) merging

PR #7394, opened 2018-06-01
Modernize lua read (part 1): C++ templating insurance

nerzhul:
Implement the float reader

Note: this has no performance penalty in release mode as functions are trivially inlined.

Idea is to modernize and securise our API using a C++ templating function to "proxy" the calls to the raw lua C API by implementing readers as high level C++ functions

### DIFF
--- a/src/script/lua_api/l_base.cpp
+++ b/src/script/lua_api/l_base.cpp
@@ -95,7 +95,7 @@ bool ModApiBase::isNaN(lua_State *L, int idx)
  * Read template functions
  */
 template<>
-float ModApiBase::read(lua_State *L, int index)
+float ModApiBase::readParam(lua_State *L, int index)
 {
 	if (isNaN(L, index))
 		throw LuaError("NaN value is not allowed.");
--- a/src/script/lua_api/l_base.cpp
+++ b/src/script/lua_api/l_base.cpp
@@ -90,3 +90,15 @@ bool ModApiBase::isNaN(lua_State *L, int idx)
 {
 	return lua_type(L, idx) == LUA_TNUMBER && std::isnan(lua_tonumber(L, idx));
 }
+
+/*
+ * Read template functions
+ */
+template<>
+float ModApiBase::read(lua_State *L, int index)
+{
+	if (isNaN(L, index))
+		throw LuaError("NaN value is not allowed.");
+
+	return (float) luaL_checknumber(L, index);
+}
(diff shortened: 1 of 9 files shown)


src/script/lua_api/l_base.cpp
```diff
@@ -90,3 +90,15 @@ bool ModApiBase::isNaN(lua_State *L, int idx)
...
+/*
+ * Read template functions
+ */
+template<>
+float ModApiBase::read(lua_State *L, int index)
+{
+	if (isNaN(L, index))
+		throw LuaError("NaN value is not allowed.");
+
+	return (float) luaL_checknumber(L, index);
```
SmallJoker: `read<T>` implies that multiple types can be read, like bool and integer. Yet it only supports floats, which lose are not accurate enough for 32-bit integers or higher.

src/script/lua_api/l_noise.cpp
```diff
@@ -437,7 +437,7 @@ int LuaPseudoRandom::create_object(lua_State *L)
 {
 	NO_MAP_LOCK_REQUIRED;
 
-	u64 seed = luaL_checknumber(L, 1);
+	u64 seed = read<float>(L, 1);
```
SmallJoker: Precision loss due to float conversion in this function.

src/script/lua_api/l_noise.cpp
```diff
@@ -534,7 +534,7 @@ int LuaPcgRandom::create_object(lua_State *L)
...
```
SmallJoker: Ditto

src/script/lua_api/l_base.cpp
nerzhul: using template function permits to prevent doing many luaReadXXX things and auto generate the function,s it's same, but it's  cleaner c+++

src/script/lua_api/l_noise.cpp
nerzhul: oops i missed that, i need to revert this change

src/script/lua_api/l_noise.cpp
nerzhul: same here

src/script/lua_api/l_base.cpp
nerzhul: currently yes we only have read<float> but i will introduce read<int>, read<u8>, read<double> and more :) this will permit to securize our Lua API calls a little bit more with a common high level interface.
also i will migrate some common tools like read<v3f> :)

nerzhul:
@SmallJoker what is missing to make it merged and permitting to continue the work ?

nerzhul:
i will finish this PR today by renaming, if no problem

PR #7396, opened 2018-06-02
World config: Make depends easier to read

HybridDog:
* Do not show an empty list but a message that there're no depends instead
* Adjust the list height to avoid the scrollbar (reverted)
* Show one big list if there're either no optional or only optional depends

2017-02-11-121451_1920x1080_scrot
2018-07-14-203929_1920x1056_scrot
[old screenshot, smaller minimum height](https://user-images.githubusercontent.com/3192173/40878761-d53d8914-6695-11e8-98e9-293f639d6950.png)

### DIFF
--- a/builtin/mainmenu/dlg_config_world.lua
+++ b/builtin/mainmenu/dlg_config_world.lua
@@ -74,23 +74,14 @@ local function get_formspec(data)
 					hard_deps_str .. ";0]" ..
 					"label[0,6;" .. fgettext("No optional dependencies") .. "]"
 			else
-				-- set the heights according to the number of list elements
-				local ratio = #hard_deps / (#soft_deps + #hard_deps)
-				-- avoid making one of the lists too short
-				local min_l = math.min(0.5,
-					math.min(#soft_deps, #hard_deps) * 0.25)
-				ratio = math.max(min_l, math.min(1.0 - min_l, ratio))
-				local h1 = 4.0 * ratio
-				local h2 = 4.0 - h1
-				local s_opt = 1.25 + h1 + 0.5
 				retval = retval ..
 					"label[0,1.25;" .. fgettext("Dependencies:") .. "]" ..
-					"textlist[0,1.75;5," .. h1 .. ";world_config_depends;" ..
+					"textlist[0,1.75;5,2.125;world_config_depends;" ..
 					hard_deps_str .. ";0]" ..
-					"label[0," .. s_opt .. ";" ..
-					fgettext("Optional dependencies:") .. "]" ..
-					"textlist[0," .. s_opt + 0.5 .. ";5," .. h2 ..
-					";world_config_optdepends;" .. soft_deps_str .. ";0]"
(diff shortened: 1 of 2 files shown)


SmallJoker:
Wuzzy loves big text, so scaling the dependency lists by a factor which is adjusted to the default scale settings won't look that pretty for other setting values.

HybridDog:
Shall I use the font size setting to calculate the minimum list size?

SmallJoker:
@HybridDog I'd prefer to limit it to two times 1/2 the size or full list size plus "No (optional) dependencies".

HybridDog:
> I'd prefer to limit it to two times 1/2 the size or full list size

I've changed it to 0.25, this is 1/2 the size of the height set when both list have the same one, shown in the new second screenshot.
I don't think text size is a problem now because of the gui_scaling setting.

SmallJoker:
I did some before/after tests, which showed me that much more time is needed to read the relevant information when scrolling through the list (key down). That's especially important when only parts of a modpack should be enabled, so that the required dependencies for each mod can be read easily.
The information used to be on the same location, whereas it's now moved up and down. However, not showing any information for modpacks [is a welcome tweak](http://irc.minetest.net/minetest-dev/2018-07-17#i_5360310).

HybridDog:
I didn't consider selecting one mod after another by continuously pressing the down key.
The start of the hard dependencies list always begins in the same position, and the hard dependencies is the relevant information, isn't it?

HybridDog:
rebased

SmallJoker:
> Note that in my opinion you are not supposed to manually search and enable all dependencies, 

That would be helpful, but for my part it's nice to have a consistent left side formatting while scrolling through the list like this:
1) Type first few characters to search
2) Navigate up/down
3) Enable mod or scroll through modpack
4) If MP: Check dependencies to decide what to enable

Changing positions and labels distract a lot there (changing things/movements always distract). It is true that the optional dependencies are less important but you still might want a recommendation for more mods to enable (which would then be compatible). The dependency list is usually large enough (5 entries) to see all at once.
Other people might care less about this, but for me this does not really improve anything.

HybridDog:
I've reverted it, a big list when there are only optional or only non-optional dependencies does not distract a lot, does it?

builtin/mainmenu/dlg_config_world.lua
```diff
@@ -42,15 +40,46 @@ local function get_formspec(data)
...
+				retval = retval ..
+					"label[0,1.25;" ..
+					fgettext("No (optional) dependencies") .. "]"
+			else
+				retval = retval ..
+					"label[0,1.25;" .. fgettext("No hard dependencies") ..
+					"]" ..
+					"label[0,1.75;" .. fgettext("Optional dependencies:") ..
+					"]" ..
+					"textlist[0,2.1;5,4;world_config_optdepends;" ..
```
SmallJoker: Label and textlist need more spacing, otherwise it looks OK now.
grafik

HybridDog:
Rebased.

PR #7410, opened 2018-06-04
Modernize lua read (part 2 & 3): C++ templating assurance

nerzhul:
Implement the boolean reader
Also remove unused & unimplemented script_error_handler
move readers to a new LuaHelper class permitting to share code between Lua interfaces

### DIFF
--- a/src/script/common/helper.cpp
+++ b/src/script/common/helper.cpp
@@ -32,7 +32,7 @@ bool LuaHelper::isNaN(lua_State *L, int idx)
  */
 template <> bool LuaHelper::readParam(lua_State *L, int index)
 {
-	return luaL_checkboolean(L, index);
+	return lua_toboolean(L, index) != 0;
 }
 
 template <> bool LuaHelper::readParam(lua_State *L, int index, const bool &dv)
--- a/src/script/common/helper.cpp
+++ b/src/script/common/helper.cpp
@@ -1,7 +1,6 @@
 /*
 Minetest
-Copyright (C) 2013 celeron55, Perttu Ahola <celeron55@gmail.com>
-Copyright (C) 2017 nerzhul, Loic Blot <loic.blot@unix-experience.fr>
+Copyright (C) 2018 nerzhul, Loic Blot <loic.blot@unix-experience.fr>
 
 This program is free software; you can redistribute it and/or modify
 it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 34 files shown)


src/script/common/helper.cpp
```diff
@@ -0,0 +1,45 @@
...
+	return lua_type(L, idx) == LUA_TNUMBER && std::isnan(lua_tonumber(L, idx));
+}
+
+/*
+ * Read template functions
+ */
+template<>
+bool LuaHelper::readParam(lua_State *L, int index)
+{
+	return lua_toboolean(L, index) != 0;
```
nerzhul: a question: should i add a throw LuaError with not a bool here to use strong bool or not ? we are lazy on checks

nerzhul:
Note, the parts are not to be squashed. I tested it with mtg without having problems, but the strict checking on bool can either fix some random bugs (undefined default bool in some calls) or permit mods to fix some implementation bugs

nerzhul:
Note to @SmallJoker : LuaError is catched by serverstep or client step and shown with a backtrace properly :)

src/script/common/helper.cpp
```diff
@@ -0,0 +1,71 @@
...
+	if (isNaN(L, index))
+		throw LuaError("NaN value is not allowed.");
+
+	return (float)luaL_checknumber(L, index);
+}
+
+template <> std::string LuaHelper::readParam(lua_State *L, int index)
+{
+	if (lua_isnil(L, index))
+		return "";
```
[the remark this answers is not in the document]
SmallJoker: That's indeed bad. I would however prefer to fix those issues instead of silently accepting it.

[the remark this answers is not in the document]
nerzhul: i remove it, and we are free to fix the engine crashes after :)

paramat:
What's the advantage of this, is lua parameter reading faster?

nerzhul:
@paramat it's exactly same performance (in release mode, negligible in debug mode). The advantage if we implement the template on all types is that permit to access to lua reader/writer in a uniform, standardized and securized way. One high level function per type to perform the call with all security/sanitize it needs. it also permits to implement easily a complex reader in the same API (imagine writeParam<Json::Value>(L, json_val), for example)

paramat:
Ok thanks, no objection from me. Unfortunately i'm not qualified to officially +1 it.

nerzhul:
@SmallJoker if you are okay, i can merge this. It seems other qualified devs are not concerned about this

nerzhul:
Approved by @SmallJoker on IRC

src/script/common/helper.cpp
```diff
@@ -0,0 +1,72 @@
...
+{
+	return lua_type(L, idx) == LUA_TNUMBER && std::isnan(lua_tonumber(L, idx));
+}
+
+/*
+ * Read template functions
+ */
+template <> bool LuaHelper::readParam(lua_State *L, int index)
+{
+	return lua_toboolean(L, index) != 0;
```
sfan5: should have a comment why `luaL_checkboolean` is not used here

src/script/common/helper.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+template <> bool LuaHelper::readParam(lua_State *L, int index, const bool &dv)
```
sfan5: clarity: dv -> default

nerzhul: default is a protected keyword in C++11 standard

src/script/common/helper.cpp
nerzhul: this function doesn't exists in Lua & has been dropped from the header, as it seems useless currently (i added this in a previous version of this PR)

src/script/common/helper.cpp
nerzhul: also dv is described in the header comments

SmallJoker: Well then `default_` and `fallback_value` would still be an option

PR #7415, opened 2018-06-05
Set ENABLE_GETTEXT to TRUE by default

Wuzzy2:
This sets the CMake option `ENABLE_GETTEXT` to `TRUE` by default.
In other words, Minetest will now build with translations enabled by default.

Gettext is currently disabled by default. Strange this wasn't noticed earlier.


nerzhul:
approved by @sfan5 on IRC

PR #7426, opened 2018-06-09
Fix narrow/utf8 difference in incoming/outcoming messages

### LINKED ISSUE #7425: Allow multibyte string.
##### Issue type
 - Feature request
 
##### Minetest version
0.4.17

##### Summary
Why does words like "äöü" just send invalid multibyte string to the chat?
Why is this disabled?
Any Reason?
It would be nice to fix/enable this like it was in 0.4.16.

##### Steps to reproduce
Write on a 0.4.17 server a "ä"

numberZero:
Fixes #7425.
Current code uses UTF-8 already in other places, including `on_chat_message` callbacks (`src/server.cpp:2921`).
@Lejo1

### DIFF
--- a/src/script/lua_api/l_server.cpp
+++ b/src/script/lua_api/l_server.cpp
@@ -74,7 +74,7 @@ int ModApiServer::l_chat_send_all(lua_State *L)
 	// Get server from registry
 	Server *server = getServer(L);
 	// Send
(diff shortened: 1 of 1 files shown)


Lejo1:
Will test it!

sfan5:
2c0d8b25dd9a2e167bcf884c07ac4369f884091e

PR #7437, opened 2018-06-11
Server: move shutdown parts to a specific shutdown state object

nerzhul:
A simple PR moving shutdown parts outside of server object to a child objet permitting to add unittests.
When the PR is review & ready i will add the required unittests

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -139,7 +139,60 @@ v3f ServerSoundParams::getPos(ServerEnvironment *env, bool *pos_exists) const
 	return v3f(0,0,0);
 }
 
+void Server::ShutdownState::reset()
+{
+	m_timer = 0.0f;
+	message.clear();
+	should_reconnect = false;
+	is_requested = false;
+}
 
+void Server::ShutdownState::trigger(float delay, const std::string &msg, bool reconnect)
+{
+	m_timer = delay;
+	message = msg;
+	should_reconnect = reconnect;
+}
+
+void Server::ShutdownState::tick(float dtime, Server *server)
+{
+	// Timed shutdown
+	static const float shutdown_msg_times[] =
+	{
+		1, 2, 3, 4, 5, 10, 15, 20, 25, 30, 45, 60, 120, 180, 300, 600, 1200, 1800, 3600
+	};
+
+	if (m_timer > 0.0f) {
+		// Automated messages
+		if (m_timer < shutdown_msg_times[ARRLEN(shutdown_msg_times) - 1]) {
+			for (u16 i = 0; i < ARRLEN(shutdown_msg_times) - 1; i++) {
+				// If shutdown timer matches an automessage, shot it
+				if (m_timer > shutdown_msg_times[i] &&
+					m_timer - dtime < shutdown_msg_times[i]) {
+					std::wstring periodicMsg = getShutdownTimerMessage();
+
+					infostream << wide_to_utf8(periodicMsg).c_str() << std::endl;
+					server->SendChatMessage(PEER_ID_INEXISTENT, periodicMsg);
+					break;
+				}
+			}
+		}
+
+		m_timer -= dtime;
+		if (m_timer < 0.0f) {
+			m_timer = 0.0f;
+			is_requested = true;
+		}
+	}
+}
+
+std::wstring Server::ShutdownState::getShutdownTimerMessage() const
+{
+	std::wstringstream ws;
+	ws << L"*** Server shutting down in "
+		<< duration_to_string(myround(m_timer)).c_str() << ".";
+	return ws.str();
+}
 
 /*
 	Server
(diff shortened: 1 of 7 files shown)


src/server.cpp
```diff
@@ -139,7 +139,60 @@ v3f ServerSoundParams::getPos(ServerEnvironment *env, bool *pos_exists) const
...
+	message = msg;
+	should_reconnect = reconnect;
+}
 
+void Server::ShutdownState::tick(float dtime, Server *server)
+{
+	// Timed shutdown
+	static const float shutdown_msg_times[] =
+	{
+		1, 2, 3, 4, 5, 10, 15, 20, 25, 30, 45, 60, 120, 180, 300, 600, 1200, 1800, 3600
```
SmallJoker: Perhaps we should re-consider the notification times? 1 second can almost not be read, unless there are network delays - and `?5` seem way to detailed. If we continue the scheme `/2` `*2/3` scheme from the end of this list, we would get an array like this: (discontinued after `20`, where it would be an unhandy `13.333` seconds)
`1, 2, 3, 4, 5, 10, 20, 40, 60, 120, 180, 300, 600, 1200, 1800, 3600`

src/server.cpp
```diff
@@ -139,7 +139,60 @@ v3f ServerSoundParams::getPos(ServerEnvironment *env, bool *pos_exists) const
...
+	};
+
+	if (m_timer > 0.0f) {
```
SmallJoker: Statement inversion and return is possible here to reduce the amount of indents. This function only handles the shutdown process and only depends on this timer.

src/server.cpp
```diff
@@ -139,7 +139,60 @@ v3f ServerSoundParams::getPos(ServerEnvironment *env, bool *pos_exists) const
...
+		// Automated messages
+		if (m_timer < shutdown_msg_times[ARRLEN(shutdown_msg_times) - 1]) {
```
SmallJoker: Do we need this check? The loop below will be executed and does nothing.

src/server.cpp
```diff
@@ -139,7 +139,60 @@ v3f ServerSoundParams::getPos(ServerEnvironment *env, bool *pos_exists) const
...
+			for (u16 i = 0; i < ARRLEN(shutdown_msg_times) - 1; i++) {
+				// If shutdown timer matches an automessage, shot it
+				if (m_timer > shutdown_msg_times[i] &&
+					m_timer - dtime < shutdown_msg_times[i]) {
```
SmallJoker: Missing indent.

src/server.cpp
```diff
@@ -139,7 +139,60 @@ v3f ServerSoundParams::getPos(ServerEnvironment *env, bool *pos_exists) const
...
+	// Timed shutdown
+	static const float shutdown_msg_times[] =
+	{
+		1, 2, 3, 4, 5, 10, 15, 20, 25, 30, 45, 60, 120, 180, 300, 600, 1200, 1800, 3600
+	};
+
+	if (m_timer > 0.0f) {
+		// Automated messages
+		if (m_timer < shutdown_msg_times[ARRLEN(shutdown_msg_times) - 1]) {
+			for (u16 i = 0; i < ARRLEN(shutdown_msg_times) - 1; i++) {
```
SmallJoker: Range-based for loop?

src/server.cpp
nerzhul: yes the loop is tiny, but it permits to not execute it :)

src/server.cpp
nerzhul: yes, using std::array can permit that, trying that after :)

src/server.cpp
nerzhul: 1 can be read :) i was based on WoW experience for those messages

src/server.cpp
nerzhul: remember it's ~15 loop iteration and tests when shutdown timer is enabled, it's useless to perform it in that case if no timer is not in the interval

src/unittest/test_server_shutdown_state.cpp
```diff
@@ -0,0 +1,117 @@
...
+51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+#include <server.h>
+#include "test.h"
+
+#include "util/string.h"
+#include "util/serialize.h"
+
+class FakeServer: public Server
```
SmallJoker: `FakeServer : public`

src/unittest/test_server_shutdown_state.cpp
```diff
@@ -0,0 +1,117 @@
...
+{
+public:
+	FakeServer(): Server("fakeworld", SubgameSpec("fakespec", "fakespec"), true,
```
SmallJoker: `FakeServer() : Server`

src/server.cpp
```diff
@@ -174,22 +227,96 @@ Server::Server(
...
+	infostream << "Server: Deinitializing scripting" << std::endl;
+	delete m_script;
+
+	// Delete detached inventories
+	for (auto &detached_inventory : m_detached_inventories) {
+		delete detached_inventory.second;
+	}
+}
+
+void Server::init()
```
SmallJoker: What's the advantage in moving this outside the constructor?

src/unittest/test_server_shutdown_state.cpp
```diff
@@ -0,0 +1,117 @@
+/*
+Minetest
+Copyright (C) 2013 celeron55, Perttu Ahola <celeron55@gmail.com>
```
SmallJoker: Incorrect copyright notice

src/server.cpp
nerzhul: If you see the code inside this function, you will see if we want to test any server we must instatiate a real server. Moving this outside permit to have the object without instantiating the whole world.
I know it's not perfect, but it's a first step to think about refactoring the server in multiple parts to permit adding tests and evolve it

PR #7473, opened 2018-06-22
Fix crash caused by Lua error during startup

red-001:
Previously the code would crash due to out of bounds access.


SmallJoker:
2nd approval [by sfan5](http://irc.minetest.net/minetest-dev/2018-06-22#i_5340349).

PR #7477, opened 2018-06-23
Android: Add aux ("use") button

### LINKED ISSUE #2873: Android users can't sprint (missing “Use” key)
Simply spoken: **Android users are pretty much unable to sprint!**

Multiple users have reported to me that there is no way to use the “Use” button (also known as “Aux1”) on Android.
So I request to add support for the “use” button on Android.

Note that some mods need the use button to work, for example, the Sprint mod by GunshipPenguin, so being able to use this button is needed.

srifqi:
Add aux ("use") button.
(Optional) Use joystick to tap aux ("use") button when out of main circle by enabling `virtual_joystick_triggers_aux` setting.
Fixes #2873

---

New button's position

Buttons' touch target


srifqi:
> The 'use' key is a multi-purpose key, so i feel should not be labelled 'use'.

I agree.

> Perhaps 'AUX' is understood internationally as being 'auxilary'?

[It is.](https://en.wiktionary.org/wiki/aux)

paramat:
I can't find a suitable icon for 'auxilary' so i feel a white 'AUX' is perhaps best. Also, no need for a surrounding square as that obscures the world more.

paramat:
Please can the AUX and zoom buttons both be on the edge of the screen with the zoom button above AUX (because it's used less)?

paramat:
This PR is good because if we, as suggested, do add another key binding to separate 'fast' from 'use' then those can be bound to joystick super-forward and the use key of this PR, instead of having this option setting.

src/defaultsettings.cpp
```diff
@@ -410,6 +410,7 @@ void set_default_settings(Settings *settings)
 	settings->setDefault("TMPFolder","/sdcard/" PROJECT_NAME_C "/tmp/");
 	settings->setDefault("touchscreen_threshold","20");
 	settings->setDefault("fixed_virtual_joystick", "false");
+	settings->setDefault("virtual_joystick_controls_use", "false");
```
paramat: Maybe 'virtual_joystick_triggers_aux'? 'controls' is a little awkward sounding and has a double meaning.

src/gui/touchscreengui.cpp
```diff
@@ -42,7 +42,8 @@ using namespace irr::core;
 const char **touchgui_button_imagenames = (const char *[]) {
 	"jump_btn.png",
 	"down.png",
-	"zoom.png"
+	"zoom.png",
+	"use_btn.png"
```
paramat: "aux_btn.png"
?

srifqi: Yes, I agree.

src/defaultsettings.cpp
srifqi: Okay.

srifqi:
I've done changing the aux button texture and repositioning the buttons.
New button layout

The behavior is still same.

paramat:
Please can anyone independently test this? Then i'll +1.
@stujones11 @Wayward1

stujones11:
Seems to work, although there is nothing in the default game I can test it with, afaik

screen

paramat:
Thanks, the zoom and AUX buttons look quite close to each other but i can't test so i'll let you optimise that. I suggest equal spacing of zoom and AUX buttons between menu and jump buttons.

paramat:
So it looks like AUX should be slightly lower?

srifqi:
@paramat Done. I've decided to move down the aux button by a little.

paramat:
:+1:

paramat:
Please can we merge this? Code is simple and it's well tested. Then it will get wider testing.

nerzhul:
okay for me

codexp:
will the menu button stay not in line with others?

srifqi:
Which button is that?

PR #7481, opened 2018-06-24
Fix mod channels crash

### LINKED ISSUE #7479: Dangling mod channel pointer.
##### Issue type
 - Bug report
 
##### Minetest version
```
5.0.0-dev
```

##### OS / Hardware
N/A

##### Summary
The Lua interface for mod channels stores a pointer to the channel, if the channel fails to open correctly (network issues, etc) the packet handler closes the channel resulting in it being deleted and the dangling pointer.

##### Steps to reproduce
Hard to reproduce under normal conditions, `-fsanitize=address` and `-fsanitize=undefined` compiler flags make it easier to do.

red-001:
Closes #7479.

### DIFF
--- a/src/script/lua_api/l_modchannels.cpp
+++ b/src/script/lua_api/l_modchannels.cpp
@@ -33,9 +33,8 @@ int ModApiChannels::l_mod_channel_join(lua_State *L)
 		return 0;
 
 	getGameDef(L)->joinModChannel(channel);
-	ModChannel *channelObj = getGameDef(L)->getModChannel(channel);
-	assert(channelObj);
-	ModChannelRef::create(L, channelObj);
+	assert(getGameDef(L)->getModChannel(channel) != nullptr);
+	ModChannelRef::create(L, channel);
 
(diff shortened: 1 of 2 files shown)


src/script/lua_api/l_modchannels.cpp
```diff
@@ -33,9 +33,8 @@ int ModApiChannels::l_mod_channel_join(lua_State *L)
 		return 0;
 
 	getGameDef(L)->joinModChannel(channel);
-	ModChannel *channelObj = getGameDef(L)->getModChannel(channel);
-	assert(channelObj);
-	ModChannelRef::create(L, channelObj);
+	assert(getGameDef(L)->isModChannelRegistered(channel));
```
SmallJoker: `assert(getGameDef(L)->getModChannel(channel));` would be possible here too because `isModChannelRegistered` is used nowhere else.

red-001: yeah it would be possible, just seemed cleaner at the moment to do it this way.

Issue #7122, opened 2018-03-11
Android: confirmation dialog is... unreable and crashes


nerzhul:
##### Issue type

 - Bug report

 
##### Minetest version

```
0.5.0 git
```

##### OS / Hardware

Operating system: Android 6.0

##### Summary

Here is the problem:

![37256354-bb4c1d36-2559-11e8-80af-a8b7b50aeb03](https://user-images.githubusercontent.com/3686677/45852276-d7ba6200-bd36-11e8-92c0-6a78b886b5ca.png)

Also when clicking on the dialog, MT crashes

##### Steps to reproduce

Build last master, connect to a 0.5 server and see MT crash

jastevenson303:
If you just change the ok button to read "OK" then it should be okay.

nerzhul:
@jastevenson303 the problem is not only the button but also the text... all should be upscaled

srifqi:
What is the screen resolution and GUI scaling?

srifqi:
Using Android 6.0.1, using APK debug build for my PR. Trying to connect to new server, it crashes (SIGSEGV) before loading dialog.
```
03-15 20:49:34.681 19970 19983 D Minetest: 2018-03-15 20:49:34: INFO[Main]: SCRIPTAPI: Initialized client game modules
03-15 20:49:34.681 19970 19983 D Minetest: 2018-03-15 20:49:34: INFO[Main]: Connecting to server at 23.28.87.79:30000
03-15 20:49:34.681 19970 19983 V Minetest: 2018-03-15 20:49:34: VERBOSE[Main]: Loading and running script *builtin*:init.lua
03-15 20:49:34.701 19970 19983 D Minetest: 2018-03-15 20:49:34: INFO[Main]: Client::peerAdded(): peer->id=1
03-15 20:49:34.711 19970 19983 D Minetest: 2018-03-15 20:49:34: INFO[Main]: Client packetcounter (20):
03-15 20:49:35.091 19970 20082 V Minetest: 2018-03-15 20:49:35: VERBOSE[ConnectionSend]: con(43/2440)RE-SENDING timed-out RELIABLE to 23.28.87.79(t/o=0.1): from_peer_id=2440, channel=0, seqnum=65501
[...]

PR #7483, opened 2018-06-24
Move touchscreen input handling to base GUIModalMenu class

stujones11:
Fixes: #7122 & #6432

This PR moves the touchscreen event translation and native input handling methods out of `GUIFormSpecMenu` into the base class `GUIModalMenu` so that all other derived modal menus can work with android devices without adding a great deal of duplicated code.

Note that the changes to `game.cpp` were necessary to avoid a crash caused by the global `g_touchscreengui` object being accessed before initialization but I can’t seem to figure out where from. However, since this is only for android then I can’t see it being a problem.

This PR involved a lot of cut & paste of existing code so there may be some code style issues I did not personally introduce. I have fixed a few obvious ones already and will be happy to fix any others that are reported.

What the PR does not do is fix the button scaling problem reported in #7122 that is a different issue altogether. It is difficult for me to debug this as I am unable to reproduce it on any of my own devices.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -1288,7 +1288,12 @@ bool Game::createClient(const std::string &playername,
 		return false;
 
 	bool could_connect, connect_aborted;
-
+#ifdef HAVE_TOUCHSCREENGUI
+	if (g_touchscreengui) {
+		g_touchscreengui->init(texture_src);
+		g_touchscreengui->hide();
+	}
+#endif
 	if (!connectToServer(playername, password, address, port,
 			&could_connect, &connect_aborted))
 		return false;
@@ -1414,7 +1419,7 @@ bool Game::initGui()
 #ifdef HAVE_TOUCHSCREENGUI
 
 	if (g_touchscreengui)
-		g_touchscreengui->init(texture_src);
+		g_touchscreengui->show();
 
 #endif
 
--- a/src/gui/guiConfirmRegistration.cpp
+++ b/src/gui/guiConfirmRegistration.cpp
@@ -157,6 +157,9 @@ void GUIConfirmRegistration::drawMenu()
 	driver->draw2DRectangle(bgcolor, AbsoluteRect, &AbsoluteClippingRect);
 
 	gui::IGUIElement::draw();
+#ifdef __ANDROID__
+	getAndroidUIInput();
+#endif
 }
 
 void GUIConfirmRegistration::closeMenu(bool goNext)
@@ -193,10 +196,14 @@ bool GUIConfirmRegistration::processInput()
 bool GUIConfirmRegistration::OnEvent(const SEvent &event)
 {
 	if (event.EventType == EET_KEY_INPUT_EVENT) {
-		if (event.KeyInput.Key == KEY_ESCAPE && event.KeyInput.PressedDown) {
+		// clang-format off
+		if ((event.KeyInput.Key == KEY_RETURN ||
+				event.KeyInput.Key == KEY_CANCEL) &&
+				event.KeyInput.PressedDown) {
 			closeMenu(false);
 			return true;
 		}
+		// clang-format on
 		if (event.KeyInput.Key == KEY_RETURN && event.KeyInput.PressedDown) {
 			acceptInput();
 			if (processInput())
@@ -239,3 +246,19 @@ bool GUIConfirmRegistration::OnEvent(const SEvent &event)
 
 	return false;
 }
+
+#ifdef __ANDROID__
+bool GUIConfirmRegistration::getAndroidUIInput()
+{
+	if (!hasAndroidUIInput() || m_jni_field_name != "password")
+		return false;
+
+	std::string text = porting::getInputDialogValue();
+	gui::IGUIElement *e = getElementFromId(ID_confirmPassword);
+	if (e)
+		e->setText(utf8_to_wide(text).c_str());
+
+	m_jni_field_name.clear();
+	return false;
+}
+#endif
--- a/src/gui/guiConfirmRegistration.cpp
+++ b/src/gui/guiConfirmRegistration.cpp
@@ -197,7 +197,7 @@ bool GUIConfirmRegistration::OnEvent(const SEvent &event)
 {
 	if (event.EventType == EET_KEY_INPUT_EVENT) {
 		// clang-format off
-		if ((event.KeyInput.Key == KEY_RETURN ||
+		if ((event.KeyInput.Key == KEY_ESCAPE ||
 				event.KeyInput.Key == KEY_CANCEL) &&
 				event.KeyInput.PressedDown) {
 			closeMenu(false);
--- a/src/gui/guiConfirmRegistration.cpp
+++ b/src/gui/guiConfirmRegistration.cpp
@@ -43,7 +43,9 @@ GUIConfirmRegistration::GUIConfirmRegistration(gui::IGUIEnvironment *env,
 		m_client(client), m_playername(playername), m_password(password),
 		m_address(address), m_aborted(aborted)
 {
+#ifdef __ANDROID__
 	m_touchscreen_visible = false;
+#endif
 }
 
 GUIConfirmRegistration::~GUIConfirmRegistration()
--- a/src/gui/guiConfirmRegistration.cpp
+++ b/src/gui/guiConfirmRegistration.cpp
@@ -43,6 +43,7 @@ GUIConfirmRegistration::GUIConfirmRegistration(gui::IGUIEnvironment *env,
 		m_client(client), m_playername(playername), m_password(password),
 		m_address(address), m_aborted(aborted)
 {
+	m_touchscreen_visible = false;
 }
 
 GUIConfirmRegistration::~GUIConfirmRegistration()
--- a/src/gui/guiPasswordChange.cpp
+++ b/src/gui/guiPasswordChange.cpp
@@ -23,7 +23,7 @@ OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 #include <IGUIButton.h>
 #include <IGUIStaticText.h>
 #include <IGUIFont.h>
-#include "keycode.h"
+
 #include "porting.h"
 #include "gettext.h"
 
--- a/src/gui/guiPasswordChange.cpp
+++ b/src/gui/guiPasswordChange.cpp
@@ -23,7 +23,8 @@ OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 #include <IGUIButton.h>
 #include <IGUIStaticText.h>
 #include <IGUIFont.h>
-
+#include "keycode.h"
+#include "porting.h"
 #include "gettext.h"
 
 const int ID_oldPassword = 256;
@@ -180,6 +181,9 @@ void GUIPasswordChange::drawMenu()
 	driver->draw2DRectangle(bgcolor, AbsoluteRect, &AbsoluteClippingRect);
 
 	gui::IGUIElement::draw();
+#ifdef __ANDROID__
+	getAndroidUIInput();
+#endif
 }
 
 void GUIPasswordChange::acceptInput()
@@ -211,10 +215,14 @@ bool GUIPasswordChange::processInput()
 bool GUIPasswordChange::OnEvent(const SEvent &event)
 {
 	if (event.EventType == EET_KEY_INPUT_EVENT) {
-		if (event.KeyInput.Key == KEY_ESCAPE && event.KeyInput.PressedDown) {
+		// clang-format off
+		if ((event.KeyInput.Key == KEY_RETURN ||
+				event.KeyInput.Key == KEY_CANCEL) &&
+				event.KeyInput.PressedDown) {
 			quitMenu();
 			return true;
 		}
+		// clang-format on
 		if (event.KeyInput.Key == KEY_RETURN && event.KeyInput.PressedDown) {
 			acceptInput();
 			if (processInput())
@@ -259,3 +267,39 @@ bool GUIPasswordChange::OnEvent(const SEvent &event)
 
 	return Parent ? Parent->OnEvent(event) : false;
 }
+
+std::string GUIPasswordChange::getNameByID(s32 id)
+{
+	switch (id) {
+	case ID_oldPassword:
+		return "old_password";
+	case ID_newPassword1:
+		return "new_password_1";
+	case ID_newPassword2:
+		return "new_password_2";
+	}
+	return "";
+}
+
+#ifdef __ANDROID__
+bool GUIPasswordChange::getAndroidUIInput()
+{
+	if (!hasAndroidUIInput())
+		return false;
+
+	gui::IGUIElement *e = nullptr;
+	if (m_jni_field_name == "old_password")
+		e = getElementFromId(ID_oldPassword);
+	else if (m_jni_field_name == "new_password_1")
+		e = getElementFromId(ID_newPassword1);
+	else if (m_jni_field_name == "new_password_2")
+		e = getElementFromId(ID_newPassword2);
+
+	if (e) {
+		std::string text = porting::getInputDialogValue();
+		e->setText(utf8_to_wide(text).c_str());
+	}
+	m_jni_field_name.clear();
+	return false;
+}
+#endif
--- a/src/gui/guiPasswordChange.cpp
+++ b/src/gui/guiPasswordChange.cpp
@@ -216,7 +216,7 @@ bool GUIPasswordChange::OnEvent(const SEvent &event)
 {
 	if (event.EventType == EET_KEY_INPUT_EVENT) {
 		// clang-format off
-		if ((event.KeyInput.Key == KEY_RETURN ||
+		if ((event.KeyInput.Key == KEY_ESCAPE ||
 				event.KeyInput.Key == KEY_CANCEL) &&
 				event.KeyInput.PressedDown) {
 			quitMenu();
--- a/src/gui/guiPasswordChange.h
+++ b/src/gui/guiPasswordChange.h
@@ -44,6 +44,13 @@ public:
 	bool processInput();
 
 	bool OnEvent(const SEvent &event);
+#ifdef __ANDROID__
+	bool getAndroidUIInput();
+#endif
+
+protected:
+	std::wstring getLabelByID(s32 id) { return L""; }
+	std::string getNameByID(s32 id);
 
 private:
 	Client *m_client;
--- a/src/gui/guiPathSelectMenu.h
+++ b/src/gui/guiPathSelectMenu.h
@@ -44,6 +44,10 @@ public:
 
 	void setTextDest(TextDest *dest) { m_text_dst = dest; }
 
+protected:
+	std::wstring getLabelByID(s32 id) { return L""; }
+	std::string getNameByID(s32 id) { return ""; }
+
 private:
 	void acceptInput();
 
--- a/src/gui/modalMenu.cpp
+++ b/src/gui/modalMenu.cpp
@@ -1,0 +1,283 @@
+/*
+Minetest
+Copyright (C) 2013 celeron55, Perttu Ahola <celeron55@gmail.com>
+Copyright (C) 2018 stujones11, Stuart Jones <stujones111@gmail.com>
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation; either version 2.1 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License along
+with this program; if not, write to the Free Software Foundation, Inc.,
+51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+#include <cstdlib>
+#include "modalMenu.h"
+#include "gettext.h"
+#include "porting.h"
+
+#ifdef HAVE_TOUCHSCREENGUI
+#include "touchscreengui.h"
+#endif
+
+// clang-format off
+GUIModalMenu::GUIModalMenu(gui::IGUIEnvironment* env, gui::IGUIElement* parent, s32 id,
+		IMenuManager *menumgr) :
+		IGUIElement(gui::EGUIET_ELEMENT, env, parent, id,
+				core::rect<s32>(0, 0, 100, 100)),
+#ifdef __ANDROID__
+		m_jni_field_name(""),
+#endif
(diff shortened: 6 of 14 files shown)


stujones11:
It seems that the linter does not like my new file :(

I guess I need to read up on the style guidelines so WIP for now but feel to test the functionality if you have an android build environment to hand ;)

paramat:
Don't worry about the linter, it has no idea of MT codestyle, once you have applied anything useful it reports you can whitelist your file by adding it to https://github.com/minetest/minetest/blob/master/util/travis/clang-format-whitelist.txt

paramat:
Essentially, if the linter reports something that our codestyle guidelines also requests, attend to it, otherwise ignore the linter. Core devs will look at the linter report and request the valid changes.

paramat:
I looked at the linter report, it's about 50% correct.
Where it asks for double-tab indents to split a long line, ignore that, unless it is in a conditional where it is needed to differentiate from the indented actions of the conditional.
At one place it asks for 3 double-indents in a row (L560), which just moves everything up against the right side causing more line splits to be needed. Very silly.
Where we have formatted similar lines with vertical alignment for readability (L611), it complains, ignore that.

nerzhul:
it doesn't block anything, it's a code refactor

src/gui/guiPathSelectMenu.h
```diff
@@ -44,6 +44,10 @@ class GUIFileSelectMenu : public GUIModalMenu
 
 	void setTextDest(TextDest *dest) { m_text_dst = dest; }
 
+protected:
+	std::wstring getLabelByID(s32 id) { return L""; }
```
nerzhul: as many values are by default identical, i think it's better to have them in parent class and override in child classes

src/gui/modalMenu.cpp
```diff
@@ -0,0 +1,285 @@
...
+	}
+#endif
+	return false;
+}
+
+#ifdef __ANDROID__
+bool GUIModalMenu::hasAndroidUIInput()
+{
+	/* no dialog shown */
+	if (m_JavaDialogFieldName == "") {
```
nerzhul: m_xxx.emtpy()

src/gui/modalMenu.cpp
```diff
@@ -0,0 +1,285 @@
...
+	}
+
+	/* still waiting */
+	if (porting::getInputDialogState() == -1) {
+		return true;
+	}
+
+	/* no value abort dialog processing */
+	if (porting::getInputDialogState() != 0) {
+		m_JavaDialogFieldName = "";
```
nerzhul: m_javaxxx.clear()

srifqi:
I tried on my phone (Samsung Galaxy J7, Android 6.0.1) and it still crashes when tapping confirmation dialog with this error below.
<details>
<summary>Crash dump</summary>

```
********** Crash dump: **********
Build fingerprint: 'samsung/j7eltexx/j7elte:6.0.1/MMB29K/J700FXXU4BQL1:user/release-keys'
pid: 3545, tid: 3576, name: MinetestNativeT  >>> net.minetest.minetest <<<
signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x18
Stack frame #00 pc 00581a5c  /data/app/net.minetest.minetest-1/lib/arm/libminetest.so (_ZN14TouchScreenGUI6ToggleEb+135): Routine std::__ndk1::vector<TouchScreenGUI::id_status, std::__ndk1::allocator<TouchScreenGUI::id_status> >::__annotate_shrink(unsigned int) const at /home/srifqi/android/android-ndk/sources/cxx-stl/llvm-libc++/include/vector:852
Stack frame #01 pc 007ea373  /data/app/net.minetest.minetest-1/lib/arm/libminetest.so (_ZN15MyEventReceiver7OnEventERKN3irr6SEventE+66): Routine Hud::drawSelectionMesh() at /home/srifqi/git/minetest/build/android/jni/../jni/src/client/hud.cpp:533
Stack frame #02 pc 009b9ba1  /data/app/net.minetest.minetest-1/lib/arm/libminetest.so (_ZN3irr14CIrrDeviceStub17postEventFromUserERKNS_6SEventE+28): Routine deflate_rle at /home/wayward_one/dev/stu/minetest/build/android/deps/irrlicht/source/Irrlicht/Android/jni/../../zlib/deflate.c:1892
Stack frame #03 pc 009b6ea7  /data/app/net.minetest.minetest-1/lib/arm/libminetest.so (_ZN3irr17CIrrDeviceAndroid11handleInputEP11android_appP11AInputEvent+530): Routine std::__ndk1::__libcpp_deallocate(void*) at /home/srifqi/android/android-ndk/sources/cxx-stl/llvm-libc++/include/new:234
Stack frame #04 pc 00b565bb  /data/app/net.minetest.minetest-1/lib/arm/libminetest.so: Routine loop at curl_fnmatch.c:?
Stack frame #05 pc 009b6fc1  /data/app/net.minetest.minetest-1/lib/arm/libminetest.so (_ZN3irr17CIrrDeviceAndroid3runEv+80): Routine std::__ndk1::__compressed_pair<Json::PathArgument*, std::__ndk1::allocator<Json::PathArgument> >::second() at /home/srifqi/android/android-ndk/sources/cxx-stl/llvm-libc++/include/memory:2184
Stack frame #06 pc 004e3cfd  /data/app/net.minetest.minetest-1/lib/arm/libminetest.so (_ZN15RenderingEngine3runEv+76): Routine Game::createClient(std::__ndk1::basic_string<char, std::__ndk1::char_traits<char>, std::__ndk1::allocator<char> > const&, std::__ndk1::basic_string<char, std::__ndk1::char_traits<char>, std::__ndk1::allocator<char> > const&, std::__ndk1::basic_string<char, std::__ndk1::char_traits<char>, std::__ndk1::allocator<char> >*, unsigned short) at /home/srifqi/git/minetest/build/android/jni/../jni/src/game.cpp:1398
Stack frame #07 pc 004eb4eb  /data/app/net.minetest.minetest-1/lib/arm/libminetest.so (_ZN4Game15connectToServerERKNSt6__ndk112basic_stringIcNS0_11char_traitsIcEENS0_9allocatorIcEEEES8_PS6_tPbSA_+2590): Routine std::__ndk1::basic_string<char, std::__ndk1::char_traits<char>, std::__ndk1::allocator<char> >::__move_assign_alloc(std::__ndk1::basic_string<char, std::__ndk1::char_traits<char>, std::__ndk1::allocator<char> >&, std::__ndk1::integral_constant<bool, true>) at /home/srifqi/android/android-ndk/sources/cxx-stl/llvm-libc++/include/string:1430
Stack frame #08 pc 004e299f  /data/app/net.minetest.minetest-1/lib/arm/libminetest.so (_ZN4Game12createClientERKNSt6__ndk112basic_stringIcNS0_11char_traitsIcEENS0_9allocatorIcEEEES8_PS6_t+158): Routine ~Game at /home/srifqi/git/minetest/build/android/jni/../jni/src/game.cpp:944
Stack frame #09 pc 004e261d  /data/app/net.minetest.minetest-1/lib/arm/libminetest.so (_ZN4Game7startupEPbbP12InputHandlerRKNSt6__ndk112basic_stringIcNS3_11char_traitsIcEENS3_9allocatorIcEEEESB_SB_PS9_tRS9_S0_P11ChatBackendRK11SubgameSpecb+508): Routine Game::readSettings() at /home/srifqi/git/minetest/build/android/jni/../jni/src/game.cpp:3962
Stack frame #10 pc 004f9ad9  /data/app/net.minetest.minetest-1/lib/arm/libminetest.so (_Z8the_gamePbbP12InputHandlerRKNSt6__ndk112basic_stringIcNS2_11char_traitsIcEENS2_9allocatorIcEEEESA_SA_SA_tRS8_R11ChatBackendS_RK11SubgameSpecb+228): Routine unordered_map at /home/srifqi/android/android-ndk/sources/cxx-stl/llvm-libc++/include/unordered_map:1260
Stack frame #11 pc 007da62b  /data/app/net.minetest.minetest-1/lib/arm/libminetest.so (_ZN14ClientLauncher3runER10GameParamsRK8Settings+5486): Routine shared_ptr at /home/srifqi/android/android-ndk/sources/cxx-stl/llvm-libc++/include/memory:4110
Stack frame #12 pc 005c3407  /data/app/net.minetest.minetest-1/lib/arm/libminetest.so (main+2766): Routine unsigned int const& std::__ndk1::min<unsigned int, std::__ndk1::__less<unsigned int, unsigned int> >(unsigned int const&, unsigned int const&, std::__ndk1::__less<unsigned int, unsigned int>) at /home/srifqi/android/android-ndk/sources/cxx-stl/llvm-libc++/include/algorithm:2646
Stack frame #13 pc 006767d7  /data/app/net.minetest.minetest-1/lib/arm/libminetest.so (android_main+134): Routine std::__ndk1::basic_string<char, std::__ndk1::char_traits<char>, std::__ndk1::allocator<char> >::__zero() at /home/srifqi/android/android-ndk/sources/cxx-stl/llvm-libc++/include/string:1319
Stack frame #14 pc 00b564c5  /data/app/net.minetest.minetest-1/lib/arm/libminetest.so: Routine loop at curl_fnmatch.c:?
Stack frame #15 pc 0003fc8b  /system/lib/libc.so (_ZL15__pthread_startPv+30)
Stack frame #16 pc 0001a3ab  /system/lib/libc.so (__start_thread+6)
```
</details>

paramat:
It fixes #7122 which is a blocker bug that makes Android unusable.

src/gui/guiPathSelectMenu.h
stujones11: These methods are pure virtual out of necessity so must be overridden by every derived class. I didn't really like doing that either but cannot see any easier way around it.

src/gui/modalMenu.cpp
stujones11: Not my original code I might point out but yes, that is better.

stujones11:
@srifqi That is very strange because the bulk of this PR deals with moving the touchsceen handling, the original crash was also caused by `TouchScreenGUI` which seem fixed for me. Really speaking, this object should not even be required until the game starts as it is only for the in-game controls, however, I cannot seem to find where it is being accessed from.

Your backtrace does provide a few clues so I will look at that again. Out of curiosity, did you build this directly from my feature branch or apply the patch to an existing build? It seems strange that I would not get this, except I am using android 7.1.1 (SG Tab-A)

srifqi:
@stujones11 Yes, I was `checkout`-ing into your branch and then do `make clean && make -j2`.

srifqi:
@stujones11 Yes, I was `git checkout`-ing into your branch and then do `make clean && make -j2`.

stujones11:
Yeah, that is usually how I test PRs, is it possible for you to bypass that (using an existing username perhaps) and check if the password change menu is working?

srifqi:
No, it is not working. I can't continue playing even if I tapped anywhere on screen. Also, when I push back button, Minetest closes and the server says that I left the game.

stujones11:
Hmm, this looks to me like something may have gone wrong with the build or the apk upload.

@nerzhul are you able to test this? @Wayward1 ?

srifqi:
Probably. I'll try to recompile when I can. Let's hope that it's just my build. :)

stujones11:
Okay, thanks. I did do a complete re-build (deps and all) in a clean env immediately prior to making this PR as I have been caught out before by weirdness like this in the build system.

stujones11:
Lint is still failing so I will have to let the reviewers tell me what I need to change, bearing in mind that the majority of this code was already in `guiFormSpecmMenu.cpp` :)

srifqi:
Yes, the GUI is working! (stujones11/minetest@e50a68526cefd2e2c57412ea2ba00fce1136fda8)

---

Touchscreen GUI shown on loading screen.
Touchscreen GUI shown on loading screen

---

> Also, when I push back button, Minetest closes and the server says that I left the game.

Maybe, you also want to fix that? ^

stujones11:
> Touchscreen GUI shown on loading screen after confirm registration.

I had completely missed that, thanks for pointing it out. It should now be fixed.

> Also, when I push back button, Minetest closes and the server says that I left the game.

Yes, this has always been an issue, afaik and should be dealt with separately (IMO) at least with the buttons now working it is not so much of a problem.

I will look into it and if I can incorporate a fix with this PR then I will, however, I think that it needn't hold up more the important fixes in the meantime, I can always make another PR to fix the back button.

stujones11:
@srifqi I believe I have fixed the the back button issue, however, I accidentally squashed the commit when cleaning up. I just needed to add a check for `KEY_CANCEL` in the event handlers.

I would appreciate it if you could re-test this :)

src/game.cpp
```diff
@@ -1287,7 +1287,13 @@ bool Game::createClient(const std::string &playername,
 		return false;
 
 	bool could_connect, connect_aborted;
-
+#ifdef HAVE_TOUCHSCREENGUI
+	if (g_touchscreengui)
+	{
```
SmallJoker: Incorrect brace placement.

src/gui/guiConfirmRegistration.cpp
```diff
@@ -236,3 +245,20 @@ bool GUIConfirmRegistration::OnEvent(const SEvent &event)
...
+
+#ifdef __ANDROID__
+bool GUIConfirmRegistration::getAndroidUIInput()
+{
+	if (!hasAndroidUIInput() || m_jni_field_name != "password")
+		return false;
+
+	std::string text = porting::getInputDialogValue();
+	gui::IGUIElement *e;
+	e = getElementFromId(ID_confirmPassword);
```
SmallJoker: `gui::IGUIElement *e = getElementFromId(ID_confirmPassword);`

src/gui/guiPasswordChange.cpp
```diff
@@ -259,3 +265,39 @@ bool GUIPasswordChange::OnEvent(const SEvent &event)
...
+
+	gui::IGUIElement *e;
+	if (m_jni_field_name == "old_password")
+		e = getElementFromId(ID_oldPassword);
+	if (m_jni_field_name == "new_password_1")
+		e = getElementFromId(ID_newPassword1);
+	if (m_jni_field_name == "new_password_2")
+		e = getElementFromId(ID_newPassword2);
+	if (e)
+	{
```
SmallJoker: Incorrect brace placement.

src/gui/guiPasswordChange.cpp
```diff
@@ -259,3 +265,39 @@ bool GUIPasswordChange::OnEvent(const SEvent &event)
...
+	return "";
+}
+
+#ifdef __ANDROID__
+bool GUIPasswordChange::getAndroidUIInput()
+{
+	if (!hasAndroidUIInput())
+		return false;
+
+	gui::IGUIElement *e;
```
SmallJoker: `= nullptr`. Otherwise you're checking against garbage below.

src/gui/guiPasswordChange.cpp
```diff
@@ -259,3 +265,39 @@ bool GUIPasswordChange::OnEvent(const SEvent &event)
...
+	if (m_jni_field_name == "new_password_1")
+		e = getElementFromId(ID_newPassword1);
+	if (m_jni_field_name == "new_password_2")
+		e = getElementFromId(ID_newPassword2);
+	if (e)
+	{
+		std::string text = porting::getInputDialogValue();
+		e->setText(utf8_to_wide(text).c_str());
+	}
+	m_jni_field_name.clear();
```
SmallJoker: Why must this be cleared?
Probably also `return true;` in the `if` above?

SmallJoker:
Looks good overall. However, consider fixing Lint by adding `// clang-format ` `on/off` to fix the build status, what it suggests is not intended.

src/gui/guiPasswordChange.cpp
stujones11: I am not sure about that, the code is based on the existing `guiFormSpecMenu` implementation https://github.com/minetest/minetest/blob/9855651c06f84f29780cf6b25a392ac841b4630c/src/gui/guiFormSpecMenu.cpp#L2252. AFAIK the return value of this is never really checked anyway, I believe the output is cleared since we currently need to poll this in the absence of proper callbacks. That should probably be fixed at some point but is well beyond the scope of this PR.

src/gui/guiPasswordChange.cpp
```diff
@@ -215,12 +215,14 @@ bool GUIPasswordChange::processInput()
 bool GUIPasswordChange::OnEvent(const SEvent &event)
 {
 	if (event.EventType == EET_KEY_INPUT_EVENT) {
+		// clang-format off
```
nerzhul: they should not be indented

src/gui/guiPasswordChange.cpp
```diff
@@ -285,15 +287,14 @@ bool GUIPasswordChange::getAndroidUIInput()
 	if (!hasAndroidUIInput())
 		return false;
 
-	gui::IGUIElement *e;
+	gui::IGUIElement *e = nullptr;
 	if (m_jni_field_name == "old_password")
 		e = getElementFromId(ID_oldPassword);
 	if (m_jni_field_name == "new_password_1")
```
nerzhul: i think we can use else_if too , as you are modifying the block

src/gui/guiPasswordChange.cpp
stujones11: As far as I can see this is a line continuation and should be double indented. The linter, however, suggests I add some spaces to indent it further. Please feel free to fix that to your liking on merge, I am done with it now :P

paramat: I agree that indented is better to visually preserve the indentation of the code block.
The linter unfortunately asks for double-tab indentation often, which is wrong.
So it's ok as-is.

src/gui/guiPasswordChange.cpp
paramat: @stujones11 any thoughts on this suggestion?

paramat:
@SmallJoker clang on/off is doen, are you +1 for this now?

src/gui/guiConfirmRegistration.cpp
```diff
@@ -193,10 +196,14 @@ bool GUIConfirmRegistration::processInput()
 bool GUIConfirmRegistration::OnEvent(const SEvent &event)
 {
 	if (event.EventType == EET_KEY_INPUT_EVENT) {
-		if (event.KeyInput.Key == KEY_ESCAPE && event.KeyInput.PressedDown) {
+		// clang-format off
+		if ((event.KeyInput.Key == KEY_RETURN ||
```
SmallJoker: Conflicts with the check in L207 below. Are you sure about removing the `KEY_ESCAPE` check here?

stujones11: Good catch, that should be `KEY_ESCAPE` not `KEY_RETURN` same in `guiPasswordChange.cpp`

paramat:
Bump for attention, is a blocker.

src/gui/guiPasswordChange.cpp
paramat: @stujones11 what's your opinion on this?

src/gui/guiPasswordChange.h
```diff
@@ -44,6 +44,13 @@ class GUIPasswordChange : public GUIModalMenu
 	bool processInput();
 
 	bool OnEvent(const SEvent &event);
+#ifdef __ANDROID__
+	bool getAndroidUIInput();
+#endif
+
+protected:
+	std::wstring getLabelByID(s32 id) { return L""; }
+	std::string getNameByID(s32 id);
```
nerzhul: this function should return const std::string & instead

stujones11: Maybe, but that code was taken from the existing implementation in `guiFormSpecMenu.h` and simply moved to the base class. 

https://github.com/minetest/minetest/blob/79e393cf4b5027eb9071df23f4834aac3f3a689f/src/gui/guiFormSpecMenu.h#L538-L539

If after 3 months you are going to come back with something like this then I feel I am wasting my time. The android build is currently borked anyway, feel free to adopt this PR if you wish to make more style changes.

stujones11: @paramat, he is probably right and I would not have minded if it had been pointed out before now. There are many places in MT code-base where constness is ignored and it is a good idea to fix these issues with new code. I will make these changes when I have time, please forgive me for feeling a bit dejected (if that's the right word)

paramat: Hey no problem, it's very understandable, contributing when dev and reviews are slow is difficult and frustrating.

src/gui/guiPathSelectMenu.h
paramat: Seems like we need another opinion on this, @sfan5 maybe?

src/gui/guiPasswordChange.h
stujones11: Yeah, frustrated, that was the word I was looking for :)

I would be willing to do a lot more if I thought it might be processed in weeks rather than months.

stujones11: > this function should return const std::string & instead

@nerzhul I started to do this then realized that we cannot return a const reference to the string literals `L""` & `""` as these are temporary objects.

stujones11:
I have fixed a small regression that I can only guess was the result of a merge conflict resolution oversight.

I have retested this PR just now with 153fb211ac2342907eb766a79c1f41824f981ab5 reverted and it is working fine on my device so I am am leaving it here. Fixing const correctness in the formspec mess is not something I want to get into, sorry.

paramat:
Thanks, and that's fine.

src/gui/guiPathSelectMenu.h
paramat: @nerzhul maybe this isn't a big issue and could be dealt with later if necessary?

stujones11:
> you can create a static const std::string member

Yes, I know that but...

> Fixing const correctness in the formspec mess is not something I want to get into, sorry.

stujones11:
An alternative to this PR could be to get rid of the unnecessary (IMO) confirm registration nonsense and somehow convert the change password menu to use a standard fromspec.

However, from a maintenance point of view, I think moving the touchscreen event handling to the base class is generally a good idea. It may even make the eventual replacement of the current 'formspec mess' an easier task for whoever might be brave (or stupid) enough to take that challenge on :)

src/gui/guiPathSelectMenu.h
nerzhul: @stujones11 lgtm

paramat:
Bump for attention, please can anyone help this along?

stujones11, nerzhul doesn't have time for Minetest at the moment unfortunately so may not be able to complete this.

stujones11:
> nerzhul doesn't have time for Minetest at the moment

In that case it is probably best you do not let the remaining android issues block a release. AFAIK he is the only one that can update the playstore release, anyway.

paramat:
Well i mean, i think he has time to update the playstore, just not much time for reviewing/developing.
However i agree that Android won't be perfect at release, especially the GUI scaling issues. I removed that issue from blockers.

sfan5:
Is this WIP or can this be merged?

sfan5:
:+1:

sfan5:
323c8600450a5ae3893de9ba1d04095589c5b06c 88ba1c70d33cfeb09b73e46d720099083925c805

PR #7488, opened 2018-06-26
Fix libgmp detection

mazocomp:
Noticed by @landryb


landryb:
Reasoning : it might work on linux where there's a 'libgmp.so' symlink under usr/lib, but OpenBSD doesnt have those symlinks, and cmake is clever enough to find our libgmp.so.X.Y when told to look for gmp. This also aligns with the other find_library() calls in cmake/Modules/*.

PR #7491, opened 2018-06-27
Log deprecated Lua function calls

SmallJoker:
Logs calls to `setpos`, `setvelocity` etc, so these can be removed in the future to drop a part of our legacy modding API together with `minetest.env:`.

This is a generic approach: (explanation using own thoughts)
1) There are currently 21 deprecated function names, some of them often used
2) The old function names are aliased to the new C++ function, `luamethod_aliased` registers the old function always below the new function. Both have the same function pointer (was only renamed)
3) Backtrace to the bad file is in all cases required to avoid spam log messages (imagine `on_step` calling `:getpos` directly)
4) All API classes differ, but the function types don't -> call the pointer
5) The pointer is either specified on-compile or by a lookup table in between. See question below for the on-compile option.

**Why not a new function for each deprecated function?**
Possible using a giant `#define NEW_DEP_API_FUNC(class, oldname, newname, newfunc)` to do the backtrace and ignoring duplicates. Plus additional lines to declare them in the header files. At the end it would save the lookup table `m_deprecated_wrappers` which has only 11 entries for `ObjectRef` (max).

**How fast is it?**
About 8 μs on my system (debug build) for each deprecated function call -> see the F6 stats, page 3: `l_deprecatedFunction`

### DIFF
--- a/src/script/lua_api/l_base.cpp
+++ b/src/script/lua_api/l_base.cpp
@@ -91,7 +91,7 @@ bool ModApiBase::registerFunction(lua_State *L, const char *name,
 std::unordered_map<std::string, luaL_Reg> ModApiBase::m_deprecated_wrappers;
 bool ModApiBase::m_error_deprecated_calls = false;
 
-int ModApiBase::l_deprecatedFunction(lua_State *L)
+int ModApiBase::l_deprecated_function(lua_State *L)
 {
 	thread_local std::vector<u64> deprecated_logged;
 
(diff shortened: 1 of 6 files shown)


src/script/lua_api/l_base.cpp
```diff
@@ -102,3 +104,80 @@ float ModApiBase::readParam(lua_State *L, int index)
 
 	return (float) luaL_checknumber(L, index);
 }
+
+std::unordered_map<std::string, luaL_Reg> ModApiBase::m_deprecated_wrappers;
+bool ModApiBase::m_error_deprecated_calls = false;
+
+int ModApiBase::l_deprecatedFunction(lua_State *L)
```
nerzhul: i tend to prefer consistent style; l_deprecated_function is better here

PR #7493, opened 2018-06-27
Fix memory leak in guiConfirmRegistration

HybridDog:


SmallJoker:
Lint is not happy about this change. It can be done manually on merge or by @HybridDog if they've got a minute.

ClobberXD:
Travis is still grumpy as LINT has failed again.

It'd be nice to have it also suggest the correct style for the erring line(s)... :)

SmallJoker:
53dd781

PR #7494, opened 2018-06-27
Fix world deletion

### LINKED ISSUE #7201: Minetest deletes the contents of outside directories instead of the symlinks that point ot them when deleting worlds
##### Issue type
 - Bug report
 
##### Minetest version
```
0.4.16
```

##### OS / Hardware

Operating system: Debian 9
CPU: Irrelevant

##### Summary
If a symbolic link is present in a world directory and Minetest deletes that world, it'll delete the contents of the symbolically linked directory and not the symlink itself. Because the symbolic link doesn't get deleted, this also prevents the world's directory from being properly deleted, so you end up with a world directory containing only the symbolic link to the freshly-emptied outside directory.

##### Steps to reproduce
0. Start a new world.
1. Symbolically link something into that world's directory for use in multiple worlds (for example, a particular `worldmods` directory).
2. Delete that world, expecting the symbolically-linked directory to still be there for use in other worlds.
3. Check your linked directory to find all the mods (or whatever) you had in the linked directory have been erased with no chance of recovery. The linked directory itself still exists, but the contents do not.

numberZero:
Fix #7201

### DIFF
--- a/src/script/lua_api/l_mainmenu.cpp
+++ b/src/script/lua_api/l_mainmenu.cpp
@@ -594,31 +594,18 @@ int ModApiMainMenu::l_create_world(lua_State *L)
 /******************************************************************************/
 int ModApiMainMenu::l_delete_world(lua_State *L)
 {
-	int worldidx	= luaL_checkinteger(L,1) -1;
-
+	int world_id = luaL_checkinteger(L,1) - 1;
 	std::vector<WorldSpec> worlds = getAvailableWorlds();
-
-	if ((worldidx >= 0) &&
-		(worldidx < (int) worlds.size())) {
-
-		WorldSpec spec = worlds[worldidx];
-
-		std::vector<std::string> paths;
-		paths.push_back(spec.path);
-		fs::GetRecursiveSubPaths(spec.path, paths, true);
-
-		// Delete files
-		if (!fs::DeletePaths(paths)) {
-			lua_pushstring(L, "Failed to delete world");
-		}
-		else {
-			lua_pushnil(L);
-		}
-	}
-	else {
+	if ((world_id < 0) || (world_id >= (int) worlds.size())) {
 		lua_pushstring(L, "Invalid world index");
+		return 1;
 	}
-	return 1;
+	const WorldSpec &spec = worlds[world_id];
+	if (!fs::RecursiveDelete(spec.path)) {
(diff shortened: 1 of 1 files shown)


src/script/lua_api/l_mainmenu.cpp
```diff
@@ -594,31 +594,18 @@ int ModApiMainMenu::l_create_world(lua_State *L)
 /******************************************************************************/
 int ModApiMainMenu::l_delete_world(lua_State *L)
 {
-	int worldidx	= luaL_checkinteger(L,1) -1;
-
+	int world_id = luaL_checkinteger(L,1) - 1;
```
sfan5: spacing around `,`

src/script/lua_api/l_mainmenu.cpp
```diff
@@ -594,31 +594,18 @@ int ModApiMainMenu::l_create_world(lua_State *L)
...
-		// Delete files
-		if (!fs::DeletePaths(paths)) {
-			lua_pushstring(L, "Failed to delete world");
-		}
-		else {
-			lua_pushnil(L);
-		}
-	}
-	else {
+	if ((world_id < 0) || (world_id >= (int) worlds.size())) {
```
sfan5: useless braces

src/script/lua_api/l_mainmenu.cpp
ClobberXD: Isn't spacing required only to the _right_ of `,`?

SmallJoker: Yes. Space *after* `,`

ClobberXD: Thank you for clarifying

nerzhul:
interesting legacy code, while we got a recursive removal function :)

red-001:
is #7433 a prerequisite for this?

numberZero:
@red-001 Ugh, forgot that... yes, for Windows™, #7433 is necessary.

src/script/lua_api/l_mainmenu.cpp
numberZero: added

src/script/lua_api/l_mainmenu.cpp
numberZero: removed

PR #7498, opened 2018-06-28
Android: Resize jump and crunch buttons' touch target

srifqi:
Resize jump and crunch buttons' touch target so that they fill empty space below them.
First discussed: https://github.com/minetest/minetest/commit/b1e58c9c35ae3eb6167a0745086cdb0ddb1dd9d7#r29473361

Before: (ignore the AUX button)
touch target (before)
After: (not final texture, read below)
touch target (after)


paramat:
How about this: Keep the buttons the same width but extend them upwards a little to make them square, then move the icons down a little to the centre of the buttons?

SmallJoker:
My only concern is that the images should be vertically centred, because that's where buttons usually are pressed.

srifqi:
@paramat @SmallJoker Done.

paramat:
Screenshot please.

paramat:
Or perhaps: Place the icons where it is comfortable to press when using a case (that may be where they are currently), create a square button around the icon with the icon at the centre, then extend the button down to the screen edge.

stujones11:
> Screenshot please.

screen

paramat:
Jump and sneak buttons look ok.
Please could you show a screenshot with the AUX button enabled?

stujones11:
> Please could you show a screenshot with the AUX button enabled?

https://github.com/minetest/minetest/pull/7477#issuecomment-401560983

paramat:
:+1:

srifqi:
Thank you. I'm still away from my computer. So, I can't provide any screenshots etc.

PR #7500, opened 2018-06-28
Fix memory leaks in mod storage

red-001:


src/script/lua_api/l_storage.cpp
```diff
@@ -57,6 +57,11 @@ StorageRef::StorageRef(ModMetadata *object):
 {
 }
 
+StorageRef::~StorageRef()
+{
+	delete m_object;
```
nerzhul: why StorageRef delete m_object ? it's passed by the constructor and seems to be owned by something else no ?

SmallJoker: `core.get_mod_storage` creates the pointer, which is then passed to the server or client which then constructs the StorageRef object with that pointer.

SmallJoker:
I added the Testing Needed label because it's not sure whether we can rely on the Lua garbage collector to destruct and unregister the class correctly from the server or client.

src/script/lua_api/l_storage.cpp
nerzhul: yeah it's on the Lua master stack, i should reverify my code to ensure it's not stored elsewhere. if not we can remove it safely

nerzhul:
It's okay, we can merge it

nerzhul:
@SmallJoker the StorageRef is deleted by the lua GC in l_storage.cpp#L74 which will destroy here the ModStorage object

PR #7538, opened 2018-07-08
Stop players on Prejoin if they are already connected.

Lejo1:
Prevents that you have to wait until everything is loaded and then you get the message that you can't join.

### DIFF
--- a/builtin/game/misc.lua
+++ b/builtin/game/misc.lua
@@ -78,7 +78,7 @@ core.register_on_leaveplayer(function(player, timed_out)
 	core.send_leave_message(player_name, timed_out)
 end)
 
-minetest.register_on_prejoinplayer(function(name)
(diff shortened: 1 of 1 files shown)


sfan5:
This is code duplication, the check that happens in C++ should just be moved.

Lejo1:
Is this on prejoin?

SmallJoker:
C++ code (top and bottom line):
https://github.com/minetest/minetest/blob/334d8c31ee96f2e6a8464c99eecf84793f598fb6/src/network/networkprotocol.h#L896-L917

nerzhul:
As session system is sensible, this should never be moved to lua. Callback order is not guaranteed, and the lua ref can be retained while the core hasn't any network session. Network session is a pure C++ thing, there is no sense to have it inside Lua, and it's shared between threads.

PR #7544, opened 2018-07-11
Allow an optional readonly base database

lhofhansl:
This is a trivial change that optionally allows for a readonly base DB that resides in a "readonly" subfolder of the world folder.
Usecases would be the following:
1. Make the primary backend dummy. Now the readonly DB can be the seed, the games can proceed and easily be reverted back.
2. The primary could also be persistent. Now we can collect "diffs" to the base DB and copy those around.

If we have a readonly DB configured we won't attempt to load from files.
I tested this with a few maps, works as expected!

This is just a concept for discussion. Should we have a lua API with this? How would that work? etc

### DIFF
--- a/doc/world_format.txt
+++ b/doc/world_format.txt
@@ -98,8 +98,15 @@ See Player File Format below.
 world.mt
 ---------
 World metadata.
-Example content (added indentation):
-  gameid = mesetint
+Example content (added indentation and - explanations):
+  gameid = mesetint             - name of the game
+  enable_damage = true          - whether damage is enabled or not
+  creative_mode = false         - whether creative mode is enabled or not
+  backend = sqlite3             - which DB backend to use for blocks (sqlite3, dummy, leveldb, redis, postgresql)
+  player_backend = sqlite3      - which DB backend to use for player data
+  readonly_backend = sqlite3    - optionally readonly seed DB (DB file _must_ be located in "readonly" subfolder)
+  server_announce = false       - whether the server is publicly announced or not
+  load_mod_<mod> = false        - whether <mod> is to be loaded in this world
 
 Player File Format
 ===================
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -1159,7 +1159,10 @@ ServerMap::ServerMap(const std::string &savedir, IGameDef *gamedef,
 	}
 	std::string backend = conf.get("backend");
 	dbase = createDatabase(backend, savedir, conf);
-
+	if (conf.exists("readonly_backend")) {
+		std::string readonly_dir = savedir + DIR_DELIM + "readonly";
+		dbase_ro = createDatabase(conf.get("readonly_backend"), readonly_dir, conf);
+	}
 	if (!conf.updateConfigFile(conf_path.c_str()))
 		errorstream << "ServerMap::ServerMap(): Failed to update world.mt!" << std::endl;
 
@@ -1230,6 +1233,8 @@ ServerMap::~ServerMap()
 		Close database if it was opened
 	*/
 	delete dbase;
+	if (dbase_ro)
+		delete dbase_ro;
 
 #if 0
 	/*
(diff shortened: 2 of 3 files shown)


lhofhansl:
I meant "base database". :)
I.e. a database as a base, with changes stored in another DB.

paramat:
Sorry, the first line of the first comment has 'readonly based DB' and i thought there was an inconsistency.

Usecases:
1. How is this different from copying a world file as a backup to revert to later? Maybe it isn't but allows to do it in-game instead of manually?
2. Does this mean having 2 world files and being able to use the 2nd for only storing changes to a world? I guess the 'changes' world would then appear as player builds floating in empty space (ignore)?

lhofhansl:
@paramat 
1. You are right that the same can be achieved by copying and restoring a world file. The advantage here is that there's more control from lua. For example an area can be restored by simply calling deleteblocks - I tried that, it's quite magical... You make a bunch of changes to a map, then call deleteblocks and it reverts to what it was before. Granted the same could in theory be done with rollback.

2. You are correct here too. The diff world is useless by itself.

lhofhansl:
Thinking about additional APIs too. Could have resetMap (or resetDatabase)... All that would be required is to disconnect and then reconnect to a dummy database.

Also not entirely clear what to do with players.

nerzhul:
i don't see the point here, you will duplicate the map storage ?

ClobberXD:
How are changes committed to the original map?

lhofhansl:
@ClobberXD They would not. The original map would always remain entirely unchanged.

ClobberXD:
Ah ok, interesting. But how would this help? I'm not able to understand the use-cases proposed...

rubenwardy:
This would help with CTF, as you could preload the schematics into the original map then use delete_area to remove any modifications. So it would result in only one map send and no need to recalculate lighting

paramat:
> Phrased in a different way... Would merging this have any detrimental effect?

Lol :) It seems this may be useful, but just wanted to say that lack of harm is not enough reasion to add something :)

ClobberXD:
@paramat: As rubenwardy said, this feature would be extremely useful for round-based games that keep resetting periodically, CTF being a classic example.

rubenwardy:
A detrimental effect could be more code to maintain. However, added lines of code isn't a good maintenance cost heuristic as where it is added does matter

lhofhansl:
> It seems this may be useful, but just wanted to say that lack of harm is not enough reasion to add something :)

Fair enough. :)

I'd definitely have to find out where best to document this!

lhofhansl:
I think this is quite useful. I could use some ideas about:
1. How/where to document this.
2. What (if any) new lua interfaces would be good for this.

SmallJoker:
@lhofhansl The documentation is [currently a stub](https://github.com/minetest/minetest/blob/master/doc/world_format.txt#L98). Would be good to have it updated (and reformatted?) at some point.
What kind of Lua API would you like to add? Map information is loaded before the mods.

lhofhansl:
Added basic documentation. Let me know what you think.

SmallJoker:
The documentation is good enough IMO.

ExeterDad:
This would be a great addition! Currently I have a database of pre-built areas that are much to big for WE and are used to "reset" the main map from time to time during restart with database commands. 
The idea of being able to "reset" while the server is running has me a bit excited.

lhofhansl:
@ExeterDad Do you think we need a new API for that?
Right now you could call `deleteblocks`, but you'd have to give it a radius and it would actually go through the (main) DB and delete all blocks.

lhofhansl:
In that case we have everything we need, and this should be ready to be merged.

lhofhansl:
Any further comments? It's a low risk change with a high potential upside.

SmallJoker:
screenshot_20180725_160006
Screenshot: Front is the readonly sqlite3 map, back: dummy backend for mapgen testing.
The PR is simple, but works exactly how it should. The loaded blocks from the readonly map can be modified, but they restore as soon `/deleteblocks` is called or they get reloaded.

PR #7548, opened 2018-07-13
Ensure that legacy mode is correctly disabled.

### LINKED ISSUE #7545: MT sliding window algorithm is ineffective
Just for kicks I setup a 100ms delay on my local network. I send observed packet transfer rates.
Turns out that the rate when down to about 5 packets/s (i.e. a RTT of 200ms). So it looks like every single packet is ack'd individually.
There's plenty of CPU on both client and server and plenty of *throughput* on the link. Just latency is high.

That is really bad!
MT should start streaming and delay acks for packets when the RTT is high. The code is clearly there, but it apparently it's not working.

I'll take a look at it during the next few days.

@nerzhul you attempted to replace MT's networking stack, perhaps you have some ideas.

lhofhansl:
Fixes #7545

For some reason legacy mode was never properly disabled and that caused the UdpPeer's sliding window size to be hardcoded to 5, which kills performance on higher latency links (as low as 20-30ms RTTs)

### DIFF
--- a/src/network/connectionthreads.cpp
+++ b/src/network/connectionthreads.cpp
@@ -1208,6 +1208,10 @@ SharedBuffer<u8> ConnectionReceiveThread::handlePacketType_Control(Channel *chan
 			m_connection->SetPeerID(peer_id_new);
 		}
 
+		// set non legacy mode locally
+		dynamic_cast<UDPPeer *>(peer)->setNonLegacyPeer();
+
+		// request the same from the remote side
 		ConnectionCommand cmd;
 
 		SharedBuffer<u8> reply(2);


src/network/connectionthreads.cpp
```diff
@@ -1208,6 +1208,10 @@ SharedBuffer<u8> ConnectionReceiveThread::handlePacketType_Control(Channel *chan
 			m_connection->SetPeerID(peer_id_new);
 		}
 
+		// set non legacy mode locally
+		dynamic_cast<UDPPeer *>(peer)->setNonLegacyPeer();
+
+		// request the same from the remote side
```
sfan5: request?
so if the remote side does not support this, wouldn't we need to omit the `setNonLegacyPeer()` call?

lhofhansl: The sending sliding window for packets is independent on the other side also have a sliding window. It simply means that that many packets can be outstanding.

... I think ...  There is *soo* much code around this.

lhofhansl: We can also just remove the whole legacy peer nonsense going forward.

sfan5: Yes, that would be much preferred.

lhofhansl: As far as I can see this was wrong from the beginning (looked through the git history). So yes, I think I think it does matter, but I'll admit I'm running out of my depth there. We had removed a bunch of the old protocol stuff a while ago, so perhaps this was left over.

Lemme amend this with a complete removal.

lhofhansl: Turned out to be OK: #7551

sfan5: It would still be useful to merge this first to make it easy to backport the fix (also so it won't be forgotten).

lhofhansl: Agreed. Will do that now.

lhofhansl:
93c0b3b8

lhofhansl:
Oh and look what I found:
```
void UDPPeer::setNonLegacyPeer()
{
	m_legacy_peer = false;
	for(unsigned int i=0; i< CHANNEL_COUNT; i++)
	{
		channels->setWindowSize(g_settings->getU16("max_packets_per_iteration"));
	}
}
```
Funny :)

This should be 
```
void UDPPeer::setNonLegacyPeer()
{
	m_legacy_peer = false;
	for(unsigned int i=0; i< CHANNEL_COUNT; i++)
	{
		channels[i].setWindowSize(g_settings->getU16("max_packets_per_iteration"));
	}
}
```

I'm just going to change that. It's obviously not intended.

lhofhansl:
Pushed that too.

lhofhansl:
I'm happy with this. Can almost not tell the difference between a 0.01ms local and 200ms delayed link.

lhofhansl:
I suppose this should be backported.

PR #7551, opened 2018-07-13
Remove old (unnecessary) legacy client handling.

lhofhansl:
See also #7548

This removes legacy client handling code that actually never worked before - it would treat all clients as legacy clients!
The largest change is just whitespace.

### DIFF
--- a/src/network/connection.cpp
+++ b/src/network/connection.cpp
@@ -663,45 +663,35 @@ void Channel::UpdateTimers(float dtime)
 
 		if (packets_successful > 0) {
 			successful_to_lost_ratio = packet_loss/packets_successful;
-		}
-		else if (packet_loss > 0)
-		{
+		} else if (packet_loss > 0) {
 			window_size = std::max(
 					(window_size - 10),
 					MIN_RELIABLE_WINDOW_SIZE);
 			done = true;
 		}
 
-		if (!done)
-		{
+		if (!done) {
 			if ((successful_to_lost_ratio < 0.01f) &&
-				(window_size < MAX_RELIABLE_WINDOW_SIZE))
-			{
+				(window_size < MAX_RELIABLE_WINDOW_SIZE)) {
 				/* don't even think about increasing if we didn't even
 				 * use major parts of our window */
 				if (reasonable_amount_of_data_transmitted)
 					window_size = std::min(
 							(window_size + 100),
 							MAX_RELIABLE_WINDOW_SIZE);
-			}
-			else if ((successful_to_lost_ratio < 0.05f) &&
-					(window_size < MAX_RELIABLE_WINDOW_SIZE))
-			{
+			} else if ((successful_to_lost_ratio < 0.05f) &&
+					(window_size < MAX_RELIABLE_WINDOW_SIZE)) {
 				/* don't even think about increasing if we didn't even
 				 * use major parts of our window */
 				if (reasonable_amount_of_data_transmitted)
 					window_size = std::min(
 							(window_size + 50),
 							MAX_RELIABLE_WINDOW_SIZE);
-			}
-			else if (successful_to_lost_ratio > 0.15f)
-			{
+			} else if (successful_to_lost_ratio > 0.15f) {
 				window_size = std::max(
 						(window_size - 100),
 						MIN_RELIABLE_WINDOW_SIZE);
-			}
-			else if (successful_to_lost_ratio > 0.1f)
-			{
+			} else if (successful_to_lost_ratio > 0.1f) {
 				window_size = std::max(
 						(window_size - 50),
 						MIN_RELIABLE_WINDOW_SIZE);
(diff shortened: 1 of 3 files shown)


lhofhansl:
Needs a small rebase after #7548 
Let's agree we want to do this, if so I'll rebase.

SmallJoker:
Also needs code style cleanups.

lhofhansl:
@SmallJoker you mean because I touched it? Fair enough. It's just code removal and whitespace changes.

nerzhul:
if you can fix it can be nice, but yeah nice job to have looked at this

src/network/connection.cpp
```diff
@@ -657,56 +657,54 @@ void Channel::UpdateTimers(float dtime,bool legacy_peer)
 			current_packet_successfull = 0;
 		}
 
-		/* dynamic window size is only available for non legacy peers */
-		if (!legacy_peer) {
-			float successfull_to_lost_ratio = 0.0;
-			bool done = false;
+		/* dynamic window size */
+		float successfull_to_lost_ratio = 0.0;
```
nerzhul: 0.0f

src/network/connection.cpp
```diff
@@ -657,56 +657,54 @@ void Channel::UpdateTimers(float dtime,bool legacy_peer)
...
+		bool done = false;
 
-			if (packets_successfull > 0) {
-				successfull_to_lost_ratio = packet_loss/packets_successfull;
```
nerzhul: successful

src/network/connection.cpp
```diff
@@ -657,56 +657,54 @@ void Channel::UpdateTimers(float dtime,bool legacy_peer)
...
+		if (packets_successfull > 0) {
+			successfull_to_lost_ratio = packet_loss/packets_successfull;
+		}
+		else if (packet_loss > 0)
+		{
+			window_size = MYMAX(
```
nerzhul: std::max

src/network/connection.cpp
```diff
@@ -657,56 +657,54 @@ void Channel::UpdateTimers(float dtime,bool legacy_peer)
...
+
+		if (!done)
+		{
+			if ((successfull_to_lost_ratio < 0.01) &&
+				(window_size < MAX_RELIABLE_WINDOW_SIZE))
+			{
+				/* don't even think about increasing if we didn't even
+				 * use major parts of our window */
+				if (reasonable_amount_of_data_transmitted)
+					window_size = MYMIN(
```
nerzhul: std::min

lhofhansl:
Lemme fix the code style and post a new version.

lhofhansl:
Rebased and fixed all code style issues I could see.

src/network/connection.cpp
```diff
@@ -627,17 +627,17 @@ void Channel::UpdatePacketTooLateCounter()
...
 }
 
-void Channel::UpdateTimers(float dtime,bool legacy_peer)
+void Channel::UpdateTimers(float dtime)
 {
 	bpm_counter += dtime;
 	packet_loss_counter += dtime;
 
-	if (packet_loss_counter > 1.0)
+	if (packet_loss_counter > 1.0f)
```
SmallJoker: Wrong bracket placement

src/network/connection.cpp
```diff
@@ -646,67 +646,65 @@ void Channel::UpdateTimers(float dtime,bool legacy_peer)
...
+		/* dynamic window size */
+		float successful_to_lost_ratio = 0.0f;
+		bool done = false;
 
-			if (packets_successfull > 0) {
-				successfull_to_lost_ratio = packet_loss/packets_successfull;
+		if (packets_successful > 0) {
+			successful_to_lost_ratio = packet_loss/packets_successful;
+		}
+		else if (packet_loss > 0)
```
SmallJoker: `} else if (...) {`

src/network/connection.cpp
```diff
@@ -646,67 +646,65 @@ void Channel::UpdateTimers(float dtime,bool legacy_peer)
...
+		{
+			window_size = std::max(
+					(window_size - 10),
+					MIN_RELIABLE_WINDOW_SIZE);
+			done = true;
+		}
+
+		if (!done)
+		{
```
SmallJoker: `if (..) {`
Same for the entire function part below.

src/network/connection.cpp
```diff
@@ -903,6 +901,10 @@ void Peer::Drop()
 UDPPeer::UDPPeer(u16 a_id, Address a_address, Connection* connection) :
 	Peer(a_address,a_id,connection)
 {
+	for(unsigned int i=0; i< CHANNEL_COUNT; i++)
```
SmallJoker: `for (size_t i = 0; i < CHANNEL_COUNT` or use the C++11 range-based "for" loop.
The brackets can be omitted, otherwise please correct the placement.

lhofhansl:
Ok... Fixed more code style issues. Let's please stop here, or have a separate PR. I'm a fan of targeted PRs, they can be tracked and reverted more easily when needed.

SmallJoker:
@lhofhansl Thanks for fixing these issues. In general, only the touched lines should be fixed, but your additional changes surely don't hurt anyone.
Which clients would needs this legacy code anyway? We bumped the minimal proto version, but I'd like to know where this was actually used in-game.

lhofhansl:
I'll admit I'm a bit confused about what is compatible with what.

lhofhansl:
ca8ec46843da054e656d2f63b23d0b1695c023da

PR #7555, opened 2018-07-14
Reduce ABM checks.

lhofhansl:
See #7280

Since is simple and low risk change, reducing the number of time blocks needs to be checked for whether ABMs needs to be run or not, by doing some simple caching.

This builds a cache of the node types in a blocks (up to ~~64~~ 32) and stores this with a cached block. Before any ABMs are checked we first check against this set.

The advantage is that cache invalidation is simple and only depends on a single block; the risk of incorrectly handling this is low.
The disadvantage is that we lose out of further optimization - for example we could cache whether the neighbor check failed, but then we'd need a more tricky protocol for cache invalidation.

Have a look. Something to discuss; but it is functioning as is.

### DIFF
--- a/src/mapblock.h
+++ b/src/mapblock.h
@@ -521,9 +521,6 @@ public:
 #ifndef SERVER // Only on client
 	MapBlockMesh *mesh = nullptr;
 #endif
-	std::unordered_set<content_t> contents;
-	bool contents_cached = false;
-	bool do_not_cache_contents = false;
 
 	NodeMetadataList m_node_metadata;
 	NodeTimerList m_node_timers;
@@ -533,6 +530,14 @@ public:
 	static const u32 zstride = MAP_BLOCKSIZE * MAP_BLOCKSIZE;
 
 	static const u32 nodecount = MAP_BLOCKSIZE * MAP_BLOCKSIZE * MAP_BLOCKSIZE;
+
+	//// ABM optimizations ////
+	// cache of content types
+	std::unordered_set<content_t> contents;
+	// true if content types are cached
+	bool contents_cached = false;
+	// true if we never want to cache content types for this block
+	bool do_not_cache_contents = false;
 
 private:
 	/*
--- a/src/mapblock.h
+++ b/src/mapblock.h
@@ -114,6 +114,8 @@ public:
 		} else if (mod == m_modified) {
 			m_modified_reason |= reason;
 		}
+		if (mod == MOD_STATE_WRITE_NEEDED)
+			contents_cached = false;
 	}
 
 	inline u32 getModified()
@@ -519,6 +521,9 @@ public:
 #ifndef SERVER // Only on client
 	MapBlockMesh *mesh = nullptr;
 #endif
+	std::set<content_t> contents;
+	bool contents_cached = false;
+	bool do_not_cache_contents = false;
 
 	NodeMetadataList m_node_metadata;
 	NodeTimerList m_node_timers;
--- a/src/mapblock.h
+++ b/src/mapblock.h
@@ -521,7 +521,7 @@ public:
 #ifndef SERVER // Only on client
 	MapBlockMesh *mesh = nullptr;
 #endif
-	std::set<content_t> contents;
+	std::unordered_set<content_t> contents;
 	bool contents_cached = false;
 	bool do_not_cache_contents = false;
 
--- a/src/mapblock.h
+++ b/src/mapblock.h
@@ -532,11 +532,11 @@ public:
 	static const u32 nodecount = MAP_BLOCKSIZE * MAP_BLOCKSIZE * MAP_BLOCKSIZE;
 
 	//// ABM optimizations ////
-	// cache of content types
+	// Cache of content types
 	std::unordered_set<content_t> contents;
-	// true if content types are cached
+	// True if content types are cached
 	bool contents_cached = false;
-	// true if we never want to cache content types for this block
+	// True if we never want to cache content types for this block
 	bool do_not_cache_contents = false;
 
 private:
--- a/src/serverenvironment.cpp
+++ b/src/serverenvironment.cpp
@@ -807,7 +807,7 @@ public:
 
 		if (block->contents_cached) {
 			bool run_abms = false;
-			for(content_t c : block->contents) {
+			for (content_t c : block->contents) {
 				if (c < m_aabms.size() && m_aabms[c]) {
 					run_abms = true;
 					break;
--- a/src/serverenvironment.cpp
+++ b/src/serverenvironment.cpp
@@ -800,7 +800,7 @@ public:
 		return active_object_count;
 
 	}
-	void apply(MapBlock *block, int &abm_checked, int &abm_run, int &abm_not_cached)
+	void apply(MapBlock *block, int &blocks_scanned, int &abms_run, int &blocks_cached)
 	{
 		if(m_aabms.empty() || block->isDummy())
 			return;
@@ -809,6 +809,7 @@ public:
 		// to see whether there are any ABMs
 		// to be run at all for this block
 		if (block->contents_cached) {
+			blocks_cached++;
 			bool run_abms = false;
 			for (content_t c : block->contents) {
 				if (c < m_aabms.size() && m_aabms[c]) {
@@ -821,9 +822,8 @@ public:
 		} else {
 			// clear any caching
 			block->contents.clear();
-			abm_not_cached++;
 		}
-		abm_checked++;
+		blocks_scanned++;
 
 		ServerMap *map = &m_env->getServerMap();
 
@@ -884,7 +884,7 @@ public:
 				}
 				neighbor_found:
 
-				abm_run++;
+				abms_run++;
 				// Call all the trigger variations
 				aabm.abm->trigger(m_env, p, n);
 				aabm.abm->trigger(m_env, p, n,
@@ -1333,9 +1333,9 @@ void ServerEnvironment::step(float dtime)
 			// Initialize handling of ActiveBlockModifiers
 			ABMHandler abmhandler(m_abms, m_cache_abm_interval, this, true);
 
-			int abm_checked = 0;
-			int abm_run = 0;
-			int abm_not_cached = 0;
+			int blocks_scanned = 0;
+			int abms_run = 0;
+			int blocks_cached = 0;
 			for (const v3s16 &p : m_active_blocks.m_abm_list) {
 				MapBlock *block = m_map->getBlockNoCreateNoEx(p);
 				if (!block)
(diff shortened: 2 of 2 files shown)


lhofhansl:
Hmm... Doesn't quite work with active objects...? They seem to pop in and out of existence.

lhofhansl:
Some minor improvements... Flew around in a few worlds, did not find a single block with more than 32 different content_types, so just caching 32 now (1/128 of the possible 4096 content types).
Also add some better stats.

Example: Just with the ethereal mod, `active_object_range` set to 7. I.e. 2084 active blocks, just 87 of those need to actually be checked:
screenshot_20180715_035236

That's a pretty significant saving... I did not see the 200ms warning at all.

lhofhansl:
More complex world with all kinds of mods (including NSSM) with lots of ABMs.

Still just 528 of the 2103 active blocks even need to be scanned.

screenshot_20180715_040143

src/mapblock.h
```diff
@@ -519,6 +521,9 @@ class MapBlock
 #ifndef SERVER // Only on client
 	MapBlockMesh *mesh = nullptr;
 #endif
+	std::set<content_t> contents;
```
nerzhul: unordered_set ?

src/serverenvironment.cpp
```diff
@@ -800,11 +800,29 @@ class ABMHandler
...
 	}
-	void apply(MapBlock *block)
+	void apply(MapBlock *block, int &abm_checked, int &abm_run, int &abm_not_cached)
 	{
 		if(m_aabms.empty() || block->isDummy())
 			return;
 
+		if (block->contents_cached) {
+			bool run_abms = false;
+			for(content_t c : block->contents) {
```
nerzhul: missing space

nerzhul:
good design, it's the design i wanted to add to mapblocks at a moment but didn't get time for it.

src/mapblock.h
lhofhansl: Oops. Yes.

lhofhansl:
Addressed comments.

Thanks on the design. :)
I'm still thinking about how to effectively cache the neighbor checks without doing a lot of work at write time and/or risking subtle bugs in the cache invalidation.

lhofhansl:
I think if/when this is merged we can increase the default active_block_range to 4... (Although I expect some opposition from @paramat :) )

lhofhansl:
With the "plantlife" modpack I find some blocks with more than 32 content types (4% of the blocks). With 64 is back down to 0.

So probably the 64 types limit was better. Or maybe it's not worth it...?
We're talking about 64 shorts, i.e. 128 bytes vs 64 bytes per cached block.

Fixer-007:
Try this world save too https://github.com/minetest/minetest/files/101456/Just.Test.zip

lhofhansl:
From the Just.Test world (I hope you didn't sneak any malicious lua scripts into that one).
Looks like 64 is still a good number...

screenshot_20180715_070309

paramat:
> I think if/when this is merged we can increase the default active_block_range to 4

I doubt it, because there will be worlds with complex built-up areas where almost all blocks are over the cache size so all blocks have ABMs running the more intensive way, these are also the areas where ABMs are working the hardest and triggering the most lua actions.
The result is that you will completely lose any performance gained from this optimisation and cause much worse performance in these areas. Better to keep range 3 and have a significant improvement in ABM performance everywhere.

My initial concern is the memory use of storing the caches, how large is a cache compared to a mapblock, i guess it's 64 added to 4096 * 2?
Cache invalidation is lightweight i assume because the cache is already loaded?

paramat:
So these caches are not saved to world DB? Or are they?

lhofhansl:
Nope. Not saved, cached the first time we try to execute ABMs. Hardly any extra work, nor any extra storage. About 140 bytes extra in RAM per cached block.

paramat:
Ok that's fine by me.

The worst case scenario is 1 player walking constantly through a world. With active block range 3, a certain block will be active for (48 * 2 + 16) / 4 = 28s so the block will be cached in the first second and the optimisation effective for the following 27s. Seems good.

src/serverenvironment.cpp
```diff
@@ -800,11 +800,29 @@ class ABMHandler
...
+			bool run_abms = false;
+			for (content_t c : block->contents) {
+				if (c < m_aabms.size() && m_aabms[c]) {
+					run_abms = true;
+					break;
+				}
+			}
+			if (!run_abms) {
+				return;
+			}
```
sfan5: brace removal

src/serverenvironment.cpp
```diff
@@ -800,11 +800,29 @@ class ABMHandler
...
 	}
-	void apply(MapBlock *block)
+	void apply(MapBlock *block, int &abm_checked, int &abm_run, int &abm_not_cached)
 	{
 		if(m_aabms.empty() || block->isDummy())
 			return;
 
+		if (block->contents_cached) {
+			bool run_abms = false;
+			for (content_t c : block->contents) {
```
sfan5: this could use a comment to easily see what it does

src/serverenvironment.cpp
```diff
@@ -818,6 +836,14 @@ class ABMHandler
 		{
 			const MapNode &n = block->getNodeUnsafe(p0);
 			content_t c = n.getContent();
+			if (!block->contents_cached && !block->do_not_cache_contents) {
+				block->contents.insert(c);
+				if (block->contents.size() > 32) {
```
sfan5: does it make sense for this to be configurable?

lhofhansl: IMHO. We should find a good default. 64 turns out to be almost universally good.

src/serverenvironment.cpp
lhofhansl: Yep!

lhofhansl:
With default settings and a mostly default world about 6ms are spent in ABMs without this PR, and about 1ms with this PR.
If I crank up active_object_range to 13, it's about 120-130ms without this PR, and about 30ms with this PR.

lhofhansl:
I will say that that was with a single player only (me). With default settings and 10 players we're talking 60ms just for the scanning when mostly nothing happens (vs less than 10ms with this PR).
So I think it's definitely worth it.

lhofhansl:
Let's get this in...?

rubenwardy:
The code looks good to me on inspection, but I'm not particularly comfortable with this area of code

src/mapblock.h
```diff
@@ -519,6 +521,9 @@ class MapBlock
 #ifndef SERVER // Only on client
 	MapBlockMesh *mesh = nullptr;
 #endif
+	std::unordered_set<content_t> contents;
+	bool contents_cached = false;
+	bool do_not_cache_contents = false;
```
SmallJoker: The variables should state that they're only for ABMs and not helpful in any other area.

src/serverenvironment.cpp
```diff
@@ -1311,8 +1345,12 @@ void ServerEnvironment::step(float dtime)
 				block->setTimestampNoChangedFlag(m_game_time);
 
 				/* Handle ActiveBlockModifiers */
-				abmhandler.apply(block);
+				abmhandler.apply(block, abm_checked, abm_run, abm_not_cached);
 			}
+			g_profiler->avg("SEnv: active blocks", m_active_blocks.m_abm_list.size());
+			g_profiler->avg("SEnv: active blocks cached", m_active_blocks.m_abm_list.size() - abm_not_cached);
```
SmallJoker: Line is too long (> 80 characters)

lhofhansl: > Why do you need the cache at all? A boolean value would do it too ...

Huh? That's the whole point of this :)
ABMs are per node-type, not per block. We also do not want to invalidate all blocks somehow when a new ABM is added. The way it is now the cache is only invalidated when the block is changed and on top the decision is block local. That makes this simple and safe.

SmallJoker: Alright, I didn't even expect that the ABMs are processed from scratch each second - thought they were a static vector of ABM trigger information per content ID.

Cazadorro:
What is ABM? what does it stand for?  Also why is `do_not_cache_contents == false` not `cache_contents = true`?

Cazadorro:
@rubenwardy  In addition to the forums and direct links to dev wiki being blocked, [I didn't realize this was domain specific to minetest]( http://lmgtfy.com/?q=abm+c%2B%2B), my bad.

src/serverenvironment.cpp
lhofhansl: @Cazadorro we can argue about the naming of the variables. :)
But please have a look at the code where they are used.

lhofhansl: @SmallJoker There're definitely more things to do. For example we could refresh the list of ABMs less frequently than we trigger them. In that case we could cache everything in a bit (as you suggest) until we refresh ABMs the next time.

Perhaps ABMs are not usually added often. So invalidating everything in that case might be OK. That would involve more complex logic with timestamps.

lhofhansl:
@Cazadorro No problem. Thanks for taking a look.

src/serverenvironment.cpp
```diff
@@ -818,6 +838,15 @@ class ABMHandler
 		{
 			const MapNode &n = block->getNodeUnsafe(p0);
 			content_t c = n.getContent();
+			// cache content types as we go
+			if (!block->contents_cached && !block->do_not_cache_contents) {
+				block->contents.insert(c);
+				if (block->contents.size() > 64) {
```
nerzhul: a thread local g_setting can be nice to use here

src/serverenvironment.cpp
```diff
@@ -800,11 +800,31 @@ class ABMHandler
 		return active_object_count;
 
 	}
-	void apply(MapBlock *block)
+	void apply(MapBlock *block, int &blocks_scanned, int &abms_run, int &blocks_cached)
 	{
 		if(m_aabms.empty() || block->isDummy())
 			return;
 
+		// check the content type cache first
```
paramat: 'Check'

src/serverenvironment.cpp
```diff
@@ -800,11 +800,31 @@ class ABMHandler
...
+		// to see whether there are any ABMs
+		// to be run at all for this block
```
paramat: Full stop at end of line.

src/mapblock.h
```diff
@@ -529,6 +531,14 @@ class MapBlock
 
 	static const u32 nodecount = MAP_BLOCKSIZE * MAP_BLOCKSIZE * MAP_BLOCKSIZE;
 
+	//// ABM optimizations ////
+	// cache of content types
```
paramat: For all comments: Capital letter at start.

src/serverenvironment.cpp
paramat: I don't see a need for a setting.

src/serverenvironment.cpp
```diff
@@ -818,6 +838,15 @@ class ABMHandler
 		{
 			const MapNode &n = block->getNodeUnsafe(p0);
 			content_t c = n.getContent();
+			// cache content types as we go
+			if (!block->contents_cached && !block->do_not_cache_contents) {
+				block->contents.insert(c);
+				if (block->contents.size() > 64) {
+					// too many different nodes... don't try to cache
```
paramat: 'Too'

paramat:
I studied the ABM code recently and this seems good, i'm close to approving.

src/serverenvironment.cpp
lhofhansl: It says "too". You mean upper case T?
It's not a full sentence, though.

src/serverenvironment.cpp
sfan5: Why not? If you have lots of RAM to spare, this would be one of the many small adjustments you could make to improve the server's responsiveness.

lhofhansl: Nah. This is just a heuristic, as long as node types for "most" blocks can be cached this is good.
A block can have at most 16^3 = 4096 different node types. Caching at most 1/64th seems good.

1. If we cache too much there's additional memory *and* additional CPU to check the cache first
2. The more different node types we have the more likely it becomes that ABMs need to be run for this block. And this check is entirely in vain.

16, 32, or 64 are good. Let's not add another config option to MT, it has *way* too many already.

src/serverenvironment.cpp
paramat: Yes capital 'T' please.

lhofhansl:
Thanks!
25cc5d1a3263aaca46f77f64b6cc20b1d2c32cac

src/serverenvironment.cpp
ClobberXD: on L805 too

PR #7557, opened 2018-07-15
Add support for per-player FOV overrides and multipliers

ClobberXD:
This PR aims to make the zoom feature more generic by allowing mods to individually override player FOVs with arbitrary values. Once this is merged, I'll work on making builtin zoom use these methods. Here are the changes:

- Add Lua API methods `ObjectRef:set_fov()`,  `ObjectRef:get_fov()`.
  - `set_fov` takes two params:
    - The FOV value.
    - A boolean to indicate whether the FOV value is a multiplier or not. Defaults to `false` - i.e. the value is an override by default.
  - `get_fov` returns a tuple containing the aforementioned data.
  - Passing 0 to `set_fov` will clear the FOV override.
- Add `PlayerFovSpec` struct, to store `f32 fov` and `bool is_multiplier`, and add getter and setter.
- Implement server->client FOV-sending using a new packet `TOCLIENT_FOV`.

****

Here's a mod to test out the new Lua API methods: **[fov_test](https://github.com/ClobberXD/fov_test)**. Check out the mod's README for the chat-commands' documentation.

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -5576,12 +5576,21 @@ This is basically a reference to a C++ `ServerActiveObject`
   `set_look_vertical`.
 * `set_look_yaw(radians)`: sets look yaw - Deprecated. Use
   `set_look_horizontal`.
-* `get_breath()`: returns players breath
-* `set_breath(value)`: sets players breath
+* `get_breath()`: returns player's breath
+* `set_breath(value)`: sets player's breath
     * values:
         * `0`: player is drowning
         * max: bubbles bar is not shown
         * See [Object properties] for more information
+* `set_fov(fov, is_multiplier)`: Sets player's FOV
+    * `fov`: FOV value.
+    * `is_multiplier`: Set to `true` if the FOV value is a multiplier.
+      Defaults to `false`.
+    * Set to 0 to clear FOV override.
+* `get_fov()`:
+    * Returns player's FOV override in degrees, and a boolean depending on whether
+      the value is a multiplier.
+    * Returns 0 as first value if player's FOV hasn't been overridden.
 * `set_attribute(attribute, value)`:  DEPRECATED, use get_meta() instead
     * Sets an extra attribute with value on player.
     * `value` must be a string, or a number which will be converted to a
--- a/src/client/camera.cpp
+++ b/src/client/camera.cpp
@@ -448,12 +448,26 @@ void Camera::update(LocalPlayer* player, f32 frametime, f32 busytime, f32 tool_r
 	if (m_camera_mode != CAMERA_MODE_FIRST)
 		m_camera_position = my_cp;
 
-	// Get FOV
+	/*
+	 * Apply server-sent FOV. If server doesn't enforce FOV,
+	 * check for zoom and set to zoom FOV.
+	 * Otherwise, default to m_cache_fov
+	 */
+
 	f32 fov_degrees;
-	// Disable zoom with zoom FOV = 0
-	if (player->getPlayerControl().zoom && player->getZoomFOV() > 0.001f) {
+	PlayerFovSpec fov_spec = player->getFov();
+	if (fov_spec.fov > 0.0f) {
+		// If server-sent FOV is a multiplier, multiply
+		// it with m_cache_fov instead of overriding
+		if (fov_spec.is_multiplier)
+			fov_degrees = m_cache_fov * fov_spec.fov;
+		else
+			fov_degrees = fov_spec.fov;
+	} else if (player->getPlayerControl().zoom && player->getZoomFOV() > 0.001f) {
+		// Player requests zoom, apply zoom FOV
 		fov_degrees = player->getZoomFOV();
 	} else {
+		// Set to client's selected FOV
 		fov_degrees = m_cache_fov;
 	}
 	fov_degrees = rangelim(fov_degrees, 1.0f, 160.0f);
--- a/src/network/clientopcodes.cpp
+++ b/src/network/clientopcodes.cpp
@@ -78,7 +78,7 @@ const ToClientCommandHandler toClientCommandTable[TOCLIENT_NUM_MSG_TYPES] =
 	{ "TOCLIENT_HP",                       TOCLIENT_STATE_CONNECTED, &Client::handleCommand_HP }, // 0x33
 	{ "TOCLIENT_MOVE_PLAYER",              TOCLIENT_STATE_CONNECTED, &Client::handleCommand_MovePlayer }, // 0x34
 	{ "TOCLIENT_ACCESS_DENIED_LEGACY",     TOCLIENT_STATE_NOT_CONNECTED, &Client::handleCommand_AccessDenied }, // 0x35
-	null_command_handler,
+	{ "TOCLIENT_FOV",                      TOCLIENT_STATE_CONNECTED, &Client::handleCommand_Fov }, // 0x36
 	{ "TOCLIENT_DEATHSCREEN",              TOCLIENT_STATE_CONNECTED, &Client::handleCommand_DeathScreen }, // 0x37
 	{ "TOCLIENT_MEDIA",                    TOCLIENT_STATE_CONNECTED, &Client::handleCommand_Media }, // 0x38
 	null_command_handler,
@@ -151,9 +151,9 @@ const ServerCommandFactory serverCommandFactoryTable[TOSERVER_NUM_MSG_TYPES] =
 	null_command_factory, // 0x14
 	null_command_factory, // 0x15
 	null_command_factory, // 0x16
-	{ "TOSERVER_MODCHANNEL_JOIN", 0, true }, // 0x17
-	{ "TOSERVER_MODCHANNEL_LEAVE", 0, true }, // 0x18
-	{ "TOSERVER_MODCHANNEL_MSG", 0, true }, // 0x19
+	{ "TOSERVER_MODCHANNEL_JOIN",    0, true }, // 0x17
+	{ "TOSERVER_MODCHANNEL_LEAVE",   0, true }, // 0x18
+	{ "TOSERVER_MODCHANNEL_MSG",     0, true }, // 0x19
 	null_command_factory, // 0x1a
 	null_command_factory, // 0x1b
 	null_command_factory, // 0x1c
--- a/src/network/clientpackethandler.cpp
+++ b/src/network/clientpackethandler.cpp
@@ -519,13 +519,22 @@ void Client::handleCommand_Movement(NetworkPacket* pkt)
 	player->movement_gravity                = g * BS;
 }
 
-void Client::handleCommand_HP(NetworkPacket* pkt)
+void Client::handleCommand_Fov(NetworkPacket *pkt)
 {
+	f32 fov;
+	bool is_multiplier;
+	*pkt >> fov >> is_multiplier;
 
+	LocalPlayer *player = m_env.getLocalPlayer();
+	player->setFov({ fov, is_multiplier });
+}
+
+void Client::handleCommand_HP(NetworkPacket *pkt)
+{
 	LocalPlayer *player = m_env.getLocalPlayer();
 	assert(player != NULL);
 
-	u16 oldhp   = player->hp;
+	u16 oldhp = player->hp;
 
 	u16 hp;
 	*pkt >> hp;
--- a/src/network/networkprotocol.h
+++ b/src/network/networkprotocol.h
@@ -199,6 +199,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 		Incremental inventory sending mode
 		Unknown inventory serialization fields no longer throw an error
 		Mod-specific formspec version
+		Player FOV override API
 */
 
 #define LATEST_PROTOCOL_VERSION 38
@@ -371,7 +372,13 @@ enum ToClientCommand
 		wstring reason
 	*/
 
-	TOCLIENT_PLAYERITEM = 0x36, // Obsolete
+	TOCLIENT_FOV = 0x36,
+	/*
+		Sends an FOV override/multiplier to client.
+
+		float fov
+		bool is_multiplier
+	*/
 
 	TOCLIENT_DEATHSCREEN = 0x37,
 	/*
--- a/src/network/serveropcodes.cpp
+++ b/src/network/serveropcodes.cpp
@@ -104,9 +104,9 @@ const ToServerCommandHandler toServerCommandTable[TOSERVER_NUM_MSG_TYPES] =
 	null_command_handler, // 0x4d
 	null_command_handler, // 0x4e
 	null_command_handler, // 0x4f
-	{ "TOSERVER_FIRST_SRP",          TOSERVER_STATE_NOT_CONNECTED, &Server::handleCommand_FirstSrp }, // 0x50
-	{ "TOSERVER_SRP_BYTES_A",        TOSERVER_STATE_NOT_CONNECTED, &Server::handleCommand_SrpBytesA }, // 0x51
-	{ "TOSERVER_SRP_BYTES_M",        TOSERVER_STATE_NOT_CONNECTED, &Server::handleCommand_SrpBytesM }, // 0x52
+	{ "TOSERVER_FIRST_SRP",                TOSERVER_STATE_NOT_CONNECTED, &Server::handleCommand_FirstSrp }, // 0x50
+	{ "TOSERVER_SRP_BYTES_A",              TOSERVER_STATE_NOT_CONNECTED, &Server::handleCommand_SrpBytesA }, // 0x51
+	{ "TOSERVER_SRP_BYTES_M",              TOSERVER_STATE_NOT_CONNECTED, &Server::handleCommand_SrpBytesM }, // 0x52
 };
 
 const static ClientCommandFactory null_command_factory = { "TOCLIENT_NULL", 0, false };
@@ -115,51 +115,51 @@ const ClientCommandFactory clientCommandFactoryTable[TOCLIENT_NUM_MSG_TYPES] =
 {
 	null_command_factory, // 0x00
 	null_command_factory, // 0x01
-	{ "TOCLIENT_HELLO",             0, true }, // 0x02
-	{ "TOCLIENT_AUTH_ACCEPT",       0, true }, // 0x03
-	{ "TOCLIENT_ACCEPT_SUDO_MODE",  0, true }, // 0x04
-	{ "TOCLIENT_DENY_SUDO_MODE",    0, true }, // 0x05
+	{ "TOCLIENT_HELLO",                    0, true }, // 0x02
+	{ "TOCLIENT_AUTH_ACCEPT",              0, true }, // 0x03
+	{ "TOCLIENT_ACCEPT_SUDO_MODE",         0, true }, // 0x04
+	{ "TOCLIENT_DENY_SUDO_MODE",           0, true }, // 0x05
 	null_command_factory, // 0x06
 	null_command_factory, // 0x07
 	null_command_factory, // 0x08
 	null_command_factory, // 0x09
-	{ "TOCLIENT_ACCESS_DENIED",     0, true }, // 0x0A
+	{ "TOCLIENT_ACCESS_DENIED",            0, true }, // 0x0A
 	null_command_factory, // 0x0B
 	null_command_factory, // 0x0C
 	null_command_factory, // 0x0D
 	null_command_factory, // 0x0E
 	null_command_factory, // 0x0F
-	{ "TOCLIENT_INIT",              0, true }, // 0x10
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
+	{ "TOCLIENT_INIT",                     0, true }, // 0x10
+	null_command_factory, // 0x11
+	null_command_factory, // 0x12
+	null_command_factory, // 0x13
+	null_command_factory, // 0x14
+	null_command_factory, // 0x15
+	null_command_factory, // 0x16
+	null_command_factory, // 0x17
+	null_command_factory, // 0x18
+	null_command_factory, // 0x19
+	null_command_factory, // 0x1A
+	null_command_factory, // 0x1B
+	null_command_factory, // 0x1C
+	null_command_factory, // 0x1D
+	null_command_factory, // 0x1E
+	null_command_factory, // 0x1F
 	{ "TOCLIENT_BLOCKDATA",                2, true }, // 0x20
 	{ "TOCLIENT_ADDNODE",                  0, true }, // 0x21
 	{ "TOCLIENT_REMOVENODE",               0, true }, // 0x22
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
+	null_command_factory, // 0x23
+	null_command_factory, // 0x24
+	null_command_factory, // 0x25
+	null_command_factory, // 0x26
 	{ "TOCLIENT_INVENTORY",                0, true }, // 0x27
-	null_command_factory,
+	null_command_factory, // 0x28
 	{ "TOCLIENT_TIME_OF_DAY",              0, true }, // 0x29
 	{ "TOCLIENT_CSM_RESTRICTION_FLAGS",    0, true }, // 0x2A
 	{ "TOCLIENT_PLAYER_SPEED",             0, true }, // 0x2B
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
+	null_command_factory, // 0x2C
+	null_command_factory, // 0x2D
+	null_command_factory, // 0x2E
 	{ "TOCLIENT_CHAT_MESSAGE",             0, true }, // 0x2F
 	null_command_factory, // 0x30
 	{ "TOCLIENT_ACTIVE_OBJECT_REMOVE_ADD", 0, true }, // 0x31
@@ -167,15 +167,15 @@ const ClientCommandFactory clientCommandFactoryTable[TOCLIENT_NUM_MSG_TYPES] =
 	{ "TOCLIENT_HP",                       0, true }, // 0x33
 	{ "TOCLIENT_MOVE_PLAYER",              0, true }, // 0x34
 	{ "TOCLIENT_ACCESS_DENIED_LEGACY",     0, true }, // 0x35
-	null_command_factory, // 0x36
+	{ "TOCLIENT_FOV",                      0, true }, // 0x36
 	{ "TOCLIENT_DEATHSCREEN",              0, true }, // 0x37
 	{ "TOCLIENT_MEDIA",                    2, true }, // 0x38
 	null_command_factory, // 0x39
-	{ "TOCLIENT_NODEDEF",                  0, true }, // 0x3a
-	null_command_factory, // 0x3b
-	{ "TOCLIENT_ANNOUNCE_MEDIA",           0, true }, // 0x3c
-	{ "TOCLIENT_ITEMDEF",                  0, true }, // 0x3d
-	null_command_factory,
+	{ "TOCLIENT_NODEDEF",                  0, true }, // 0x3A
+	null_command_factory, // 0x3B
+	{ "TOCLIENT_ANNOUNCE_MEDIA",           0, true }, // 0x3C
+	{ "TOCLIENT_ITEMDEF",                  0, true }, // 0x3D
+	null_command_factory, // 0x3E
 	{ "TOCLIENT_PLAY_SOUND",               0, true }, // 0x3f
 	{ "TOCLIENT_STOP_SOUND",               0, true }, // 0x40
 	{ "TOCLIENT_PRIVILEGES",               0, true }, // 0x41
@@ -203,12 +203,12 @@ const ClientCommandFactory clientCommandFactoryTable[TOCLIENT_NUM_MSG_TYPES] =
 	{ "TOCLIENT_MODCHANNEL_MSG",           0, true }, // 0x57
 	{ "TOCLIENT_MODCHANNEL_SIGNAL",        0, true }, // 0x58
 	{ "TOCLIENT_NODEMETA_CHANGED",         0, true }, // 0x59
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
-	null_command_factory,
+	null_command_factory, // 0x5A
+	null_command_factory, // 0x5B
+	null_command_factory, // 0x5C
+	null_command_factory, // 0x5D
+	null_command_factory, // 0x5E
+	null_command_factory, // 0x5F
 	{ "TOSERVER_SRP_BYTES_S_B",            0, true }, // 0x60
 	{ "TOCLIENT_FORMSPEC_PREPEND",         0, true }, // 0x61
 };
--- a/src/script/lua_api/l_object.cpp
+++ b/src/script/lua_api/l_object.cpp
@@ -1249,6 +1249,37 @@ int ObjectRef::l_set_look_yaw(lua_State *L)
 	return 1;
 }
 
+// set_fov(self, degrees[, is_multiplier])
+int ObjectRef::l_set_fov(lua_State *L)
+{
+	NO_MAP_LOCK_REQUIRED;
+	ObjectRef *ref = checkobject(L, 1);
+	RemotePlayer *player = getplayer(ref);
+	if (!player)
+		return 0;
+
+	player->setFov({ static_cast<f32>(luaL_checknumber(L, 2)), readParam<bool>(L, 3) });
+	getServer(L)->SendPlayerFov(player->getPeerId());
+
+	return 0;
+}
+
+// get_fov(self)
+int ObjectRef::l_get_fov(lua_State *L)
+{
+	NO_MAP_LOCK_REQUIRED;
+	ObjectRef *ref = checkobject(L, 1);
+	RemotePlayer *player = getplayer(ref);
+	if (!player)
+		return 0;
+
+	PlayerFovSpec fov_spec = player->getFov();
+	lua_pushnumber(L, fov_spec.fov);
+	lua_pushboolean(L, fov_spec.is_multiplier);
+
+	return 2;
+}
+
 // set_breath(self, breath)
 int ObjectRef::l_set_breath(lua_State *L)
 {
@@ -1962,6 +1993,8 @@ luaL_Reg ObjectRef::methods[] = {
 	luamethod(ObjectRef, set_look_vertical),
 	luamethod(ObjectRef, set_look_yaw),
 	luamethod(ObjectRef, set_look_pitch),
+	luamethod(ObjectRef, get_fov),
+	luamethod(ObjectRef, set_fov),
 	luamethod(ObjectRef, get_breath),
 	luamethod(ObjectRef, set_breath),
 	luamethod(ObjectRef, get_attribute),
--- a/src/script/lua_api/l_object.h
+++ b/src/script/lua_api/l_object.h
@@ -215,6 +215,9 @@ private:
 	// add_player_velocity(self, {x=num, y=num, z=num})
 	static int l_add_player_velocity(lua_State *L);
 
+	// get_fov(self)
+	static int l_get_fov(lua_State *L);
+
 	// get_look_dir(self)
 	static int l_get_look_dir(lua_State *L);
 
@@ -231,6 +234,9 @@ private:
 
 	// get_look_yaw2(self)
 	static int l_get_look_horizontal(lua_State *L);
+
+	// set_fov(self, degrees, is_multiplier)
+	static int l_set_fov(lua_State *L);
 
 	// set_look_vertical(self, radians)
 	static int l_set_look_vertical(lua_State *L);
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -1861,6 +1861,16 @@ void Server::SendMovePlayer(session_t peer_id)
 	Send(&pkt);
 }
 
+void Server::SendPlayerFov(session_t peer_id)
+{
+	NetworkPacket pkt(TOCLIENT_FOV, 4 + 1, peer_id);
+
+	PlayerFovSpec fov_spec = m_env->getPlayer(peer_id)->getFov();
+	pkt << fov_spec.fov << fov_spec.is_multiplier;
+
+	Send(&pkt);
+}
+
 void Server::SendLocalPlayerAnimations(session_t peer_id, v2s32 animation_frames[4],
 		f32 animation_speed)
 {
--- a/src/client/client.h
+++ b/src/client/client.h
@@ -193,6 +193,7 @@ public:
 	void handleCommand_ActiveObjectRemoveAdd(NetworkPacket* pkt);
 	void handleCommand_ActiveObjectMessages(NetworkPacket* pkt);
 	void handleCommand_Movement(NetworkPacket* pkt);
+	void handleCommand_Fov(NetworkPacket *pkt);
 	void handleCommand_HP(NetworkPacket* pkt);
 	void handleCommand_Breath(NetworkPacket* pkt);
 	void handleCommand_MovePlayer(NetworkPacket* pkt);
--- a/src/client/game.cpp
+++ b/src/client/game.cpp
@@ -1083,7 +1083,7 @@ void Game::run()
 		//    RenderingEngine::run() from this iteration
 		//  + Sleep time until the wanted FPS are reached
 		limitFps(&draw_times, &dtime);
-		
+
 		// Prepare render data for next iteration
 
 		updateStats(&stats, draw_times, dtime);
@@ -2332,7 +2332,7 @@ void Game::toggleFullViewRange()
 void Game::checkZoomEnabled()
 {
 	LocalPlayer *player = client->getEnv().getLocalPlayer();
-	if (player->getZoomFOV() < 0.001f)
+	if (player->getZoomFOV() < 0.001f || player->getFov().fov > 0.0f)
 		m_game_ui->showTranslatedStatusText("Zoom currently disabled by game or mod");
 }
 
--- a/src/player.h
+++ b/src/player.h
@@ -32,6 +32,12 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #define PLAYERNAME_ALLOWED_CHARS "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-_"
 #define PLAYERNAME_ALLOWED_CHARS_USER_EXPL "'a' to 'z', 'A' to 'Z', '0' to '9', '-', '_'"
 
+struct PlayerFovSpec
+{
+	f32 fov;
+	bool is_multiplier;
+};
+
 struct PlayerControl
 {
 	PlayerControl() = default;
@@ -178,6 +184,16 @@ public:
 	void setWieldIndex(u16 index);
 	u16 getWieldIndex() const { return m_wield_index; }
 
+	void setFov(const PlayerFovSpec &spec)
+	{
+		m_fov_spec = spec;
+	}
+
+	const PlayerFovSpec &getFov() const
+	{
+		return m_fov_spec;
+	}
+
 	u32 keyPressed = 0;
 
 	HudElement* getHud(u32 id);
@@ -187,10 +203,12 @@ public:
 
 	u32 hud_flags;
 	s32 hud_hotbar_itemcount;
+
 protected:
 	char m_name[PLAYERNAME_SIZE];
 	v3f m_speed;
 	u16 m_wield_index = 0;
+	PlayerFovSpec m_fov_spec = { 0.0f, false };
 
 	std::vector<HudElement *> hud;
 private:
--- a/src/server.h
+++ b/src/server.h
@@ -336,6 +336,7 @@ public:
 	void SendInventory(PlayerSAO *playerSAO, bool incremental);
 	void SendMovePlayer(session_t peer_id);
 	void SendPlayerSpeed(session_t peer_id, const v3f &added_vel);
+	void SendPlayerFov(session_t peer_id);
 
 	void sendDetachedInventories(session_t peer_id, bool incremental);
 
@@ -515,7 +516,6 @@ private:
 	/*
 		Variables
 	*/
-
 	// World directory
 	std::string m_path_world;
 	// Subgame specification
@@ -575,7 +575,6 @@ private:
 	/*
 		Threads
 	*/
-
 	// A buffer for time steps
 	// step() increments and AsyncRunStep() run by m_thread reads it.
 	float m_step_dtime = 0.0f;
@@ -590,14 +589,14 @@ private:
 	/*
 		Time related stuff
 	*/
-
 	// Timer for sending time of day over network
 	float m_time_of_day_send_timer = 0.0f;
 	// Uptime of server in seconds
 	MutexedVariable<double> m_uptime;
+
 	/*
-	 Client interface
-	 */
+	 	Client interface
+	*/
 	ClientInterface m_clients;
 
 	/*


ClobberXD:
It did not occur to me that adding extra white-spaces to align lines in a nicer way will make LINT unhappy... 

Fixed. :)

lhofhansl:
How should be integrate this best with the zoom stuff?
Perhaps we simply allow to set the fov, and zoom is just a boolean controlling whether the server is allow to retrieve more distant blocks...?

ClobberXD:
That sounds like a good idea, since there are situations where only the fov change is required, and loading distant blocks is unnecessary...

paramat:
> That sounds like a good idea, since there are situations where only the fov change is required, and loading distant blocks is unnecessary...

I agree, you may want to set a very narrow FOV for an effect but not want world 2000 nodes away being generated and/or loaded.

TMcSquared:
I just had a thought, someone could use this feature to animate the FOV for a nausea-type effect

ClobberXD:
Just a _gentle_ reminder: The PR is still WIP since the documentation hasn't been updated yet, but the actual code is complete. I can start working on the documentation only when the code has been approved, and nothing needs to be majorly modified. :)

ClobberXD:
Has anyone reviewed this PR yet? Is there anything that needs to be changed in a major way? I'll start adding the documentation now; if something needs change, will update the relevant docs accordingly...

ClobberXD:
Updated `lua_api.txt`, but I'm still not sure about whether or not to bump the latest protocol version. I'm sure the minimum version needn't be bumped since only new packets have been added, preserving backwards compat. with version 36.

doc/lua_api.txt
```diff
@@ -4883,12 +4883,18 @@ This is basically a reference to a C++ `ServerActiveObject`
...
-* `set_breath(value)`: sets players breath
+* `get_breath()`: returns player's breath
+* `set_breath(value)`: sets player's breath
     * values:
         * `0`: player is drowning
         * max: bubbles bar is not shown
         * See Object properties for more information
+* `set_fov(fov)`: sets player's FOV
+    * `fov`: Accepts values from 1 to 160, in degrees.
+    * Use `clear_fov` to reset player's FOV to its default
```
SmallJoker: Why not just specify that `0` (or negative values) will clear the FOV overwrite? `clear_fov` will be superfluous in that case.

src/camera.cpp
```diff
@@ -454,7 +454,12 @@ void Camera::update(LocalPlayer* player, f32 frametime, f32 busytime, f32 tool_r
 	if (player->getPlayerControl().zoom && player->getZoomFOV() > 0.001f) {
 		fov_degrees = player->getZoomFOV();
 	} else {
-		fov_degrees = m_cache_fov;
+		// Default to m_cache_fov if server-sent FOV == 0
+		f32 fov_received = player->getFOV();
+		if (!fov_received)
```
SmallJoker: Please compare against `== 0` since it's a float. It would be fine for integers but here it might show possible imprecision issues.

src/client.h
```diff
@@ -193,6 +193,7 @@ class Client : public con::PeerHandler, public InventoryManager, public IGameDef
 	void handleCommand_ActiveObjectRemoveAdd(NetworkPacket* pkt);
 	void handleCommand_ActiveObjectMessages(NetworkPacket* pkt);
 	void handleCommand_Movement(NetworkPacket* pkt);
+	void handleCommand_Fov(NetworkPacket* pkt);
```
SmallJoker: Wrong pointer placement.

src/network/clientopcodes.cpp
```diff
@@ -31,7 +31,7 @@ const ToClientCommandHandler toClientCommandTable[TOCLIENT_NUM_MSG_TYPES] =
 	{ "TOCLIENT_ACCEPT_SUDO_MODE",        TOCLIENT_STATE_CONNECTED, &Client::handleCommand_AcceptSudoMode}, // 0x04
 	{ "TOCLIENT_DENY_SUDO_MODE",          TOCLIENT_STATE_CONNECTED, &Client::handleCommand_DenySudoMode}, // 0x05
 	null_command_handler, // 0x06
-	null_command_handler, // 0x07
+	{ "TOCLIENT_FOV",					  TOCLIENT_STATE_CONNECTED, &Client::handleCommand_Fov }, // 0x07
```
SmallJoker: Wrong indents. Please line up with spaces like above.

src/network/clientpackethandler.cpp
```diff
@@ -488,13 +488,21 @@ void Client::handleCommand_Movement(NetworkPacket* pkt)
 	player->movement_gravity                = g * BS;
 }
 
-void Client::handleCommand_HP(NetworkPacket* pkt)
+void Client::handleCommand_Fov(NetworkPacket* pkt)
```
SmallJoker: Wrong pointer placement.

src/network/networkprotocol.h
```diff
@@ -246,6 +246,12 @@ enum ToClientCommand
 	/*
 		Signals client that sudo mode auth failed.
 	*/
+	TOCLIENT_FOV = 0x07,
+	/*
+		Sends an FOV value to client.
```
SmallJoker: `Send a FOV value to the player`

src/network/serveropcodes.cpp
```diff
@@ -101,12 +101,12 @@ const ToServerCommandHandler toServerCommandTable[TOSERVER_NUM_MSG_TYPES] =
...
-	null_command_handler, // 0x4d
+	null_command_handler, // 0x4d	
 	null_command_handler, // 0x4e
 	null_command_handler, // 0x4f
-	{ "TOSERVER_FIRST_SRP",          TOSERVER_STATE_NOT_CONNECTED, &Server::handleCommand_FirstSrp }, // 0x50
-	{ "TOSERVER_SRP_BYTES_A",        TOSERVER_STATE_NOT_CONNECTED, &Server::handleCommand_SrpBytesA }, // 0x51
-	{ "TOSERVER_SRP_BYTES_M",        TOSERVER_STATE_NOT_CONNECTED, &Server::handleCommand_SrpBytesM }, // 0x52
+	{ "TOSERVER_FIRST_SRP",          		TOSERVER_STATE_NOT_CONNECTED, &Server::handleCommand_FirstSrp }, // 0x50
+	{ "TOSERVER_SRP_BYTES_A",        		TOSERVER_STATE_NOT_CONNECTED, &Server::handleCommand_SrpBytesA }, // 0x51
+	{ "TOSERVER_SRP_BYTES_M",        		TOSERVER_STATE_NOT_CONNECTED, &Server::handleCommand_SrpBytesM }, // 0x52
```
SmallJoker: Please revert these changes, same for L104

src/network/serveropcodes.cpp
```diff
@@ -120,7 +120,7 @@ const ClientCommandFactory clientCommandFactoryTable[TOCLIENT_NUM_MSG_TYPES] =
 	{ "TOCLIENT_ACCEPT_SUDO_MODE",  0, true }, // 0x04
 	{ "TOCLIENT_DENY_SUDO_MODE",    0, true }, // 0x05
 	null_command_factory, // 0x06
-	null_command_factory, // 0x07
+	{ "TOCLIENT_FOV",				0, true }, // 0x07
```
SmallJoker: Wrong indents. Please line up like above.

src/remoteplayer.h
```diff
@@ -101,6 +101,9 @@ class RemotePlayer : public Player
 		*clouds = m_sky_clouds;
 	}
 
+	void setFov(const f32 fov) { m_fov = fov; }
+	const f32 getFov() const { return m_fov; }
```
SmallJoker: These don't seem to be necessary. Just pass the FOV value to the network packet sending function (like breath, HP).

src/script/lua_api/l_object.cpp
```diff
@@ -1174,6 +1174,47 @@ int ObjectRef::l_set_look_yaw(lua_State *L)
 	return 1;
 }
 
+// clear_fov(self)
+int ObjectRef::l_clear_fov(lua_State *L)
+{
+	NO_MAP_LOCK_REQUIRED;
+	ObjectRef *ref = checkobject(L, 1);
+	RemotePlayer *player = getplayer(ref);
+	if (player == NULL) return 0;
```
SmallJoker:
``` C++
if (!player)
	return 0;
```
Same for all other touched and new functions.

src/script/lua_api/l_object.cpp
```diff
@@ -1174,6 +1174,47 @@ int ObjectRef::l_set_look_yaw(lua_State *L)
...
+
+	f32 fov = luaL_checknumber(L, 2);
+	getServer(L)->setPlayerFov(player, fov);
+
+	return 1;
```
SmallJoker: Does not return anything. Return 0.

doc/lua_api.txt
paramat: I agree, 0 is good.

paramat:
Where is the option for controlling zoom causing distant-world loading?

ClobberXD:
@paramat This does not cause distant-world loading at all. _Only_ the FOV is changed. Should this behaviour be changed? If yes, how?

src/network/networkprotocol.h
ClobberXD: `an FOV value` is the phonetically correct alternative. Also, L247 right above also uses present tense (`Signals`) instead of an imperative tone (`Signal`). Should that be changed too?

src/client.h
ClobberXD: Shall I replace all occurrences in that file?

doc/lua_api.txt
ClobberXD: Alright, removed `clear_fov`

src/script/lua_api/l_object.cpp
ClobberXD: I wasn't sure about what the return value was supposed to be. Fixed, thanks

src/remoteplayer.h
ClobberXD: Only server->client FOV sending has been implemented, which is why I added these to `RemotePlayer` to keep track of the player's FOV.

ClobberXD: Adding a client->server FOV sending will require another server->client "request" packet, which is why it was not implemented

src/network/serveropcodes.cpp
ClobberXD: Indents fixed

src/network/networkprotocol.h
SmallJoker: Well, then. Keep it as-is.

paramat: "A field of view" is correct, using 'an' with a non-vowel is bad English.

paramat: "Signals to client that sudo mode auth failed." is clearer.

paramat:
Ok i see. It's ok then. I see no need for this to cause distant world loading.

src/network/networkprotocol.h
ClobberXD: Sorry for this minor nit-pick, but articles like 'a' and 'an' are used on the basis of phonetic vowels. In the case, "FOV" sounds like _eff-oh-vee_ and since the first sound is a vowel sound, 'an' is used here. :)

e.g. "I'll reach my home _an_ hour later"

ClobberXD: See [this relevant question](https://ell.stackexchange.com/questions/86665/use-of-an-for-a-word-not-starting-with-a-vowel) on ell.stackexchange.com

src/remoteplayer.h
```diff
@@ -159,6 +162,7 @@ class RemotePlayer : public Player
 	float m_chat_message_allowance = 5.0f;
 	u16 m_message_rate_overhead = 0;
 
+	f32 m_fov;
```
SmallJoker: Needs initializing. How about moving this property to `class Player`?

src/script/lua_api/l_object.cpp
```diff
@@ -212,8 +217,10 @@ int ObjectRef::l_right_click(lua_State *L)
 	ObjectRef *ref2 = checkobject(L, 2);
 	ServerActiveObject *co = getobject(ref);
 	ServerActiveObject *co2 = getobject(ref2);
-	if (co == NULL) return 0;
-	if (co2 == NULL) return 0;
+	if (!co)
+		return 0;
+	if (!co)
```
SmallJoker: `if (!co2)` or revert these changes.

src/script/lua_api/l_object.cpp
```diff
@@ -1167,20 +1202,52 @@ int ObjectRef::l_set_look_yaw(lua_State *L)
...
+// get_fov(self)
+int ObjectRef::l_get_fov(lua_State *L)
+{
+	NO_MAP_LOCK_REQUIRED;
+	ObjectRef *ref = checkobject(L, 1);
+	RemotePlayer *player = getplayer(ref);
+	if (!player)
+		return 0;
+
+	f32 fov = getServer(L)->getPlayerFov(player);
```
SmallJoker: Useless function. Replace with `player->getFov();`

src/server.cpp
```diff
@@ -2922,6 +2930,19 @@ std::set<std::string> Server::getPlayerEffectivePrivs(const std::string &name)
 	return privs;
 }
 
+void Server::setPlayerFov(RemotePlayer *player, const f32 fov)
```
SmallJoker: Please use `player->setFov(fov); SendPlayerFov(player->getPeerId());` like it's done similar with other setter&send functions (`set_pos`, `set_hp`).

src/script/lua_api/l_object.cpp
SmallJoker: Yes, a separate cleanup PR would be nice.

ClobberXD:
Addressed comments; rebased

ClobberXD:
Keeping in mind the recent discussions on Degrees vs Radians, I suggest this stays as-is, since unlike vectors and rotations, degrees is the most natural way to express FOV.

paramat:
Yes i agree with degrees for this, zoom FOV and zoom settings are also degrees.

Wuzzy2:
I don't get this PR. This would mean that FOV is no longer a client setting, only the server is supposed to get control over your FOV?

Also, this would be useless if someone use a hacked, err, modified client to set whatever FOV they like. :D

ClobberXD:
Using the provided API methods, mods would be able to **override** client's FOV (set by the client itself). Passing a value of 0 to `set_fov` would nullify the override, and the FOV is reset to client's original value

ClobberXD:
@SmallJoker You have done some considerable amount of reviewing to this PR. Would you like to suggest any more changes/improvements?

ClobberXD:
Rebased :)

SmallJoker: Note that the `(var == 0)` --> `(!var)`, incorrect pointer placements, and other such code-style issues exist throughout the touched file(s), and you asked me to fix them in a later PR.

ClobberXD:
#5797 would be the best solution, but considering that the feature would need SSCSM to actually be useful...

paramat:
>  using modifiers here doesn't seem to be a good idea.

Agreed, modifiers would be silly.

ClobberXD:
I did see that message, and I thought I'll fix it in a follow-up PR which would just fix code-style in a couple of files. But I understand that the touched code should follow the correct code-style, even if the rest of the file currently don't. Thanks for pointing that out :)

doc/lua_api.txt
```diff
@@ -5119,12 +5119,16 @@ This is basically a reference to a C++ `ServerActiveObject`
...
   `set_look_horizontal`.
-* `get_breath()`: returns players breath
-* `set_breath(value)`: sets players breath
+* `get_breath()`: returns player's breath
+* `set_breath(value)`: sets player's breath
     * values:
         * `0`: player is drowning
         * max: bubbles bar is not shown
-        * See [Object properties] for more information
+        * See Object Properties for more information
```
pauloue: Capitalization should match header (lowercase properties)

src/network/clientpackethandler.cpp
```diff
@@ -515,7 +515,7 @@ void Client::handleCommand_Movement(NetworkPacket* pkt)
 	player->movement_gravity                = g * BS;
 }
 
-void Client::handleCommand_Fov(NetworkPacket* pkt)
+void Client::handleCommand_Fov(NetworkPacket *pkt)
```
pauloue: You could make the same change in the header file

doc/lua_api.txt
ClobberXD: Oops, that happened while fixing a merge conflict. Good catch.

src/network/clientpackethandler.cpp
ClobberXD: Done

ClobberXD:
Fixed pointer placement in `client.h`, improved `get_fov` documentation, and squashed.

ClobberXD:
This PR intends to allow the Lua API to override the FOV of players, in a consistent manner. From that perspective, overrides themselves are inconsistent, because we're essentially trying to force a fixed FOV value on all clients, irrespective of their default FOV. e.g. a mod sets FOV of all clients to 90. But this value can be more zoomed for some, and more wide-angled for others (like me - my default FOV is 130).

With that in mind, I feel overrides should be replaced by modifiers. If this sounds good, we can decide on the implementation.

@SmallJoker Please let me know what you're opinion is. I'm proposing a major change in the PR after you've approved; so, apologies.

paramat:
> because we're essentially trying to force a fixed FOV value on all clients, irrespective of their default FOV.

The mod can get the player's FOV first to decide what to do.
> With that in mind, I feel overrides should be replaced by modifiers.

:-1: :-1: :-1: no need, overkill and complex. Most of the MT API sets fixed values, please let's not start adding combinable modifiers to everything, this is insane.

paramat:
Mod-set player FOV is a low priority feature, so doesn't justify a complex implementation.

Most uses of this feature will want to set a fixed FOV for certain intended effects, not multiply a player's existing FOV by using modifiers.
FOV is a player preference or a specific special effect, it is not something like physics overrides where potions are consumed to multiply a player aspect by a certain amount.
> This PR intends to allow the Lua API to override the FOV of players, in a consistent manner. From that perspective, overrides themselves are inconsistent, because we're essentially trying to force a fixed FOV value on all clients, irrespective of their default FOV

This logic can be applied to many APIs, and would lead to combinable modifiers for many APIs, which we cannot do, and is unnecessary, almost all of the APIs work fine as they are.
There has to be a critically important need to use combinable modifiers.

Overriding FOV consistently (that is, mulitplying the player's FOV instead of setting a fixed value) shouldn't be the intention here, and is not needed.

Please keep things simple and consider the bugs caused and maintenance required. Please read both of celeron55's roadmaps if you haven't already (linked in the 'contributing' document). Simplicity is a primary philosophy of MT.
We also recently, generally, have fewer active core devs than ever, and each of those have less free time than ever. We are overwhelmed and already can't keep up with contributions. We have to be very careful about adding completely unnecessary fancy features and luxuries.

ClobberXD:
@paramat: Understood. I'll drop the idea of modifiers, and leave it to the mods to come up with their own workarounds. :)

> The mod can get the player's FOV first to decide what to do.

That isn't currently possible with this PR. The `ObjectRef::get_fov` method returns 0 if the player's FOV hasn't been overridden. (See https://github.com/minetest/minetest/pull/7557/files#diff-1305560bd8befb32862f0feeefabd02eR5132).

I think the method should return two values:

- The current FOV of player, irrespective of whether it has been overridden.
- A boolean that return `true` if the FOV has been overridden (otherwise, mods have no way of knowing this).

ClobberXD:
Getting the default FOV requires the client to send the value over to the server on join. Another packet, yay! :)

But I consider this feature incomplete without the client->server packet - the server must know the client's FOV at any given time which will ensure that this feature is quite robust.

ClobberXD:
I'm very much against adding another new packet just to report the client's FOV to the server on join. An obvious alternative would be to send the FOV along with an existing packet. Please let me know which is the best choice among the following to send the FOV with:

- `TOSERVER_INIT`
- `TOSERVER_INIT2`
- `TOSERVER_CLIENT_READY`

`TOSERVER_CLIENT_READY` is probably the way to go, IMHO, as it isn't too early and is sent after the server-client connection has been established (I think). But do let me know whether I'm wrong, or whether there are better options than the aforementioned packets.

nerzhul:
no, CLIENT_READY is here just for session change. Don't mix the attributes.

ClobberXD:
Ok, thanks for the info. Any other packets to which the FOV value can be added to?

paramat:
> That isn't currently possible with this PR. The ObjectRef::get_fov method returns 0 if the player's FOV hasn't been overridden.

Hmm that's unfortunate. But i don't think it's essential, setting a fixed FOV regardless of client default FOV is fine.
> I'm very much against adding another new packet just to report the client's FOV to the server on join.

I agree, that's unnecessary overkill.
> An obvious alternative would be to send the FOV along with an existing packet

:-1: unnecessary and messy.
From IRC:
> Setting a client's FOV to a fixed value without taking their current FOV into consideration won't be of much help in providing a uniform experience to all clients.

A uniform experience for all clients is acheived by setting a fixed value on all clients.
> there's no way to provide the experience of a 4x scope (for example) to all players in a consistent, uniform manner.

This approach scopes or telescopes doesn't make sense. They should not multiply the player's default zoom, they should be a fixed experience for all players.
A player's default FOV is something independent for general gameplay. Once a player uses a gun scope or a telescope or binoculars, that should act the same way for all players, same field of view, same magnification. Because that's how they act in real life, every human will experience the same field of view and magnification.
For special effects, the modder will want the effect to be the same for all players.

So far i can see no significant need for mod-set FOV to modify player's FOV or even know what it is, so a simple implementation without client to server packets or modifiers is fine.

What sfan5 wrote on IRC today is nonsense:
We can't add modifiers to many APIs, and they are unnecessary for almost all APIs. So far only physics overrides is known to need them for certain. Maybe one or two more may benefit, but most of the APIs work fine.
Every added feature adds bugs and requires constant maintenance.
We shouldn't add completely unnecessary features, this is obvious.

paramat:
Hmm i just thought of something.

Lets say a mod adds guns with magnifying scopes. Smaller FOV (more magnification) creates more advantage for a player. So if scope FOV depends on player default FOV you end up with players with unmatched advantages. Some players will intentionally adjust their default FOV to get a higher power scope.
This will enable client cheating and exploits through adjustment of default FOV.

So setting a fixed value for all players will be necessary for many uses, i think most uses will require fixed. Are there many uses that require modifiers? Let's not assume, there has to be an actual need to justify doing this.

So if modifiers are added there still has to be a way to set a fixed value, especially because of the below.
> Once this is merged, I'll work on making builtin zoom use these methods instead

Zoom loads distant world, so it is even more critical that no player can gain advantage through adjusting their default FOV and therefore zoom FOV.
Binoculars and telescope items are meant to provide a fixed FOV, fixed field of view, fixed magnification (for equal benefit for all players) and a fixed distant world loading range, so will need to set a fixed FOV for all players.
I also can't see any need for modifiers for zoom FOV.

ClobberXD:
> A uniform experience for all clients is acheived by setting a fixed value on all clients.

Providing a uniform experience for all clients can't be achieved by setting a fixed value - the resultant  FOV has to be a product of default FOV and a constant. As I said in https://github.com/minetest/minetest/pull/7557#issuecomment-477587657:

> e.g. a mod sets FOV of all clients to 90. But this value can be zoomed in for some, and zoomed out for others (like me - my default FOV is 130).

paramat:
> Providing a uniform experience for all clients can't be achieved by setting a fixed value - the resultant FOV has to be a product of default FOV and a constant. As I said in

You're repeating what you stated, but i've explained clearly why, for most uses, fixed is what will be wanted and needed. Gun scopes, binoculars and telescopes have fixed field of view, fixed magnification, they cannot multiply a player's FOV because that gives certain players unfair advantages. You would be enabling clientside cheating.
>  a mod sets FOV of all clients to 90. But this value can be zoomed in for some, and zoomed out for others

That is a usecase that requires a FOV multiplier, but is a low priority usecase of, for example, a 'potion that increases FOV', such a thing won't be used much, it's a bizarre usecase. Setting FOV will be used mostly for usecases which require a fixed FOV.
The usecases that require multiplication of player FOV are, i think, too low priority to justify the complexity and maintenance.
Your logic is based on a bizarre usecase and you're not addressing what i explained or the more likely usecases.

At the least, the ability to set a fixed FOV will be absolutely essential.

ClobberXD:
> Gun scopes, binoculars and telescopes have fixed field of view, fixed magnification, ...

And they're used by human eyes with a fixed client-side FOV, IRL.

> they cannot multiply a player's FOV because that gives certain players unfair advantages. You would be enabling clientside cheating.

I agree the player would have unfair zoom advantages if they set their FOV to a lower-than-normal value, but they'd also be at a severe disadvantage while not zooming. I think this should be left to the mods/games to decide. Also, their apparent short-term advantage while zooming also has no negative effect on the server as distant map generation/loading does not occur.

paramat:
> And they're used by human with a fixed client-side FOV, IRL.

That's illogical and irrelevant.
Just because a player has adjustable FOV doesn't mean a scope should multiply the player's FOV, there's no logic in that.
In-game, there's no point to a scope multiplying the player's FOV, because scopes should obviously have fixed abilities, as they do so in real life.
> I agree the player would have unfair zoom advantages if they set their FOV to a lower-than-normal value, but they'd also be at a severe disadvantage while not zooming

That doesn't make the first point ok, you can't ignore the cheating point in this way.
A player would temporarily set their FOV to a small or smaller value while using the feature, to gain the advantage, especially when, as might happen, we get in-game adjustable FOV.
Even without intentional changing of FOV, the normal differences in player FOV themselves would create unfair advantages when multiplied.
> I think this should be left to the mods/games to decide.

Yes this is fine of course, but to do so the option to set a fixed FOV has to be there. Are you planning this?
> Also, their apparent short-term advantage while zooming also has no negative effect on the server as distant map generation/loading does not occur.

Yes, distant-world loading doesn't occur, but distant world already loaded in the client (because the player travels) will still be viewable.
Viewing distant world isn't the only advantage, viewing what is loaded in the client at higher magnification is too.
///////////////////////////////

Please can you explain the usecases that require a FOV multiplier? So far none have been explained. Maybe there are some good ones i haven't thought of which might convince me.
Good usecases are required for a new feature.
///////////////////////////////

Also, i might have misunderstood:
> I feel overrides should be replaced by modifiers.

What is your intention? Do you mean a FOV multiplier that works as 'physics overrides' do now, or, combinable modifiers as proposed for physics overrides by raymoo in their PR?

Due to 'physics overrides' being as they are (multipliers) and 'modifiers' being proposed as combinable multipliers, the word 'modifier' suggested to me combinable multipliers as proposed for physics overrides by raymoo in their PR. Did i misunderstand? If so i'm sorry and i was over-concerned.
A non-combinable multiplier for player FOV may be fairly simple.

The ability to set a fixed value too is essential as explained. As long as this is possible, mods will of course use that to avoid cheating, then the cheating issue doesn't matter.
Setting a fixed value should be the primary feature as it has more usecases, using a multiplier is the secondary functionality.

ClobberXD:
> Do you mean a FOV multiplier that works as 'physics overrides' do now, or, combinable modifiers as proposed for physics overrides by raymoo in their PR?

Initially I did intend to implement combinable modifiers.

> A non-combinable multiplier for player FOV may be fairly simple.

Not what I intended, but this sounds like a great compromise between the two. The ability to set a fixed FOV would be left untouched, of course, and setting multipliers could be done by adding another method (e.g. `ObjectRef:set_fov_multiplier`). Multipliers also don't require the client to send their FOV to the server, adding to the mess of packets. I assume it's fairly simple to use the same packet and add another variable which would indicate whether the value is a fixed override or a multiplier.

The only small nit is to figure out how the fixed override and the multiplier would interact - would one of them be ignored if both are set? If so, which one?

paramat:
There's a good case for combinable modifiers for physics override, but i think FOV is of too low importance to justify that amount of complexity. An option to apply a non-combinable multiplier is ok for me.
> The ability to set a fixed FOV would be left untouched, of course

Awesome.
> how the fixed override and the multiplier would interact

It seems to me that the most recent setting of FOV, either absolute or a multiplier, has to be applied, erasing whatever came before, otherwise the feature won't work.
Absolute and multiplier won't be applied simulataneously, and if they are not simultaneous then they are possibly from different mods, if anything is ignored then the feature doesn't function.

ClobberXD:
> It seems to me that the most recent setting of FOV, either absolute or a multiplier, has to be applied, erasing whatever came before

Makes sense. On it.

ClobberXD:
Looks like I have to re-add the `clear_fov` API method because clearing server-sent FOV might not be as straight-forward using `set_fov` anymore. :)

ClobberXD:
Updated the PR to support FOV multipliers. Updated docs, the PR's first post, and my test mod `fov_test`.

FOV multipliers are implemented using pgimeno's formula (`2 * atan(tan(m_cache_fov / 2) * factor)`) instead of direct multiplication with `m_cache_fov`. See:

- http://irc.minetest.net/minetest-dev/2019-04-07#i_5525861
- http://irc.minetest.net/minetest-dev/2019-04-07#i_5526029

> Looks like I have to re-add the clear_fov API method

Not required, tbh - the second param is optional anyway.

paramat:
> FOV multipliers are implemented using pgimeno's formula (2 * atan(tan(m_cache_fov / 2) * factor)) instead of direct multiplication with m_cache_fov

Sorry, i forgot to comment on that earlier ... although technically correct, this is not a good implementation for the feature. The setting is FOV in degrees, not width of view frustrum, so the multiplication should be FOV in degrees too, otherwise this mixes 2 different measurements of FOV and it gets confusing.
All MT FOV units are angle in degrees, and modders work with that with no problems, so it makes no sense to start worrying about view frustrum width.
It also means extremely large values have to be used when dealing with a very wide FOV.
If FOV 180 degrees was possible, the view frustrum value would be infinite, this shows how FOV in degrees is better and simpler, and more intuitive to work with.
:-1:

ClobberXD:
Um, so... you want me to replace that formula with just `multiplier * m_cache_fov`... right?

paramat:
Yes.

paramat:
p_gimeno wrote:
> When you're talking zoom factor, you're not talking FOV angle factor, but projected image size factor, and that's calculated with the formula I posted.

> The formula I gave would yield a zoom factor (which is what FOV is all about) of 2.0x.

But we're not talking zoom factor or technically correct 'field of view', that's the initial assumption that has caused the unnecessary complexity. This is simply a multiplier for client view frustrum angle.
> Multiplying the angle by a factor is essentially useless for wide FOV angles, because the result is visually unpredictable. If a user has a wide FOV configured in their client, setting a multiplier of, say, 0.5 will have a radically different result than if they have a smaller FOV. It does not produce uniform results.

It is predictable obviously, because modders work with FOV as a view frustrum angle and will understand what will happen to various client FOVs when a simple multiplier is applied. They will expect and understand the 'radically different result'.
What makes it complex and confusing is changing all this to some idea of 'zoom factor' or technically correct 'field of view', when the underlying FOV is not that. The problem is the setting name 'FOV', it should have been called 'frustrum angle' or something.

Modders and players use and understand FOV in the way it already behaves, as a view frustrum angle, it makes no sense to make FOV behave differently when multipliers are applied.
Modders will complain: "Why does client FOV * 1.5 not behave the same as FOV 60 to FOV 90?".
> and have an extra function e.g. set_zoom that sets the zoom factor relative to the current setting.

This isn't about 'zoom factor' though.
Even our current zoom works with FOV being a view frustrum angle, not real-world technically correct 'field of view'.
ANAND wrote:
> I also can't come up with usecases for FOV multipliers apart from fixed zoom factor like weapon optics, binoculars, or telescopes

Weapon optics, binoculars, or telescopes do not apply a fixed zoom factor (i assume this means a multiplier), they apply an absolute FOV value obviously. This is your misunderstanding. I have a physics degree and have studied optics at degree level so i may have a good understanding.

The binoculars in MTG apply an absolute FOV of 10 degrees, which applies an absolute magnification.
In the real world, when you look through such optics, there is a fixed FOV, a fixed magnification, for all people independent of how wide a FOV their eyes may have when not looking through them. Everyone who looks through such optics see exactly the same FOV and magnification.

Also as i have explained, if such optics applied a multiplier to what the client is using, different player will have different abilities when using the optics, leading to unfairness and cheating.

So in fact FOV multipliers are a minority usecase, you still haven't explained a usecase for them, there probably is but it will be a minority. Absolute FOV is the primary uscase for weapon optics, binoculars, and telescopes.
p_gimeno wrote:
> made my calculations: from a 160° FOV to an 80° FOV the zoom is 6.75X. From 80° FOV to 40° FOV the zoom is 2.3x. The first zoom factor is hugely disproportionate.

The fact it is disproportionate is irrelevant, this is exactly what the modder is used to and expects, and they will alter their mod code to take this into account. Thinking about this as a 'zoom factor' is the mistake.

It's best that an engine API is simpler by moving the complexity into mods, because this way the API becomes more flexible for various uses, as well as reducing code bloat, maintenance and core dev work.

ClobberXD:
> Weapon optics, binoculars, or telescopes do not apply a fixed zoom factor (i assume this means a multiplier), they apply absolute FOV and magnification.

How can they apply absolute FOV _and_ magnification at the same time? I assume you mean absolute multipliers. And how is magnification different from zoom factor?

> The binoculars in MTG apply an absolute FOV of 10 degrees, which applies an absolute magnification.

That's bad...

> In the real world, when you look through such optics, there is a fixed FOV, a fixed magnification,

Fixed magnification, not fixed FOV. e.g. a 5x binocular provides 5x magnification, which would make the object appear 500% bigger.

> Absolute FOV is the primary uscase for weapon optics, binoculars, and telescopes.

Apologies, if I misunderstood, but how in the world do these use absolute FOV? They provide **fixed magnification**, like 5x, 10x. Would the final FOV not vary depending on the viewer?

ClobberXD:
Related: #4700

Btw, someone please review, approve _and_ merge ;)

doc/lua_api.txt
```diff
@@ -5260,12 +5260,21 @@ This is basically a reference to a C++ `ServerActiveObject`
...
         * max: bubbles bar is not shown
         * See [Object properties] for more information
+* `set_fov(fov, is_multiplier)`: Sets player's FOV
+    * `fov`: FOV value.
+    * `is_multiplier`: Set to `true` if the FOV value is a multiplier.
+      Defaults to `false`.
+    * Set to 0 to clear FOV override.
+* `get_fov()`:
+    * Returns player's FOV override in degrees, and a boolean depending on whether
+    the value is a multiplier.
```
```suggestion
      the value is a multiplier.
```

src/network/clientpackethandler.cpp
```diff
@@ -519,13 +519,22 @@ void Client::handleCommand_Movement(NetworkPacket* pkt)
...
 {
+	f32 fov;
+	bool is_multiplier;
+	*pkt >> fov >> is_multiplier;
+
+	LocalPlayer *player = m_env.getLocalPlayer();
+	player->setFov(fov, is_multiplier);
+}
 
+void Client::handleCommand_HP(NetworkPacket* pkt)
```
DS-Minetest: You could
```suggestion
void Client::handleCommand_HP(NetworkPacket *pkt)
```

src/player.cpp
```diff
@@ -74,6 +74,8 @@ Player::Player(const char *name, IItemDefManager *idef):
 
 	hud_hotbar_itemcount = HUD_HOTBAR_ITEMCOUNT_DEFAULT;
 
+	m_fov = 0.0;
```
DS-Minetest:
```suggestion
	m_fov = 0.0f;
```

src/client/camera.cpp
```diff
@@ -454,7 +454,19 @@ void Camera::update(LocalPlayer* player, f32 frametime, f32 busytime, f32 tool_r
 	if (player->getPlayerControl().zoom && player->getZoomFOV() > 0.001f) {
 		fov_degrees = player->getZoomFOV();
 	} else {
-		fov_degrees = m_cache_fov;
+		// Apply server-sent FOV. If server doesn't
+		// enforce client's FOV, default to m_cache_fov
+		auto fov_spec = player->getFov();
+		if (fov_spec.first > 0) {
```
DS-Minetest:
```suggestion
		if (fov_spec.first > 0.0f) {
```

src/script/lua_api/l_object.h
```diff
@@ -229,6 +232,9 @@ class ObjectRef : public ModApiBase {
 	// get_look_yaw2(self)
 	static int l_get_look_horizontal(lua_State *L);
 
+	// set_fov(self, degrees)
```
DS-Minetest:
```suggestion
	// set_fov(self, degrees, is_multiplier)
```
or
```suggestion
	// set_fov(self, degrees, bool)
```
? Idk.

src/player.cpp
ClobberXD: As per C++11 standards, `m_fov` should actually be defined inline, as the value is a constant.

ClobberXD: Fixed, thanks

ClobberXD:
Rebased.

paramat:
> How can they apply absolute FOV and magnification at the same time?

> Fixed magnification, not fixed FOV

They simply do. In the real world, anyone who looks through optics sees the same field of view (however that is defined) and the same magnification.
> That's bad...

No, see above, it's realistic. Also because anything else would, as explained multiple times, allow cheating and cause differing advantages for different players. You are ignoring and dismissing this important issue.
> They provide fixed magnification, like 5x, 10x. Would the final FOV not vary depending on the viewer?

No, because in the real world humans cannot alter the density of rods and cones on their retina at will. So in MT, to be realistic (and to avoid cheating and unfairness), you cannot use a multiplier when using optics.
> Continuing with my argument, it's absolute FOV that has minimal (if any at all) usecases, as most mods would tend to set multipliers.

No, i have explained why it's the opposite. Very few servers will use a mod that enables cheating and gives all players differing abilites when using optics.
> considering that this feature is essential for mimicking real-life optics.

It's not, FOV multipliers do not mimick real life optics at all, realistic optics require absolute FOV.
///////

I can understand why you think as you do, it's a tricky and complex thing to consider, but it is a misunderstanding.

The optional use of a multiplier may be useful, but there hasn't been any valid usecases stated for it yet. It would be good to see some but you have stated you can't think of any apart from the invalid one.
However i'm not too bothered because i suspect there are some usecases for a FOV multiplier, and it is false by default, so i won't disapprove the inclusion of a multiplier.

My only request is to remove the unnecessary and confusing FOV transformation maths suggested by p_gimeno, and to use a simple multiplication of FOV. After all, this is a multiplier and the documentation states it is a multiplier of FOV, so it should be that.
:-1: until that is done.

doc/lua_api.txt
```diff
@@ -5261,12 +5261,21 @@ This is basically a reference to a C++ `ServerActiveObject`
...
-* `set_breath(value)`: sets players breath
+* `get_breath()`: returns player's breath
+* `set_breath(value)`: sets player's breath
     * values:
         * `0`: player is drowning
         * max: bubbles bar is not shown
         * See [Object properties] for more information
+* `set_fov(fov, is_multiplier)`: Sets player's FOV
+    * `fov`: FOV value.
+    * `is_multiplier`: Set to `true` if the FOV value is a multiplier.
```
paramat: It's not a mulitplier as long as it uses p_gimeno's transformation maths.

src/client/camera.cpp
```diff
@@ -454,7 +454,19 @@ void Camera::update(LocalPlayer* player, f32 frametime, f32 busytime, f32 tool_r
 	if (player->getPlayerControl().zoom && player->getZoomFOV() > 0.001f) {
 		fov_degrees = player->getZoomFOV();
 	} else {
-		fov_degrees = m_cache_fov;
+		// Apply server-sent FOV. If server doesn't
+		// enforce client's FOV, default to m_cache_fov
+		auto fov_spec = player->getFov();
+		if (fov_spec.first > 0.0f) {
+			// If server-sent FOV is a multiplier, multiply
+			// it with m_cache_fov instead of overriding
```
paramat: Currently it's not a mulitpler.

src/client/camera.cpp
```diff
@@ -454,7 +454,19 @@ void Camera::update(LocalPlayer* player, f32 frametime, f32 busytime, f32 tool_r
...
+			if (fov_spec.second)
+				fov_degrees = 2 * atan(tan(m_cache_fov / 2) * fov_spec.first);
```
paramat: fov_degrees = m_cache_fov * fov_spec.first;
(i think)

paramat:
> Once this is merged, I'll work on making builtin zoom use these methods.

Sorry, but while you misunderstand, you cannot be trusted to work on zoom, as that is even more critical as it causes the server to load distant world.
That is best left to lhofhansl and me as we have the most experience with zoom.

src/client/camera.cpp
paramat: In MT 'FOV' is 'view frustrum angle in degrees'.
p_gimeno's transformation is based on considering FOV as something else that is not 'view frustrum angle in degrees'.

paramat:
http://irc.minetest.net/minetest-dev/2019-04-07#i_5525861

11:59 p_gimeno   what units are used for FOV? is it an angle?
11:59 ANAND      degrees
12:00 p_gimeno   in that case, direct multiplication is not the best idea
12:01 ANAND      What problems do you foresee, and what do you suggest as a solution?
12:01 p_gimeno   atan(tan(FOV)*factor) may work
12:02 ANAND      Ok, I'll try that out
12:02 ANAND      What would be the issue with direct multiplication, though?
12:03 p_gimeno   imagine you have a FOV of 120° (pretty wide) - half that is 60°, it's not really double the zoom, it's more than that
12:04 ANAND      How so? Isn't that exactly double the zoom?
12:04 p_gimeno   I assume you want the applied zoom factor to be the same at any FOV
12:04 p_gimeno   no...
12:05 ANAND      Correct
12:05 p_gimeno   ok, imagine the ZOOMED view is 100°. What should the NON-ZOOMED view be?
12:05 ANAND      Erm... the player's default FOV?
12:05 p_gimeno   (I mean, walking backwards)
12:06 p_gimeno   basically, the problem is that the actual zoom factor applies to the projected image, not to the angle
12:06 p_gimeno   I can try to draw diagrams to explain
12:06 ANAND      I see
12:06 ANAND      That'd be really helpful
12:35 p_gimeno   http://www.formauri.es/personal/pgimeno/temp/FOV-Diagram-2X-zoom-is-not-0.5X-angle.png
12:36 p_gimeno   that's what you expect from a 2X zoom: the small rectangle will be zoomed to occupy the full screen
12:36 p_gimeno   however, note the angles are not half of each other
12:38 p_gimeno   they approximate half of each other as the initial angle decreases, though; if the initial FOV is e.g. 30° you're not going to note a difference. But if it's wider, you might.
12:40 rubenwardy we're now at 360 contributors according to Github 🎉
12:40 ANAND      🎉
12:41 ANAND      p_gimeno: I'll try out the atan(tan()) method then
12:41 ANAND      It's still somewhat strange that 60 isn't the half of 120 :P
12:41 ANAND      Thanks for your efforts :)
12:42 p_gimeno   the problem is thinking that a FOV expressed as an angle is reasonable, it is not
12:43 p_gimeno   FOV has always been expressed in length units with respect to a certain size film, typically in mm with respect to a 35mm film
12:44 p_gimeno   then a 64mm FOV is really double a 32mm FOV
12:45 ANAND      Hmmmm...
12:46 p_gimeno   a similar problem happens in some raycasting tutorials where they use angles to raycast successive pixel columns, and they get deformation and don't understand why - https://love2d.org/forums/viewtopic.php?f=4&t=85624
12:49 p_gimeno   there's sooo many wrong design decisions in minetest, I've lost count of them. That's just another one.
12:50 sfan5      isn't that every software
12:51 p_gimeno   yeah but rarely so prominent and with so many implications as in minetest
12:59 p_gimeno   [0407 14:36:20] <p_gimeno> however, note the angles are not half of each other <--- wait, I should have said "note the angles are not equal"
13:04 p_gimeno   THESE angles are not half of each other: http://www.formauri.es/personal/pgimeno/temp/FOV-Diagram-2X-zoom-is-not-0.5X-angle-fixed.png

The diagram:
FOV-Diagram-2X-zoom-is-not-0 5X-angle-fixed

Here p_gimeno is talking about 'field of view' meaning 'the in-world width of the viewable part of a plane at a certain distance in front of the player'. Their transformation maths (currently in the PR) is a multiplier for 'the in-world width of the viewable part of a plane at a certain distance in front of the player', it is not a multiplier for MT 'FOV' which is 'the angle of the view frustrum'.

12:42 p_gimeno   the problem is thinking that a FOV expressed as an angle is reasonable, it is not
12:43 p_gimeno   FOV has always been expressed in length units with respect to a certain size film, typically in mm with respect to a 35mm film
12:49 p_gimeno   there's sooo many wrong design decisions in minetest, I've lost count of them. That's just another one.

p_gimeno's transformation is based on what they think MT's 'FOV' value should be, instead on what it actually is. So the suggestion is invalid.
///////

If a modder uses a multiplier for FOV they expect it to multiply FOV.

For example, if a player has FOV 60 deg, and the modder applies a multiplier of 1.5, the modder expects the player to end up with a FOV of 90 deg, but with the current transformation that doesn't happen.

ClobberXD:
I give up. I'll change it to simple multiplication. Thanks.

paramat:
Thanks, no objection from me.

ClobberXD:
Thanks, paramat. @SmallJoker Would you be able to update/dismiss your review?

paramat:
Replying to your IRC comment: I have no more changes to suggest. I can't approve this because i personally find in-game FOV changes irritating and visually uncomfortable (celeron55 finds them nauseating). I don't think this is of much use and don't think it is worth adding. Unfortunately MC's stupid use of FOV changes has created a demand for it.
However, i can't quite justify disapproving based on the above, so i'm neutral.

ClobberXD:
I understand. Thanks for your opinion and reviews though. :)

jastevenson303:
Not everybody feels the way you do, @paramat.

ClobberXD:
Rebased. I've also added `set_fov = true` to `minetest.features`

ClobberXD:
I just realized that there's no way to check if the client supports FOV overriding. Bumping the protocol version might help, though. That way, servers can enforce strict proto version checking and/or make mods query the client's proto version during runtime.

jastevenson303:
+1 for strict protocol version check.

paramat:
> Not everybody feels the way you do, paramat.

Well, obviously. There's no point adding a comment that does nothing other than resent the fact that someone has an opinion you disagree with.
Please refrain from the sarcastic remarks too.

Moth-Tolias:
I hope I'm not making too much a fool of myself here, but - would there be a way to disable this in the client? that way folks like paramat need never encounter it

ClobberXD:
> I hope I'm not making too much a fool of myself here, but - would there be a way to disable this in the client? that way folks like paramat need never encounter it

I'm afraid that's not possible. This is to provide a fair, unified experience. If clients are given the option to opt out of server-sent FOV overrides, they'd be essentially cheating, by using their default FOV while the other players are made to use a much more zoomed-out FOV, for example. As a result, they would be able to see distant things more clearly than other players who haven't opted out server-sent FOV overrides.

jastevenson303:
zoom/fov really is a client-side preference, however.

You wouldn't set a player's mouse sensitivity server-side, and if you did
you wouldn't restrict their ability to choose for themselves.

ClobberXD:
Bumped protocol version. Mods can now check if a client is aware of server-sent FOV overrides, and can perform alternate actions if not, for example.

benrob0329:
Actually overriding mouse sensitivity would be useful for slowing down or locking the player's camera.

jastevenson303:
Yes but when that action is complete, their own mouse sensitivity
preference is restored.  You wouldn't lock them on a particular setting
because of your "refined taste," for example.

ClobberXD:
Rebased, and renamed

src/server.cpp
```diff
@@ -1866,6 +1866,16 @@ void Server::SendMovePlayer(session_t peer_id)
 	Send(&pkt);
 }
 
+void Server::SendPlayerFov(session_t peer_id)
+{
+	NetworkPacket pkt(TOCLIENT_FOV, 0, peer_id);
```
DS-Minetest:
```suggestion
	NetworkPacket pkt(TOCLIENT_FOV, 4 + 1, peer_id);
```

ClobberXD: Thanks

paramat:
Concerning IRC discussion about 'zoom', please don't involve zoom in this PR, leave it to a later PR.

To be clear, this PR has nothing to do with 'zoom', that shouldn't be stated in the 1st post or on IRC. 'zoom' is an independent FOV applied when pressing the 'zoom' key. This PR allows mods to set the normal non-'zoom' FOV, this can result in something similar to 'zoom' but it's very different.

Any work on 'zoom' in future has to be discussed and carefully thought through before starting.

ClobberXD:
I'm planning to make a separate issue/PR for re-implementing zoom using `set_fov`. This PR is finished and I won't be adding anything else to it.

ClobberXD:
Rebased. I've lost count of the number of times I've rebased this PR until now... :(

The incremental inv. sending commit already bumped the protocol version to 38, so this PR will bump it up to 39 instead.

builtin/game/features.lua
```diff
@@ -13,6 +13,7 @@ core.features = {
 	object_use_texture_alpha = true,
 	object_independent_selectionbox = true,
 	httpfetch_binary_data = true,
+	set_fov = true,
```
sfan5: this can be detected using `if player.set_fov then` (assuming you have a playerref at hand)
so this doesn't need to be here

src/client/camera.cpp
```diff
@@ -454,7 +454,19 @@ void Camera::update(LocalPlayer* player, f32 frametime, f32 busytime, f32 tool_r
...
+		auto fov_spec = player->getFov();
+		if (fov_spec.first > 0.0f) {
+			// If server-sent FOV is a multiplier, multiply
+			// it with m_cache_fov instead of overriding
+			if (fov_spec.second)
+				fov_degrees = m_cache_fov * fov_spec.first;
+			else
+				fov_degrees = fov_spec.first;
+		} else {
+			fov_degrees = m_cache_fov;
```
sfan5: Shouldn't this apply with the zoom key pressed, too?

src/network/networkprotocol.h
```diff
@@ -361,7 +362,14 @@ enum ToClientCommand
 		wstring reason
 	*/
 
-	TOCLIENT_PLAYERITEM = 0x36, // Obsolete
+	// TOCLIENT_PLAYERITEM = 0x36, Obsolete
```
sfan5: just remove the comment here, it's been gone for a long time

src/script/lua_api/l_object.cpp
```diff
@@ -1249,6 +1249,40 @@ int ObjectRef::l_set_look_yaw(lua_State *L)
 	return 1;
 }
 
+// set_fov(self, degrees)
```
sfan5: `set_fov(self, degrees[, is_multiplier])`

src/client/camera.cpp
ClobberXD: I'm not sure how set_fov would work with the existing zoom functionality, which is why I left it as it is (in hindsight, probably not the best thing to do)

- If server sends an FOV override, **and** player presses zoom key, which one overrides the other?
- If server-sent FOV is a multiplier, what would happen then? Doubling the existing zoom, for example, would be messed up.

src/script/lua_api/l_object.cpp
ClobberXD: Fixed, thanks

src/network/networkprotocol.h
ClobberXD: Done

builtin/game/features.lua
ClobberXD: True, removed.

src/client/camera.cpp
sfan5: > If server sends an FOV override, and player presses zoom key, which one overrides the other?

the server

> If server-sent FOV is a multiplier, what would happen then? Doubling the existing zoom, for example, would be messed up.

no issue with that, the fov is still clamped to a sane range

ClobberXD: > > If server sends an FOV override, and player presses zoom key, which one overrides the other?
> 
> the server

Uh... ok

> > If server-sent FOV is a multiplier, what would happen then? Doubling the existing zoom, for example, would be messed up.
> 
> no issue with that, the fov is still clamped to a sane range

FOV is clamped between `1.0f` and `160.0f`. Sane? Sure. Playable? Nope.

But if the server has the authority to override client-side zoom, then this isn't a problem. Thanks for clarifying.

sfan5: >But if the server has the authority to override client-side zoom, then this isn't a problem. 

Yes, that was the idea here.
If the server does stupid things that impact zoom functionality or make the game unplayable it's not our problem. The server has full control and full responsibility.

ClobberXD: I've juggled the code around a bit to give priority to server-sent overrides first, then client-side zoom, and then default to client's preferred FOV:

```cpp
f32 fov_degrees;
auto fov_spec = player->getFov();
if (fov_spec.first > 0.0f) {
	// If server-sent FOV is a multiplier, multiply
	// it with m_cache_fov instead of overriding
	if (fov_spec.second)
		fov_degrees = m_cache_fov * fov_spec.first;
	else
		fov_degrees = fov_spec.first;
} else if (player->getPlayerControl().zoom && player->getZoomFOV() > 0.001f) {
	// Player requests zoom, apply zoom FOV
	fov_degrees = player->getZoomFOV();
} else {
	// Set to client's selected FOV
	fov_degrees = m_cache_fov;
}
```

Now, zoom is only checked if the server-sent FOV is `0.0f`.

ClobberXD:
@sfan5 I've tested the new FOV checking order (see https://github.com/minetest/minetest/pull/7557#discussion_r319729836), and it works perfectly. I've used the `fov_test` mod linked in the first post. I've added another chat-command to the mod, to allow testers to verify MT's FOV checking order.

I've also made the `Zoom currently disabled by game or mod` status message appear when the server enforces a certain FOV, not allowing the client to use the zoom feature.

@SmallJoker Code has changed quite a bit, please re-review if possible. Thanks.

I'll also try to add a couple of unittests, if possible.

ClobberXD:
Rebased. Would be great to have this PR merged for 5.1.0

paramat:
Hi, i removed the milestone back in May because that was soon after the release of 5.0.1 and milestones are (as i understand) primarily used when approaching a release to highlight PRs/issues that must be merged/attended to before release. Or, more rarely, features where the core devs have agreed that they really want the feature merged for the release.

I'm not keen on the casual use of milestones for saying 'it would be nice to have this merged for next release' because that is automatically the case for every PR, so is somewhat redundant. This use of milestones tends to result in them being too casually added in large numbers, only to be removed again when approaching the release and realising a merge probably won't happen in time. These often then get re-added as milesetones for the following release and it starts all over again. it just creates a redundant mess of labels.

ClobberXD:
@paramat Thanks for clarifying. But wasn't this PR agreed to be merged for 5.1.0? At least that's what nerzhul said. Wasn't this a joint decision?

paramat:
A PR is not agreed to be merged until it has 2 approvals, however, we might have agreed it would be good to merge it for 5.1.0 if possible. However, nerzhul added this to the milestone in Dec 2018 which was long before 5.0.0, that's too far off to decide it will be merged for 5.1.0, which is why i removed the milestone, and therefore i expect he did that casually and alone.
If we had agreed we really wanted this in 5.1.0 the milestone would have been re-added, maybe together with a comment.

ClobberXD:
Alright, makes sense. Thanks again for clearing up things. :)

PR #7578, opened 2018-07-22
Disable selection box for dead players

ClobberXD:
Closes #7510, see this issue comment for details - https://github.com/minetest/minetest/issues/7510#issuecomment-401535819

**Untested**

### DIFF
--- a/src/content_cao.cpp
+++ b/src/content_cao.cpp
@@ -343,7 +343,7 @@ GenericCAO::~GenericCAO()
 bool GenericCAO::getSelectionBox(aabb3f *toset) const
 {
 	if (!m_prop.is_visible || !m_is_visible || m_is_local_player
-			|| !m_prop.pointable) {
+			|| !m_prop.pointable || m_hp == 0) {
 		return false;
 	}
 	*toset = m_selection_box;
--- a/src/content_sao.cpp
+++ b/src/content_sao.cpp
@@ -813,7 +813,7 @@ bool LuaEntitySAO::getCollisionBox(aabb3f *toset) const
 
 bool LuaEntitySAO::getSelectionBox(aabb3f *toset) const
 {
-	if (!m_prop.is_visible || !m_prop.pointable) {
+	if (!m_prop.is_visible || !m_prop.pointable || m_hp == 0) {
 		return false;
 	}
 


ClobberXD:
@nerzhul, assuming that you're referring to the _selection_ box, would you prefer it to stretch horizontally instead of being completely disabled?

paramat:
:-1: 
The collisionbox remains and is upright, so the selectionbox should match it.
We could change both to be horizontal on death but that doesn't seem worth doing for such a minor issue and for the very short time before the player respawns.

SmallJoker:
> for the very short time before the player respawns.

Please keep in mind that the players might be afk (app running in the background), so that their respawn happens after a few minutes. The issue is that it blocks the "view" to the other players, especially when they're all at the same spot. This happens on spawns, team bases and in an intensive CTF combat, where dead selection boxes can be used as a wall to not get damaged.

paramat:
It's not likely they will be afk when they die, combat is the least likely time to be afk. It's also unlikely for a player to go afk when they die instead of hitting respawn.
However also, if the dead player has an upright collisionbox then an upright selectionbox should also remain. The player may visually be lying on the ground but they are effectively dead and standing up.
I think this minor issue should be dealt with in games.

paramat:
This is a sloppy hack and adds a condition to a heavily-used engine function, for a minor issue only relevant to PVP.
This is actually a game issue, games 'should' (ideally) alter the player's boxes to be horizontal when they die. Not saying we will do this in MTG but maybe. CTF servers will likely be using custom games so can do it there.

IhrFussel:
> It's not likely they will be AFK when they die

Maybe you should ask some longtime server owners about that before making such an assumption. I experience that on my server A LOT.

It's not uncommon at all to die while idling:

- Monster spawned nearby and attacks
- Server is 100% PvP and player attacks
- Player places lava or makes you drown or causes some other environmental damage to you
- Server has a hunger mod and you starve to death

Those are just a few possibilities.

Ezhh:
Just to confirm - dead AFKers are pretty common. 

Expecting games like CTF or others to handle this separately also potentially adds another barrier to creating games. A fix within MTG may well be the best answer, given that many other games build from it.

ClobberXD:
Interesting perspectives... looks like implementing this should be done in a game, considering the opinions presented.

Closing, thanks for your time. :)

paramat:
Ok, sorry, it seemed likely to me but just not in the intensive battle situation that was used as an example. But i guess i should have realised there are many sloppy players :)
Perhaps we will fix this in MTG.

nerzhul:
@ClobberXD i don't think collisionbox not matching model orientation is a game problem, it's more a client problem or server problem. Both orientation must match.

paramat:
No it is actually the game that defines the collisionbox for a particular model, see https://github.com/minetest/minetest_game/blob/master/mods/player_api/init.lua
The issue is that the MTG player lying down is just one animation of the model but the collisionbox is defined for all of the animations. So there's nothing to be done in the engine, it's a game issue.

paramat:
Yes sorry, i don't consider it a minor issue anymore.
As a penance i will make sure this gets fixed in MTG, i'm getting there:

screenshot_20180724_133143

screenshot_20180724_133143

ClobberXD:
Nice, but shouldn't it be elongated?

paramat:
I tried that at first, but the box is fixed to world axes and the player can lay in any direction, so it has to be square, it could be smaller of course for a balance. Maybe i'll choose the diagonal size of the player. PR coming soon in MTG.

rubenwardy:
if the box is set dynamically, you could use maths to work out the max width and height. Then maybe make a slight bit smaller

ClobberXD:
> the box is fixed to world axes and the player can lay in any direction, so it has to be square

Ah I see... that'd make proportional boxes more difficult.

paramat:
Directionally calculated: nah, MT is already full of elongated objects with square boxes =)
WIP Game PR https://github.com/minetest/minetest_game/pull/2184

PR #7586, opened 2018-07-26
Fix build on gcc 5.0

zeuner:
The master branch is supposed to build on gcc 4.8 and higher. However, the patches in this pull request (adding some includes only) were necessary for building on different Linux-based platforms when using the vanilla gcc 5.0 compiler from the gcc developers.


zeuner:
Great, thanks!

original message follows:

zeuner:
Great!

And thanks for the hint. I have yet to determine what real benefit it
would bring, though.

original message follows:

PR #7587, opened 2018-07-27
Implement on_wielditem_change callback registration

### LINKED ISSUE #1377: on_wielditem_change Lua function
Today I discovered stu's fantastic 3D-Armor mod. Which doesn't only add an armor system that reduces damage, but also the long awaited features of showing worn armor and the wielded item on the player model, allowing other players to see what you are wearing or holding!

https://forum.minetest.net/viewtopic.php?f=11&t=4654

Although the armor system works well and efficiently, engine limitations make it difficult to find a good way of handling wielded items. The server needs to periodically check when a player changed the selected item on the hotbar, which is laggy and costly.

Please add a Lua function that gets triggered when the item held by a player changes, and reports the new and possibly old item. It should typically execute when the player selects another hotbar slot which contains another item, or if an item was added / removed to / from the selected hotbar slot.

Visible wielded items are difficult to implement in a quick and cheap way without this. Even if it's for 3D-Armor or any other mod that implements them, an engine function to make it possible is essential in my opinion. Here's an example of what I'm thinking:

minetest.on_wielditem_change(player, old_item, new_item)
    mod.change_wielditem_appearance(player:get_player_name(), new_item)
end

ClobberXD:
This PR implements the following global callback:

```lua
register_on_wielditem_change(player, old_item, old_wield_index, new_item, new_wield_index)
```

- `player`: An ObjectRef. Refers to player whose wielditem changed.
- `old_item`: An ItemStack holding the previous wielditem of `player`.
- `old_wield_index`: Wield-index of the old item.
- `new_item`: An ItemStack holding the current wielditem of `player`.
- `new_wield_index`: Wield-index of the new item.

`on_wielditem_change` callbacks are invoked whenever the current wielded item is changed by using the scroll wheel, or by pressing on the (numerical) hot-keys.

****

Here's some testing code:
```lua
minetest.register_on_wielditem_change(function(player, old_item, old_idx, new_item, new_idx)
	minetest.chat_send_all("\n********\n" ..
		"on_wielditem_change:" ..
		"\n\tname = " .. player:get_player_name() ..
		"\n\told_item = " .. old_item:to_string() ..
		"\n\tnew_idx = " .. old_idx ..
		"\n\tnew_item = " .. new_item:to_string() ..
		"\n\tnew_idx = " .. new_idx ..
		"\n********\n")
end)
```

Closes #1377. Ready for review. Tested; works.

### DIFF
--- a/src/network/serverpackethandler.cpp
+++ b/src/network/serverpackethandler.cpp
@@ -840,16 +840,15 @@ void Server::handleCommand_PlayerItem(NetworkPacket* pkt)
 	session_t peer_id = pkt->getPeerId();
 	RemotePlayer *player = m_env->getPlayer(peer_id);
 
-	if (player == NULL) {
-		errorstream <<
-			"Server::ProcessData(): Canceling: No player for peer_id=" <<
+	if (!player) {
+		errorstream << "Server::ProcessData(): Canceling: No player for peer_id=" <<
 			peer_id << " disconnecting peer!" << std::endl;
 		DisconnectPeer(peer_id);
 		return;
 	}
 
 	PlayerSAO *playersao = player->getPlayerSAO();
-	if (playersao == NULL) {
+	if (!playersao) {
 		errorstream <<
 			"Server::ProcessData(): Canceling: No player object for peer_id=" <<
 			peer_id << " disconnecting peer!" << std::endl;
@@ -857,11 +856,23 @@ void Server::handleCommand_PlayerItem(NetworkPacket* pkt)
 		return;
 	}
 
-	u16 item;
+	u16 old_wield_idx, new_wield_idx;
+	ItemStack old_itemstack, new_itemstack;
 
-	*pkt >> item;
+	*pkt >> new_wield_idx;
 
-	playersao->getPlayer()->setWieldIndex(item);
+	// Get current wielded item and its idx
+	old_wield_idx = player->getWieldIndex();
+	player->getWieldedItem(&old_itemstack, nullptr);
+
+	player->setWieldIndex(new_wield_idx);
+
+	// Get new wielded item and run on_wielditem_change callbacks
+	// Clamp the new wield idx by calling getWieldIndex instead of
+	// passing it directly
+	player->getWieldedItem(&new_itemstack, nullptr);
+	getScriptIface()->on_wielditem_change(playersao, old_itemstack, old_wield_idx,
+		new_itemstack, player->getWieldIndex());
 }
 
 void Server::handleCommand_Respawn(NetworkPacket* pkt)
@@ -1668,8 +1679,8 @@ void Server::handleCommand_SrpBytesM(NetworkPacket* pkt)
 
 	if (client->chosen_mech != AUTH_MECHANISM_SRP &&
 			client->chosen_mech != AUTH_MECHANISM_LEGACY_PASSWORD) {
-		actionstream << "Server: got SRP _M packet, while auth"
-			<< "is going on with mech " << client->chosen_mech << " from " 
+		actionstream << "Server: got SRP _M packet, while auth "
+			<< "is going on with mech " << client->chosen_mech << " from "
 			<< addr_s << " (wantSudo=" << wantSudo << "). Denying." << std::endl;
 		if (wantSudo) {
 			DenySudoAccess(peer_id);
--- a/src/script/cpp_api/s_player.cpp
+++ b/src/script/cpp_api/s_player.cpp
@@ -235,6 +235,30 @@ void ScriptApiPlayer::on_authplayer(const std::string &name, const std::string &
 	runCallbacks(3, RUN_CALLBACKS_MODE_FIRST);
 }
 
+void ScriptApiPlayer::on_wielditem_change(ServerActiveObject *player,
+	const ItemStack &old_item, u16 old_wield_idx,
+	const ItemStack &new_item, u16 new_wield_idx)
+{
+	SCRIPTAPI_PRECHECKHEADER
+
+	// Get core.registered_on_wielditem_change callbacks
+	lua_getglobal(L, "core");
+	lua_getfield(L, -1, "registered_on_wielditem_change");
+
+	// First param: player object
+	objectrefGetOrCreate(L, player);
+
+	// Params 2 & 3: Old wielded item and its wield index
+	LuaItemStack::create(L, old_item);
+	lua_pushnumber(L, old_wield_idx);
+
+	// Params 4 & 5: New wielded item and its wield index
+	LuaItemStack::create(L, new_item);
+	lua_pushnumber(L, new_wield_idx);
+
+	runCallbacks(5, RUN_CALLBACKS_MODE_FIRST);
+}
+
 void ScriptApiPlayer::pushMoveArguments(
 		const MoveAction &ma, int count,
 		ServerActiveObject *player)
--- a/builtin/game/register.lua
+++ b/builtin/game/register.lua
@@ -612,6 +612,7 @@ core.registered_can_bypass_userlimit, core.register_can_bypass_userlimit = make_
 core.registered_on_modchannel_message, core.register_on_modchannel_message = make_registration()
 core.registered_on_player_inventory_actions, core.register_on_player_inventory_action = make_registration()
 core.registered_allow_player_inventory_actions, core.register_allow_player_inventory_action = make_registration()
+core.registered_on_wielditem_change, core.register_on_wielditem_change = make_registration()
 
 --
 -- Compatibility for on_mapgen_init()
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -4531,11 +4531,17 @@ Call these functions only at load time!
       once with revoker being the player name, and again with revoker being nil.
 * `minetest.register_can_bypass_userlimit(function(name, ip))`
     * Called when `name` user connects with `ip`.
-    * Return `true` to by pass the player limit
+    * Return `true` to bypass the player limit
 * `minetest.register_on_modchannel_message(function(channel_name, sender, message))`
     * Called when an incoming mod channel message is received
     * You should have joined some channels to receive events.
     * If message comes from a server mod, `sender` field is an empty string.
+* `minetest.register_on_wielditem_change(function(player, old_item, old_idx, new_item, new_idx))`
+    * Called when a player selects a wield item, either by using the scroll wheel,
+      or by pressing the corresponding hotbar key.
+    * `player` is an ObjectRef.
+    * `old_item` and `new_item` are ItemStacks.
+    * `old_idx` and `new_idx` are wield indices of the old and new items.
 
 Setting-related
 ---------------
--- a/src/script/cpp_api/s_player.h
+++ b/src/script/cpp_api/s_player.h
@@ -52,6 +52,9 @@ public:
 	void on_playerReceiveFields(ServerActiveObject *player,
 			const std::string &formname, const StringMap &fields);
 	void on_authplayer(const std::string &name, const std::string &ip, bool is_success);
+	void on_wielditem_change(ServerActiveObject *player,
+		const ItemStack &old_item, u16 old_wield_idx,
+		const ItemStack &new_item, u16 new_wield_idx);
 
 	// Player inventory callbacks
 	// Return number of accepted items to be moved


ClobberXD:
I thought a globalstep-based callback is much more resource-hungry. Besides, this callback is for functions that run immediately when a player changes their current wield-item, and for that very reason, update intervals of once per second or similar make this callback useless. This could get excessively triggered, I agree, but is useful nonetheless. Here is a use-case from the top of my mind:
- Imagine a sniper-rifle mod (which would incidentally be released in a week or so), that allows the player to toggle a "scope view" using the secondary mouse button (right-click). But if a player changes their wielditem while scope view is enabled, they won't be able to exit the scope view since right-clicks won't trigger `onSecondaryUse`/`onPlace` callbacks of the original wielditem.

Isn't a globalstep function to check for wielditem changes unnecessarily expensive performance-wise? Also, it isn't really a callback if it's constantly called every server-step...

ClobberXD:
I've been wanting to do the first one. The second idea sounds good too; I'll have to think up a way to do that though, sounds complicated...

ClobberXD:
> Is it possible that the player scrolls so fast from item A to B then to C that the server only gets a callback with from A to C?

Would someone be able to give me some guidance on how this could be done? Are the core-devs fine with this feature?

> It would be more useful to have the hotbar numbers. (Eg. a mod wants to set an other item to the position in the inventory.)

This seems to be a little out of the scope of this PR, as it requires modifications in many other parts of the code. This PR is just about adding an `on_wielditem_change` callback, but I could try making another PR for it once this is in.

DS-Minetest:
> > Is it possible that the player scrolls so fast from item A to B then to C that the server only gets a callback with from A to C?
> 
> Would someone be able to give me some guidance on how this could be done? Are the core-devs fine with this feature?

This was actually not meant as feature request but as question on how reliable the callback is.
I'm sorry for causing confusion.

ClobberXD:
But that is a very good thing to have. I just don't know how to do it... :)

paramat:
> Is it possible that the player scrolls so fast from item A to B then to C that the server only gets a callback with from A to C?

This is a good point. Many players scroll back and forth fast through the hotbar to get to the item they actually want. Whether to include the items in between or not is a problem. Triggering all callbacks will result in rapid calls, mostly for items the player never wanted to select. How to ignore the items in between is another problem.

ClobberXD:
> This is a good point. Many players scroll back and forth fast through the hotbar to get to the item they actually want. Whether to include the items in between or not is a problem. Triggering all callbacks will result in rapid calls, mostly for items the player never wanted to select. How to ignore the items in between is another problem.

Yes, I'd really like to do that. It's just that I can't figure out how... :)

ClobberXD:
Rebased, and added documentation.

jastevenson303:
I know this may be unrelated, maybe, but I'd just like to add that it's only possible to get a player's wield list and index, you can't set this in a server-side mod.  I don't know about client-side.

Do you have any thoughts on this @ClobberXD?  Or on the possibility of setting wield index with this PR?  If a player scrolls too fast, or if a callback is missed (do they get missed?) because the server didn't get the message, can the server then set the wield index itself somehow?  Sorry if this is too off-topic.

ClobberXD:
I'd like that feature as well, but I'm afraid it's unrelated to this PR.

src/script/cpp_api/s_player.cpp
```diff
@@ -228,6 +228,21 @@ void ScriptApiPlayer::on_auth_failure(const std::string &name, const std::string
 	runCallbacks(2, RUN_CALLBACKS_MODE_FIRST);
 }
 
+void ScriptApiPlayer::on_wielditem_change(ServerActiveObject *player,
+		const ItemStack old_item,
```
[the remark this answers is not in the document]
ClobberXD: Fixed

ClobberXD:
Skipping the callbacks for the hotbar indexes in the middle only concerns scrollbar. I could implement a timer of sorts, and run the callbacks only if there's at least t seconds (would 0.2 s be correct?) between the current change and the previous change.

paramat:
@sofar i'm assuming that's an approval and added label.

paramat:
Looks like a code approval to me, not a concept approval =)

ClobberXD:
> Looks like a code approval to me, not a concept approval =)

Oops, that what I meant. But a code approval still isn't the same as a "Tested. Everything works as it should" type of approval. :)

(Don't mind me, I'm arguing over nothing... :P)

jastevenson303:
I think the intermittent slots won't get dropped on a fast scroll.

ClobberXD:
@nerzhul Could you update your review?

ClobberXD:
I've re-implemented this PR to make use of a timer, and only run callbacks if the item has been selected for at least 0.5s. The code compiles alright, but I haven't got the time to test it yet. Also, this PR has become relatively more complex, and I don't trust myself to verify that everything is fine and working. The more testers and reviewers checking out this PR, the lesser the chance of bugs and other side-effects occurring in the code. So, everyone is welcome to test this PR. :)

@sofar and @nerzhul: Please update or dismiss your reviews when you can, as this PR has changed quite a bit. :)

src/content_sao.cpp
```diff
@@ -990,6 +990,15 @@ void PlayerSAO::getStaticData(std::string * result) const
 
 void PlayerSAO::step(float dtime, bool send_recommended)
 {
+	if (m_wielditem_timer_active) {
+		if (m_wielditem_interval.step(dtime, 0.5f)) {
```
nerzhul: merge conditions

src/content_sao.cpp
```diff
@@ -990,6 +990,15 @@ void PlayerSAO::getStaticData(std::string * result) const
...
+			ItemStack old_item = getOldWieldedItem();
```
nerzhul: const ref for both

src/content_sao.cpp
```diff
@@ -1420,6 +1429,23 @@ void PlayerSAO::setWieldIndex(int i)
 	}
 }
 
+ItemStack PlayerSAO::getOldWieldedItem() const
+{
+	return m_old_wielded_item;
+}
+
+void PlayerSAO::setOldWieldedItem(ItemStack item)
```
nerzhul: const ref

src/content_sao.cpp
ClobberXD: Both? Which is the other one here?

ClobberXD: Nvm, figured it out. :)

src/content_sao.cpp
ClobberXD: I forgot to push this, lol

src/content_sao.cpp
ClobberXD: Done.

ClobberXD:
I've noticed that while the timer works, the old item passed to the callback is the one which was selected before the new item. This shouldn't be the case. Here's why:

- Fill the hotbar with unique items.
- Select the first slot, if it isn't selected already.
- Quickly scroll to the eighth slot.
- The callback isn't called for all the items in between, which is good.
- But notice that the old item is the one in the seventh slot, instead of the first.

I'm sure the fix is quite simple - I'll look into it.

src/content_sao.cpp
```diff
@@ -1440,6 +1448,23 @@ void PlayerSAO::setWieldIndex(int i)
...
+void PlayerSAO::setOldWieldedItem(const ItemStack &item)
+{
+	m_old_wielded_item = item;
+}
+
+void PlayerSAO::resetWielditemTimer()
+{
+	m_wielditem_interval = IntervalLimiter();
+	m_wielditem_timer_active = true;
+	infostream << "PlayerSAO::resetWielditemTimer()" << std::endl;
```
sfan5: I don't quite see what this timer is for

ClobberXD: The timer is for ensuring that the on_wielditem_change callbacks are run only if the item has been selected for more than 0.5s. This prevents the callback from being invoked for every item when scrolling from say, slot 1 to slot 8. The timer_active boolean exists to make the timer run only once after a wielditem change, to prevent the callback from being run every 0.5s.

ClobberXD:
Currently, the callback is run only when a different item has been selected for at least 0.5s. This is to prevent intermittent wield index changes to trigger the callback (e.g. while scrolling to a different slot). But the concept of using a timer doesn't feel very reliable and robust to me, which is why I was sorta staying away from this PR. Is there a better way to implement this PR?

sorcerykid:
I've long been waiting to see this PR merged. I could use the functionality for one of my mods which allows for executing chat command macros just by switching to a special Command Block item in the hotbar. The timer in my view would be hindrance, since there would always be a brief delay before executing the macro.

One possible solution would be to add an optional property to tools, nodes, and craftitems that specifies the delay before triggering the on_wielditem_change callback. By default it would be 0.5. Another idea is to only impose a delay when the scrollwheel is being used. However, that might require a change to the client-server protocol.

ClobberXD:
> I've long been waiting to see this PR merged. I could use the functionality for one of my mods which allows for executing chat command macros just by switching to a special Command Block item in the hotbar. The timer in my view would be hindrance, since there would always be a brief delay before executing the macro.

I see, interesting.

> One possible solution would be to add an optional property to tools, nodes, and craftitems that specifies the delay before triggering the on_wielditem_change callback. By default it would be 0.5.

Per-item delay? Sounds overkill, IMHO. :thinking:

Also, if the item's delay is set to a much lower value, scrolling through this item to select another item will trigger the callback for this item too, which I believe shouldn't happen.

> Another idea is to only impose a delay when the scrollwheel is being used. However, that might require a change to the client-server protocol.

This is an interesting idea, and I did actually think about it. But as you mention, there's no backwards-compatible way to do this, which made me use the current approach.

Currently, the problem is that the timer-based approach isn't a very solid implementation, as scrolling slowly can trigger the callbacks for each item in-between. This is why I'm not comfortable with the current approach; and this is also why I'm not pushing for this to be merged, although I very well understand how many modders want this functionality.

Thanks for your support, @sorcerykid. Feel free to post other approaches to this feature.

sfan5:
Why not let modders filter events themselves (if they want to do so)?

sorcerykid:
What about instead of a global callback registration, each item definition has its own on_wield() and on_unwield() callbacks? Since most items wouldn't define either such callback, then they could simply be ignored. For 99% of wield item changes therefore nothing would happen and that should alleviate most concerns about excessive callback flooding.

Those items that do have an on_wield() callback of course could then probably just implement their own timer by using minetest.after or something similar. Just a thought.

ClobberXD:
> Why not let modders filter events themselves (if they want to do so)?

Hmm... ditch the timer altogether and run callbacks for *all* wielditem changes? That was the original state of this PR, and was modified to drop the in-between changes while scrolling by using a timer. :)

But I can see the benefit in the suggestion. Mods would be in-charge of not overloading the server by registering a heavy callback that runs for all the wielditems.

> What about instead of a global callback registration, each item definition has its own on_wield() and on_unwield() callbacks? Since most items wouldn't define either such callback, then they could simply be ignored. For 99% of wield item changes therefore nothing would happen and that should alleviate most concerns about excessive callback flooding.
>
> Those items that do have an on_wield() callback of course could then probably just implement their own timer by using minetest.after or something similar. Just a thought.

Although mods would lose the ability to register one generic callback for multiple wielditem change events, this is actually a fantastic idea! I can see benefits in having both the callbacks. The global one can be a more generic type of callback that isn't limited to a very specific item, while the per-item callback can be used for item-specific code like toggling scopes for sniper rifles (this was one of the original use-cases for this functionality). Ideally, this and sfan5's previous suggestion should be everything a mod needs, to implement all sorts of code. :+1:

Thanks for your inputs, @sfan5 and @sorcerykid. Much appreciated. :)

sfan5:
> Mods would be in-charge of not overloading the server by registering a heavy callback that runs for all the wielditems.

I dislike this too, but this is already the case with many other functions and it gives modders the biggest possible freedom.

ClobberXD:
True. `register_on_wielditem_change` can be compared to `register_on_dignode` in terms of its frequency. Should I add a warning to the documentation to make it very clear that this callback will run very frequently?

pandaro:
> What about instead of a global callback registration, each item definition has its own on_wield() and on_unwield() callbacks? Since most items wouldn't define either such callback, then they could simply be ignored. For 99% of wield item changes therefore nothing would happen and that should alleviate most concerns about excessive callback flooding.

This is the way

ClobberXD:
PR #9862 implements the per-item `on_wield` callback.

ClobberXD:
I'm re-doing this PR from scratch(-ish), as the PR wasn't working earlier due to reasons like outdated code and botched rebases. Should be ready for review tomorrow.

Btw, can someone confirm if `RUN_CALLBACKS_MODE_FIRST` is the right `runCallbacks` mode for my purpose? This callback doesn't check for return values.

ClobberXD:
### Updates

- Callback runs for all wielditem changes, even for instantaneous ones (while scrolling, for example), and the PR is so much simpler now.
- All extraneous code-style fixes have been reverted, as they unnecessarily bloated the PR.
- Callback now also passes the old and new wield indices. See https://github.com/minetest/minetest/pull/9862#issuecomment-628568646 for the rationale behind this decision.
- PR has been tested - works as expected; ready for review.
- PR summary and testing code has been updated in the first post.

ClobberXD:
Rebased; merge conflict resolved.

ClobberXD:
Could this be added to the 5.3.0 milestone, if possible?

src/network/serverpackethandler.cpp
```diff
@@ -857,11 +857,21 @@ void Server::handleCommand_PlayerItem(NetworkPacket* pkt)
...
+	ItemStack old_itemstack, new_itemstack;
 
-	*pkt >> item;
+	// Get current wielded item and its idx
+	old_wield_idx = playersao->getWieldIndex();
+	playersao->getWieldedItem(&old_itemstack);
 
-	playersao->getPlayer()->setWieldIndex(item);
+	// Deserialize new wield idx from packet
+	*pkt >> new_wield_idx;
```
sfan5, line 868: The deserialization should happen before any actions

src/network/serverpackethandler.cpp
```diff
@@ -857,11 +857,21 @@ void Server::handleCommand_PlayerItem(NetworkPacket* pkt)
...
+	playersao->getPlayer()->setWieldIndex(new_wield_idx);
```
sfan5, line 869:
```suggestion
	player->setWieldIndex(new_wield_idx);
```

src/script/cpp_api/s_player.cpp
```diff
@@ -235,6 +235,30 @@ void ScriptApiPlayer::on_authplayer(const std::string &name, const std::string &
...
+	SCRIPTAPI_PRECHECKHEADER
+
+	// Get core.registered_on_wielditem_change callbacks
+	lua_getglobal(L, "core");
+	lua_getfield(L, -1, "registered_on_wielditem_change");
+
+	// First param: player object
+	objectrefGetOrCreate(L, player);
+
+	// Params 2 & 3: Old wielded item and its wield index
```
sfan5, line 251: This doesn't match the documentation, did you forget to update it?

src/server/player_sao.h
```diff
@@ -184,6 +184,8 @@ class PlayerSAO : public UnitSAO
 	void unlinkPlayerSessionAndSave();
 	std::string generateUpdatePhysicsOverrideCommand() const;
 
+	ItemStack getItemFromIndex(std::string, int i) const;
```
sfan5, line 187: unused

src/network/serverpackethandler.cpp
sfan5: Also since `new_wield_idx` is directly given to mods the server should clamp it to the allowed range.

src/network/serverpackethandler.cpp
ClobberXD: Ahaha, good catch! :D

src/network/serverpackethandler.cpp
ClobberXD: > The deserialization should happen before any actions

May I know why? The existing wield idx needs to be assigned to `old_wield_idx` first, and only then can the current wield idx be updated. If you insist, I could deserialize the new idx into a temp u16 before doing anything else, but it feels unnecessary to me, tbh.

I'll look into clamping the new idx, thanks for the heads-up.

src/network/serverpackethandler.cpp
ClobberXD: A couple of other `playersao` calls can be changed to directly calling from `player` too, as the SAO class simply wraps around the respective `Player` methods, which I'm guessing is for convenience.

src/network/serverpackethandler.cpp
ClobberXD: `Player::setWieldIndex` clamps the passed idx. I could just copy over the same code, but that'd result in code duplication (albeit only for two lines). Shall I go ahead? Maybe you have a better idea?

src/script/cpp_api/s_player.cpp
ClobberXD: Indeed. Updating docs...

src/server/player_sao.h
ClobberXD: Wha... I don't even remember adding this, lol.

src/script/cpp_api/s_player.cpp
ClobberXD: Done

src/server/player_sao.h
ClobberXD: Done

src/network/serverpackethandler.cpp
ClobberXD: Done

src/network/serverpackethandler.cpp
sfan5: > May I know why?

In this case it's irrelevant/minor, but you typically only want to do anything after grabbing all necessary data from the packet (and thus verifying that it's complete)

> Maybe you have a better idea?

You could call `getWieldIndex()` after setting the new index to get the clamped value.

pandaro:
As I just suggested for [this](https://github.com/minetest/minetest/pull/9862) PR, I want to understand if it is possible for `on_wielditem_change` to `return` the `new_tool` as `Itemstack`. I personally want to use it with `set_tool_capabilities` to modify the capabilities of a tool based on some user-related values.
I don't know if this is possible or if I can get the same result in another way, and I hope I have explained myself sufficiently clearly.
thanks

sfan5:
You can achieve that using `set_wielded_item` regardless of whether these callbacks take a return value.

pandaro:
> You can achieve that using `set_wielded_item` regardless of whether these callbacks take a return value.

[here](https://github.com/minetest/minetest/pull/9862#issuecomment-638476583) the code that i will run. if anyone can help me to find what i'm misunderstand will be loved for long time

ClobberXD:
@pandaro Replacing `return itemstack` with `player:set_wielded_item(itemstack)` should work.

src/network/serverpackethandler.cpp
ClobberXD: You're right, it makes sense to ensure the packet is intact before doing anything else. And yeah, `getWieldIndex()` would work too.

ClobberXD: Done!

src/script/cpp_api/s_player.cpp
sfan5: Still missing...

ClobberXD:
@sfan5 I've attended to all of your review comments. I've re-tested the PR and it works.

sfan5:
What about `old_idx` and `new_idx` here?
pic

src/script/cpp_api/s_player.cpp
ClobberXD: I'm absolutely sure I updated this... o_O

ClobberXD: Ah, I didn't stage the change before committing, and switched to another branch... :facepalm: 

Updated docs, sorry. :)

sfan5:
Note that if this merged before #9862, it would make sense if #9862 were reimplemented on top of the existing API in builtin.
If it isn't, then this still makes sense but _this_ PR should be changed to do that.

pandaro:
Just retested, all work fine.
thank you @ClobberXD

ClobberXD:
@sfan5 Do you mean that #9862 should use this callback behind the scenes if this PR is merged first? That sounds good from a unified API viewpoint, but would it be performant enough?

Implementing global wielditem change callbacks using per-item `on_(un)wield` callbacks (which I assume is what you suggested, if #9862 is going to be merged first) is *probably* a bad idea because the global callbacks are going to be run for *all* wielditem changes and it needs to be fast. Also, I don't see a straight-forward way to implement this alongside actual `on_wield` callbacks defined by mods. Do let me know if I misunderstood you. :)

sfan5:
> Do you mean that #9862 should use this callback behind the scenes if this PR is merged first?

Yes.

> That sounds good from a unified API viewpoint, but would it be performant enough?

If anything it would be more performant because you only have to cross the C++ -> Lua boundary once.

> Implementing global wielditem change callbacks using per-item on_(un)wield callbacks

No that's now what I meant.
If the other PR is merged first, this PR should make the required changes to implement `on_(un)wield` ontop of the global callback. If this PR is merged first, the other PR should make those changes.

ClobberXD:
> If anything it would be more performant because you only have to cross the C++ -> Lua boundary once.

That's interesting... I'd love to profile both variations and see which implementation is faster, just for personal interest. :)
 
> No that's not what I meant.
> If the other PR is merged first, this PR should make the required changes to implement `on_(un)wield` ontop of the global callback. If this PR is merged first, the other PR should make those changes.

Ahh, makes sense. Yeah, I definitely misunderstood you. This sounds good. :+1:

PR #7592, opened 2018-07-28
get_node_drops: Make empty drop return empty table

tenplus1:
This stops get_node_drops from returning an empty string when using drop = "" and instead returns the supposed empty table instead {}


SmallJoker:
Whereas the actual issue is yet not known, it's surely good to strip this empty ItemStack.

PR #7595, opened 2018-07-28
Document default values for general cmake options

### LINKED ISSUE #7489: CMake default options
Hi!

Can you please explicitly say which values are default to each option at README.md (if all dependencies are present)?

Also are all ENABLE_* options for both client and server?

pauloue:
And fix a typo

Closes #7489


mazocomp:
Thank you!

PR #7598, opened 2018-07-29
Fix sky objects not rendering with ogles

### LINKED ISSUE #3155: Moon, Sun and Stars on Android
Nobody noticed, but they really do not have to Android. This is bug.

stujones11:
Fixes #3155 and possibly #7573

Note that the primitive type `scene::EPT_QUADS` is not implemented by Irrlicht ES yet so I have used a triangle indexed list instead. We could produce the full quad by drawing two triangles but the effect is lost at such low resolution it is difficult to tell the difference IMO.

screen1
screen

### DIFF
--- a/src/sky.cpp
+++ b/src/sky.cpp
@@ -44,7 +44,11 @@ Sky::Sky(s32 id, ITextureSource *tsrc):
 
 	video::SMaterial mat;
 	mat.Lighting = false;
+#ifdef __ANDROID__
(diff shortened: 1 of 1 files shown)


srifqi:
I can confirm that this fixes 3155. I'm still not sure about 7573.
sunrise
^ sunrise
sunset
^sunset

lhofhansl:
Nice!

PR #7615, opened 2018-08-03
Save debug.txt to build dir for RUN_IN_PLACE build

pauloue:
For the RUN_IN_PLACE build, debug.txt gets saved to the current working directory, usually my home dir. Shouldn't it go in the build dir along with everything else?

Tested.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -530,11 +530,8 @@ static bool read_config_file(const Settings &cmd_args)
 
 static void init_log_streams(const Settings &cmd_args)
 {
-#if RUN_IN_PLACE
-	std::string log_filename = DEBUGFILE;
-#else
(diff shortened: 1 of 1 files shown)


ClobberXD:
Isn't the current working dir `<minetest>/bin`?

SmallJoker:
For me, the working dir was always `bin` but I agree that changing this to a fixed location would be better.

pauloue:
For me the working dir was `~` because I had the executable symlinked to `~/bin/mtdev`, which is in my `$PATH` so I can run it from anywhere.

SmallJoker:
> Shouldn't it go in the build dir along with everything else?

Well, with this PR the debug.txt file is now saved to `minetest_dir/` directly; but not into the `bin` directory.

pauloue:
Right, I meant the minetest directory. `bin` should be just for the binary.

PR #7618, opened 2018-08-04
feature check for commit ac4884c3d677ef51bcce47f95ccd49d874022692

zeuner:
Commit [1] introduced the possibility to set the selectionbox of an entity independently from its collisionbox. However, there seems to be no way to check for this engine feature (otherwise having to fall back to setting the collisionbox), so I currently have to work around it by comparing version numbers (which is error-prone).

I therefore suggest to allow the feature to be checked through minetest.features, by checking for the key "independent_selectionbox".

[1] https://github.com/minetest/minetest/commit/ac4884c3d677ef51bcce47f95ccd49d874022692

### DIFF
--- a/builtin/game/features.lua
+++ b/builtin/game/features.lua
@@ -3,7 +3,6 @@
 core.features = {
 	glasslike_framed = true,
 	nodebox_as_selectionbox = true,
-	independent_selectionbox = true,
 	chat_send_player_param3 = true,
 	get_all_craft_recipes_works = true,
 	use_texture_alpha = true,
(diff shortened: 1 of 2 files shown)


rubenwardy:
should be

`entity_independent_selectionbox`

builtin/game/features.lua
```diff
@@ -3,6 +3,7 @@
 core.features = {
 	glasslike_framed = true,
 	nodebox_as_selectionbox = true,
+	independent_selectionbox = true,
```
SmallJoker: Should be moved to the bottom of the list since all other were added in a chronological order.

zeuner:
@SmallJoker: Ok, no problem. Did not know about the chronological
order.

original message follows:

zeuner:
GTHT

original message follows:

zeuner:
Great, thanks!

original message follows:

zeuner:
I'm fine with that.

Is there any naming convention for the feature table contributors
should be aware of?

original message follows:

PR #6153, opened 2017-07-19
Windows: Cpack wix installer

adrido:
This allows CMake/CPack to use the Windows Installer XML (WIX) to create a Installer package.

The Installer package is only available if its **not** a `RUN_IN_PLACE` build. For that the zip package is generated.

Todo:
- [ ] ~~Uninstaller should clean the cache dir.~~ *
- [ ] ~~Question: should uninstaller also remove the minetest.conf?~~ *
- [x] Each sub-game should have its own component
    - [x] "minimal" sub-game should may not installed by default only optional

*= Applications are installed for all users, while config file and cache is per user. Removing this while uninstalling is neither a good idea nor supported by MSI.
Related: #4931, #3596
Fixes: #4528, #3802, #4697
### Download:
[Minetest-0.5.0-win32.msi.zip](https://github.com/minetest/minetest/files/1777352/Minetest-0.5.0-win32.msi.zip)
[Minetest-0.5.0-win64.msi.zip](https://github.com/minetest/minetest/files/1777354/Minetest-0.5.0-win64.msi.zip)

The installer here is packed in a zip, because Github does not allow MSI files.

### Screenshots:
grafik
grafik
grafik
The minimal subgame have to be manually selected.
grafik
grafik
grafik

PR #7629, opened 2018-08-08
Run Minetest update checker on startup

SmallJoker:
~However, the version checking is not optimal (relies on `MINOR.MAJOR.PATCH`) and GitHub seems to limit the API page calls ([source](http://irc.minetest.net/minetest-hub/2018-08-08#i_5375273)). `core.get_version` should probably also be extended by a `is_dev_build` field to get around the ugly detection.~

See https://github.com/minetest/minetest/pull/7629#issuecomment-656786622 for the current status and testing instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,7 +261,7 @@ General options and their default values:
     ENABLE_SYSTEM_GMP=ON       - Use GMP from system (much faster than bundled mini-gmp)
     ENABLE_SYSTEM_JSONCPP=ON   - Use JsonCPP from system
     RUN_IN_PLACE=FALSE         - Create a portable install (worlds, settings etc. in current directory)
-    ENABLE_UPDATE_CHECKER=TRUE - Whether to enable the update checker by default
+    ENABLE_UPDATE_CHECKER=TRUE - Whether to enable update checks by default
     USE_GPROF=FALSE            - Enable profiling using GProf
     VERSION_EXTRA=             - Text to append to version (e.g. VERSION_EXTRA=foobar -> Minetest 0.4.9-foobar)
     ENABLE_TOUCH=FALSE         - Enable Touchscreen support (requires support by IrrlichtMt)
--- a/README.md
+++ b/README.md
@@ -261,6 +261,7 @@ General options and their default values:
     ENABLE_SYSTEM_GMP=ON       - Use GMP from system (much faster than bundled mini-gmp)
     ENABLE_SYSTEM_JSONCPP=ON   - Use JsonCPP from system
     RUN_IN_PLACE=FALSE         - Create a portable install (worlds, settings etc. in current directory)
+    ENABLE_UPDATE_CHECKER=TRUE - Whether to enable the update checker by default
     USE_GPROF=FALSE            - Enable profiling using GProf
     VERSION_EXTRA=             - Text to append to version (e.g. VERSION_EXTRA=foobar -> Minetest 0.4.9-foobar)
     ENABLE_TOUCH=FALSE         - Enable Touchscreen support (requires support by IrrlichtMt)
--- a/android/native/jni/Android.mk
+++ b/android/native/jni/Android.mk
@@ -102,6 +102,7 @@ LOCAL_CFLAGS += \
 	-DVERSION_MINOR=${versionMinor} \
 	-DVERSION_PATCH=${versionPatch} \
 	-DVERSION_EXTRA=${versionExtra} \
+	-DDEVELOPMENT_BUILD=${developmentBuild} \
 	$(GPROF_DEF)
 
 ifdef USE_BUILTIN_LUA
--- a/android/native/jni/Android.mk
+++ b/android/native/jni/Android.mk
@@ -98,7 +98,6 @@ LOCAL_CFLAGS += \
 	-DUSE_SOUND=1                   \
 	-DUSE_LEVELDB=0                 \
 	-DUSE_GETTEXT=1                 \
-	-DENABLE_UPDATE_CHECKER=0       \
 	-DVERSION_MAJOR=${versionMajor} \
 	-DVERSION_MINOR=${versionMinor} \
 	-DVERSION_PATCH=${versionPatch} \
--- a/android/native/jni/Android.mk
+++ b/android/native/jni/Android.mk
@@ -98,6 +98,7 @@ LOCAL_CFLAGS += \
 	-DUSE_SOUND=1                   \
 	-DUSE_LEVELDB=0                 \
 	-DUSE_GETTEXT=1                 \
+	-DENABLE_UPDATE_CHECKER=0       \
 	-DVERSION_MAJOR=${versionMajor} \
 	-DVERSION_MINOR=${versionMinor} \
 	-DVERSION_PATCH=${versionPatch} \
--- a/builtin/mainmenu/dlg_version_info.lua
+++ b/builtin/mainmenu/dlg_version_info.lua
@@ -34,15 +34,15 @@ local function version_info_formspec(data)
 
 	local fs = {
 		"formspec_version[3]",
-		"size[10,5.5]",
+		"size[12.8,7]",
 		"style_type[label;textcolor=#0E0]",
-		"label[0.375,0.775;", core.formspec_escape(title), "]",
-		"textarea[0.375,1.375;9.25,2.75;;;",
+		"label[0.5,0.8;", core.formspec_escape(title), "]",
+		"textarea[0.4,1.6;12,3.4;;;",
 			core.formspec_escape(message), "]",
-		"container[0.375,4.375]",
-		"button[0,0;2.92,0.75;version_check_visit;", fgettext("Visit website"), "]",
-		"button[3.16,0;2.92,0.75;version_check_remind;", fgettext("Later"), "]",
-		"button[6.33,0;2.92,0.75;version_check_never;", fgettext("Never"), "]",
+		"container[0.4,5.8]",
+		"button[0.0,0;4.0,0.8;version_check_visit;", fgettext("Visit website"), "]",
+		"button[4.5,0;3.5,0.8;version_check_remind;", fgettext("Later"), "]",
+		"button[8.5.5,0;3.5,0.8;version_check_never;", fgettext("Never"), "]",
 		"container_end[]",
 	}
 
--- a/builtin/mainmenu/dlg_version_info.lua
+++ b/builtin/mainmenu/dlg_version_info.lua
@@ -1,0 +1,157 @@
+--[[
+Minetest
+Copyright (C) 2018-2020 SmallJoker
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation; either version 2.1 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License along
+with this program; if not, write to the Free Software Foundation, Inc.,
+51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+]]
+
+local function version_info_formspec(data)
+	local title = fgettext("A new Minetest version is available")
+	local message =
+		fgettext("Current version: $1\nNew version: $2\n" ..
+				"Visit $3 to find out how to get the newest version to stay up to date" ..
+				" with the features and bugfixes.",
+			core.get_version().string, data.new_version, data.url)
+
+	local fs = {
+		"formspec_version[3]",
+		"size[10,5.5]",
+		"style_type[label;textcolor=#0E0]",
+		"label[0.375,0.775;", core.formspec_escape(title), "]",
+		"textarea[0.375,1.375;9.25,2.75;;;",
+			core.formspec_escape(message), "]",
+		"container[0.375,4.375]",
+		"button[0,0;2.92,0.75;version_check_visit;", fgettext("Visit website"), "]",
+		"button[3.16,0;2.92,0.75;version_check_remind;", fgettext("Later"), "]",
+		"button[6.33,0;2.92,0.75;version_check_never;", fgettext("Never"), "]",
+		"container_end[]",
+	}
+
+	return table.concat(fs, "")
+end
+
+local function version_info_buttonhandler(this, fields)
+	if fields.version_check_remind then
+		-- Only wait for the check interval
+		core.settings:set("update_last_known", "")
+		this:delete()
+		return true
+	end
+	if fields.version_check_never then
+		core.settings:set("update_last_checked", "disabled")
+		this:delete()
+		return true
+	end
+	if fields.version_check_visit then
+		core.open_url(this.data.url)
+		this:delete()
+		return true
+	end
+
+	return false
+end
+
+local function create_version_info_dlg(new_version, url)
+	assert(type(new_version) == "string")
+	assert(type(url) == "string")
+
+	local retval = dialog_create("version_info",
+		version_info_formspec,
+		version_info_buttonhandler,
+		nil)
+
+	retval.data.new_version = new_version
+	retval.data.url = url
+
+	return retval
+end
+
+local function get_current_version_code()
+	-- Format: Major.Minor.Patch
+	-- Convert to MMMNNNPPP
+	local cur_string = core.get_version().string
+	local cur_major, cur_minor, cur_patch = cur_string:match("^(%d+).(%d+).(%d+)")
+
+	if not cur_patch then
+		core.log("error", "Failed to read version numbers (invalid tag format?)")
+		return
+	end
+
+	return (cur_major * 1000 + cur_minor) * 1000 + cur_patch
+end
+
+local function on_version_info_received(json)
+	local known_update = tonumber(core.settings:get("update_last_known")) or 0
+
+	-- Format: MMNNPPP (Major, Minor, Patch)
+	local new_number = json.latest.version_code
+	if not new_number then
+		core.log("error", "Failed to read version numbers (invalid tag format?)")
+		return
+	end
+
+	local cur_number = get_current_version_code()
+	if new_number <= known_update or new_number < cur_number then
+		return
+	end
+
+	-- Also consider updating from 1.2.3-dev to 1.2.3
+	if new_number == cur_number and not core.get_version().is_dev then
+		return
+	end
+
+	core.settings:set("update_last_known", tostring(new_number))
+
+	local tabs = ui.find_by_name("maintab")
+	tabs:hide()
+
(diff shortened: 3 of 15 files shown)


rubenwardy:
This, unfortunately, doesn't help with the biggest offender - Ubuntu. I've heard their policy was to remove update notifications and such from programs (although I have seen one in firefox). Maybe they won't notice? :D

src/defaultsettings.cpp
```diff
@@ -286,6 +286,9 @@ void set_default_settings(Settings *settings)
 	settings->setDefault("mono_font_size", font_size_str);
 	settings->setDefault("contentdb_url", "https://content.minetest.net");
 
+	settings->setDefault("update_information_url", "https://api.github.com/repos/minetest/minetest/releases");
```
rubenwardy: this is a bad idea, it would be better to put a file on minetest.net imo

SmallJoker: Not redirect, but cache the JSON file.
```
$ curl -i https://api.github.com/repos/minetest/minetest/releases
HTTP/1.1 200 OK
Server: GitHub.com
Date: Wed, 08 Aug 2018 16:24:19 GMT
Content-Type: application/json; charset=utf-8
Content-Length: 92175
Status: 200 OK
X-RateLimit-Limit: 60
X-RateLimit-Remaining: 52  <-- issue is here
X-RateLimit-Reset: 1533746023
```

sfan5: Since the code barely uses any of the 92 KB that JSON file is big, we should definitely host a replacement ourselves.

rubenwardy: Yeah, agreed. Probably easiest to just add a file directly to https://github.com/minetest/minetest.github.io

builtin/mainmenu/dlg_version_info.lua
```diff
@@ -0,0 +1,125 @@
...
+--51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+
+
+local function version_info_formspec(data)
+	return (
+		"size[9,4.5,true]" ..
+		"textarea[0.5,0.5;8.5,3;;" ..
+		fgettext("A new Minetest version was found!") .. ";" ..
+		fgettext("Current version: $1\nNew version: $2\n" ..
+			"Visit $3 to find out how to get the newest version to stay up to date" ..
```
ClobberXD: Adding another new line before `Visit $3 to...` will make it more readable, I think

builtin/mainmenu/dlg_version_info.lua
```diff
@@ -0,0 +1,125 @@
...
+		"textarea[0.5,0.5;8.5,3;;" ..
+		fgettext("A new Minetest version was found!") .. ";" ..
+		fgettext("Current version: $1\nNew version: $2\n\n" ..
+			"Visit $3 to find out how to get the newest version to stay up to date" ..
+			" with the features and bugfixes.",
+			core.get_version().string, data.new_version,
+			"https://www.minetest.net/downloads/") .. "]" ..
+		"checkbox[1.5,3;version_check_never;" ..
+			fgettext("Never notify me again") .. ";false]" ..
+		"button[1.5,4;2,0.5;version_check_seen;" .. fgettext("Ok") .. "]" ..
```
ClobberXD: `OK`

Calinou:
> This, unfortunately, doesn't help with the biggest offender - Ubuntu. I've heard their policy was to remove update notifications and such from programs, but I can't find a citation for this (although I have seen one in firefox). Maybe they won't notice? :D

That is indeed the policy of most distributions (including Debian). It makes sense from their point of view, as users are not supposed to download Minetest directly from minetest.net according to them.

rubenwardy:
> That is indeed the policy of most distributions (including Debian and Ubuntu). It makes sense from their point of view, as users are not supposed to download Minetest directly from minetest.net according to them.

It would be nice if they also didn't provide an outdated version of a networked program :/

ClobberXD:
> Is it really necessary to add one more configuration option just for this? We have a lot of them already.

I feel that the user should have control over how often MT checks for updates, and I don't think one extra drop-down would do any harm. :)

rubenwardy:
I don't think there is a need for the setting, personally

src/defaultsettings.cpp
nerzhul: or we should just use the github API correctly ?

https://developer.github.com/v3/repos/releases/#get-the-latest-release

nerzhul:
Just use the github last release API instead of full release API:

src/defaultsettings.cpp
rubenwardy: We should not rely on the GitHub API as we may not use it forever. We could use the Github API in Jekyll to automatically make a JSON file with the latest release, this would be future proof

rubenwardy:
> We should not rely on the GitHub API as we may not use it forever. We could use the Github API in Jekyll to automatically make a JSON file with the latest release, this would be future proof

Fixer-007:
"was found" somewhere in the dungeons of the internet?

builtin/mainmenu/dlg_version_info.lua
```diff
@@ -22,7 +22,7 @@ local function version_info_formspec(data)
 	return (
 		"size[9,4.5,true]" ..
 		"textarea[0.5,0.5;8.5,3;;" ..
-		fgettext("A new Minetest version was found!") .. ";" ..
+		fgettext("New Minetest version is available") .. ";" ..
```
ClobberXD: "A new Minetest version..." maybe? Sounds better and more like a full sentence... :)

orwell96:
The "never notify me again" tickbox is a bit misleading IMO, it should read "don't notify me on future updates" or so. I wouldn't expect from such a checkbox that it disabled notifications for future updates.
Also, update notifications should be disabled in builds that are neither run_in_place nor installed by the wix installer #6153, by setting the default setting this way

MrChiantos:
I think not good, I suggest : Add a news column in main menu is not possible ?

SmallJoker:
@orwell96 
> is a bit misleading IMO, it should read "don't notify me on future updates" 

That's also a possibility, will adjust the PR accordingly.
> update notifications should be disabled in builds that are neither run_in_place nor installed by the wix installer #6153, by setting the default setting this way

Why? Referring the users to the official download page covers all (or at least should) platforms which we support. Official repositories tend to be outdated quickly and people can tick the checkbox once on the initial startup to never get a notification again.

@MrChiantos This is possible but wouldn't provide enough new and helpful information since the (half/)yearly updates would be likely the only content.

t4im:
How about setting the defaults via cmake flag, with it defaulting to no updates?

* During release builds you can flip this on, so that the distributed prebuild releases come with the check enabled
* Distributions building their own releases can enable or disable updatechecks according to their policies
* Anyone building a dev build from source would automatically skip it (if you build from git, you are already in full control of the version you are building anyway, especially if you intentionally build an older version for testing)

This would then work as wanted without depending on having the right minetest.conf ready or additional version information passed around.

SmallJoker:
@t4im Thanks for your suggestion. I added the CMake config `UPDATE_CHECKER_DEFAULT` which allows distributions to disable the checker by default. Defaulting to `false` for in-development builds seems however wrong, because there are people who use the Windows development builds from the forums or Linux users which take the binaries from the daily PPA.

t4im:
Well, anyone doing the windows development builds and daily ppa builds would flip this on as well.

The idea behind the default was that someone cloning the git repository would have a default of no-updates without having to opt-out. But anyone downloading a prebuild binary would have it enabled by the person doing the build flipping the updates on.

Fixer-007:
BTW, what is the difference between "OK" and "Remind me later"?

DS-Minetest:
I think, "Ok" means something like "OK, I understood, don't tell me about this version again.".

builtin/mainmenu/dlg_version_info.lua
```diff
@@ -0,0 +1,131 @@
...
+		this:delete()
+		return true
+	end
+
+	return false
+end
+
+
+function create_version_info_dlg(new_version)
+	assert(type(new_version) == "string" and new_version ~= "")
```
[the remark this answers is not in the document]
SmallJoker: Looks like this additional check became superfluous during the progress of developing this feature.

nerzhul:
i think it can be good to embed the update API in server list or another such program which will cache the github API calls for 1h or more, and control the backend API in case of at a point a migration outside github will be done (i don't say in the next year, but maybe at a point github will have too many problems).

Lejo1:
One thing about the big Problem Ubuntu:
In the Software Center, the Place where most players install Minetest from, there is one outdated Minetest and one which seems to be the stable ppa... so why we can’t use a ppa for Minetest? There are also some unofficial builds so it shouldn’t be so hard to add a build.

Calinou:
@Lejo1 The issue is not providing up-to-date unofficial builds, it's that distribution repositories are frozen. We can't do anything about this :slightly_frowning_face:

Lejo1:
But there is an stable auto updating Minetest build on the software center...

But I think it uses snap as Package Manager.

Calinou:
@Lejo1 With Snaps/Flatpaks, the concern of providing updates is still on the side of the package maintainer, not the application developer. This is why in-application update checks should still be disabled in this case.

Lejo1:
Sure but why not for Windows?
The Precompiler could filter the check out.

Calinou:
@Lejo1 That's what I said in my comment above :slightly_smiling_face:

kilbith:
The link in the formspec could use a `hypertext` action with the use of `core.open_url`.

nerzhul:
@minetest/engine , is this a feature we want or not ? 
ServerList should have the latest available version, and we may enable this feature only on MacOSX & on Windows. A compile time option will be needed for Linux

rubenwardy:
Sounds good to me.

When https://github.com/minetest/minetest.github.io/pull/204 is merged, you'll be able to get a JSON file from `https://www.minetest.net/release_info.json`:

```json
{
    "latest": {
        "version": "5.2.0",
        "date": "2020-04-05"
    }
}
```

nerzhul:
i suggest to have a integer version for easier comparison. Protocol is not intended to be bump on each release.
FreeBSD and some other softwares implements thing like this:
10.2.1 => 100201
it can be easy code side to perform the check instead of reimplementing version parsing

appgurueu:
> i suggest to have a integer version for easier comparison. Protocol is not intended to be bump on each release.
> FreeBSD and some other softwares implements thing like this:
> 10.2.1 => 100201
> it can be easy code side to perform the check instead of reimplementing version parsing

I agree, but I think the integer number should have 3 instead of two digits:

5.3.0 => 005003000

Calinou:
@appgurueu We're not going to exceed version 99 for any of these components…

rubenwardy:
We will reach 5.100.0 in 2068, at our current rate of development

rubenwardy:
Please update this with support for the new website API, and also support for opening an arbitrary URL using core.open_url

```json
{
    "latest": {
        "version": "5.2.0",
        "date": "2020-04-05",
        "protocol_version": 39,
		"url": "https://www.minetest.net/downloads/"
    }
}
```

In the future, I'd like to add direct links to installers if using Windows

rubenwardy:
also should use `core.get_http_api()`

SmallJoker:
@rubenwardy Heh, both features are way newer than this PR. I'd love to update the PR when an unlimited HTTP fetch mirror for this JSON file is available.

rubenwardy:
Merged, see https://www.minetest.net/release_info.json

rubenwardy:
Thanks

check_new_version should probably run on an AsyncWorker thread, however

SmallJoker:
**Updates**

grafik

 * Update checker now uses the minetest.net json file
 * Adjusted internal numeric format
 * Use URL provided by the JSON file

**How to test**

1) Set up a local HTTP server and point your minetest.conf to use that JSON file
2) Set the version to "5.4.0" and "5004000" in the file
3) Start Minetest, new version is detected
4) Remove or change `update_last_` to test all buttons

LoneWolfHT (approved):
Tested all buttons, works :+1:

README.md
```diff
@@ -240,7 +240,7 @@ General options and their default values:
     ENABLE_SYSTEM_GMP=ON       - Use GMP from system (much faster than bundled mini-gmp)
     ENABLE_SYSTEM_JSONCPP=OFF  - Use JsonCPP from system
     OPENGL_GL_PREFERENCE=LEGACY - Linux client build only; See CMake Policy CMP0072 for reference
-    RUN_IN_PLACE=FALSE         - Create a portable install (worlds, settings etc. in current directory)
+    UPDATE_CHECKER_DEFAULT=TRUE - Default behavior of the version update checker
```
rubenwardy, line 243: What does true/false mean?  Perhaps the setting should be `ENABLE_UPDATE_CHECKER`

rubenwardy:
Here's a branch with asynchronous URL fetching, and a better formspec   https://github.com/rubenwardy/minetest/tree/update_info

image

I don't know why the kerning is weird

kilbith:
Why not using hypertext? I know this is marked as experimental, but then it can be adapted if the element changes.

EDIT: nevermind. The URL should be clickable and the versions should be mono+bold. The buttons are not needed, just make it a checkbox (checked by default) « Remind me later ».

rubenwardy:
Closing this as it's not been updated in a while. Feel free to reopen and apply my changes or fix the issues yourself

kilbith:
This is a good feature. Hope to see it reopened.

rubenwardy:
Another option is that I could submit my branch as another PR, or push it into this PR. I didn't want to do that without Krock checking it over though, as it complicates reviews

SmallJoker:
@rubenwardy Do as you wish. It's been long enough that I forgot what I wrote anyway.

rubenwardy:
One comment to resolve: https://github.com/minetest/minetest/pull/7629#discussion_r499193932

nerzhul:
clearly a nice improvement to add

erlehmann:
If enabled by default, this seems like a privacy issue. Minetest should ask for consent before!

**I think that the code allows the server to crash a client repeatedly via crafted JSON response.**

You can probably solve that issue by checking the response against a proper grammar, e.g. JSON Schema.

erlehmann:
Granted, there exist way easier methods to make a Minetest client fall into a null pointer, but none of the existing ones can target *all* of the clients.

sfan5:
> I think that the code allows the server to crash a client repeatedly via crafted JSON response.

<b>Substantiate your claims or bugger off.</b>

(Made bold for emphasis because you seem to like that)

erlehmann:
@sfan5 set the `version_code` value to some number that makes the update dialog display and the `url` value to `<item name=>`.

builtin/mainmenu/dlg_version_info.lua
```diff
@@ -0,0 +1,157 @@
...
+			"Visit $3 to find out how to get the newest version to stay up to date" ..
+			" with the features and bugfixes.",
+		core.get_version().string, data.new_version, data.url)
+	}
+
+	local fs = {
+		"formspec_version[3]",
+		"size[10,5.5]",
+		"hypertext[0.375,0.375;9.25,3.75;;",
+			core.formspec_escape(table.concat(body, "")), "]",
```
appgurueu, line 34: As @erlehmann pointed out, `formspec_escape` isn't enough here. You have to consider Hypertext escaping as well. You can also omit the empty string as second `table.concat` parameter.

appgurueu (changes requested):

builtin/mainmenu/dlg_version_info.lua
rubenwardy: This was intentional, as it allows the backend to include markup. I don't believe there's anything security sensitive in the mark up language, although it could be used to crash the client

rubenwardy: body contains markup, we don't want to escape that

builtin/mainmenu/dlg_version_info.lua
```diff
@@ -0,0 +1,157 @@
...
+		return
+	end
+
+	core.settings:set("update_last_checked", tostring(time_now))
+
+	core.handle_async(function(params)
+		local http = core.get_http_api()
+		return http.fetch_sync(params)
+	end, { url = url }, function(result)
+		local json = result.succeeded and core.parse_json(result.data)
```
sfan5, line 148: would parse_json throw an error if the input is invalid?
or just return nil/false/...?

android/native/jni/Android.mk
```diff
@@ -98,6 +98,7 @@ LOCAL_CFLAGS += \
 	-DVERSION_MINOR=${versionMinor} \
 	-DVERSION_PATCH=${versionPatch} \
 	-DVERSION_EXTRA=${versionExtra} \
+	-DDEVELOPMENT_BUILD=${developmentBuild} \
```
sfan5, line 101: missing `ENABLE_UPDATE_CHECKER` define _somewhere_ (can also be `config.h` for android)

src/defaultsettings.cpp
```diff
@@ -336,6 +336,13 @@ void set_default_settings()
 	settings->setDefault("contentdb_flag_blacklist", "nonfree, desktop_default");
 #endif
 
+	settings->setDefault("update_information_url", "https://www.minetest.net/release_info.json");
+#if ENABLE_UPDATE_CHECKER
+	settings->setDefault("update_last_checked", "");
+#else
+	settings->setDefault("update_last_checked", "disabled");
+#endif
+	settings->setDefault("update_last_known", "");
```
sfan5, line 345: don't think it's necessary to set empty defaults

sfan5:
Actually the variable is merely missing here:
https://github.com/minetest/minetest/blob/6df69f9b5bea5b438498bbae7e9873bb99434439/android/native/build.gradle#L13-L17

builtin/mainmenu/dlg_version_info.lua
SmallJoker: > * On failure outputs an error message and returns `nil`

It prints to `errorstream`.

SmallJoker:
Nice catch, thank you @sfan5.
Checks on Android are now disabled, as requested.

Niklp09 (approved):
works:)

sfan5:
I pushed a commit that fixes some potential issues (and moves a few lines around for no reason) and tested the feature. In principle this has my +1.

Just two things:
* can we make the dialog a bit bigger? It doesn't need to be so tiny
* what happens if the http request takes a long time and the player is already in e.g. the world creation menu?
  * the dialog should probably simply be not shown (it can be shown when the next check happens)

sfan5:
The following should work to detect if we're still in the main menu (and not some dialog):
`if not ui.childlist["maintab"].hidden then`

Testing code:
```diff
--- a/builtin/mainmenu/tab_about.lua
+++ b/builtin/mainmenu/tab_about.lua
@@ -141,7 +141,10 @@ return {
        end,
        cbf_button_handler = function(this, fields, name, tabdata)
                if fields.homepage then
-                       core.open_url("https://www.minetest.net")
+                       core.handle_async(function(_) os.execute("sleep 5") end, "",
+                       function()
+                               print(dump(_G.ui))
+                       end)
                end
 
                if fields.share_debug then
```
click the homepage button, switch to some other part of the menu and see what the internal state looks like

sfan5 (approved):
Works.

Issue #7602, opened 2018-07-31
Add interact acknowledgements


HybridDog:
##### Issue type

 - Feature request
 
##### Minetest version

```
Minetest 5.0.0-dev-74d27642 (Linux)
Using Irrlicht 1.8.4
BUILD_TYPE=RELEASE
RUN_IN_PLACE=0
USE_GETTEXT=0
USE_SOUND=1
USE_CURL=1
USE_FREETYPE=1
USE_LUAJIT=1
STATIC_SHAREDIR="/usr/local/share/minetest"
```

##### OS / Hardware

Operating system:linux
CPU:intel

##### Summary

Currently node dig and placement prediction only work as long as the client does not receive a mapblock sent before the server incorporated the interaction. I've emulated a slow connection on localhost (as described [there](https://gist.github.com/trongthanh/1196596)) and placed nodes on a local server. First the nodes appear because of placement prediction, then all but the first one disappear and after that slowly one after another node re-appears. I could reproduce the [skyblock nightmare](https://github.com/minetest/minetest/issues/7098#issuecomment-371767557).
The similar problem with digging is reported there: #7098

As rubenwardy mentioned, acknowledgements could be used to solve the problem.
An acknowledgement could contain a byte which tells the interaction command (dig, place, punch, etc.), the position and for digging and placing a flag tells whether digging or placing succeeded. Unused values (such as punching acknowledgement) are reserved for future use.
After placing a node and doing the placement prediction, the client could save the position, the current node in the mapblock, the predicted node, a boolean indicating whether it's the newest prediction for the position and the time into some data structure. After receiving a mapblock, the client replays the predictions in the mapblock if it's still possible (e.g. still buildable_to) and the current node values are updated. After receiving an acknowledgement, the corresponding prediction is removed and if the newest prediction boolean is true and the current and predicted nodes are the same, the node is changed to the current node.
If an acknowledgement arrives before the mapblock, a wrongly predicted node can already be removed.

The flag which tells if digging or placing succeeded can be used to improve protection:
If the interaction did not succeed, disallow placing, resp. digging, in the position for ack_delay * 2 + 100 milliseconds where ack_delay is the RTT or the time elapsed since prediction.
This makes it more difficult to dig through protected walls with a fast tool because if you need to dig two nodes and one of them reappears, you have to wait a time long enough for the server to tell the client that the other dug node should also reappear and be locked for that time.

PR #7630, opened 2018-08-08
Digging and placing: Send nodes instead of mapblocks to undo prediction

HybridDog:
The first commit disables mapblock sending when it is not necessary.
For example, when a player places something, the server executes minetest.set_node (or something similar); the changed nodes are sent via a TOCLIENT_ADDNODE, which already undoes the prediction.
Therefore, the mapblock does not need to be sent back to the client.
However, if a node is protected, placement fails and the client has to be informed that its placement prediction is wrong.
Nodes removal due to digging is a similar example.

The second commit changes mapblock sending to sending a TOCLIENT_ADDNODE packet when possible. (it is not possible when the node contains metadata and not done in rare edge cases)
Sending single node changes instead of whole mapblocks mitigates the disappearing predicted nodes problem, which is annoying especially at a skyblocks server.
I'm not sure how often the mapblocks were actually sent because mapblock sending is done with `SetBlockNotSent`, so if a player places many nodes, he/she may receive only one mapblock for all of the changed nodes together. Furthermore, according to thomasrudin, there seem to be situations where the placement prediction is not undone for some reason. 

Related:
* https://github.com/minetest/minetest/issues/7098#issuecomment-410433402
* adding interact acknowledgements (this breaks minetest stable compatibility) would be a fix for #7098 (instead of mitigations) (see #7602)
* #9349 

#
You may want to test it yourself:
You can temporary change the wood node definition to test different situations
```Lua
minetest.override_item("default:wood", {
	node_placement_prediction = "default:stone",
	node_dig_prediction = "default:cobble",
	can_dig = function() return false end,
	on_place = function(_,_, pt)
		minetest.set_node(pt.under, {name="default:mese"})
	end,
})
```
To simulate a slow internet connection on Linux, you can execute this (found [there](https://gist.github.com/trongthanh/1196596)):
```Sh
sudo tc qdisc add dev lo root handle 1: htb default 12
sudo tc class add dev lo parent 1:1 classid 1:12 htb rate 16kbps ceil 32kbps
sudo tc qdisc add dev lo root netem delay 600ms
```

### DIFF
--- a/src/network/serverpackethandler.cpp
+++ b/src/network/serverpackethandler.cpp
@@ -1025,17 +1025,31 @@ void Server::handleCommand_Interact(NetworkPacket *pkt)
 		return;
 	}
 
+	// Get pointed to node (undefined if not POINTEDTYPE_NODE)
+	v3s16 p_under = pointed.node_undersurface;
+	v3s16 p_above = pointed.node_abovesurface;
+
+	// The client only predicts a node when placing or digging a node
+	bool may_have_prediction = pointed.type == POINTEDTHING_NODE &&
+		(action == INTERACT_DIGGING_COMPLETED || action == INTERACT_PLACE);
+
 	if (playersao->isDead()) {
 		actionstream << "Server: NoCheat: " << player->getName()
 				<< " tried to interact while dead; ignoring." << std::endl;
-		if (pointed.type == POINTEDTHING_NODE) {
-			// Re-send block to revert change on client-side
-			RemoteClient *client = getClient(pkt->getPeerId());
-			v3s16 blockpos = getNodeBlockPos(pointed.node_undersurface);
-			client->SetBlockNotSent(blockpos);
-		}
-		// Call callbacks
 		m_script->on_cheat(playersao, "interacted_while_dead");
+		if (may_have_prediction) {
+			// Re-send the whole block(s) to undo placement or dig prediction
+			RemoteClient *client = getClient(pkt->getPeerId());
+			v3s16 blockpos_u = getNodeBlockPos(p_under);
+			client->SetBlockNotSent(blockpos_u);
+			if (action == INTERACT_DIGGING_COMPLETED)
+				// Dig prediction -> send only under
+				return;
+			// Placement prediction -> also send above
+			v3s16 blockpos_a = getNodeBlockPos(p_above);
+			if (blockpos_a != blockpos_u)
+				client->SetBlockNotSent(blockpos_a);
+		}
 		return;
 	}
 
@@ -1045,10 +1059,6 @@ void Server::handleCommand_Interact(NetworkPacket *pkt)
 
 	// Update wielded item
 	playersao->getPlayer()->setWieldIndex(item_i);
-
-	// Get pointed to node (undefined if not POINTEDTYPE_NODE)
-	v3s16 p_under = pointed.node_undersurface;
-	v3s16 p_above = pointed.node_abovesurface;
 
 	// Get pointed to object (NULL if not POINTEDTYPE_OBJECT)
 	ServerActiveObject *pointed_object = NULL;
@@ -1077,20 +1087,23 @@ void Server::handleCommand_Interact(NetworkPacket *pkt)
 		Make sure the player is allowed to do it
 	*/
 	if (!checkPriv(player->getName(), "interact")) {
+		// This may only happen while revoking interact before the client
+		// receives its change of privileges
 		actionstream << player->getName() << " attempted to interact with " <<
 				pointed.dump() << " without 'interact' privilege" << std::endl;
 
-		// Re-send block to revert change on client-side
-		RemoteClient *client = getClient(pkt->getPeerId());
-		// Digging completed -> under
-		if (action == INTERACT_DIGGING_COMPLETED) {
-			v3s16 blockpos = getNodeBlockPos(floatToInt(pointed_pos_under, BS));
-			client->SetBlockNotSent(blockpos);
-		}
-		// Placement -> above
-		else if (action == INTERACT_PLACE) {
-			v3s16 blockpos = getNodeBlockPos(floatToInt(pointed_pos_above, BS));
-			client->SetBlockNotSent(blockpos);
+		if (may_have_prediction) {
+			// Re-send the whole block(s) to undo placement or dig prediction
+			RemoteClient *client = getClient(pkt->getPeerId());
+			v3s16 blockpos_u = getNodeBlockPos(p_under);
+			client->SetBlockNotSent(blockpos_u);
+			if (action == INTERACT_DIGGING_COMPLETED)
+				// Dig prediction -> send only under
+				return;
+			// Placement prediction -> also send above
+			v3s16 blockpos_a = getNodeBlockPos(p_above);
+			if (blockpos_a != blockpos_u)
+				client->SetBlockNotSent(blockpos_a);
 		}
 		return;
 	}
@@ -1108,10 +1121,13 @@ void Server::handleCommand_Interact(NetworkPacket *pkt)
 		float d = playersao->getEyePosition().getDistanceFrom(pointed_pos_under);
 
 		if (!checkInteractDistance(player, d, pointed.dump())) {
-			// Re-send block to revert change on client-side
-			RemoteClient *client = getClient(pkt->getPeerId());
-			v3s16 blockpos = getNodeBlockPos(floatToInt(pointed_pos_under, BS));
-			client->SetBlockNotSent(blockpos);
+			if (may_have_prediction) {
+				// Re-send the whole block to undo placement or dig prediction
+				RemoteClient *client = getClient(pkt->getPeerId());
+				v3s16 blockpos = getNodeBlockPos(
+					floatToInt(pointed_pos_under, BS));
+				client->SetBlockNotSent(blockpos);
+			}
 			return;
 		}
 	}
@@ -1276,20 +1292,20 @@ void Server::handleCommand_Interact(NetworkPacket *pkt)
 				}
 			}
 
-			/* Actually dig node */
+			if (!is_valid_dig || n.getContent() == CONTENT_IGNORE)
+				return;
 
-			if (is_valid_dig && n.getContent() != CONTENT_IGNORE)
-				m_script->node_on_dig(p_under, n, playersao);
+			// Actually dig node
+			m_script->node_on_dig(p_under, n, playersao);
 
-			v3s16 blockpos = getNodeBlockPos(floatToInt(pointed_pos_under, BS));
-			RemoteClient *client = getClient(pkt->getPeerId());
-			// Send unusual result (that is, node not being removed)
-			if (m_env->getMap().getNode(p_under).getContent() != CONTENT_AIR) {
-				// Re-send block to revert change on client-side
-				client->SetBlockNotSent(blockpos);
-			}
-			else {
-				client->ResendBlockIfOnWire(blockpos);
+			/*
+				If the node is still the same, e.g. because of protection,
+				re-send the whole block to undo dig prediction
+			*/
+			MapNode newnode = m_env->getMap().getNode(p_under);
+			if (newnode.getContent() == n.getContent()) {
+				v3s16 blockpos = getNodeBlockPos(p_under);
+				getClient(pkt->getPeerId())->SetBlockNotSent(blockpos);
 			}
 		}
 	} // action == INTERACT_DIGGING_COMPLETED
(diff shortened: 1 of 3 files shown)


src/network/serverpackethandler.cpp
```diff
@@ -1080,18 +1080,20 @@ void Server::handleCommand_Interact(NetworkPacket* pkt)
...
-		RemoteClient *client = getClient(pkt->getPeerId());
-		// Digging completed -> under
+		actionstream << player->getName() << " attempted to interact with "
+			<< pointed.dump() << " without 'interact' privilege" << std::endl;
+		// This may only happen while revoking interact before the client
+		// receives its change of privileges
+		// Re-send the whole block to undo placement or dig prediction
 		if (action == 2) {
+			// Digging completed -> under
+			RemoteClient *client = getClient(pkt->getPeerId());
```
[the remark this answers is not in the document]
HybridDog: It's only used for digging and placing action.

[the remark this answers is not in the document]
ClobberXD:
```
	if (!checkPriv(player->getName(), "interact")) {
		..
		if (action == 2) {
			// Digging completed -> under
			RemoteClient *client = getClient(pkt->getPeerId());
			..
		}
		else if (action == 3) {
			// Placement -> above
			RemoteClient *client = getClient(pkt->getPeerId());
			..
		}
		return;
	}
```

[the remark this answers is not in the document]
ClobberXD: I feel that it's better to move the declaration of `client` before the `if` and `else` blocks, since it's used in both places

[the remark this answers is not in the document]
HybridDog: ok.

HybridDog:
I need help with these points: What should be done if
* the interact distance was too far?
* a player dug/placed something without interact privilege?
* a dead player dug/placed something?

src/server.cpp
```diff
@@ -2151,6 +2151,20 @@ void Server::sendAddNode(v3s16 p, MapNode n, u16 ignore_id,
 	}
 }
 
+void Server::sendCurrentNode(session_t peer_id, v3s16 p, MapNode n)
+{
+	if (m_env->getMap().getNodeMetadata(p)) {
```
SmallJoker: Does not check whether it actually contains metadata - only whether the container exists (which is always the case).

HybridDog: I firstly used `m_env->getMap().getNodeMetadata(p).empty()`, but I got memory access errors, so I noticed that the function returns meta or NULL:
https://github.com/minetest/minetest/blob/3a95d760e97d16fcc499aee8df1be4fe583c8168/src/nodemetadata.cpp#L203-L209
https://github.com/minetest/minetest/blob/3a95d760e97d16fcc499aee8df1be4fe583c8168/src/map.cpp#L963-L980 

When testing with a chest it worked correctly.

src/network/serverpackethandler.cpp
```diff
@@ -1277,21 +1279,19 @@ void Server::handleCommand_Interact(NetworkPacket* pkt)
...
-			}
-			else {
-				client->ResendBlockIfOnWire(blockpos);
-			}
+			/*
+				If the node is still the same, e.g. because of protection,
+				send it to the client to undo dig prediction
+			*/
+			MapNode newnode = m_env->getMap().getNodeNoEx(p_under);
+			if (newnode.getContent() == n.getContent())
```
sfan5: doesn't work with e.g. `node_dig_prediction = "the:very_same_node"`,
the server needs to apply the "full" dig prediction here and only send if the result differs

src/network/serverpackethandler.cpp
```diff
@@ -1319,34 +1319,65 @@ void Server::handleCommand_Interact(NetworkPacket* pkt)
...
+			m_emerge->enqueueBlockEmerge(peer_id,
+				getNodeBlockPos(p_under), false);
+			m_emerge->enqueueBlockEmerge(peer_id,
+				getNodeBlockPos(p_above), false);
+			return;
 		}
+
+		// The actual placing
+		if (!m_script->item_OnPlace(item, playersao, pointed))
+			// The on_place function was missing
```
sfan5: inaccurate comment

src/network/serverpackethandler.cpp
```diff
@@ -1319,34 +1319,65 @@ void Server::handleCommand_Interact(NetworkPacket* pkt)
...
+					send under if the node didn't change
+				ii) Server predicts above
+					Client's under prediction must have already been undone
+			2. Client predicts above
+				i) Server predicts above
+					send above if the node didn't change
+				ii) Server predicts under
+					send above if the nodes didn't change
+
+			However, client prediction is unknown:
```
sfan5: ?

src/network/serverpackethandler.cpp
HybridDog: I don't understand you. When the node differs, nothing needs to be sent because a TOCLIENT_REMOVENODE or something similar is already sent because of digging.
If the node is not changed, the dig prediction of the client may needs to be undone.
If dig prediction is disabled (or set to the same node), sending the current node is not required, but I think explicitly disabling is not needed because it's an edge case.

src/network/serverpackethandler.cpp
```diff
@@ -1080,19 +1079,20 @@ void Server::handleCommand_Interact(NetworkPacket* pkt)
...
-		else if (action == 3) {
-			v3s16 blockpos = getNodeBlockPos(floatToInt(pointed_pos_above, BS));
+		actionstream << player->getName() << " attempted to interact with "
+			<< pointed.dump() << " without 'interact' privilege" << std::endl;
+		// This may only happen while revoking interact before the client
+		// receives its change of privileges
+		// Re-send the whole block to undo placement or dig prediction
+		if (action == 2 || action == 3) {
+			RemoteClient *client = getClient(peer_id);
+			v3s16 blockpos;
```
nerzhul: v3s16 blockpos = getNodeBlockPos(action == 2 ? p_under : p_above);

src/network/serverpackethandler.cpp
```diff
@@ -1109,10 +1109,13 @@ void Server::handleCommand_Interact(NetworkPacket* pkt)
...
 		float d = player_pos.getDistanceFrom(pointed_pos_under);
 		if (!checkInteractDistance(player, d, pointed.dump())) {
-			// Re-send block to revert change on client-side
-			RemoteClient *client = getClient(pkt->getPeerId());
-			v3s16 blockpos = getNodeBlockPos(floatToInt(pointed_pos_under, BS));
-			client->SetBlockNotSent(blockpos);
+			if (action == 2 || action == 3) {
+				// Re-send the whole block to undo placement or dig prediction
+				RemoteClient *client = getClient(peer_id);
+				v3s16 blockpos = getNodeBlockPos(
```
nerzhul: inline with next function call

src/network/serverpackethandler.cpp
```diff
@@ -1203,10 +1205,11 @@ void Server::handleCommand_Interact(NetworkPacket* pkt)
 				infostream << "Server: Not finishing digging: Node not found."
 						<< " Adding block to emerge queue."
 						<< std::endl;
-				m_emerge->enqueueBlockEmerge(pkt->getPeerId(),
+				m_emerge->enqueueBlockEmerge(peer_id,
 					getNodeBlockPos(p_above), false);
 			}
 
+			ItemStack playeritem = playersao->getWieldedItemOrHand();
```
nerzhul: why ? it's seems unused in next code

src/network/serverpackethandler.cpp
```diff
@@ -1277,21 +1279,19 @@ void Server::handleCommand_Interact(NetworkPacket* pkt)
 				}
 			}
 
-			/* Actually dig node */
+			if (!is_valid_dig || n.getContent() == CONTENT_IGNORE)
```
SmallJoker: Resend the block is it's not a valid dug node -> cheat prevention.

src/network/serverpackethandler.cpp
sfan5: Why is the client prediction "unknown"?

src/network/serverpackethandler.cpp
sfan5: Hm, that's unexpected but okay though...

src/network/serverpackethandler.cpp
sfan5: well if you think this edge case is not important then so be it

src/network/serverpackethandler.cpp
HybridDog: This code actually does not work right: For placement under and above need to be sent.
For a long time interacting without interact hasn't been possible (except the rare edge case while revoking interact), I guess mapblock sending can be removed here.

src/network/serverpackethandler.cpp
HybridDog: This is also one of the three points where I'm not sure if a mapblock or anything should be sent: https://github.com/minetest/minetest/pull/7630#issuecomment-412029621

src/network/serverpackethandler.cpp
HybridDog: Thanks, I forgot to undo this temporary change.

src/network/serverpackethandler.cpp
HybridDog: It's a synchronisation problem: The server only tests if the node at p_under is buildable to when the interact packet arrives, but the client may has put the node in an old version of the mapblock, or even put the node onto a wrongly predicted one. 
If the client puts two nodes, the second one onto the first one, and only the second one is placed, the server assumes that it's put in p_under (air is buildable_to, on_place actually puts the node there), whereas the client assumes that it's put on p_above (the firstly predicted node is not buildable_to).

src/network/serverpackethandler.cpp
HybridDog: I've changed it.

lhofhansl:
If I understand this correctly this avoids the server sending blocks that the client has already modified anyway, i.e. to the one client that has made the change.

I wonder how much improvement we'd see with many players, since the other clients need to be informed of the changes anyway.

HybridDog:
> I wonder how much improvement we'd see with many players, since the other clients need to be informed of the changes anyway.

The other clients who did not do the interact do not get a whole mapblock additionally, each of them gets a TOCLIENT_ADDNODE (or -REMOVE) packet. I don't know how exactly metadata changes are sent, I think it's mapblock sending.
Digging and placing happens frequently in minetest, so I don't think sending whole mapblocks is justified.
I think nobody had a look at the code, it was there when c55 added it many years ago:
https://github.com/minetest/minetest/commit/f8c2f1cf54c3328abf994d146858b47029c98b3d
https://github.com/minetest/minetest/commit/6a0388bb4bedc1b1c6318d7bfebaf1ec67ccf94e#diff-ad60d65b34e16a3319296bb5d683acd6R2977

SmallJoker:
@nerzhul Are you satisfied with the changes which were done since the last review?

src/server.cpp
```diff
@@ -2151,6 +2151,35 @@ void Server::sendAddNode(v3s16 p, MapNode n, u16 ignore_id,
...
+		client->SetBlockNotSent(blockpos_a);
+}
+
+void Server::sendCurrentNode(session_t peer_id, v3s16 p, MapNode n)
+{
+	if (m_env->getMap().getNodeMetadata(p)) {
+		// If the node contains metadata, the whole block needs to be sent
+		RemoteClient *client = getClient(peer_id);
+		v3s16 blockpos = getNodeBlockPos(p);
+		client->SetBlockNotSent(blockpos);
```
SmallJoker: Now this would be (more or less) just `sendMetadataChanged({ blockpos })`

HybridDog:
I searched for minetest skyblock server and the first video I found already shows the problem: https://www.youtube.com/watch?v=NUr5PuLxfLk&t=353

HybridDog:
Could you add the Bugfix label?

thomasrudin:
Testing this on my server as of now, will give feedback in a few days...

thomasrudin:
I'm reverting to the latest stable `master` again:

### Observations
* :+1: Even with 24 players on and lag in the second-range block placing works pretty decent/reliable (no disappearing nodes)
* :+1: Maybe i'm biased but the game feels pretty snappy with your changes
* :-1: Sadly there were issues with blocks placed but only to see them gone after a logout/login (i noticed a few myself and other players reported the same issue)

### Lag stats
Here the lag stats with a full server:
(i don't know if your commits made a change but i have the feeling the lag is better now)

https://monitoring.minetest.land/d/YUpouLmWk/overview?orgId=1&from=1560268784137&to=1560271981734
</img>

### Summary

I don't think thats _production_ ready yet but it is certainly a (big!) step in the right direction.
I'm open for testing again if there are improvements :smile:

DS-Minetest:
Tested in singleplayer with simulated slow internet connection. I wasn't able to create ghosting nodes.

HybridDog:
How can I use the util/wireshark/minetest.lua script in wireshark to view the packets and calculate statistics on the mapblock sending frequency and bytes per second?

thomasrudin:
> How can I use the util/wireshark/minetest.lua script in wireshark to view the packets and calculate statistics on the mapblock sending frequency and bytes per second?

I don't think wireshark is a good tool for that..

**If you are familiar with prometheus and grafana:**
I used a fork of mine (https://github.com/thomasrudin-mt/minetest/tree/prometheus) to debug for PR #8686
You can pretty easy extract metrics and display/query/export them: https://pandorabox.io/grafana/d/ifuy9KSZz/minetest-engine?orgId=1

**If not**: how about some simple `cout <<` statements?

**Another option**: i _think_ i saw some builtin metric collection, no idea how to use that though :/

HybridDog:
I recently built a house on the pandorabox server and nodes dis- and reappeared during placement very often (issue #7098).

> Sadly there were issues with blocks placed but only to see them gone after a logout/login (i noticed a few myself and other players reported the same issue)

I don't know how this happens. 
Is it possible that in minetest network packets arrive in the wrong order?

thomasrudin:
> I recently built a house on the pandorabox server and nodes dis- and reappeared during placement very often (issue #7098).

I saw that issue a few times during high load (high globalstep cpu usage).
Haven't been able to reproduce that reliably but it happens more often when the `Receive()` method gets called fewer times (https://github.com/minetest/minetest/blob/master/src/server.cpp#L102) due to more work in the `AsyncRunStep()` (https://github.com/minetest/minetest/blob/master/src/server.cpp#L102).

</img>

Source: https://pandorabox.io/grafana/d/ifuy9KSZz/minetest-engine?orgId=1&from=1565747594347&to=1565749532565

I'm opening an issue for that if i have a repro, but i had the same issue when the Pathfinder had hickups and used 100ms per globalstep :confused: 

> Is it possible that in minetest network packets arrive in the wrong order or get lost?

I suspect the problem is in the client->server packet transit not the reverse (digging received late due to laggy package processing)

HybridDog:
rebased.

rubenwardy:
Nice, I like this

> Is it possible that in minetest network packets arrive in the wrong order or get lost?

You need to make sure the packets are sent reliably - I'm not how Minetest does this, but you probably need to opt-in

src/server.cpp
```diff
@@ -2341,6 +2341,35 @@ void Server::sendMetadataChanged(const std::list<v3s16> &meta_updates, float far
...
+		return;
+	}
+	NetworkPacket pkt(TOCLIENT_ADDNODE, 6 + 2 + 1 + 1 + 1, peer_id);
+	pkt << p << n.param0 << n.param1 << n.param2 << (u8) 0;
+	Send(&pkt);
+}
```
sfan5, line 2371: have you considered making these two functions lambdas inside `handleCommand_Interact`?

HybridDog: I have changed them to lambda expressions now. Thanks for the advice.

rubenwardy:
> Sadly there were issues with blocks placed but only to see them gone after a logout/login (i noticed a few myself and other players reported the same issue)

Is this still a problem?

Pevernow:
@rubenwardy Can this PR be used only by changing the base?

If you only need to change the base, I can help improve this pr

Desour:
In case this PR gets adopted at some point:
Related:  #10410

PR #7633, opened 2018-08-09
Compute clarity per vertex instead of per fragment

numberZero:
This should improve performance significantly. See [my comment](https://github.com/minetest/minetest/issues/4335#issuecomment-411903979) in another issue for some data.

### DIFF
--- a/client/shaders/nodes_shader/opengl_fragment.glsl
+++ b/client/shaders/nodes_shader/opengl_fragment.glsl
@@ -3,12 +3,12 @@ uniform sampler2D normalTexture;
 uniform sampler2D textureFlags;
 
 uniform vec4 skyBgColor;
-uniform float fogDistance;
 uniform vec3 eyePosition;
 
 varying vec3 vPosition;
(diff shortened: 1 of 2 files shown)


Fixer-007:
No fps boost for me on i5-3450 / HD 2500 in that heavy jungle fly straight for 3 minutes test (literally within stat error).

lhofhansl:
Unfortunately I also see no difference.

I'd expect it depends on the view distance, and on the resolution used.

You could imagine a scenario of a far view range where you'd end up with more vertexes than fragments. (Fragments are function of the output device - pixels, vertexes are a function of the complexity of the scene... Just to state the obvious :) ).

numberZero:
Oh well, I use view range of 30–70 nodes usually...
So nvm, this is probably not the right thing to do.

lhofhansl:
It's never entirely clear to me what's best done at the fragment and what's best at the vertex level.
If you have a super high-res screen and short view range this could be significant. If we even do LOD then do more things at the vertex level also would be beneficial.

PR #7651, opened 2018-08-16
Simpler software inventorycube

### LINKED ISSUE #5296: Android: Missing/wrong cart texture

numberZero:
Fixes #5296, closes #7486: this is mostly the same, with a bit different and more predictable rendering. Look at cart and ghoststone:
screenshot_20180817_162640

Screenshot of previous version:
screenshot_20180817_001238


SmallJoker:
I see what you mean. They look squashed and more like a 3/4 slab than an entire 1x1x1 cube.

numberZero:
Updated&rebased. Now should look better.

src/client/tile.cpp
```diff
@@ -775,205 +770,145 @@ void TextureSource::rebuildImagesAndTextures()
...
-			errorstream<<"TextureSource::generateTextureFromMesh(): "
-				<<"EVDF_RENDER_TO_TARGET not supported."<<std::endl;
-			warned = true;
+	/* Draw top image.
+	 *
+	 * Left edge spans lines `0` to `size - 1`, and whole image spans
+	 * lines from `0` to `2 * size - 2` inclusive.
+	 */
+	source = lock_image(top);
+	for (int v = 0; v < size; v++) {
```
SmallJoker: GCC 7.3.0:
```
src/client/tile.cpp:843:20: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
  for (int v = 0; v < size; v++) {
                  ~~^~~~~~
```

SmallJoker:
comparison
```Lua
minetest.register_on_mods_loaded(function()
for name, def in pairs(core.registered_nodes) do
	local can_replace = false
	if def.drawtype == "normal"
			and type(def.tiles) == "table" then
		can_replace = true
	end
	if can_replace then
		for i, tile in pairs(def.tiles) do
			if type(tile) ~= "string" then
				can_replace = false
				break
			end
		end
	end
	if can_replace then
		if #def.tiles == 1 or name == "bones:bones" then
			core.override_item(name, {
				-- Top, left, right
				inventory_image = core.inventorycube(def.tiles[1], def.tiles[1], def.tiles[1])
			})
		elseif #def.tiles == 3 then
			core.override_item(name, {
				-- Top, left, right
				inventory_image = core.inventorycube(def.tiles[1], def.tiles[3], def.tiles[3])
			})
		elseif #def.tiles == 6 then
			core.override_item(name, {
				-- Top, left, right
				inventory_image = core.inventorycube(def.tiles[1], def.tiles[6], def.tiles[3])
			})
		end
	end
end
end)
```

The proportions are good now, but the textures seem to have a low resolution.

paramat:
SmallJoker in your images the inventory cube image exceeds it's border.

numberZero:
> * PR 7651 = `core.inventorycube`
> * Master = software rendered node cubes (look right)

@SmallJoker Actually, `core.inventorycube` is what does software (in this PR) rendering. It takes three images and draws a cube. The resulting image may then be used as an inventory image.

Default node images, on the other hand, aren’t actually images. They are meshes, and they are rendered (by the GPU) right in the inventory window, without any intermediate stage. That’s why they don’t get distorted when you resize window: they are rendered each frame, always at the correct size.

SmallJoker:
@numberZero Would it theoretically be possible to generate `core.inventorycube` with a higher resolution to reduce the distorted effect? That they exceed the image button borders not pretty either but it's less of a concern.

numberZero:
@SmallJoker I think that won’t help, but will try.

> they exceed the image button borders

Just like rails, because I don’t leave any margins on the texture. I can, if desired, just by using a larger texture (it is non-power-of-2 anyway; well, won’t that be a problem on Android?).

paramat:
Compared to #7486 this looks much better because of the cube proportions are a better match for the common mesh-rendered cubes. It also looks less distorted than #7486.
I would certainly prefer the cube images do not exceed the inventory borders.

paramat:
@SmallJoker what mod does this image come from?:

comparison

Note that other items exceed the borders too, so this seems a mod formspec error, as long as it looks ok in inventory and MTG creative inventory it's ok.

paramat:
So assuming that image is irrelevant this can probably be merged.

paramat:
GUI scaling setting set to 1, game window sizes from tiny to HD1080.
Here the acacia trunk node uses inventory cube also.

screenshot_20180905_084840

screenshot_20180905_084640

screenshot_20180905_084328

At all window sizes the new inventory cube image is consistent in size and proportion with the mesh rendered cubes, a little distorted yes but not much, good enough.
As long as the code is good i think this can be merged.

paramat:
Now see inventory cube in current master, again cart and acacia tree:

screenshot_20180905_085834

The proportions are different to the rendered mesh cubes and they are also distorted, possibly look worse than this PR.

paramat:
'row 0-15' terminal messages will need removing of course.

SmallJoker:
> @SmallJoker what mod does this image come from?:

@paramat Once with [this mod](https://github.com/minetest/minetest/pull/7651#issuecomment-414044106) and once without (make sure it depends on `default`). Then invest 5 minutes into GIMP to chop the before/after images correctly, add labels to the images and use them as layers for a new gif file export. This is much more helpful than showing a side-by-side comparison.

> Note that other items exceed the borders too, so this seems a mod formspec error, as long as it looks ok in inventory and MTG creative inventory it's ok.

Right, I didn't notice that. Not an issue is this PR.

@numberZero Afaik bad non-power-of-2 texture scaling was only a problem on OpenGL 1.4 or lower; increasing the resolution for the final image would be great because this blurry appearance is not nice at all.

paramat:
Yeah i don't know why your images are blurry, i don't get that.

paramat:
Can anyone test on Android?

paramat:
Retested PR on desktop PC with a changed 'GUI scaling', with no scaling filter and both scaling filters, and with various window sizes. Images still look fine and not blurry, i don't know how that blurriness occurred @SmallJoker

SmallJoker:
@paramat I think you've missed that I used the `core.inventorycube` function to generate those images. The software inventory cube drawing code from this PR works fine, but the formspec function needs fixing somewhen too (although could be done in a separate PR).

paramat:
My testing was of a node with:
```
inventory_image = minetest.inventorycube("default_acacia_tree_top.png",
		"default_acacia_tree.png", "default_acacia_tree.png"),
```
How is that different to your code which does:
```
			core.override_item(name, {
				-- Top, left, right
				inventory_image = core.inventorycube(def.tiles[1], def.tiles[6], def.tiles[3])
			})
```
?

paramat:
Currently we're wondering how SmallJoker gets fuzziness and i don't, please can anyone else test?

numberZero:
@SmallJoker Try [a patch](https://gist.github.com/numberZero/f77ec029b86dd31b00d3b70c06907fa5) to this PR to render in double resolution (NB: the last hunk is to save images in /tmp, you may or may not want that).

Also, here are some images rendered:

master | PR | PR+patch
--|--|--
 |  | 
 |  | 

Sadly my current (Intel) GPU forces filtering...

paramat:
That's good, but surely the issue to fix is image buttons, since this PR works fine for inventory. Then we can avoid this additional resolution, which seems excessive.

paramat:
Unified inventory with cart and acacia tree using inventorycube:

screenshot from 2018-09-24 20-45-56

^ Big game window

screenshot from 2018-09-24 20-47-05

^ 1024x600 window

In both the images are clear and fine. No idea why SmallJoker gets fuzziness.

paramat:
I may be able to +1 this due to my testing but would like another core dev to check the code as i don't completely understand it.

paramat:
@sfan5 tested this PR (the low res version) and got fuzziness using unified inventory.

stujones11:
I have tested this PR on PC and android, here are the results:

**Android before**
before_android
**Android after**
after_android
**PC before**
before_pc
**PC after**
after_pc

So this does fix the inverted minecart issue for me on android and I can not tell any difference with the PC version. However, I also seem to recall that non power-of-2 resolutions can be problematic for some devices or perhaps older versions? What I do know is that in other areas of the android porting, we go to great lengths to ensure power-of-2 numbers so this certainly needs more android testers.

paramat:
The high-res solution may well be justified now. My coment was out of date.

numberZero:
> The high-res solution may well be justified now. My coment was out of date.

Pushed the patch.

paramat:
I'll retest.

paramat:
Tested MTG creative inventory and unified inventory mod at all possible game window sizes. Looks fine.

src/client/tile.cpp
```diff
@@ -775,205 +770,125 @@ void TextureSource::rebuildImagesAndTextures()
...
+	}));
+
+	// It must be divisible by 4, to let everything work correctly.
+	// But it is a power of 2, so being at least 4 is the same.
+	if (size <= 4)
+		size = 4;
+
+	// We don't need *that* large cube textures... I suppose, at least -- numzero
+	if (size >= 64)
+		size = 64;
```
SmallJoker: `rangelim` would fit better for these two checks (or `clamp`?)

src/client/tile.cpp
```diff
@@ -775,205 +770,125 @@ void TextureSource::rebuildImagesAndTextures()
...
+			std::initializer_list<v2s16> offsets) -> void {
+		u32 brightness = core::clamp<u32>(256 * shade_factor, 0, 256);
+		u32 const *source = lock_image(image);
+		for (u16 v = 0; v < size; v++) {
+			for (u16 u = 0; u < size; u++) {
+				video::SColor pixel(*source);
+				applyShadeFactor(pixel, brightness);
+				s16 x = xu * u + xv * v + x1;
+				s16 y = yu * u + yv * v + y1;
+				for (auto const &off: offsets)
```
SmallJoker: `for (auto const &off : offsets)`

stujones11:
I am still a little bit concerned about the non power-of-2 texture. While this has never been an issue for my device, that does not mean it will work for everyone.

https://github.com/minetest/minetest/blob/5f0c969485577c8fbe35b54505e37bace0516a84/src/guiscalingfilter.cpp#L117-L119

numberZero:
Found that OpenGL ES 1 doesn’t support NPOT textures by itself, that’s an extension.

numberZero:
Found that OpenGL ES 1 doesn’t support NPOT textures by itself, that’s an extension. But the code knows that and scales the image to the next power of 2, if necessary, using `Align2Npot2`.

PR #7182, opened 2018-03-30
Made client-side translations take fallback strings from .en.tr files.

MineWitherMC:
Absolutely trivial thing, fixes #7171. Still not sure why no one thought of this feature :)

PR #7657, opened 2018-08-19
Load translations after media download, add fallback translations

SmallJoker:
~`gettext("LANG_CODE")` returned an empty string, so this PR will now use the `setlocale` output which does not seem to follow any C standard. If you find something more reliable, please add a comment.~
Note to myself: Update the `locale` directory next time when working with gettext.

**What this PR does**
 * ~Now detects the language correctly in most cases~
 * `.en.tr` as fallback for mods which aren't translated entirely (or in another language)
 * Language loading priorities, first `en`, then `pt` and `pt_BR` for the final overwrite if such translations were found. This language priority list may be extended to 255 entries for a custom fallback order
 * Improved documentation

**Testing**
[Here's](https://github.com/SmallJoker/upgrade_packs) a rather simple mod with a few translated strings. Copy&paste the files (or rename) for your language code.

Implements #7171
Replaces PR #7182

### DIFF
--- a/src/network/clientpackethandler.cpp
+++ b/src/network/clientpackethandler.cpp
@@ -132,12 +132,12 @@ void Client::handleCommand_AuthAccept(NetworkPacket* pkt)
 					<< m_recommended_send_interval<<std::endl;
 
 	// Reply to server
-	std::string lang = gettext("LANG_CODE");
-	if (lang == "LANG_CODE")
-		lang = "";
+	auto &locale = getLanguageCodes();
+	NetworkPacket resp_pkt(TOSERVER_INIT2, 1);
+	resp_pkt << (u8)locale.size();
+	for (auto &code : locale)
+		resp_pkt << code;
 
-	NetworkPacket resp_pkt(TOSERVER_INIT2, sizeof(u16) + lang.size());
-	resp_pkt << lang;
 	Send(&resp_pkt);
 
 	m_state = LC_Init;
--- a/src/network/clientpackethandler.cpp
+++ b/src/network/clientpackethandler.cpp
@@ -132,11 +132,9 @@ void Client::handleCommand_AuthAccept(NetworkPacket* pkt)
 					<< m_recommended_send_interval<<std::endl;
 
 	// Reply to server
-	auto &locale = getLanguageCodes();
-	NetworkPacket resp_pkt(TOSERVER_INIT2, 1);
-	resp_pkt << (u8)locale.size();
-	for (auto &code : locale)
-		resp_pkt << code;
+	const std::vector<std::string> &locale = getLanguageCodes();
+	NetworkPacket resp_pkt(TOSERVER_INIT2, 0);
+	resp_pkt << locale;
 
 	Send(&resp_pkt);
 
--- a/src/network/networkpacket.h
+++ b/src/network/networkpacket.h
@@ -22,6 +22,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include <ctime>
 #include "util/pointer.h"
 #include "util/numeric.h"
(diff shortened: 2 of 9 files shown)


sfan5:
> so this PR will now use the setlocale output which does not seem to follow any C standard

clearly the better alternative here is to <i>make</i> `gettext("LANG_CODE")` return something

src/network/serverpackethandler.cpp
```diff
@@ -287,9 +287,18 @@ void Server::handleCommand_Init2(NetworkPacket* pkt)
 	m_clients.event(pkt->getPeerId(), CSE_GotInit2);
 	u16 protocol_version = m_clients.getProtocolVersion(pkt->getPeerId());
 
-	std::string lang;
-	if (pkt->getSize() > 0)
-		*pkt >> lang;
+	// Receive the client locale
+	std::vector<std::string> locale;
+	u8 locale_count = 0;
+	try {
```
nerzhul: i suggest a more interesting thing, learn NetworkPacket to serialize/deserialize vectors by adding operators << and >>

SmallJoker: Thanks for the suggestion. It is indeed a much better solution to have this part as a generic part of NetworkPacket.

src/network/networkpacket.h
```diff
@@ -118,6 +119,32 @@ class NetworkPacket
 	NetworkPacket &operator>>(video::SColor &dst);
 	NetworkPacket &operator<<(video::SColor src);
 
+	template<typename T> NetworkPacket &operator >> (std::vector<T> &dst)
+	{
+		u16 count;
```
nerzhul: i think we really should use u32, 2 bytes for a such thing is nothing, breaking protocol because we send 80k elements is :)

src/network/networkpacket.h
```diff
@@ -118,6 +119,32 @@ class NetworkPacket
...
+		*this >> count;
+
+		while (count-- > 0) {
```
nerzhul: is underflow possible if count == 0 ?

SmallJoker: No, `count--` ensures that the value before decreasing is checked against `> 0`.

src/network/networkpacket.h
SmallJoker: Lists longer than 65k entries will be rare, but I will extend it just for the case.

SmallJoker: C99 ensures that `size_t` must be at least a `u16`.
http://irc.minetest.net/minetest-dev/2018-08-21#i_5382824

SmallJoker:
@nerzhul Your comments are handled for a long time already. Could you please update your review?

src/network/clientpackethandler.cpp
```diff
@@ -132,12 +132,10 @@ void Client::handleCommand_AuthAccept(NetworkPacket* pkt)
 					<< m_recommended_send_interval<<std::endl;
 
 	// Reply to server
-	std::string lang = gettext("LANG_CODE");
-	if (lang == "LANG_CODE")
-		lang = "";
+	const std::vector<std::string> &locale = getLanguageCodes();
+	NetworkPacket resp_pkt(TOSERVER_INIT2, 0);
+	resp_pkt << locale;
```
nerzhul: network breakage

SmallJoker: Indeed. This PR was created during the unstable network phase. Will have to add guards around this code.

Wuzzy2:
If this implemens #7171 as described, then thumbs down :-1: due to same reason as posted here:

https://github.com/minetest/minetest/issues/7171#issuecomment-537276219

SmallJoker:
RIP

PR #7677, opened 2018-08-25
Add whereis chat-command

ClobberXD:
- Syntax: `/whereis [<name>]`
- Returns coordinates of `name` with 3-digit precision
- If no `name` provided, returns position of caller
- Intended for use by moderators only. (Should this be changed?)

Tested all possible possible inputs. Works as expected.


Ezhh:
Just to note: This is very easy to abuse on certain types of server, to the extent I've had to disable the version provided by the IRC mods. It definitely should not be added without requiring a priv (and I'm not sure it's needed in builtin at all).

builtin/game/chatcommands.lua
```diff
@@ -1056,3 +1056,23 @@ core.register_chatcommand("kill", {
 		return handle_kill_command(name, param == "" and name or param)
 	end,
 })
+
+core.register_chatcommand("whereis", {
+	params = "[<name>]",
+	description = "Show location of yourself or another player",
+	privs = {kick = true, ban = true},
```
Ezhh: Why these privs? Moderators on my survival server often tell me they don't want privs that allow unfair gameplay advantages. Being able to locate a player would give them this.

ClobberXD: Good point. But it's still provided by the IRC mod without any restrictions, and I feel that it should definitely have some sort of restriction. Should I make it require `server` like `/kill`?

BluebirdGreycoat:
Probably better to just add a new priv, `whereis`. Since privileges represent capability, not rank.

There's no inherent reason this would be bad for a survival server. I've often wanted a version of whereis that, in addition to showing player coordinates, would also show me names of nearby players, and whose protected area they are currently in.

In fact maybe I'll go code that right now ....

tenplus1:
While I wouldn't give out the position of a player using /whereis incase of foul play like teleporting to their locations, you could simply calculate and show how many blocks they are away from one another, and have the position shown only to 'server' priv owners.

ClobberXD:
Updated to require `server` priv

BluebirdGreycoat:
test

Not accessible to players, of course.

Megaf:
I think the addition of this command is valid and welcomed.
As long as the priv whereis/locate/whatever is required to run it.
Codewise it seems alright.

ClobberXD:
Updated command to only require the `server` priv if a param (player name in this case) has explicitly been passed to `whereis`. So now, `server` isn't required if a player uses `whereis` without any params (i.e. know their own coords)

Ezhh:
So now we double the number of ways players can get their location, even when the server may not want players to know it (think mazes) as well.

ClobberXD:
Good point, again. So a separate priv (`whereis`?) would be the way to go?

BluebirdGreycoat:
I think a separate priv is good.

@Ezhh, players can already display their coordinates in the F5 debug info. If there is a way to turn this off server-side, I would love to know. ;-)

Ezhh:
@BluebirdGreycoat In general I agree with you, but my understanding is that only commands that will be universally wanted no matter the type of game should go into builtin. 

(And yes, I know F5 can do that. I also want a way to prevent F5 doing that as well because I'd love to be able to disable coords from showing  under certain conditions.)

> I think a separate priv is good.

A separate priv would probably be best for this if it is added.

Ezhh:
(Also as an aside - it's not really opt in as such, because it will have the knock on effect of changing IRC mod behaviour, and then on servers where I do want people able to use whereis, I have  to take actions (and so opt out) to prevent what this causes.)

Megaf:
I don't really get the negativism against this command...

What I mean is, it will only be available to admins and moderators, **they can already perhaps even teleport to the player. So why is it such a huge deal to see player coords?**

**Perhaps making teleport priv required to use this command would make everyone happy?**

paramat:
> So now, server isn't required if a player uses whereis without any params (to know their own coords)

No thanks :) I feel either 'server' priv only or a new dedicated priv.

ClobberXD:
Reverted previous change that allowed players without the server priv to know their own coords. It's also worth noting that @nerzhul approved this PR at this stage only, and so his review is still valid.

Ezhh:
@paramat You've had two server admins here so far say they prefer a separate priv. It would at least have some use for me on one of my servers that way, (though I still feel builtin is the wrong place for this command anyway).

paramat:
Yeah i'm staying neutral on the priv issue now.

DS-Minetest:
Only usable with server priv, this chatcommand is not really useful. Moderators normally won't have the server priv afaik.
Needing the teleport priv would be acceptable imo but there's the problem that a server might want to allow its players to know the positions of each other.
What about giving players the ability to hide?

BluebirdGreycoat:
> What about giving players the ability to hide with a chatcommand?

That's definitely getting into mod territory. IMO, if added, it should be kept minimalist. We do have `minetest.override_chatcommand` if anyone needs something more for their server.

ClobberXD:
From what I can tell, one thing still remains to be decided upon: `server` or a separate `whereis` priv?

DS-Minetest:
<details>

ClobberXD wrote:
>So a separate priv (whereis?) would be the way to go?

Megaf wrote:
>Perhaps making teleport priv required to use this command would make everyone happy?

Paramat wrote:
>We really should avoid adding privs unless there is a significant need
</details>

Imo needing teleport priv would be a good compromise.
But do we need this chatcommand at all?
<details><summary>Let's see what the most here think.</summary>

Emote
- :+1: for server priv
- :-1: for not wanting this command be added at all
- :Hooray: for no priv needed
- :smile: for extra whereis priv
- :confused: for teleport priv
- :heart: if you hate me for doing this

(Take as many as you want if you are not sure.)
</details>

Megaf:
Teleport priv would be by far the most logical to have in order use whereis.

ClobberXD:
Btw, the `teleport` priv only allows the player to teleport themselves (`bring` is required to teleport other players). So I'm not too convinced about making this command require `teleport`.

I've added a `whereis` priv, which I feel gives admins the freedom to selectively allow players other than themselves to use this priv on an opt-in basis.

sorcerykid:
As a server operator, I'm of the opinion that only essential chat commands should be added to builtin, anything else should be left to mods. This isn't a bad idea in principle, but I'm just not convinced it really qualifies as essential, particularly since it necessitates yet another privilege.

Incidentally, the JT2 server has a /where command (part of my xcommands mod) that shows position,  distance, direction, and relative elevation of other players and is available to regardless of rank. But I also limited it to a range of 1000 meters to prevent abuse. Of course the range is unlimited for moderators and administrators. I've found this to be a happy medium in terms of both utility and security.

image

https://forum.minetest.net/viewtopic.php?p=303848#p303848

paramat:
There seems to be a feeling that this isn't important enough to add to builtin, and server owners tend to provide their own versions of this, so it seems perhaps best to not merge this.

Wuzzy2:
With `whereis` cmd and priv, I am neutral.
With `whereis` cmd and `server` priv, I am opposed (don't bloat `server`, it already has too many features at once).

PR #7683, opened 2018-08-28
Add commonly used data structures to builtin

HybridDog:
Implementation of #7490

### DIFF
--- a/builtin/common/datastructures.lua
+++ b/builtin/common/datastructures.lua
@@ -1,0 +1,317 @@
+local stack_mt
+stack_mt = {
+	__index = {
+		push = function(self, v)
+			self.n = self.n+1
+			self[self.n] = v
+		end,
+		pop = function(self)
+			local v = self[self.n]
+			self[self.n] = nil
+			self.n = self.n-1
+			return v
+		end,
+		is_empty = function(self)
+			return self.n == 0
+		end,
+		size = function(self)
+			return self.n
+		end,
+		clone = function(self, copy_element)
+			local stack
+			if copy_element then
+				stack = {n = self.n, true}
+				for i = 1, self.n do
+					stack[i] = copy_element(self[i])
+				end
+			else
+				stack, n = self:to_table()
+				stack.n = n
+			end
+			setmetatable(stack, stack_mt)
+			return stack
+		end,
+		to_table = function(self)
+			local t = {}
+			for i = 1, self.n do
+				t[i] = self[i]
+			end
+			return t, self.n
+		end,
+		to_string = function(self, value_tostring)
+			if self.n == 0 then
+				return "empty stack"
+			end
+			value_tostring = value_tostring or tostring
+			local t = {}
+			for i = 1, self.n do
+				t[i] = value_tostring(self[i])
+			end
+			return self.n .. " elements; bottom to top: " ..
+				table.concat(t, ", ")
+		end,
+	}
+}
+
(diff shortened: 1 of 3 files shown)


raymoo:
Summary of worst case time complexity of possible implementations:

Old code (n = number of jobs in queue, k = number of ready jobs):
- Add a job: O(1)
- Run and remove all ready jobs: O(kn), which is potentially O(n^2)

This code:
- Add a job: O(log n)
- Run and remove all ready jobs: O(k log n), which is potentially O(n log n)

Linear-scanning priority queue that fixes the performance bug in the current implementation:
- Add a job: O(1)
- Run and remove all ready jobs: O(n)

So depending on the actual workload, the binary heap may have better or worse performance than a linear-scanning priority queue implementation that is better than the current one, but the difference in practice will be small because there are normally not that many jobs in the queue (less than a thousand).

HybridDog:
I assume that the globalstep is executed a lot more frequently than adding or removing a job.
With linear-scanning priority queue searching for an expired job is O(n), whereas with the binary heap it is O(1).
If one job gets added and executed, it's O(1 + n) = O(n) with linear-scanning and O(log(n) + log(n)) = O(log(n)) with the binary heap.
Anyway, the commit is just an example usage of a provided priority queue data structure.

builtin/common/datastructures.lua
```diff
@@ -0,0 +1,317 @@
...
+				return v
+			end
+			-- source is empty, swap it with sink
+			self.source, self.sink = self.sink, self.source
+			self.n_out = self.n_in
+			self.n_in = 0
+			local v = self.source[1]
+			self.source[1] = nil
+			self.p_out = 2
+			return v
```
DS-Minetest: You could swap the uppermost element with the current one and set the uppermost one to `nil` and when you reach the top you can make the source empty from above.

builtin/common/datastructures.lua
```diff
@@ -0,0 +1,317 @@
...
+			for i = 1, self.n do
+				t[i] = value_tostring(self[i])
+			end
+			return self.n .. " elements; bottom to top: " ..
+				table.concat(t, ", ")
+		end,
+	}
+}
+
+function core.create_stack(data)
```
DS-Minetest: Why not `input, n` and then set `n = n or input.n`? Wouldn't that be more efficient?

HybridDog: core.create_stack does not need to be as efficient as possible, it's only invoked when creating a stack, an extendable parameter list passed as table is more desirable.

builtin/common/datastructures.lua
HybridDog: This doesn't improve performance, does it?

DS-Minetest: It could maybe save memory.
It was probably a bad suggestion. >_<

HybridDog:
I've removed the usage examples because I put them only for demonstration purposes.

SmallJoker:
Even though these are pretty basic operations, wouldn't a library mod fit better? I think only a few mods could actually make use of these structures.

HybridDog:
It is possible to copy-paste the datastructures code to a mod (or use git subtree), so I don't know if they are important enough to be implemented directly in minetest. You can close this if a mod (or copied code) fits better.

SmallJoker:
The code is there. Contributors may pick it up at any time they like.
For now this PR first best into a (library) mod.

PR #7684, opened 2018-08-29
 Add core.remove_detached_inventory

SmallJoker:
Adds the possibility to remove detached inventories after initialization.
This PR will also drop the backwards compatibility to make proper use of the `NetworkPacket` operators.

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -2475,33 +2475,42 @@ void Server::sendRequestedMedia(session_t peer_id,
 
 void Server::sendDetachedInventory(const std::string &name, session_t peer_id)
 {
-	if(m_detached_inventories.count(name) == 0) {
-		errorstream<<FUNCTION_NAME<<": \""<<name<<"\" not found"<<std::endl;
-		return;
+	const auto &inv_it = m_detached_inventories.find(name);
+	const auto &player_it = m_detached_inventories_player.find(name);
+
+	if (player_it == m_detached_inventories_player.end() ||
+			player_it->second.empty()) {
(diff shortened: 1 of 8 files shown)


src/server.cpp
```diff
@@ -2475,33 +2475,41 @@ void Server::sendRequestedMedia(session_t peer_id,
...
-	Inventory *inv = m_detached_inventories[name];
-	std::ostringstream os(std::ios_base::binary);
+	const auto &inv_it = m_detached_inventories.find(name);
+	const auto &player_it = m_detached_inventories_player.find(name);
 
-	os << serializeString(name);
-	inv->serialize(os);
+	if (player_it == m_detached_inventories_player.end() ||
+			player_it->second.empty()) {
+		// ok. no restriction
```
sfan5: i wouldn't call these "restrictions"
<pre>
		// Send to everyone
</pre>

src/server.cpp
```diff
@@ -2475,33 +2475,41 @@ void Server::sendRequestedMedia(session_t peer_id,
...
+	} else {
+		RemotePlayer *p = m_env->getPlayer(player_it->second.c_str());
+		if (!p)
+			return; // Player is offline
 
-	// Make data buffer
-	std::string s = os.str();
+		if (peer_id != PEER_ID_INEXISTENT && peer_id != p->getPeerId())
+			return; // Send to nobody
```
sfan5: this is hard to grasp on first sight
<pre>
			return; // Caller requested send to a different player, so don't send.
</pre>

PR #7688, opened 2018-09-01
Add disable_jump to liquids and ladders

SmallJoker:
Implements feature requested in #4494

Jumping is now no longer possible when the group `disable_jump = 1` is set. Jumping at the bottom of liquids may look a bit weird, but is actually because the node below can be used to jump.

**Testing code**
```Lua
-- river water for comparison
core.override_item("default:water_source", {
	groups = {
		disable_jump = 1,
		water = 3, liquid = 3, puts_out_fire = 1, cools_lava = 1
	}
})

core.override_item("default:water_flowing", {
	groups = {
		disable_jump = 1,
		water = 3, liquid = 3, puts_out_fire = 1,
		not_in_creative_inventory = 1, cools_lava = 1
	}
})

core.override_item("default:leaves", {
	groups = {
		disable_jump = 1,
		cracky = 3
	}
})
-- steel ladder for comparison
core.override_item("default:ladder_wood", {
	groups = {
		disable_jump = 1,
		cracky = 3
	}
})
```

### DIFF
--- a/src/client/localplayer.cpp
+++ b/src/client/localplayer.cpp
@@ -309,7 +309,7 @@ void LocalPlayer::move(f32 dtime, Environment *env, f32 pos_max_d,
 			collision_info->push_back(colinfo);
 
 			if (colinfo.type != COLLISION_NODE ||
-					colinfo.new_speed.Y != 0 ||
+					colinfo.axis != COLLISION_AXIS_Y ||
 					(could_sneak && m_sneak_node_exists))
 				continue;
 
@@ -435,20 +435,15 @@ void LocalPlayer::move(f32 dtime, Environment *env, f32 pos_max_d,
 	/*
 		Check properties of the node on which the player is standing
 	*/
-	const ContentFeatures &f_under = nodemgr->get(map->getNodeNoEx(m_standing_node));
-	const ContentFeatures &f_above = nodemgr->get(map->getNodeNoEx(
-		floatToInt(m_position + v3f(0, 0.2f * BS, 0), BS)));
+	const ContentFeatures &f = nodemgr->get(map->getNodeNoEx(m_standing_node));
 
 	// Determine if jumping is possible
-	m_disable_jump = itemgroup_get(f_under.groups, "disable_jump");
+	m_disable_jump = itemgroup_get(f.groups, "disable_jump");
 	m_can_jump = ((touching_ground && !is_climbing)
 			|| sneak_can_jump) && !m_disable_jump;
 
-	// Disallow swimming upwards
-	m_disable_jump |= itemgroup_get(f_above.groups, "disable_jump");
-
 	// Jump key pressed while jumping off from a bouncy block
-	if (m_can_jump && control.jump && itemgroup_get(f_under.groups, "bouncy") &&
+	if (m_can_jump && control.jump && itemgroup_get(f.groups, "bouncy") &&
 		m_speed.Y >= -0.5 * BS) {
 		float jumpspeed = movement_speed_jump * physics_override_jump;
 		if (m_speed.Y > 1) {
--- a/src/client/localplayer.cpp
+++ b/src/client/localplayer.cpp
@@ -170,8 +170,8 @@ void LocalPlayer::move(f32 dtime, Environment *env, f32 pos_max_d,
 		std::vector<CollisionInfo> *collision_info)
 {
 	if (!collision_info || collision_info->empty()) {
-		// Node below the feet, update each ClientEnvironment::step()
-		m_standing_node = floatToInt(m_position, BS) - v3s16(0, 1, 0);
+		// Node at feet position, update each ClientEnvironment::step()
+		m_standing_node = floatToInt(m_position, BS);
 	}
 
 	// Temporary option for old move code
@@ -320,6 +320,7 @@ void LocalPlayer::move(f32 dtime, Environment *env, f32 pos_max_d,
 			if (is_first || len < distance) {
 				m_standing_node = colinfo.node_p;
 				distance = len;
+				is_first = false;
 			}
 		}
 	}
@@ -434,15 +435,20 @@ void LocalPlayer::move(f32 dtime, Environment *env, f32 pos_max_d,
 	/*
 		Check properties of the node on which the player is standing
 	*/
-	const ContentFeatures &f = nodemgr->get(map->getNodeNoEx(m_standing_node));
+	const ContentFeatures &f_under = nodemgr->get(map->getNodeNoEx(m_standing_node));
+	const ContentFeatures &f_above = nodemgr->get(map->getNodeNoEx(
+		floatToInt(m_position + v3f(0, 0.2f * BS, 0), BS)));
+
 	// Determine if jumping is possible
-	m_can_jump = (touching_ground && !in_liquid && !is_climbing)
-			|| sneak_can_jump;
-	if (itemgroup_get(f.groups, "disable_jump"))
-		m_can_jump = false;
+	m_disable_jump = itemgroup_get(f_under.groups, "disable_jump");
+	m_can_jump = ((touching_ground && !is_climbing)
+			|| sneak_can_jump) && !m_disable_jump;
+
+	// Disallow swimming upwards
+	m_disable_jump |= itemgroup_get(f_above.groups, "disable_jump");
 
 	// Jump key pressed while jumping off from a bouncy block
-	if (m_can_jump && control.jump && itemgroup_get(f.groups, "bouncy") &&
+	if (m_can_jump && control.jump && itemgroup_get(f_under.groups, "bouncy") &&
 		m_speed.Y >= -0.5 * BS) {
 		float jumpspeed = movement_speed_jump * physics_override_jump;
 		if (m_speed.Y > 1) {
@@ -637,18 +643,14 @@ void LocalPlayer::applyControl(float dtime, Environment *env)
 				setSpeed(speedJ);
 				m_client->getEventManager()->put(new SimpleTriggerEvent(MtEvent::PLAYER_JUMP));
 			}
-		}
-		else if(in_liquid)
-		{
-			if(fast_climb)
+		} else if (in_liquid && !m_disable_jump) {
+			if (fast_climb)
 				speedV.Y = movement_speed_fast;
 			else
 				speedV.Y = movement_speed_walk;
 			swimming_vertical = true;
-		}
-		else if(is_climbing)
-		{
-			if(fast_climb)
+		} else if (is_climbing && !m_disable_jump) {
+			if (fast_climb)
 				speedV.Y = movement_speed_fast;
 			else
 				speedV.Y = movement_speed_climb;
@@ -1035,13 +1037,20 @@ void LocalPlayer::old_move(f32 dtime, Environment *env, f32 pos_max_d,
 	/*
 		Check properties of the node on which the player is standing
 	*/
-	const ContentFeatures &f = nodemgr->get(map->getNodeNoEx(getStandingNodePos()));
+	const ContentFeatures &f_under = nodemgr->get(map->getNodeNoEx(
+		getStandingNodePos()));
+	const ContentFeatures &f_above = nodemgr->get(map->getNodeNoEx(
+		floatToInt(m_position + v3f(0, 0.2f * BS, 0), BS)));
+
 	// Determine if jumping is possible
-	m_can_jump = touching_ground && !in_liquid;
-	if (itemgroup_get(f.groups, "disable_jump"))
-		m_can_jump = false;
+	m_disable_jump = itemgroup_get(f_under.groups, "disable_jump");
+	m_can_jump = touching_ground && !m_disable_jump;
+
+	// Disallow swimming upwards
+	m_disable_jump |= itemgroup_get(f_above.groups, "disable_jump");
+
 	// Jump key pressed while jumping off from a bouncy block
-	if (m_can_jump && control.jump && itemgroup_get(f.groups, "bouncy") &&
+	if (m_can_jump && control.jump && itemgroup_get(f_under.groups, "bouncy") &&
 			m_speed.Y >= -0.5 * BS) {
 		float jumpspeed = movement_speed_jump * physics_override_jump;
 		if (m_speed.Y > 1) {
--- a/src/client/localplayer.cpp
+++ b/src/client/localplayer.cpp
@@ -892,6 +892,12 @@ void LocalPlayer::old_move(f32 dtime, Environment *env, f32 pos_max_d,
 		pos_max_d, m_collisionbox, player_stepheight, dtime,
 		&position, &m_speed, accel_f);
 
+	// Positition was slightly changed; update standing node pos
+	if (touching_ground)
+		m_standing_node = floatToInt(m_position - v3f(0, 0.1f * BS, 0), BS);
+	else
+		m_standing_node = floatToInt(m_position, BS);
+
 	/*
 		If the player's feet touch the topside of any node, this is
 		set to true.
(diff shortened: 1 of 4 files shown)


paramat:
Why disable jump for climbable nodes, is there a need or use? Does that stop climbing upwards or just stop jumping off?

Wuzzy2:
I have tested this. Nice!

Works fine for ladders. A ladder that you can only climb down, but not up? Funny. :-) This will be interesting for puzzle games, like @sofar's Inside The Box. :D

Liquids behave a bit weird. Yes, you cannot rise in liquids, this part works fine. But if you are at the bottom of the water (in your test mod), you will rise a tiny bit. Not enough to climb a node, but it sure looks weird, I think it should be fixed. Otherwise fine.

Please update `lua_api.txt` about the special meaning of `disable_jump=1` for ladders and liquids.

Maybe instead of `disable_jump=1`, this should use `disable_ascent=1` for a more fitting group name. You don't really “jump” in liquids and ladders.

---
One idea (optional):
I feel there should be a group that works like this, but for the downwards direction as well. I.e. `disable_descent=1`. For ladders, this woud prevent climbing down, for liquids, this would prevent *manual* (!) sinking. The reason for this is consistency. If you can prevent going up, there should also be a way to prevent going down.

You don't have to do this for me to approve, I just think that with a `disable_descent`, this would feel more “complete”. :)

SmallJoker:
> But if you are at the bottom of the water (in your test mod), you will rise a tiny bit. Not enough to climb a node, but it sure looks weird, I think it should be fixed.

Post above:
> Jumping at the bottom of liquids may look a bit weird, but is actually because the node below can be used to jump.

The player uses the node at its feet to repel from the ground, the liquid viscosity then causes the player to slow down quickly. Intended behaviour.
----
I agree that `disable_ascent` and `disable_descent` would be more fitting group names. I will add the required compatibility code to the item registration and override part soon.

Wuzzy2:
Thanks, awesome!

The minijump on sea floor is not a big deal and I can accept it as-is. It's not really *wrong* anyway.

paramat:
What's the usecase of no ascent in climbable nodes? Why is that included here? Just wondering.

Wuzzy2:
One-way climbable blocks for puzzle games. But I didn't have any concrete use case in mind. It's really more about symmetry than anything else.

src/localplayer.cpp
```diff
@@ -1038,13 +1040,20 @@ void LocalPlayer::old_move(f32 dtime, Environment *env, f32 pos_max_d,
...
+		getStandingNodePos()));
+	const ContentFeatures &f_above = nodemgr->get(map->getNodeNoEx(
+		floatToInt(m_position + v3f(0, 0.2f * BS, 0), BS)));
+
 	// Determine if jumping is possible
-	m_can_jump = touching_ground && !in_liquid;
-	if (itemgroup_get(f.groups, "disable_jump"))
-		m_can_jump = false;
+	m_disable_jump = itemgroup_get(f_under.groups, "disable_jump");
+	m_can_jump = touching_ground && !m_disable_jump;
```
paramat: Seems a possible change in behaviour in old_move (very dangerous), previously `m_can_jump` depended on `in_liquid` also.

paramat:
Looking at the code, also disabling climbing just uses an already calculated bool, so does no harm.

I'm uncomfortable with the changes to 'old_move' in this PR. They seem to possibly change behaviour and since this is a new feature i'm not sure about including it in old move.

src/localplayer.cpp
```diff
@@ -170,8 +170,8 @@ void LocalPlayer::move(f32 dtime, Environment *env, f32 pos_max_d,
 		std::vector<CollisionInfo> *collision_info)
 {
 	if (!collision_info || collision_info->empty()) {
-		// Node below the feet, update each ClientEnvironment::step()
-		m_standing_node = floatToInt(m_position, BS) - v3s16(0, 1, 0);
+		// Node at feet position, update each ClientEnvironment::step()
+		m_standing_node = floatToInt(m_position, BS);
```
paramat: Why change this, was it a bug?

src/localplayer.cpp
```diff
@@ -327,6 +327,7 @@ void LocalPlayer::move(f32 dtime, Environment *env, f32 pos_max_d,
 			if (is_first || len < distance) {
 				m_standing_node = colinfo.node_p;
 				distance = len;
+				is_first = false;
```
paramat: Bugfix?

src/localplayer.cpp
```diff
@@ -438,15 +439,20 @@ void LocalPlayer::move(f32 dtime, Environment *env, f32 pos_max_d,
...
+		floatToInt(m_position + v3f(0, 0.2f * BS, 0), BS)));
+
 	// Determine if jumping is possible
-	m_can_jump = (touching_ground && !in_liquid && !is_climbing)
-			|| sneak_can_jump;
-	if (itemgroup_get(f.groups, "disable_jump"))
-		m_can_jump = false;
+	m_disable_jump = itemgroup_get(f_under.groups, "disable_jump");
+	m_can_jump = ((touching_ground && !is_climbing)
+			|| sneak_can_jump) && !m_disable_jump;
```
paramat: Seems a possible change in behaviour in old_move (very dangerous), previously m_can_jump depended on in_liquid also.

src/localplayer.cpp
SmallJoker: So did the new sneak code. Surely, I can undo these changes but in this case the `disable_ jump` in liquids will only work for people who use the new sneak code.

src/localplayer.cpp
SmallJoker: In liquids, the player usually does not touch ground. And when they do, it seems more realistic so that they can  push off the ground - also in liquids which don't allow moving up when floating.

src/localplayer.cpp
SmallJoker: Yes, but ingame it doesn't matter too much because `move()` is called multiple times per dtime, whereas the second call corrected this issue.

src/localplayer.cpp
SmallJoker: Why is it obviously wrong? The player's feet are now at `0,0,0` and this code is relevant when there are no collisions (falling down, liquids). Imagine the player is floating on a 1/16th slab (`free_move` enabled) - the current code will return the feet position (slab node) instead of the one below (dirt). This was _obviously wrong_ before.

Wuzzy2:
I agree, legacy movement code should not be touched for this. Can you just implement this feature without touching old_move at all? As far I understand, old_move is legacy and does not need touching anyway.

SmallJoker:
@Wuzzy2 Well sure, that's doable. But then I would also note in the Lua API that this group's full functionality is only available to new_move users.

src/localplayer.cpp
paramat: Hmmm .. well, it's a minor feature, not much of a loss. I think it's best to not alter old_move.

src/localplayer.cpp
```diff
@@ -438,15 +439,20 @@ void LocalPlayer::move(f32 dtime, Environment *env, f32 pos_max_d,
 	/*
 		Check properties of the node on which the player is standing
 	*/
-	const ContentFeatures &f = nodemgr->get(map->getNodeNoEx(m_standing_node));
+	const ContentFeatures &f_under = nodemgr->get(map->getNodeNoEx(m_standing_node));
+	const ContentFeatures &f_above = nodemgr->get(map->getNodeNoEx(
+		floatToInt(m_position + v3f(0, 0.2f * BS, 0), BS)));
```
paramat: Is it possible to avoid checking the nodedefs of 2 nodes instead of 1?

src/localplayer.cpp
paramat: I don't think the realism of pushing off ground when in liquid is worth pursuing, liquid resistance will make such jumps pointless anyway. I'd rather not change this behaviour if it's not essential to the feature.

FaceDeer: If this isn't supported then I won't be able to use this feature to create a "poison gas" liquid that the player can move through as if it were air, as I had hoped. A player standing in non-breathable gas should still be able to jump normally if they're standing on a solid surface.

src/localplayer.cpp
paramat:
```
	/*
		Check properties of the node on which the player is standing
	*/
	const ContentFeatures &f = nodemgr->get(map->getNodeNoEx(m_standing_node));
	// Determine if jumping is possible
```
Yeah the point should be slightly below feet position, somewhere in code (lighting probably) we use a tiny offset to move the point below the edge of a node.

paramat:
I have several requests before i can remove my -1, see line comments.

src/localplayer.cpp
SmallJoker: According to the collision detection code, distances up to 0.15m away from the next colliding face is considered as "ground".
https://github.com/minetest/minetest/blob/622e2235ee700cf92843d2e99a47e36fd4508cf0/src/collision.cpp#L581

Subtracting any value from the position will cause wrong values for thin slabs. Nevertheless, this is only a default value which is overwritten as soon there is any collision. `m_standing_node` is only used *after* the collision detection and the standing node calculation.
Would you please be so nice and show me a location in the code where the standing node position is required to be exactly a bit below the player's feet? They could also stand on the bottom of a ladder, where the ground node doesn't matter at all.

src/localplayer.cpp
paramat: To clarify i'm not necessarily opposed to this aspect, just unsure and being cautious.
Is it possible to make a liquid that the player moves as normal within, assuming jumping is possible within it?
Anyway i'm mostly neutral on this aspect.

src/localplayer.cpp
paramat: Ok thanks for the explanation, this is fine as it is then.

paramat:
2nd commit seems good.
My only remaining concern is the changes to `old_move()`, i don't think it's worth the risk, adding features and bugfixing have a large chance of changing behaviour. So i'm :-1: unless it's left untouched.
Note the feature request had no core dev support for 2 years so is low priority, i'm a little bemused by the scale of changes being made for such a specialist feature :)

src/localplayer.cpp
SmallJoker: Pushing off from the ground will also be possible in regular liquids, causing overall a little faster elevation time (negligible). I know that it alters the current movement behaviour _slightly_ - compare it in a testing build if you think it would be concern,

SmallJoker:
> i'm a little bemused by the scale of changes being made for such a specialist feature :)

After all it was you who requested these changes indirectly. I'm just coding. If you want fewer changed lines, then please adopt this PR to show me that it's possible. I could also leave the old move code untouched if you tell me so. The feature will then be limited to the new move code only.

paramat:
> the scale of changes being made

I think there is a misunderstanding, by that i'm referring to the initial PR, i'm not complaining about coding you have done since to attend to my concerns, which i appreciate of course.
> After all it was you who requested these changes indirectly

If you are referring to the initial PR, i have never requested this feature be coded, i was -1 at first and then neutral, see #4494 No other core dev supported the feature over 2 years, but i assume you support the feature now?
Maybe there has been a misunderstanding, did you think i supported the feature?

For `old_move()`, i feel the feature is not important enough to be worth the risk of changing the behaviour.

paramat:
Ah, maybe it's because i commented: "Any core dev support?", when i do that i'm not expressing support for a request.

SmallJoker:
@paramat Ah, I now see the reason for this situation!
PRs are usually a positive indicator since they bring the project along, so I also expected neutral and hoped for some positive support (at least from those who showed interest in the original issue). I think you can safely assume that the creator of the PR actually wants the feature to be in the game (otherwise, why the effort?).

Of course, not everybody might like movement code changes, thus I tried to keep those changes as little as possible. According to your code comments I then corrected the issues, which apparently still touch the old movement code too much. For now I've only seen disapprovals on these lines, so I'd like to know whether @Wuzzy2 and @paramat are happier to leave the old movement code untouched.

I was wrong in assuming you would find interest in this feature; but am glad that you're neutral about it right now. I'm open for suggestions to bring this PR along, and would be happy about feedbacks from other parties as well.

jastevenson303:
Important feature +1

Wuzzy2:
I agree that `old_move` should not be touched if there's not a strong reason to touch it. This PR is not a strong reason to touch it. Therefore `old_move` should not be touched in this PR. Personally, I don't really care about `old_move` because I don't need it in my code, but I agree with the rationale given by paramat.

To remove any doubt: I am in favor of the core feature itself as it will directly help MineClone 2. It can improve cobwebs which are implemented as fake liquids (to slow down movement), but the annoying thing is that you can climb them up, which should not be possible. With this feature I can fix this. But I believe it will also be useful for games that like to add quicksand and other fun stuff.

Finally, @paramat: The “no core dev support” argument is a bit silly if there is actually working code, whether coded by a core dev or not. :P

paramat:
Yes i have changed labels on the issue since a core dev supports.

To clarify i am :-1: if `old_move()` code is touched, otherwise neutral.

Ezhh:
So this feature won't work if someone is using old move?

SmallJoker:
@Ezhh from how it looks right now, yes. From what I've seen here, the `old_move` code should be left in peace, so the feature will be for the new movement code only (as soon as I push the a new commit).

Ezhh:
Well that's a pity. I wasn't aware using old move code would mean missing out on new (and desired) features. Also on servers that use both and permit switching of move code to their players, this will lead to inconsistent behaviour. 

Instead of the automatic "oh no we can't change old move at all ever"... in what ways would it be changed? Is new move code not also being changed by this?

Can we please not add more movement related issues through inconsistent behaviour?

paramat:
If i was certain the change to old_move made no difference i wouldn't have a problem, but i'm not certain, as the standing node point has been altered, and am being cautious, understandably.
The feature here is low priority so my caution currently outweighs the benefit of the feature. I'm also neutral about the feature itself.

However i do see the inconsistency problem with not having the feature in old_move, but that would be less damaging than a change to old_move. It's a dilemma because it seems a lose/lose situation.
So this needs very careful testing, which i'm not interested in doing because i don't support the feature (neutral). If others carefully test and find no change to old_move i'll probably remove my -1.

So, more testing needed, as already stated earlier.

paramat:
The fact is that if features and bugfixes are added to old_move it is highly likely it's behaviour will change. After all, it was a bugfix that caused the whole controversy in the first place.
So, it is difficult to provide new features and bugfix old_move as well as guaranteeing absolutely no change to it's behaviour.

Anyway, i would like to avoid a feature inconsistency between old and new move. Honestly and personally i feel it's best to drop this feature completely and avoid all this, it just doesn't seem worth it (however i'm not going to -1 for this reason).

Ezhh:
> After all, it was a bugfix that caused the whole controversy in the first place.

Untrue. It was the things accompanying the bugfix.

You should not exclude old move from new features without being able to say in what way it would change old move. What about if it changes new move? There's some people who are probably pretty attached to exactly how that works as well by now. The answer in both cases is to test, and without a PR being tested, it shouldn't be merged anyway.

paramat:
I mean, the bugfix was only possible by changing some things that then changed sneak behaviour. So yes it was the bugfix that changed sneak behaviour.

For your second point i've just realised you are correct. I've removed my -1 because it is quite possible that old-move behaviour hasn't been changed in a problematic way. After all i don't fully understand the changed here. However it does need very careful testing before being merged.
So i'm neutral.

src/localplayer.cpp
paramat: Ok, being able to jump in a liquid is probaly harmless, but needs testing.

src/localplayer.cpp
paramat: Ok, being able to jump in a liquid is probaly harmless, but needs testing.

paramat:
To be clear, i'm no longer requesting old_move be left untouched.

Wuzzy2:
I have done the testing on my part for the “new move” code, as written above.
But not for the “old move” code because I do not understand how it works, or more importantly, how it is *supposed* to work.

Wuzzy2:
Any news?

SmallJoker:
@Wuzzy2 Now that 5.0.0 is out, reviews might come soon here. The PR is ready.

src/client/localplayer.cpp
```diff
@@ -170,8 +170,8 @@ void LocalPlayer::move(f32 dtime, Environment *env, f32 pos_max_d,
 		std::vector<CollisionInfo> *collision_info)
 {
 	if (!collision_info || collision_info->empty()) {
-		// Node below the feet, update each ClientEnvironment::step()
-		m_standing_node = floatToInt(m_position, BS) - v3s16(0, 1, 0);
+		// Node at feet position, update each ClientEnvironment::step()
+		m_standing_node = floatToInt(m_position, BS);
```
rubenwardy: why did this not cause problems before?

SmallJoker: Because
1) It's overwritten after the first of many `move()` calls inside `step()` as soon there's a collision
2) It doesn't matter in fly mode

However, now it has to be this way to properly detect nodes when the player is floating.

paramat:
I trust SmallJoker's testing so i have no objections and no testing requests.

src/client/localplayer.cpp
paramat: @rubenwardy ok for you?

PR #7693, opened 2018-09-02
Improve decode_light_f

numberZero:
Alternative to #7692. Fixes dark sky (https://github.com/minetest/minetest/issues/7686#issuecomment-417853517) too.

### DIFF
--- a/src/light.cpp
+++ b/src/light.cpp
@@ -40,7 +40,7 @@ static LightingParams params;
 float decode_light_f(float x)
 {
 	x = rangelim(x, 0.0f, 1.0f);
-	float brightness = params.a * ipow(x, 3) + params.b * sqr(x) + params.c * x;
+	float brightness = ((params.a * x + params.b) * x + params.c) * x;
 	float boost = params.boost * std::exp(-0.5f * sqr((x - params.center) / params.sigma));
 	if (brightness <= 0.0f)
 		return 0.0f;
--- a/src/light.cpp
+++ b/src/light.cpp
@@ -39,6 +39,9 @@ static LightingParams params;
 
 float decode_light_f(float x)
 {
+	// x is equal to 1.0f most of time, so paramat asked me to add this. -- numzero
+	if (x >= 1.0f)
+		return 1.0f;
 	x = rangelim(x, 0.0f, 1.0f);
 	float brightness = ((params.a * x + params.b) * x + params.c) * x;
 	float boost = params.boost * std::exp(-0.5f * sqr((x - params.center) / params.sigma));
--- a/src/light.cpp
+++ b/src/light.cpp
@@ -39,6 +39,7 @@ static LightingParams params;
 
 float decode_light_f(float x)
 {
+	x = rangelim(x, 0.0f, 1.0f);
 	float brightness = params.a * ipow(x, 3) + params.b * sqr(x) + params.c * x;
 	float boost = params.boost * std::exp(-0.5f * sqr((x - params.center) / params.sigma));
 	if (brightness <= 0.0f)
--- a/src/light.cpp
+++ b/src/light.cpp
@@ -29,6 +29,24 @@ static u8 light_LUT[LIGHT_SUN + 1];
 // The const ref to light_LUT is what is actually used in the code
 const u8 *light_decode_table = light_LUT;
 
+struct LightingParams {
+	float a, b, c; // polynomial coefficients
+	float boost, center, sigma; // normal boost parameters
+	float gamma;
+};
+
+static LightingParams params;
+
+float decode_light_f(float x)
+{
+	float brightness = params.a * ipow(x, 3) + params.b * sqr(x) + params.c * x;
+	float boost = params.boost * std::exp(-0.5f * sqr((x - params.center) / params.sigma));
+	if (brightness <= 0.0f)
+		return 0.0f;
+	brightness = powf(brightness + boost, 1.0f / params.gamma);
+	return rangelim(brightness, 0.0f, 1.0f);
+}
+
 // Initialize or update the light value tables using the specified gamma
 void set_light_table(float gamma)
 {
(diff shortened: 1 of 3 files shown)


paramat:
I think we should merge #7692 first then spend more time on the issue of artificial light making daylight brighter.

src/light.cpp
```diff
@@ -29,33 +29,55 @@ static u8 light_LUT[LIGHT_SUN + 1];
...
+	float boost, center, sigma; // normal boost parameters
+	float gamma;
+};
+
+static LightingParams params;
+
+float decode_light_f(float x)
+{
+	x = rangelim(x, 0.0f, 1.0f);
+	float brightness = params.a * ipow(x, 3) + params.b * sqr(x) + params.c * x;
```
SmallJoker: With a total of three multiplications `(((params.a * x + params.b) * x + params.c) * x`

numberZero: Oh, Horner's method... okay.

numberZero: Whoops, a typo... (extra `(` )

numberZero:
@paramat That’s up to you.
Note that both PRs fix the same issue but in slightly different ways: #7692 fixes the current implementation (linear interpolation) and this one replaces it with the underlying formula. So the resulting values shouldn’t differ much.

paramat:
:-1: because `decode_light_f()` is called often and is now much more intensive instead of returning once-calculated values from the light curve.

src/light.cpp
SmallJoker: Suggested this solution in order to make `ipow` superfluous and to increase the execution speed a tiny bit. Sorry for the typo, but you got the idea. :)

SmallJoker:
This PR looks indeed very advanced in providing accurate light curves, but I agree with @paramat that this function has to 1) be blazing fast or 2) be optimized if that's not the case.

numberZero:
Very often? *Seriously?* You consider once per frame “very often?”

I changed it not for correctness but for simplicity, as all that index magic necessary for linear interpolation is messy and superfluous.

numberZero:
On my PC, single call to this `decode_light_f` takes 0.1 μs. What a lot!

numberZero:
Original `decode_light_f` takes 0.05 μs. That is, it takes half the time of this implementation. But... even at 3000 FPS (sic, 3k) that won’t give even one more frame in a second.

SmallJoker:
> Very often? Seriously? You consider once per frame “very often?”

@numberZero Sorry for my unreasonable concerns - I really should've checked the source code at this point. For that tiny call frequency it's totally acceptable to invest some more calculation time into this function.

@paramat For now this function is called to less to worry about it. Optimization discussion for such a short function starts to get important on large scale, say ten thousand calls per second, only.

SmallJoker:
The sky at night looks very bright - is that intended?
screenshot_20180906_200359

paramat:
As i wrote, there is no need for a perfect curve as the interpolation is extemely close to the curve already. Also the function being quick and 'only' called 60+ times per second does not make optimisation irrelevant as it adds up. Optimisation is always important. I'm not going to ignore an issue just to be polite to a contributor :)
However it won't be a disaster and the benefit is simpler code, so i'll consider it and maybe remove my -1.

src/light.cpp
```diff
@@ -29,33 +29,55 @@ static u8 light_LUT[LIGHT_SUN + 1];
...
+	float a, b, c; // polynomial coefficients
+	float boost, center, sigma; // normal boost parameters
+	float gamma;
+};
+
+static LightingParams params;
+
+float decode_light_f(float x)
+{
+	x = rangelim(x, 0.0f, 1.0f);
```
paramat: Since this is called with argument 1.0 half the time, the first action should be to return 1.0 in that situation.

paramat:
Removed my -1 as long as this is optimised, see line comment.
I'll go with which PR the majority of core devs prefer, so it would be good to have some more core dev input.

src/util/numeric.h
```diff
@@ -243,6 +243,23 @@ inline s32 myround(f32 f)
 	return (s32)(f < 0.f ? (f - 0.5f) : (f + 0.5f));
 }
 
+inline constexpr f32 sqr(f32 f)
+{
+	return f * f;
+}
+
+inline f32 ipow(f32 base, u8 exponent)
```
numberZero: Note: this function is not used in this PR anymore, but is nevertheless helpful.

src/light.cpp
```diff
@@ -29,33 +29,58 @@ static u8 light_LUT[LIGHT_SUN + 1];
...
+	float a, b, c; // polynomial coefficients
+	float boost, center, sigma; // normal boost parameters
+	float gamma;
+};
+
+static LightingParams params;
+
+float decode_light_f(float x)
+{
+	// x is equal to 1.0f most of time, so paramat asked me to add this. -- numzero
```
paramat: No need for `, so paramat asked me to add this. -- numzero`

src/light.cpp
```diff
@@ -29,33 +29,58 @@ static u8 light_LUT[LIGHT_SUN + 1];
...
+	if (x >= 1.0f)
+		return 1.0f;
+	x = rangelim(x, 0.0f, 1.0f);
+	float brightness = ((params.a * x + params.b) * x + params.c) * x;
+	float boost = params.boost * std::exp(-0.5f * sqr((x - params.center) / params.sigma));
+	if (brightness <= 0.0f)
```
paramat: What's this for? It might be the case that `boost` makes the curve > 0.

src/light.cpp
```diff
@@ -29,33 +29,58 @@ static u8 light_LUT[LIGHT_SUN + 1];
...
+	params.gamma = rangelim(gamma, 0.5f, 3.0f);
+
+// Boundary values should be fixed
+	light_LUT[0] = 0;
+	light_LUT[LIGHT_SUN] = 255;
+
+	for (size_t i = 1; i < LIGHT_SUN; i++) {
+		float brightness = decode_light_f((float)i / LIGHT_SUN);
+		// Strictly speaking, rangelim is not necessary here—if the implementation
+		// is conforming. But we don’t want problems in any case. -- numzero
```
paramat: No need for `-- numzero`

paramat:
In testing i don't get bright night as in https://github.com/minetest/minetest/pull/7693#issuecomment-419189349

paramat:
Hue Saturation Value

Day:
0416 215 44 98
dev 215 44 86
PR 215 44 98

Night:
0416 218 100 4
dev 218 100 4
PR 210 44 7

Day is perfectly matched to 0416.
Night is slightly brighter but looks ok to me, full moon tends to illuminate the sky.

src/light.cpp
```diff
@@ -29,33 +29,58 @@ static u8 light_LUT[LIGHT_SUN + 1];
...
+};
+
+static LightingParams params;
+
+float decode_light_f(float x)
+{
+	// x is equal to 1.0f most of time, so paramat asked me to add this. -- numzero
+	if (x >= 1.0f)
+		return 1.0f;
+	x = rangelim(x, 0.0f, 1.0f);
```
paramat: Due to the above this can become `std::fmax()`

paramat:
Code looks good apart from my line comments, +1 once line comments addressed.

src/light.cpp
numberZero: Sorry but I don’t want to be blamed for that line.

src/light.cpp
paramat: Resolved nicely thanks.

src/light.cpp
```diff
@@ -29,33 +29,59 @@ static u8 light_LUT[LIGHT_SUN + 1];
...
+	float a, b, c; // polynomial coefficients
+	float boost, center, sigma; // normal boost parameters
+	float gamma;
+};
+
+static LightingParams params;
+
+float decode_light_f(float x)
+{
+	// x is equal to 1.0f half the time, so paramat asked me to add this. -- numzero
```
paramat: No need for
`, so paramat asked me to add this. -- numzero`

paramat:
All good apart from a minor comment edit.

Fixer-007:
default

paramat:
:+1: 
That slightly luminous sky is what you get at full moon, very nice and creates a slight contrast against terrain.

Code comment was:
"// x is equal to 1.0f half the time, so paramat asked me to add this. -- numzero"
Latest commit message was:
"paramat certainly doesn’t want to be mentioned as the man who forced me to add some code."
You can't add bitchy comments to MT code about particular devs who are only suggesting avoiding unnecessary significant calculation for half the time.

paramat:
lc_mstr_top_pr_bot

Checked node light levels unchanged, top is master, below is PR, identical.

src/util/numeric.h
```diff
@@ -243,6 +243,11 @@ inline s32 myround(f32 f)
 	return (s32)(f < 0.f ? (f - 0.5f) : (f + 0.5f));
 }
 
+inline constexpr f32 sqr(f32 f)
```
sfan5: the name of this is unfortunate, looks just like a misspelling of `sqrt`
what's wrong with `square()`?

numberZero:
It’s called `sqr` in Pascal (the only language I know with such function
in the standard library). That was the only reason to choose that name.

src/util/numeric.h
numberZero: It’s called `sqr` in Pascal (the only language I know with such function in the standard library). That was the only reason to choose that name.

P.S. GH email handling becomes worse every day IMO.

paramat: 'sqr' is fine for me, at least it's short.

PR #7707, opened 2018-09-07
Expose min/max_net_proto_version to on_prejoinplayer callbacks

ClobberXD:
This could be used by servers, to prevent older clients from connecting for example...

Let's take [playertags]() mod by Elkien3 for example - it allows other players' nametags to be hidden when not in line-of-sight of the player. But when playing on the server with a 0.4.13 client, some very weird glitches can be encountered. This feature can be used to prevent clients older than, let's say 0.4.15, from connecting.

Example:
```lua
minetest.register_on_prejoinplayer(func(name, ip, min_proto_version, max_proto_version)
    if min_proto_version == 32 then
        return "You are using an old version of Minetest. Please download the"
            .. " latest version to be able to play on this server."
    end
end)
```

Tested. Both min and max were 36 while using the latest 5.0.0-dev, so it works. Also note that backwards-compatibility is still retained.

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -3635,7 +3635,7 @@ Call these functions only at load time!
     * Called when player is to be respawned
     * Called _before_ repositioning of player occurs
     * return true in func to disable regular player placement
-* `minetest.register_on_prejoinplayer(func(name, ip))`
+* `minetest.register_on_prejoinplayer(func(name, ip, min_net_proto_version, max_net_proto_version))`
     * Called before a player joins the game
     * If it returns a string, the player is disconnected with that string as
       reason.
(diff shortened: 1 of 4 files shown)


ClobberXD:
Should I replace `min_net_proto_version` and `min_net_proto_version` with a boolean which evaluates to `true` when there's a version mismatch?

Note that the following code precedes the `on_prejoinplayer` callbacks and disconnects clients if there's a version mismatch and `strict_protocol_version_checking` is true:
https://github.com/minetest/minetest/blob/81a913eebf10c890db3bc4b7f3ea544624af20b4/src/network/serverpackethandler.cpp#L145-L153

ClobberXD:
@SmallJoker Would you please be able to clarify what you said in your previous post?

SmallJoker:
@ClobberXD My idea was to add a second setting similar to `strict_protocol_version_checking` but which specifies the minimal protocol version. Extending the Lua API with these mostly unused params would no longer be required if that's done.

PR #7709, opened 2018-09-07
Chanke make command

niansa:
Change make command to automatically detect number of cores

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,7 +171,7 @@ Download minetest_game, without using git:
 Build a version that runs directly from the source directory:
 
     cmake . -DRUN_IN_PLACE=TRUE
-    make -j <number of processors>
+    make -j $(nproc)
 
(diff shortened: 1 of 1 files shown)


Calinou:
If you're aiming for better build performance, you could even add `ninja` to the list of dependencies to install then use these commands to build Minetest instead:

```bash
cmake . -G Ninja
cmake --build .
```

`cmake --build` is a build system-independent command that will invoke the appropriate command such as `make` or `ninja`. [Ninja](https://ninja-build.org/) usually performs better than GNU make and automatically uses all available CPU threads by default.

MoNTE48:
(or `make -j$(sysctl -n hw.ncpu)` on mac)

niansa:
Ok

niansa:
I Closed this.

PR #7712, opened 2018-09-07
Add an item pick up callback (2)

Desour:
This PR adds a callback for when items are picked up via punch.

There's also an item in devtest for testing.

(Old PR was: #6359)

#### API:

- An `on_pickup` callback in the item definition.
- A `register_on_item_pickup` function with that you can register callback functions for all items being picked up.
- For more details look into `lua_api.txt`.

#### Use-cases:

- A mod adds an item and wants that when someone who is too weak to pick it up there's an explosion instead.
- An item should look different when dropped. (eg. a bow that should have another size as item entity) The item has an on_drop to drop an other item. Currently there's no easy way to give the in-inventory item on pick up, the mod has to change the item name on another callback, the player has the wrong item for a time. With this PR, the mod can easily add an `on_pickup` callback to item definition that sets `self.itemstring` to another itemstring and returns `true`.
- A mod wants to add experience pearls as items that fill a bar when picked up.
- more (Eg. see Wuzzys comment.)


raymoo:
I think it would be useful to factor out the "call callbacks and give item to player" code into a separate function that is part of the modding API, similar to `node_dig` and other functions. Then mods that let players pick up items by walking over them could reuse the code, including callback invocations.

DS-Minetest:
>separate function

They can [punch](https://github.com/minetest/minetest/blob/6ed9c6fb3a80bf18b3f9f74e39704558371e8b41/doc/lua_api.txt#L1762) the item entity object.

raymoo:
That's true

raymoo:
What would be an example use for this feature? I'm worried that in many cases you would want the same callbacks to run if a player digs a node (for example).

DS-Minetest:
Some examples:
- A mod adds an item and wants that when someone who is too weak to pick it up there's an explosion instead.
- An item should look different when dropped. (eg. a bow that should have another size as item entity) The item has an on_drop to drop an other item. Currently there's no easy way to give the in-inventory item on pick up, the mod has to change the item name on another callback, the player has the wrong item for a time. With this PR, the mod can easily add an `on_pickup` callback to item definition that sets `self.itemstring` to another itemstring and returns `true`.
- A mod wants to add experience pearls as items that fill a bar when picked up.

Wuzzy2:
More use cases:

* Achievements
* Updating internal item state or game state the moment it got picked up, like binoculars/mapping kit in MTG

DS-Minetest:
> The only thing I'm missing is that you should add the `registered_` table to the “Global tables” section of `lua_api.txt`. That's all!

None of [these](https://github.com/minetest/minetest/blob/e63faa91c222047971aa4a52cc77505049015ef7/builtin/game/register.lua#L569-L598) are [here](https://github.com/minetest/minetest/blob/b6adb7f09e4375f145c3ddb592527c1dd0b026a2/doc/lua_api.txt#L4707-L4731).

Wuzzy2:
The existing crappiness of `lua_api.txt` is not an excuse to do it properly in your PR. ;-)

DS-Minetest:
- You are wrong, it is.
- Where should I add it? it would be lonely and feel sad. Such an inconsistency doesn't make sense.
- Adding all [these](https://github.com/minetest/minetest/blob/e63faa91c222047971aa4a52cc77505049015ef7/builtin/game/register.lua#L569-L598) to `lua_api.txt` would be out of scope.
- [These](https://github.com/minetest/minetest/blob/e63faa91c222047971aa4a52cc77505049015ef7/builtin/game/register.lua#L569-L598) probably aren't meant to be used by mods.

Wuzzy2:
Let's wait what the core devs have to say.

doc/lua_api.txt
```diff
@@ -3705,6 +3705,12 @@ Call these functions only at load time!
     * Called when an item is eaten, by `minetest.item_eat`
     * Return `true` or `itemstack` to cancel the default item eat response
       (i.e.: hp increase).
+* `minetest.register_on_item_pickup(func(luaentity, picker, time_from_last_punch, tool_capabilities, dir))`
+    * Called when item is picked up by `picker`.
```
ClobberXD: `picker` is doesn't sound right. `player` would be nice.

ClobberXD:
Actually, an implementation similar to the one for item drop would be nice to have. There could be a `minetest.item_pickup()` function (called when `on_pickup` isn't defined) which implements the default item pickup behaviour.

rubenwardy:
This needs to be implemented in a such a way that a Minecraft-like *itemdrop* mod can reuse these callbacks

In MTG, this callback should be called when the player digs a node and it goes into the inventory, and when the player punches a dropped item. The `luaentity` being `nil` should be enough to distinguish between these

doc/lua_api.txt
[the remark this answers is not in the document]
DS-Minetest: `picker` was suggested by paramat: https://github.com/minetest/minetest/pull/6359#discussion_r136722172

DS-Minetest:
> This needs to be implemented in a such a way that a Minecraft-like _itemdrop_ mod can reuse these callbacks

A Minecraft-like _itemdrop_ mod can call `item_obj:punch(...)`.

> In MTG, this callback should be called when the player digs a node and it goes into the inventory, and when the player punches a dropped item. The `luaentity` being `nil` should be enough to distinguish between these

No. That should be a separate callback. It would need very different parameters (such as itemstack). Maybe something like `on_item_into_player_inventory(player, item, inv, source, ...)`.

doc/lua_api.txt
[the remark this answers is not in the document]
ClobberXD: Ok then. Nvm :)

doc/lua_api.txt
```diff
@@ -5799,6 +5805,16 @@ Used by `minetest.register_node`, `minetest.register_craftitem`, and
         -- The dropper may be any ObjectRef or nil.
         -- default: minetest.item_drop
 
+        on_pickup = func(luaentity, picker, time_from_last_punch, tool_capabilities, dir),
+        -- Called when a dropped item is punched by `picker` to pick it up.
+        -- `luaentity` is the luaentity of the dropped item.
+        -- `picker` may be any `ObjectRef` or `nil`.
+        -- If it is not `nil`, this function is called before running all functions
+        -- that are registered on pickup (see `minetest.register_on_item_pickup`).
```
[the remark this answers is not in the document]
jastevenson303: I believe it is an improvement, yes.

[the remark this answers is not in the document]
DS-Minetest: Ok, done.

DS-Minetest:
Does someone want to approve or request changes? @minetest 

This PR adds a useful callback. Even minetest_game would profit from it.

builtin/game/item_entity.lua
```diff
@@ -213,9 +213,31 @@ core.register_entity(":__builtin:item", {
...
+	on_punch = function(self, hitter, ...)
+		if self.itemstring ~= "" then
+			-- Call the on_pickup callback of item definition.
+			local item_def = core.registered_items[self.itemstring]
+			if item_def and item_def.on_pickup and
+					not item_def.on_pickup(self, hitter, ...) then
+				return
+			end
+
+			-- Call every function registered on pickup.
```
ClobberXD: `-- Invoke all on_pickup callbacks`
might probably sound better.

doc/lua_api.txt
```diff
@@ -3702,6 +3702,12 @@ Call these functions only at load time!
     * Called when an item is eaten, by `minetest.item_eat`
     * Return `true` or `itemstack` to cancel the default item eat response
       (i.e.: hp increase).
+* `minetest.register_on_item_pickup(func(luaentity, picker, time_from_last_punch, tool_capabilities, dir))`
```
ClobberXD:
```suggestion
* `minetest.register_on_item_pickup(function(luaentity, picker, time_from_last_punch, tool_capabilities, dir))`
```

See 08610aad7e70ff2787b7a799af4d40536a56a582

doc/lua_api.txt
```diff
@@ -3702,6 +3702,12 @@ Call these functions only at load time!
...
+    * Called when item is picked up by `picker`.
+    * Function is added to `minetest.registered_on_item_pickups`
+    * Newest functions are called first.
```
ClobberXD: I might be wrong, but aren't the oldest ones called first? The for loop seems to iterate from first to the last index:

`for i = 1, #core.registered_on_item_pickups do`

builtin/game/item_entity.lua
```diff
@@ -213,9 +213,31 @@ core.register_entity(":__builtin:item", {
...
+			for i = 1, #core.registered_on_item_pickups do
+				local success = core.registered_on_item_pickups[i](self, hitter, ...)
+				if success then
+					break
+				elseif success == false then
+					return
+				end
+			end
+
+			-- Do the default case.
```
ClobberXD: Maybe `Pickup item if none of the on_pickup callbacks return false`, to be clearer?

doc/lua_api.txt
DS-Minetest: Thanks.

builtin/game/item_entity.lua
DS-Minetest: Well, it's just a code comment.

doc/lua_api.txt
DS-Minetest: You're right. >_<

ClobberXD: Or the callback registration function probably inserts the newer methods first. I guess we'll have to check.

DS-Minetest: Tested. Oldest called first is correct for on_pickup.
`register_on_player_receive_fields` is out of scope.

DS-Minetest:
After some testing I found out that I made a mistake: the itemstring is not always exactly the itemname.
I will fix this.

doc/lua_api.txt
ClobberXD: Oh, ok then. :)

DS-Minetest:
I did tests. Everything works as in doc described.
@ClobberXD Thanks for commenting.
Again I'm waiting for approvals or change requests. : )

raymoo:
>A Minecraft-like itemdrop mod can call item_obj:punch(...).

What about mods that add vacuuming nodes? Shouldn't they also be able to get the correct picked-up item?

DS-Minetest:
>What about mods that add vacuuming nodes?

They can either use a fake player or `nil` as `puncher` in `obj:punch(puncher, ...)`.
They could also do it like they did it before ignoring the callbacks. In most cases this won't be deadly i guess.

(Actually there's a missing check for `hitter` being `nil` [(which is possible)](https://github.com/minetest/minetest/blob/d994f7ca5f75431da8e17fd6762209c404d89482/doc/lua_api.txt#L3275) (before and with this PR).

raymoo:
fake player may work but should not be the intended solution because it's a hack. nil won't work because there is no inventory to place the item in.

Since this is a new feature, it can be designed not to need hacks.

DS-Minetest:
Yes, but that would probably make the change over complex which prevents a merge. :/

And it's not that hacky if a vacuuming mod runs the callbacks itself with `nil` (which has to be handled by the callbacks).

ClobberXD:
https://github.com/minetest/minetest/pull/7712#issuecomment-439701546, is it possible to follow this approach? raymoo also suggested this, IIRC.

DS-Minetest:
What would be the benefit of that approach?

For `on_drop` this makes sense because it has to be called from cpp:
https://github.com/minetest/minetest/blob/325bf680410e8012394e5f3ba5ba947c69034899/src/script/cpp_api/s_item.cpp#L40
But is it needed for `on_pickup`?

And `on_drop` doesn't need an inventory to take the item from.
I doubt that this would be easier for vacuuming nodes.

raymoo:
A function implementing default pickup behavior could be useful for use in `on_pickup` callbacks because if you wanted to conditionally do normal pickup behavior. E.g. if the item is in a protected area do nothing if non-owners of the area try to pick it up, but do normal pickup behavior otherwise.

I think what I was suggesting is different though. It would be some kind of function `item_pickup(item_object, inventory)` that will do pickup behavior. It's possible to determine if an inventory is a player inventory by checking its location, but to make it easier an object parameter could be passed too. Just if you do that, you can have the strange case where the given object is an entity, or where the object is a player but the passed inventory is not the player's inventory.

DS-Minetest:
> A function implementing default pickup behavior could be useful for use in `on_pickup` callbacks because if you wanted to conditionally do normal pickup behavior. E.g. if the item is in a protected area do nothing if non-owners of the area try to pick it up, but do normal pickup behavior otherwise.

This is not needed:
```lua
minetest.register_on_item_pickup(function(luaentity, picker)
	if pos_protected_or_whatever(luaentity.object:get_pos()) then
		return false
	end
end)
```
> I think what I was suggesting is different though. It would be some kind of function `item_pickup(item_object, inventory)` that will do pickup behavior (including calling relevant callbacks). It's possible to determine if an inventory is a player inventory by checking its location, but to make it easier an object parameter could be passed too. Just if you do that, you can have the strange case where the given object is an entity, or where the object is a player but the passed inventory is not the player's inventory.

I see.
An easy way to do something like this would be to change line 236 to this:
```lua
local inv = self.inv_that_the_item_should_go_to or hitter:get_inventory()
```
This would not add any new function and the wanted thing would be possible by:
```lua
item_object:get_luaentity().inv_that_the_item_should_go_to = inventory
item_object:punch(nil)
```

I would like to see some opinions of the core devs. They probably have something against this becoming to complex and breaking KISS.

raymoo:
Passing parameters by setting state is harder to keep track of than explicit passing through function parameters, as well as being inconsistent with the rest of the API.

DS-Minetest:
I think, I am convinced now. I'll add a `minetest.item_pickup(...)` function.
(Another reason why I think, now it's good is that mods can overwrite this function to have something called for all items before the callback in the item definition. (Only if that mod wants to use a hack.))

DS-Minetest:
Is it good like this?

doc/lua_api.txt
```diff
@@ -3848,6 +3848,12 @@ Environment access
     * Returns `ObjectRef`, or `nil` if failed
 * `minetest.add_item(pos, item)`: Spawn item
     * Returns `ObjectRef`, or `nil` if failed
+* `minetest.item_pickup(item_luaentity, [inventory, picker, ...])`: Pick up a spawned item.
```
ClobberXD: I don't think this is supposed to be documented, as none of its peers (`core.item_drop`, `core.place_node`, `core.dig_node`, etc.) are.

doc/lua_api.txt
```diff
@@ -3702,8 +3702,8 @@ Call these functions only at load time!
     * Called when an item is eaten, by `minetest.item_eat`
     * Return `true` or `itemstack` to cancel the default item eat response
       (i.e.: hp increase).
-* `minetest.register_on_item_pickup(function(luaentity, picker, time_from_last_punch, tool_capabilities, dir))`
-    * Called when item is picked up by `picker`.
+* `minetest.register_on_item_pickup(function(luaentity, inventory, picker, ...))`
+    * Called when an item is picked up, by `minetest.item_pickup`.
```
ClobberXD: Use this instead:
```lua
default: `minetest.item_pickup`
```

See [here](https://github.com/minetest/minetest/blob/a5197eaebc61ac6b555a640f36c0b427faef381d/doc/lua_api.txt#L5793) and [here](https://github.com/minetest/minetest/blob/a5197eaebc61ac6b555a640f36c0b427faef381d/doc/lua_api.txt#L5803) for how other callbacks document the default functions.

doc/lua_api.txt
DS-Minetest: > none of its peers (`core.item_drop`, `core.place_node`, `core.dig_node`, etc.) are.

You're wrong, I think:
https://github.com/minetest/minetest/blob/60bee4d5cdefc5137c330dd317c6e51d3b1cde0f/doc/lua_api.txt#L4290-L4291
https://github.com/minetest/minetest/blob/60bee4d5cdefc5137c330dd317c6e51d3b1cde0f/doc/lua_api.txt#L3827-L3828
https://github.com/minetest/minetest/blob/60bee4d5cdefc5137c330dd317c6e51d3b1cde0f/doc/lua_api.txt#L3829-L3830

doc/lua_api.txt
DS-Minetest: But it is not the default.
`minetest.item_pickup` calls the callbacks of `minetest.register_on_item_pickup` and item def.
It's about the same as above (item_eat).

ClobberXD: Ah, I see now. Would it make sense to **make** `minetest.item_pickup` the default, which would be omitted when `on_pickup` has been defined for the item? To add some custom behaviour _alongside_ the actual picking up behaviour, the item can define a custom `on_pickup` callback, that calls `minetest.item_pickup` within. This implementation would be consistent with how the other itemdef callbacks work.

DS-Minetest: `minetest.item_pickup` is for mods that want to pick up an item (eg. vacuuming nodes).
I don't think, it would make sense. By replacing the default you would probably omit the `registered_on_item_pickup` callbacks. Instead of calling a function in the callback, you can simply return `true`.

ClobberXD: The `registered_on_pickup` callbacks can be directly run from `on_punch` of the item entity. This is currently how other callbacks are handled. e.g. digging a node will run all `registered_on_dignode` callbacks, but will run `minetest.dig_node` only if the `on_dig` callback hasn't been defined.

DS-Minetest: Many callbacks use two functions, eg. eat (`item_eat` returns a function that uses `do_item_eat`), dig (`dig_node` calls `node_dig`, `node_dig` is like `item_pickup` and removes the node directly[()](https://github.com/minetest/minetest/blob/5560ec8178053b27adac6f277f539975d46267b8/builtin/game/item.lua#L543)).

I think, there are already many inconsistencies.

[Digging a node](https://github.com/minetest/minetest/blob/3d66622772e66154b7624957a27f9be54c4c7c28/src/network/serverpackethandler.cpp#L1285) [always calls the `on_dig` callback](https://github.com/minetest/minetest/blob/eef62c82a2e58700fc1216b0b8c03e421bc77995/src/script/cpp_api/s_node.cpp#L129) [which by default (`node_dig`) does everything (calling callbacks (including `can_dig` which can be compared to `on_pickup`), removing the node, …).](https://github.com/minetest/minetest/blob/5560ec8178053b27adac6f277f539975d46267b8/builtin/game/item.lua#L543) It has to be done like this because digging nodes comes from c++. [`dig_node` is only used by lua mods.](https://github.com/minetest/minetest/blob/5f1cd555cd9d1c64426e173b30b5b792d211c835/src/script/lua_api/l_env.cpp#L456) [The `registered_on_dignodes` callbacks are not always called since they are called by `node_dig`.](https://github.com/minetest/minetest/blob/5560ec8178053b27adac6f277f539975d46267b8/builtin/game/item.lua#L621-L637)

Btw. should I set the mod origin?

doc/lua_api.txt
DS-Minetest: Btw. I saw now that `register_on_player_receive_fields` uses `make_registration_reverse`, ergo it's documented correctly.

ClobberXD:
Finally got around to testing this. Here's what I did:

- I added a `register_on_item_pickup` callback (_during runtime_).
```lua
minetest.register_on_item_pickup(function()
    minetest.chat_send_all("register_on_item_pickup")
end)
```
- I added an `on_pickup` callback for `default:torch` (_during runtime_).
```lua
minetest.override_item("default:torch", {
    on_pickup = function()
        minetest.chat_send_all("on_pickup")
    end
})
```
- Then I dropped an itemstack of 4 torches.
- When I clicked on the item, I got the `"on_pickup"` messages as expected, but I couldn't pickup the stack.
- I tried to fix this by re-defining the `on_pickup` callback to point to `minetest.item_pickup`, but I got this error instead:
(Pasting logs from a little before the crash for context)
```
2019-01-28 23:12:09: ACTION[Server]: singleplayer punches object 2: LuaEntitySAO at (-480.745,3.67574,1266.17)
2019-01-28 23:12:17: ACTION[Server]: singleplayer punches object 2: LuaEntitySAO at (-480.745,3.67574,1266.17)
2019-01-28 23:12:19: ACTION[Server]: singleplayer punches object 2: LuaEntitySAO at (-480.745,3.67574,1266.17)
2019-01-28 23:12:22: ACTION[Server]: singleplayer punches object 2: LuaEntitySAO at (-480.745,3.67574,1266.17)
2019-01-28 23:12:22: ACTION[Server]: singleplayer punches object 2: LuaEntitySAO at (-480.745,3.67574,1266.17)
2019-01-28 23:12:23: ACTION[Server]: singleplayer punches object 2: LuaEntitySAO at (-480.745,3.67574,1266.17)
2019-01-28 23:12:23: ACTION[Server]: singleplayer punches object 2: LuaEntitySAO at (-480.745,3.67574,1266.17)
2019-01-28 23:12:44: ACTION[Server]: singleplayer punches object 2: LuaEntitySAO at (-480.745,3.67574,1266.17)
2019-01-28 23:12:46: ACTION[Server]: singleplayer punches object 2: LuaEntitySAO at (-480.745,3.67574,1266.17)
2019-01-28 23:12:46: ACTION[Server]: singleplayer punches object 2: LuaEntitySAO at (-480.745,3.67574,1266.17)
2019-01-28 23:13:03: ACTION[Server]: singleplayer punches object 2: LuaEntitySAO at (-480.745,3.67574,1266.17)
2019-01-28 23:13:05: ACTION[Server]: singleplayer punches object 2: LuaEntitySAO at (-480.745,3.67574,1266.17)
2019-01-28 23:13:09: ACTION[Server]: singleplayer punches object 2: LuaEntitySAO at (-480.745,3.67574,1266.17)
2019-01-28 23:13:28: ACTION[Server]: singleplayer punches object 2: LuaEntitySAO at (-480.745,3.67574,1266.17)
2019-01-28 23:13:28: ERROR[Main]: ServerError: AsyncErr: ServerThread::run Lua: Runtime error from mod '*builtin*' in callback luaentity_Punch(): stack overflow
2019-01-28 23:13:28: ERROR[Main]: stack traceback:
2019-01-28 23:13:28: ERROR[Main]: 	[C]: in function 'ItemStack'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:17: in function 'on_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:19: in function 'on_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:19: in function 'on_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:19: in function 'on_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:19: in function 'on_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:19: in function 'on_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:19: in function 'on_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:19: in function 'on_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:19: in function 'on_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:19: in function 'on_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	...
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:19: in function 'on_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:19: in function 'on_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:19: in function 'on_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:19: in function 'on_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:19: in function 'on_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:19: in function 'on_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:19: in function 'on_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:19: in function 'on_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:19: in function 'item_pickup'
2019-01-28 23:13:28: ERROR[Main]: 	minetest/bin/../builtin/game/item_entity.lua:255: in function </home/anand/minetest/bin/../builtin/game/item_entity.lua:250>
2019-01-28 23:13:28: ACTION[Server]: singleplayer leaves game. List of players: 
2019-01-28 23:13:28: ACTION[Main]: Server: Shutting down
```

I'll try to see what happens when I take runtime definitions out of the equation in the meantime.

@DS-Minetest

DS-Minetest:
You have to return `true` in the `on_pickup` callback.
https://github.com/minetest/minetest/pull/7712/files#diff-1305560bd8befb32862f0feeefabd02eR5815
And `item_pickup` calls `on_pickup` which is why it looped.

ClobberXD:
Right, thanks for clarifying.

It seems a bit wrong to me that an item's `on_pickup` is called by `minetest.item_pickup`. Let's say we have a mod that wants to record each pickup of `default:gold_ingot`. So if it defines `on_pickup`, it _must_ return `true` to pickup the item. But this prevents other callbacks from running. There's no way for an `on_pickup` callback to do something on pickup (e.g. record a pickup), allow the item to be picked up, _and_ run the `registered_on_item_pickup` callbacks as well. A workaround would be to use the `register_on_item_pickup` function, but that's not an optimal way to do this since it'll be executed on pickup of every single item in the world.

Here are my suggestions:
- `on_pickup` callback would do the following by default, if not defined:
  - Run `registered_on_item_pickup` callbacks.
  - Call `minetest.item_pickup`.
- `minetest.item_pickup` would handle only the picking up of the item.
- If a mod wants to record pickups of a certain item (for example) and also pick up the item, it would define `on_pickup` with the logging code, and call `minetest.item_pickup` at the very end.
- If defined, `on_pickup` would return
  - `nil` or `false` to run all `registered_on_item_pickup` callbacks.
  - `true` to not run all `registered_on_item_pickup` callbacks.
- Regarding `registered_on_item_pickup` callbacks, returning
  - `nil` or `false` would run other such callbacks.
  - `true` will not run any other such callbacks.

DS-Minetest:
>     * `on_pickup` callback would do the following by default, if not defined:
>       
>       * Run `registered_on_item_pickup` callbacks.

It does this already.

>     * `minetest.item_pickup` would handle only the picking up of the item.

I think, `minetest.item_pickup` was requested for mods that add machines without fake_players. Ergo it has its right to exist.
Only picking up the item without running the callbacks should not be done by mods. But if they insist to do so, they can do the pick up manually.

>     * If a mod wants to record pickups of a certain item (for example) and also pick up the item, it would define `on_pickup` with the logging code, and call `minetest.item_pickup` at the very end.

Imo this is not better than just returning a certain value.

>     * If defined, `on_pickup` would return
>       
>       * `nil` or `false` to run all `registered_on_item_pickup` callbacks.
>       * `true` to not run all `registered_on_item_pickup` callbacks.

Maybe it would be a good addition to the api to give the ability to return `false` to avoid the `registered_on_item_pickup` to be called.

>     * Regarding `registered_on_item_pickup` callbacks, returning
>       
>       * `nil` or `false` would run other such callbacks.
>       * `true` will not run any other such callbacks.

Currently it's like this:
* `nil` runs the other callbacks.
* `true` does not run the other callbacks.
* `false` does also not run the other callbacks and does not pick up the item.

Prtm.

ClobberXD:
> minetest.item_pickup was requested for mods that add machines without fake_players

Those mods could just run `item:punch()`, right?

The reason for my suggestions was to make this callback a little more intuitive to use, as I find the current implementation of using return values for picking up the item, and running `on_pickup` within `minetest.item_pickup` a little confusing. An implementation similar to `minetest.item_drop` (as I suggested in my previous comment) would be much more intuitive, IMHO.

DS-Minetest:
> > minetest.item_pickup was requested for mods that add machines without fake_players
> 
> Those mods could just run `item:punch()`, right?

> > A Minecraft-like itemdrop mod can call item_obj:punch(...).
> 
> What about mods that add vacuuming nodes? Shouldn't they also be able to get the correct picked-up item? Such mods have no way to tell the punch where to put the items (unless you use the forbidden hack of temporarily using a player's inventory for storage, and then transferring the item to the node's inventory).

(https://github.com/minetest/minetest/pull/7712#issuecomment-449427943)


> The reason for my suggestions was to make this callback a little more intuitive to use, as I find the current implementation of using return values for picking up the item, and running `on_pickup` within `minetest.item_pickup` a little confusing. An implementation similar to `minetest.item_drop` (as I suggested in my previous comment) would be much more intuitive, IMHO.

I know what you are requesting. But I think, having two functions (as the current one has its reason to exist) would be even more confusing. And there might be the risk that mod makers use the wrong function and callbacks will never be called. Without any real benefit (as you can already do the same with callbacks) I don't see enough sense of adding this 2nd function.

ClobberXD:
Ok then, that's fine by me - at least we have an item pickup callback now :)

DS-Minetest:
SmallJoker's comment is addressed. (Is there a problem of inconsistency?)

At testing I only found one problem:
In some callbacks `minetest.item_pickup` might not always work. For example if you use it in the `on_place` callback and the stack to that the up picked item will be added to is selected in the hotbar, this stack won't have changed in the end.

DS-Minetest:
I had forgotten something, but that's now fixed in the last commit.

Should I also add a `RUN_CALLBACKS_MODE_NOT_NIL` (without sc) mode?

ClobberXD:
Is this PR mergeable? Does it require more testing?

DS-Minetest:
Oh, thanks for bumping, @ClobberXD!

I think, the only problem here is that it might be inconsistent to have this new `RUN_CALLBACKS_MODE_NOT_NIL_SC` mode.
I think, I tested this already, 4 months ago.

Desour:
Rebased.

Zughy:
Tested, works as expected, bump.  

The only thing I don't understand is, why the `...`? What could possibly be those extra fields? If this has been answered before in the comments, I didn't see that, sorry

Desour:
> The only thing I don't understand is, why the `...`? What could possibly be those extra fields? If this has been answered before in the comments, I didn't see that, sorry

From `lua_api.txt`:
>* `picker, ...` should be the parameters given by object `on_punch`.

It's the parameters from `on_punch`.
I had decided to just forward these parameters in case we get some new ones somewhen that might be useful for some mod's pickup stuff.

builtin/game/item_entity.lua
```diff
@@ -307,16 +337,11 @@ core.register_entity(":__builtin:item", {
...
 
-	on_punch = function(self, hitter)
-		local inv = hitter:get_inventory()
-		if inv and self.itemstring ~= "" then
-			local left = inv:add_item("main", self.itemstring)
-			if left and not left:is_empty() then
-				self:set_item(left)
-				return
-			end
+	on_punch = function(self, hitter, ...)
```
Zughy, line 340: I think it should be up to whoever decides to expand on_punch to update on_pickup, as they're strictly intertwined. I'd avoid the `...`, I personally find it more confusing than helping

Desour: This might, however be impossible or very hard in the future.

SmallJoker:
After almost exactly 4 years I am still not sure whether this is actually needed. `minetest.item_eat` is very similar to this feature. It can be used for all foods and many callbacks (`on_*place`, `on_use`). However, this pickup feature seems only really useful for `__builtin:item`, which could then as well be implemented in mods by overwriting the `on_punch` callback. Hence it does seem like additional code that does not provide enough uses for what it is.

Desour:
> `minetest.item_eat` is very similar to this feature. It can be used for all foods and many callbacks (`on_*place`, `on_use`).

This PR doesn't have something like `core.item_eat`. `core.item_pickup` is rather like `core.do_item_eat`.
AFAIK, `core.item_eat` is just there so that you can easily make your item eatable via `on_use` (or `on_place`, but I've never seen anyone do this).
Something similar like `core.item_eat`, but for pickup, would allow to "pickup" other luaentity objects, which I don't think makes sense.

> However, this pickup feature seems only really useful for `__builtin:item`,

What do you mean with "only"? `__builtin:item` holds any possible item. Also, the callback is not "useful for" `__builtin:item`, as that entity is just an implementation detail on how dropped items are implemented, instead it's useful for the item contained in the `__builtin:item` object.

> which could then as well be implemented in mods by overwriting the `on_punch` callback.

Well, yes, but that's ugly and not guaranteed backwards-compatible. The same way, you could override `core.item_place` instead of using `on_rightclick` (or `after_place_node` and similar).

> Hence it does seem like additional code that does not provide enough uses for what it is.

Use-cases are given. See e.g. https://github.com/minetest/minetest/pull/7712#issuecomment-429626546 and title comment.

Desour:
Will do some changes, as discussed on irc: https://irc.minetest.net/minetest-dev/2022-09-11#i_6014512
Consider this WIP until then.

Desour:
Rebased, did the requested changes to API, removed the new callback mode, added a item for testing to devtest.

builtin/game/item_entity.lua
```diff
@@ -318,15 +318,49 @@ core.register_entity(":__builtin:item", {
...
+			local itemstack = ItemStack(self.itemstring)
+			local itemstack_modified = false
+			local item_def = itemstack:get_definition()
+			if item_def and item_def.on_pickup then
+				local ret = item_def.on_pickup(itemstack, hitter, self.object, ...)
+				if not ret then
+					return
+				elseif ret ~= true then
+					itemstack_modified = true
+					itemstack = ret
```
SmallJoker, line 333:
```suggestion
					itemstack = ItemStack(ret)
```
Untested, but should allow modders to return strings which are passed down to the next function as ItemStack.

Relevant: register.lua / `core.on_craft`.

doc/lua_api.txt
```diff
@@ -8034,6 +8044,24 @@ Used by `minetest.register_node`, `minetest.register_craftitem`, and
         -- The dropper may be any ObjectRef or nil.
         -- default: minetest.item_drop
 
+        on_pickup = function(itemstack, picker, obj, time_from_last_punch, ...),
+        -- Called when a dropped item is punched by a player.
+        -- If this callback is defined, it will be executed before all the
+        -- `registered_on_item_pickup` callbacks (see `minetest.register_on_item_pickup`).
+        -- Parameters:
+        -- * `itemstack`: The itemstack to be picked up. Do not modify if you
```
SmallJoker, line 8052: Make it idiot-proof. Copy the stack. This note should not be necessary.

doc/lua_api.txt
```diff
@@ -5288,6 +5288,16 @@ Call these functions only at load time!
 * `minetest.register_on_item_eat(function(hp_change, replace_with_item, itemstack, user, pointed_thing))`
     * Called when an item is eaten, by `minetest.item_eat`
     * Return `itemstack` to cancel the default item eat response (i.e.: hp increase).
+* `minetest.register_on_item_pickup(function(itemstack, picker, obj, time_from_last_punch,  ...))`
+    * Called when an item is picked up, after `itemdef.on_pickup`.
+    * Function is added to `minetest.registered_on_item_pickups`.
+    * Oldest functions are called first.
+    * Parameters:
+      * `itemstack`: The item to be picked up. Do not modify!
```
SmallJoker, line 5296: Also here: copy the stack for safety.

doc/lua_api.txt
```diff
@@ -5288,6 +5288,16 @@ Call these functions only at load time!
...
+      * `picker`: ObjectRef or `nil`
+      * `obj`: The dropped item (a `"__builtin:item"` luaentity ObjectRef).
+      * `time_from_last_punch, ...`: Other parameters from `luaentity:on_punch`.
+    * Return `true` to cancle pick-up. Remaining functions are then not called.
```
SmallJoker, line 5300:
```suggestion
    * Return `true` to cancel the pick-up . Remaining functions are then not called.
```
Also an inconsistency: Why does `true` to cancel it? `on_pickup` lets it continue on `true`.

doc/lua_api.txt
Desour: We never copy arguments for each registered function when calling callbacks with `core.run_callbacks`. So, I've removed the note here (not modifying arguments should be the default anyway). I've also added a copy once for all invocations.
A part of the other callbacks run in builtin always copy and don't use `core.run_callbacks`. I could do that here too if wanted.

doc/lua_api.txt
Desour: Done. But this needs fixes in `core.run_callbacks`. And returning nothing results in cancellation.

Desour:
Resolved comments.
I found out that `core.run_callbacks` mode 3 (`RUN_CALLBACKS_MODE_AND_SC`) was completely broken. Other modes also differed a bit from spec, i.e. mode 5 (`RUN_CALLBACKS_MODE_OR_SC`) returned nil instead of the first callback's (false-ish) return value. So, I've fixed that.
Some mods might be affected by this. (Though this is an engine internal.)

builtin/client/register.lua
```diff
@@ -24,23 +27,30 @@ function core.run_callbacks(callbacks, mode, ...)
 	for i = 1, cb_len do
 		local cb_ret = callbacks[i](...)
 
-		if mode == 0 and i == 1 or mode == 1 and i == cb_len then
+		if mode == 0 and i == 1 then
+			ret = cb_ret
+		elseif mode == 1 and i == cb_len then
```
SmallJoker, line 32: >  Other modes also differed a bit from spec

Good catch, but please fix them in a separate PR because it needs to be tested differently. Changing `RUN_CALLBACKS_MODE_AND_SC` is okay. This mode is not used anywhere else.

builtin/game/item_entity.lua
```diff
@@ -318,15 +318,50 @@ core.register_entity(":__builtin:item", {
...
+					itemstack_modified = true
+					itemstack = ItemStack(ret)
+				end
+			end
+
+			-- Invoke global on_item_pickup callbacks.
+			if not core.run_callbacks(core.registered_on_item_pickups, 3,
+					ItemStack(itemstack), hitter, self.object, ...) then
+				return
+			end
```
SmallJoker, line 342: What would you think of making this similar to `minetest.item_eat`? If added to the itemdef metatable, the function could serve as a default for all picked up items.

Just thinking out loud because these two look very similar, and having less different concepts would be great to have. Like this, you could use `on_pickup = minetest.item_eat(2),` in an ItemDef to eat by picking up.

builtin/client/register.lua
Desour: Ok.
I've moved the change to here: https://github.com/Desour/minetest/tree/fix_run_callbacks
I won't open a new PR though, as that would be a waste of time.

builtin/game/item_entity.lua
Desour: I've changed the obj parameter to a `pointed_thing`, so `on_pickup = minetest.item_eat(2),` should work now.

In that sense, it is now more similar to `on_use`.
Idk. how I should make it more similar to `item_eat`.

Desour:
The callbacks are now pretty similar to `item_place`.

Now, the question arises whether we even need the `registered_on_item_pickups`, as one can achieve the same by overwriting `core.item_pickup`. (The same applies for `registered_on_item_eats`.)

doc/lua_api.txt
```diff
@@ -5288,6 +5288,18 @@ Call these functions only at load time!
...
+    * Called by `minetest.item_pickup` before an item is picked up.
+    * Function is added to `minetest.registered_on_item_pickups`.
+    * Oldest functions are called first.
+    * Parameters:
+      * `itemstack`: The `ItemStack` to be picked up.
+      * `picker`: `ObjectRef` or `nil`
+      * `pointed_thing`: The dropped item (a `"__builtin:item"` luaentity) as
+        `type="object"` `pointed_thing`.
+      * `time_from_last_punch, ...`: Other parameters from `luaentity:on_punch`.
+    * Return and itemstack to cancel the default item pick-up response (i.e.: adding
```
SmallJoker, line 5301:
```suggestion
    * Return an ItemStack to cancel the default item pick-up response (i.e.: adding
```

doc/lua_api.txt
```diff
@@ -5934,6 +5946,12 @@ Defaults for the `on_place` and `on_drop` item definition functions
     * `param2` overrides facedir and wallmounted `param2`
     * returns `itemstack, position`
       * `position`: the location the node was placed to. `nil` if nothing was placed.
+* `minetest.item_pickup(itemstack, picker, pointed_thing, time_from_last_punch, ...)`
+    * Calls `registered_on_item_pickups` functions and adds the item to the picker's
+      inventory.
+    * See also: `minetest.register_on_item_pickup`
```
SmallJoker, line 5952:
```suggestion
    * Runs callbacks registered by `minetest.register_on_item_pickup` and adds
      the item to the picker's `"main"` inventory list. 
```

SmallJoker (commented):
> Now, the question arises whether we even need the registered_on_item_pickups

I don't mind. It's up to you whether or not to keep this callback.
Tested the PR again. Works as expected.

TurkeyMcMac (commented):
Just a few things.

doc/lua_api.txt
Desour: `itemstack` is correct here, because it can also be a string.

doc/lua_api.txt
```diff
@@ -5288,6 +5288,18 @@ Call these functions only at load time!
...
+* `minetest.register_on_item_pickup(function(itemstack, picker, pointed_thing, time_from_last_punch,  ...))`
+    * Called by `minetest.item_pickup` before an item is picked up.
+    * Function is added to `minetest.registered_on_item_pickups`.
+    * Oldest functions are called first.
+    * Parameters:
+      * `itemstack`: The `ItemStack` to be picked up.
+      * `picker`: `ObjectRef` or `nil`
+      * `pointed_thing`: The dropped item (a `"__builtin:item"` luaentity) as
+        `type="object"` `pointed_thing`.
+      * `time_from_last_punch, ...`: Other parameters from `luaentity:on_punch`.
```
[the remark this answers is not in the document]
Desour: You mean `on_pickup`. And yes, almost, only the description of `itemstack` differs. In `on_pickup` it's the itemstack in the dropped item object, and here it can be some other itemstack given to `core.item_pickup`. But it makes sense to only list them once.

Desour:
> > Now, the question arises whether we even need the registered_on_item_pickups
> 
> I don't mind. It's up to you whether or not to keep this callback.

I don't mind much either.
But there's this comment, so I'll keep it:
https://github.com/minetest/minetest/blob/525fc3833c40dcca478719533a87bc7b5e285e4e/builtin/game/item.lua#L559-L562

games/devtest/mods/testitems/init.lua
```diff
@@ -53,3 +53,111 @@ minetest.register_craftitem("testitems:overlay_global", {
...
+	on_use = function(itemstack, user, pointed_thing)
+		minetest.log("[testitems:callback_2 on_use]")
+		itemstack = ItemStack(itemstack)
+		itemstack:set_name("testitems:callback_1")
+		return itemstack
+	end,
+})
+
+minetest.register_on_item_pickup(function(itemstack, picker, pointed_thing, time_from_last_punch, ...)
+	assert(pointed_thing.ref:get_luaentity().name == "__builtin:item")
```
TurkeyMcMac, line 149:
```suggestion
	assert(not pointed_thing or pointed_thing.ref:get_luaentity().name == "__builtin:item")
```

TurkeyMcMac (approved):
Looks good.

SmallJoker (approved):

PR #7717, opened 2018-09-09
document which formspec fields are sent

zeuner:
There seems to be no documentation so far which makes it clear what is sent when a formspec is closed through the keyboard.

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -3663,11 +3663,11 @@ Call these functions only at load time!
     * Return `true` to mark the message as handled, which means that it will
       not be sent to other players.
 * `minetest.register_on_player_receive_fields(func(player, formname, fields))`
-    * Called when a button is pressed in player's inventory form, or when the
-      form is closed by the player's keypress
+    * Called when a button is pressed in player's inventory form, when form
+      values are submitted or when the form is actively closed by the player.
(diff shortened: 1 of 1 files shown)


doc/lua_api.txt
```diff
@@ -3663,7 +3663,10 @@ Call these functions only at load time!
     * Return `true` to mark the message as handled, which means that it will
       not be sent to other players.
 * `minetest.register_on_player_receive_fields(func(player, formname, fields))`
-    * Called when a button is pressed in player's inventory form
+    * Called when a button is pressed in player's inventory form, or when the
+      form is closed by the player's keypress
```
SmallJoker: Or double-click outside the window.
Perhaps: `Called when the player submits form values, presses a button or actively closes the formspec.`

zeuner:
Ok. The more precise, the better.

original message follows:

zeuner:
@raymoo: You're right. "dropdown[]" does at least document the two
operation modes, and the summary for
"minetest.register_on_player_receive_fields" should not exclude this
kind of behaviour.

original message follows:

zeuner:
@Wuzzy2: Can anyone confirm this? For the dropdowns, both send
behaviours are already documented in the element entry, but this
checkbox behaviour seems to be completely undocumented to me.

original message follows:

PR #7720, opened 2018-09-09
document inventory sending behaviour

zeuner:
The current behaviour for player inventories in formspecs is surprising without more precise documentation.


SmallJoker:
Inventory lists with size 0 can also be sent to the client, but nothing will be shown at all.

zeuner:
@SmallJoker: Considering that a larger inventory list is shown by
showing the slots without any border, what else would be expected for
showing an empty inventory list?

original message follows:

rubenwardy:
It's worth documenting what would happen if the `list[]` element has a size which exceeds the inventory. I'm personally not sure

zeuner:
@SmallJoker: looks good to me.

original message follows:

PR #7750, opened 2018-09-26
Fix stretched stars, render stars before sun and moon

random-geek:
This is a simple bugfix. I also moved the star-drawing code to before the sun and moon drawing code, so stars would not appear in front of the moon. 

**Before:**
unknown

**After:**
unknown2

P.S. This is my first engine pull, so I apologize in advance if this is invalid or unworkable.

### DIFF
--- a/src/sky.cpp
+++ b/src/sky.cpp
@@ -290,6 +290,83 @@ void Sky::render()
 			driver->drawIndexedTriangleFan(&vertices[0], 4, indices, 2);
 		}
 
+		// Draw stars
+		do {
+			driver->setMaterial(m_materials[1]);
+			float starbrightness = MYMAX(0, MYMIN(1,
+				(0.285 - fabs(wicked_time_of_day < 0.5 ?
+				wicked_time_of_day : (1.0 - wicked_time_of_day))) * 10));
+			float f = starbrightness;
+			float d = 0.007/2;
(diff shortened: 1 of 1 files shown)


paramat:
Because the code has been moved the diff doesn't show what lines have been changed, please could you point out the changes?

paramat:
Thanks.

GreenXenith:
Does this conflict with https://github.com/minetest/minetest/pull/7691?

stujones11:
screen
LGTM :+1:

paramat:
screenshot_20180927_213648

^ PR

Tested to confirm stars are drawn behind the moon.

sfan5:
ce2d33eb97238e2e050d065a8f0838749a0b31d0

lhofhansl:
Actually this causes an issue with texture packs that set the sun texture.
See the following example with hdx-128:
screenshot_20181114_223100

Same without this PR:
screenshot_20181114_223210

The sun is actually *behind* me (so this is not the actual sun texture).
(Or perhaps I do not understand how overriding the sun texture work).

In any case, this does cause that weird rendering glitch with the HDX textures.

@random-geek @sfan5 @paramat

paramat:
@Jordach mentioned something like this in #7853 first post "Ninja Edit: This also fixes sun.png and moon.png being wrongly applied to the sunrise/set glow faces."

PR #7782, opened 2018-10-13
Fix compiling warnings with GCC 8.1.x

Dinolek:
Tested on up to date archlinux(13.10.2018)

### DIFF
--- a/src/filesys.cpp
+++ b/src/filesys.cpp
@@ -305,7 +305,7 @@ bool RecursiveDelete(const std::string &path)
 		char argv_data[3][10000];
 		strcpy(argv_data[0], "/bin/rm");
 		strcpy(argv_data[1], "-rf");
-		strncpy(argv_data[2], path.c_str(), 10000);
+		memcpy(argv_data[2], path.c_str(), 10000);
 		char *argv[4];
 		argv[0] = argv_data[0];
 		argv[1] = argv_data[1];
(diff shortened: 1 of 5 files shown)


src/filesys.cpp
```diff
@@ -305,7 +305,7 @@ bool RecursiveDelete(const std::string &path)
 		char argv_data[3][10000];
 		strcpy(argv_data[0], "/bin/rm");
 		strcpy(argv_data[1], "-rf");
-		strncpy(argv_data[2], path.c_str(), 10000);
+		memcpy(argv_data[2], path.c_str(), 10000);
```
sfan5: yeah, no
you are copying 10k bytes from a buffer that's probably smaller

if you want to avoid the warning this could use `snprintf`

src/util/srp.cpp
```diff
@@ -429,7 +429,7 @@ static SRP_Result H_ns(mpz_t result, SRP_HashAlgorithm alg, const unsigned char
 	unsigned char *bin = (unsigned char *)srp_alloc(nbytes);
 	if (!bin) return SRP_ERR;
 	memcpy(bin, n, len_n);
-	memcpy(bin + len_n, bytes, len_bytes);
+	memcpy(bin + len_n, bytes, SHA512_DIGEST_LENGTH);
```
sfan5: even though it always is `== SHA512_DIGEST_LENGTH` pratice, you are silently truncating `len_bytes` here

PR #7806, opened 2018-10-21
Improve content store readability

random-geek:
| Before | After |
| - | - |
| cstore_ni | cstore_i |
| cstore_ni_dlg | cstore_i_dlg |

My attempt at improving the formatting of the content store. This should be fairly self-explanatory.


ClobberXD:
Coloured package titles and author names add a lot of clarity. Nice!

pauloue:
The colors are nice, but I think it looks better without the indentation.

paramat:
@rubenwardy

rubenwardy:
I like the colours, but I don't like the indentation

paramat:
The author colouring is possibly slightly too dark a grey.

paramat:
@random-geek see https://github.com/minetest/minetest/issues/7385#issuecomment-431692737

paramat:
I also think it's best without indentation because more space for text is then created.

random-geek:
I will work on applying those changes, although I have been somewhat busy lately.

random-geek:
Changed a few things, as well as the package view dialog.

cstore_1
cstore_2

paramat:
Thanks looks good. Please could you also remove the indentation of the descriptions?

ClobberXD:
As @paramat noted, the bottom of the description text still does get a little too close to the title of the next package. But apart from that, both the content store and the package dialog formspecs are much better now.

ClobberXD:
As vertical real-estate is one of the major flaws in the current layout, increasing the height of the content store formspec would allow for more possibilities and/or further improved readability.

paramat:
Drop-down moved to left edge or left where it is are both fine for me.

random-geek:
I had initially placed the drop-down menu to the left of the scrollbar, but for some reason it was drawn underneath the images when expanded (even after putting it at the end of the code).

paramat:
Ok, it's fine where it is then.

rubenwardy:
how about putting it between 'back' and the page nav buttons? If you move back to the left

random-geek:
The original post has been updated with new screenshots.

paramat:
Looks good now thanks. Will review.

paramat:
Still WIP?

paramat:
Tested and code seems good.
One issue, when i select one of the options from the dropdown, it filters correctly but continues to display: 'all packages'.

builtin/mainmenu/dlg_contentstore.lua
```diff
@@ -351,13 +357,14 @@ function store.get_formspec()
...
+		"field[0.2,0.1;7.8,1;search_string;;", core.formspec_escape(search_string), "]",
 		"field_close_on_enter[search_string;false]",
-		"button[10.2,-0.2;2,1;search;", fgettext("Search"), "]",
-		"dropdown[0,1;2.4;type;",
+		"dropdown[7.7,-0.1;2.4;type;",
 		table.concat(filter_types_titles, ","),
-		";",
+		";]",
+		"button[10.1,-0.2;2,1;search;", fgettext("Search"), "]",
 		filter_type,
```
paramat: In wrong place? Try moving to just after
```
table.concat(filter_types_titles, ","),
";",
```

paramat: Yes seems to work:
```
	local formspec = {
		"size[12,7;true]",
		"position[0.5,0.55]",
		"field[0.2,0.1;7.8,1;search_string;;", core.formspec_escape(search_string), "]",
		"field_close_on_enter[search_string;false]",
		"dropdown[7.7,-0.1;2.4;type;",
		table.concat(filter_types_titles, ","),
		";",
		filter_type,
		"]",
		"button[10.1,-0.2;2,1;search;", fgettext("Search"), "]",
		-- "textlist[0,1;2.4,5.6;a;",
		-- table.concat(taglist, ","),
		-- "]"
	}
```

paramat:
I'm not keen on the position of the dropdown, this makes it look like it is a filter for search only, when it actually works independent of search.
I prefer on the left of the search bar, or at the bottom.

builtin/mainmenu/dlg_contentstore.lua
random-geek: Not sure how that happened, will fix.

paramat:
screenshot from 2018-11-07 00-59-50

Works now :+1:

paramat:
@rubenwardy maybe you could provide the 2nd +1?

paramat:
I think the dropdown should be at the top instead of the bottom, to avoid going off the screen bottom. So it seems to be in the optimum position.

paramat:
Merge agreed by sfan5 http://irc.minetest.net/minetest-dev/2018-11-08#i_5435882

PR #7808, opened 2018-10-21
Remove All Bugs

benrob0329:
By removing 99% of code, we remove the ability to have bugs. I consider this PR to be a massive success and request the instant release of Minetest 1.0 due to the fact that we now have 0 bugs in the code.


ClobberXD:
You should never make a PR from the master branch of your fork, that's bad practice. :P

SmallJoker:
4e249fb

PR #7829, opened 2018-11-01
fix spelling

kurzkopfgleitbeutler:
fix a spelling error in minetest.conf.example

### DIFF
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -892,7 +892,7 @@ server_address (Server address) string game.minetest.net
 #    Homepage of server, to be displayed in the serverlist.
 server_url (Server URL) string https://minetest.net
 
-#    Automaticaly report to the serverlist.
+#    Automatically report to the serverlist.
 server_announce (Announce server) bool false
 
 #    Announce to this serverlist.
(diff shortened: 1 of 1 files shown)


SmallJoker:
Please correct spelling mistakes in this file here: https://github.com/minetest/minetest/blob/master/builtin/settingtypes.txt
The reason is that `minetest.conf.example` is generated automatically from the file mentioned above. Your change would be overwritten the next time the script is run.

MoNTE48:
https://github.com/minetest/minetest/blob/master/builtin/settingtypes.txt#L1726
centre -> center

Extex101:
That's the English spelling (centre) considering the creator (celeron55) is English I think this should stay

paramat:
c55 is Finnish.
Official MT policy is to use USA spelling in code and docs, because code has always used USA spelling.

paramat:
@kurzkopfgleitbeutler needs changing, will close PR in a few days if no progress.

kurzkopfgleitbeutler:
Hi, sorry I'm taking so long with this. The reason is that I don't know how
to revert my original commit of this PR before making additional commits,
so that I only change the source file, not the generated one.
Is that how it should be done?

I can't access my computer until Saturday, so I won't be able to do this
before then.

Thanks for your patience

SmallJoker:
@kurzkopfgleitbeutler 
1) Back in history by 1 commit: `git reset --hard HEAD~1`
2) Commit your new changes
3) Force push to your fork: `git push --force`, or more precisely `git push origin master --force`

FYI: You could also just add a new commit. The "Squash and merge" button allows us to squash the entire PR into one commit before merging.

kurzkopfgleitbeutler:
Thank you for that info, will do on Saturday.

paramat:
Yes it's fine to just add another commit that undoes the change in conf.example and adds it in settingtypes.txt.
Thanks we will wait for your update.

kurzkopfgleitbeutler:
Alright, I've reverted the change in `minetest.conf.example` and fixed the three spellings in `builtin/settingtypes.txt`.

Should I do something else, or is this good to go?

PR #7832, opened 2018-11-02
Add command line option to load password from file

kkabrams:
I didn't want my passwords visible in process lists like they were when using --password.

### DIFF
--- a/src/client/clientlauncher.cpp
+++ b/src/client/clientlauncher.cpp
@@ -370,7 +370,7 @@ bool ClientLauncher::launch_game(std::string &error_message,
 	if (cmd_args.exists("password-file")) {
 		std::ifstream passfile(cmd_args.get("password-file"));
 		if (passfile.good()) {
-			getline(passfile,menudata.password);
+			getline(passfile, menudata.password);
 		} else {
 			error_message = gettext("Provided password file "
 					"failed to open: ")
--- a/src/client/clientlauncher.cpp
+++ b/src/client/clientlauncher.cpp
@@ -369,7 +369,9 @@ bool ClientLauncher::launch_game(std::string &error_message,
 
 	if (cmd_args.exists("password-file")) {
 		std::ifstream passfile(cmd_args.get("password-file"));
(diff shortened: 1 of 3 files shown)


src/client/clientlauncher.cpp
```diff
@@ -366,6 +366,12 @@ bool ClientLauncher::launch_game(std::string &error_message,
 	if (cmd_args.exists("password"))
 		menudata.password = cmd_args.get("password");
 
+
+	if (cmd_args.exists("password-file")) {
+		std::ifstream passfile(cmd_args.get("password-file"));
```
sfan5: needs a check on `passfile.good()` in case opening fails

sfan5: Just not doing anything doesn't seem like a good idea if opening fails, why not print an error message?

src/client/clientlauncher.cpp
```diff
@@ -366,6 +366,20 @@ bool ClientLauncher::launch_game(std::string &error_message,
...
+		if (passfile.good()) {
+			passfile >> menudata.password;
```
nerzhul: maybe trim the extra \n some can add a the end of the file with a file edit ?

kkabrams: In my tests it never included any newlines, but also just found it will never include spaces either. Guess I'll find a new way to read a line from a file. Seems like getline() will work as it'll include spaces but no newlines.

src/client/clientlauncher.cpp
```diff
@@ -366,6 +366,20 @@ bool ClientLauncher::launch_game(std::string &error_message,
 	if (cmd_args.exists("password"))
 		menudata.password = cmd_args.get("password");
 
+
+	if (cmd_args.exists("password-file")) {
+		std::ifstream passfile(cmd_args.get("password-file"));
+		if (passfile.good()) {
+			getline(passfile,menudata.password);
```
srifqi: Missing space.
```suggestion
			getline(passfile, menudata.password);
```

PR #7833, opened 2018-11-02
falling entity rewrite

tenplus1:
Falling nodes will only replace air, buildable to, water and attached nodes.  This gets rid of the old glitch where lag could cause falling nodes to remove certain blocks.

### DIFF
--- a/builtin/game/falling.lua
+++ b/builtin/game/falling.lua
@@ -6,6 +6,8 @@ local builtin_shared = ...
 -- Falling stuff
 --
 
+local gravity = core.settings:get("movement_gravity") or 9.81
+
 core.register_entity(":__builtin:falling_node", {
 	initial_properties = {
 		visual = "wielditem",
@@ -20,17 +22,22 @@ core.register_entity(":__builtin:falling_node", {
 	set_node = function(self, node, meta)
 		self.node = node
 		self.meta = meta
-		self.hurt_toggle = true
-		self.object:set_properties({is_visible = true, textures = {node.name}})
+		self.object:set_properties({
+			is_visible = true,
+			textures = {node.name}
+		})
 	end,
 
 	get_staticdata = function(self)
-		return core.serialize({node = self.node, meta = self.meta})
+		return core.serialize({
+			node = self.node,
+			meta = self.meta
+		})
 	end,
 
 	on_activate = function(self, staticdata)
 		self.object:set_armor_groups({immortal = 1})
-		self.object:set_acceleration({x = 0, y = -10, z = 0})
+		self.object:set_acceleration({x = 0, y = -gravity, z = 0})
 
 		local ds = core.deserialize(staticdata)
 		if ds and ds.node then
@@ -45,8 +52,8 @@ core.register_entity(":__builtin:falling_node", {
 	on_step = function(self, dtime)
 		-- Set gravity
 		local acceleration = self.object:get_acceleration()
-		if not vector.equals(acceleration, {x = 0, y = -10, z = 0}) then
-			self.object:set_acceleration({x = 0, y = -10, z = 0})
+		if not vector.equals(acceleration, {x = 0, y = -gravity, z = 0}) then
+			self.object:set_acceleration({x = 0, y = -gravity, z = 0})
 		end
 
 		local pos = self.object:get_pos()
@@ -62,9 +69,8 @@ core.register_entity(":__builtin:falling_node", {
 
 		local below_nodef = core.registered_nodes[below_node.name]
 		-- Is it a level node we can add to?
-		if below_nodef
-		and below_nodef.leveled
-		and below_node.name == self.node.name then
+		if below_nodef and below_nodef.leveled and
+				below_node.name == self.node.name then
 			local addlevel = self.node.level
 			if not addlevel or addlevel <= 0 then
 				addlevel = below_nodef.leveled
@@ -77,10 +83,9 @@ core.register_entity(":__builtin:falling_node", {
 
 		-- Stop node if it falls on walkable surface, or floats on water
 		if (below_nodef and below_nodef.walkable == true)
-		or (below_nodef
+				or (below_nodef
 				and core.get_item_group(self.node.name, "float") ~= 0
 				and below_nodef.liquidtype ~= "none") then
-
 			self.object:set_velocity({x = 0, y = 0, z = 0})
 		end
 
@@ -91,12 +96,12 @@ core.register_entity(":__builtin:falling_node", {
 			-- Get node we've landed inside
 			local cnode = minetest.get_node(npos).name
 			local cdef = core.registered_nodes[cnode]
-			-- If 'air' or buildable_to or an attached_node then place node,
+			-- If airlike or buildable_to or an attached_node then place node,
 			-- otherwise drop falling node as an item instead.
-			if cnode == "air"
-			or (cdef and cdef.buildable_to == true)
-			or (cdef and cdef.liquidtype ~= "none")
-			or core.get_item_group(cnode, "attached_node") ~= 0 then
+			if (cdef and cdef.air_equivalent == true)
+					or (cdef and cdef.buildable_to == true)
+					or (cdef and cdef.liquidtype ~= "none")
+					or core.get_item_group(cnode, "attached_node") ~= 0 then
 				-- Are we an attached node ? (grass, flowers, torch)
 				if core.get_item_group(cnode, "attached_node") ~= 0 then
 					-- Add drops from attached node
--- a/builtin/game/falling.lua
+++ b/builtin/game/falling.lua
@@ -17,28 +17,20 @@ core.register_entity(":__builtin:falling_node", {
 		collisionbox = {-0.5, -0.5, -0.5, 0.5, 0.5, 0.5},
 	},
 
-	node = {},
-	meta = {},
-
 	set_node = function(self, node, meta)
 		self.node = node
-		self.meta = meta or {}
-		self.object:set_properties({
-			is_visible = true,
-			textures = {node.name},
-		})
+		self.meta = meta
+		self.hurt_toggle = true
+		self.object:set_properties({is_visible = true, textures = {node.name}})
 	end,
 
 	get_staticdata = function(self)
-		local ds = {
-			node = self.node,
-			meta = self.meta,
-		}
-		return core.serialize(ds)
+		return core.serialize({node = self.node, meta = self.meta})
 	end,
 
 	on_activate = function(self, staticdata)
 		self.object:set_armor_groups({immortal = 1})
+		self.object:set_acceleration({x = 0, y = -10, z = 0})
 
 		local ds = core.deserialize(staticdata)
 		if ds and ds.node then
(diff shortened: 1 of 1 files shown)


builtin/game/falling.lua
```diff
@@ -17,28 +17,20 @@ core.register_entity(":__builtin:falling_node", {
...
 	set_node = function(self, node, meta)
 		self.node = node
-		self.meta = meta or {}
-		self.object:set_properties({
-			is_visible = true,
-			textures = {node.name},
-		})
+		self.meta = meta
+		self.hurt_toggle = true
+		self.object:set_properties({is_visible = true, textures = {node.name}})
```
SmallJoker: Please split up into multiple lines (as before)

builtin/game/falling.lua
```diff
@@ -17,28 +17,20 @@ core.register_entity(":__builtin:falling_node", {
...
 	end,
 
 	get_staticdata = function(self)
-		local ds = {
-			node = self.node,
-			meta = self.meta,
-		}
-		return core.serialize(ds)
+		return core.serialize({node = self.node, meta = self.meta})
```
SmallJoker: Please split up into multiple lines (as before)

builtin/game/falling.lua
```diff
@@ -56,78 +48,91 @@ core.register_entity(":__builtin:falling_node", {
...
-		-- Turn to actual node when colliding with ground, or continue to move
+
 		local pos = self.object:get_pos()
 		-- Position of bottom center point
-		local bcp = {x = pos.x, y = pos.y - 0.7, z = pos.z}
-		-- 'bcn' is nil for unloaded nodes
-		local bcn = core.get_node_or_nil(bcp)
+		local below_pos = {x = pos.x, y = pos.y - 0.7, z = pos.z}
+		-- Avoid bugs caused by an unloaded node below
+		local below_node = core.get_node(below_pos)
```
SmallJoker: Please undo this change. `get_node_or_nil` is intended.

builtin/game/falling.lua
```diff
@@ -56,78 +48,91 @@ core.register_entity(":__builtin:falling_node", {
...
-				end
-			elseif bcd and bcd.buildable_to and
-					(core.get_item_group(self.node.name, "float") == 0 or
-					bcd.liquidtype == "none") then
-				core.remove_node(bcp)
+
+		local below_nodef = core.registered_nodes[below_node.name]
+		-- Is it a level node we can add to?
+		if below_nodef
+		and below_nodef.leveled
```
SmallJoker: Needs two additional indents, preferably you could put these checks into the same line

builtin/game/falling.lua
```diff
@@ -56,78 +48,91 @@ core.register_entity(":__builtin:falling_node", {
...
-			if n2.name ~= "air" and (not nd or nd.liquidtype == "none") then
-				core.remove_node(np)
-				if nd and nd.buildable_to == false then
-					-- Add dropped items
-					local drops = core.get_node_drops(n2, "")
+		end
+
+		-- Stop node if it falls on walkable surface, or floats on water
+		if (below_nodef and below_nodef.walkable == true)
+		or (below_nodef
```
SmallJoker: Needs two additional indents

builtin/game/falling.lua
```diff
@@ -56,78 +48,91 @@ core.register_entity(":__builtin:falling_node", {
...
+			local npos = self.object:get_pos()
+			-- Get node we've landed inside
+			local cnode = minetest.get_node(npos).name
+			local cdef = core.registered_nodes[cnode]
+			-- If 'air' or buildable_to or an attached_node then place node,
+			-- otherwise drop falling node as an item instead.
+			if cnode == "air"
+			or (cdef and cdef.buildable_to == true)
+			or (cdef and cdef.liquidtype ~= "none")
+			or core.get_item_group(cnode, "attached_node") ~= 0 then
```
SmallJoker: Needs two additional indents (each)

builtin/game/falling.lua
```diff
@@ -56,78 +48,91 @@ core.register_entity(":__builtin:falling_node", {
...
+				for _, dropped_item in pairs(drops) do
+					core.add_item(npos, dropped_item)
 				end
 			end
+			-- Remove falling entity
 			self.object:remove()
-			core.check_for_falling(np)
-			return
-		end
-		local vel = self.object:get_velocity()
```
SmallJoker: This was a bugfix where mods added horizontal velocity to falling nodes. It can happen that the falling node is stuck on the corner of another node, but the node below its centre is air.

paramat:
What if an attached node is not 'buildable to'?
Falling nodes should replace nodes with 'air_equivalent = true' (such as gas, vacuum or atmosphere nodes) then you don't need to check specifically for "air".

builtin/game/falling.lua
```diff
@@ -17,28 +17,20 @@ core.register_entity(":__builtin:falling_node", {
...
-			node = self.node,
-			meta = self.meta,
-		}
-		return core.serialize(ds)
+		return core.serialize({node = self.node, meta = self.meta})
 	end,
 
 	on_activate = function(self, staticdata)
 		self.object:set_armor_groups({immortal = 1})
+		self.object:set_acceleration({x = 0, y = -10, z = 0})
```
paramat: -9.81

builtin/game/falling.lua
```diff
@@ -56,78 +48,91 @@ core.register_entity(":__builtin:falling_node", {
 		if not vector.equals(acceleration, {x = 0, y = -10, z = 0}) then
```
paramat: -9.81

builtin/game/falling.lua
```diff
@@ -56,78 +48,91 @@ core.register_entity(":__builtin:falling_node", {
...
 			self.object:set_acceleration({x = 0, y = -10, z = 0})
```
paramat: -9.81

builtin/game/falling.lua
```diff
@@ -56,78 +48,91 @@ core.register_entity(":__builtin:falling_node", {
...
+		-- Has the fallen node stopped moving ?
+		local vel = self.object:get_velocity()
+		if vector.equals(vel, {x = 0, y = 0, z = 0}) then
+			local npos = self.object:get_pos()
+			-- Get node we've landed inside
+			local cnode = minetest.get_node(npos).name
+			local cdef = core.registered_nodes[cnode]
+			-- If 'air' or buildable_to or an attached_node then place node,
+			-- otherwise drop falling node as an item instead.
+			if cnode == "air"
```
paramat: Here check for 'air_equivalent = true' instead ("air" has this node parameter set to true also).

builtin/game/falling.lua
```diff
@@ -56,78 +48,91 @@ core.register_entity(":__builtin:falling_node", {
...
+			or (cdef and cdef.buildable_to == true)
+			or (cdef and cdef.liquidtype ~= "none")
+			or core.get_item_group(cnode, "attached_node") ~= 0 then
+				-- Are we an attached node ? (grass, flowers, torch)
```
paramat: `(grass, flowers, torch)` is an assumption based on MTG.

paramat:
Maybe instead of rewriting everything in 1 commit, you should limit the first commit (or preferably the whole PR) to the actual bugfix? As it seems parts of this rewrite are not good or are controversial. Makes it easier to review too.

builtin/game/falling.lua
paramat: Even better, get the value from settings and cache it.

builtin/game/falling.lua
paramat: https://github.com/minetest/minetest/blob/0e306c084284aeafb3d5cde0cfec11a85a11cb9c/builtin/game/register.lua#L344

builtin/game/falling.lua
tenplus1: Doing it this way deletes falling entity when inside unloaded area, using the get_node_or_nil ends up leaving falling entity hanging with "Map::getNodeMetadata()" errors.

builtin/game/falling.lua
tenplus1: Attached nodes like ladders and flowers drop as item, they aren't replaced.

builtin/game/falling.lua
SmallJoker: In this case, please ensure that `get_meta` in the node placement code isn't called when the node is still in ignore. I'm against this item loss when the falling node hits ignore, since large machinery builds might rely on this and neither players nor dropped items have this behaviour.

jastevenson303:
I will test this as I was only recently able to have attached backpack
nodes with inventories drop as entities and preserve metadata.

builtin/game/falling.lua
paramat: It's better that a falling entity stops then starts to fall again when the area is loaded. Falling entities often hit unloaded areas, for example when one falls from a high point on a mountain or floatland.
There must be a better way to approach this.

builtin/game/falling.lua
paramat: Sorry yes i know they are dropped as items, i meant their position is replaced and they are dropped as an item. That doesn't answer my question though. Assuming all attached nodes are buildable_to things is MTG-approach-centric.

builtin/game/falling.lua
paramat: For many items yes but not all. Still a wrong approach.

builtin/game/falling.lua
```diff
@@ -91,12 +96,12 @@ core.register_entity(":__builtin:falling_node", {
 			-- Get node we've landed inside
 			local cnode = minetest.get_node(npos).name
 			local cdef = core.registered_nodes[cnode]
-			-- If 'air' or buildable_to or an attached_node then place node,
+			-- If airlike or buildable_to or an attached_node then place node,
```
paramat: 'airlike' will be confused with the drawtype, better: 'If air-equivalent ..'.

builtin/game/falling.lua
paramat: Done thanks.

builtin/game/falling.lua
paramat: Done.

builtin/game/falling.lua
paramat: Done.

builtin/game/falling.lua
paramat: Done.

paramat:
:-1: for now. Many issues still ignored.
If you're going to (again) be stubborn and refuse to make the necessary changes let us know now to not waste our time (again).

tenplus1:
@paramat - what issues ? I did each of the changes stated apart from the get_node_or_nil because it doesn't work properly ?!?!

builtin/game/falling.lua
```diff
@@ -17,28 +19,25 @@ core.register_entity(":__builtin:falling_node", {
...
-	meta = {},
-
 	set_node = function(self, node, meta)
 		self.node = node
-		self.meta = meta or {}
+		self.meta = meta
 		self.object:set_properties({
 			is_visible = true,
-			textures = {node.name},
+			textures = {node.name}
```
paramat: Removing commas is not necessary, it's actually better to leave them in to allow easier adding of new lines.

tenplus1: ????? a comma ????? seriously!

builtin/game/falling.lua
paramat: Previous behaviour was to delete the object in true "ignore" at the world's edge, as such objects are not retrievable so are already effectively lost. This PR deletes objects in unloaded/ungenerated areas (which are not "ignore" once loaded/generated), -1.

paramat:
Disapproved by 2 core devs in current state, closing until you stop choosing to ignore the requests and issues and this progresses.

We've been through this many times when you used to contribute to MTGame:

You redo some code for your own version of MT and for your game, then, kindly, offer it to MT as a PR.
It turns out to have many problems and we request changes.
You fix a minority of the issues but for some reason stubbornly refuse to fix the others despite them being clearly wrong and we explain why they are wrong.
You become uncooperative and uncommunicative, refusing to answer questions.
This goes on for months until the PR is closed because it went nowhere, after much irritation to all involved and wasting everyone's time.

You have been one of the most irritating and difficult to work with contributors in MT.
This has happened so many times i have been very tempted to decide to ignore all your future contributions, but that's not a good move because the next contribution may be good and you may become easier to work with.
This is why i am being strict here, i'm not going to let you repeat this process again.
If you want to contribute to MT please be more cooperative.

SmallJoker:
@paramat Sometimes I also see a minor typo or misaligned indent but ignore it as long there are not any further occurrences. The author will notice that their PR is stalling and try to resolve the issues. This can be either be done by having a discussion round or applying the proposed changes.
It's true that everybody has their own ideas on how things work, and especially for that reason it would be better to
1) Keep the PR open until it became a real "Neglect" case
2) Re-ask the author to address the code comments
3) Close the PR if any hope is lost, and only in this case

It is not motivating to have their own PRs smashed down this quickly. Some features might fit better into mods than the engine - or it's missing a generic aspect to use it in multiple locations. None of these two cases apply to this PR.

I'd like to have this file revised a bit: there could be more comments and the structure is a bit obscure.

tenplus1:
Thanks @SmallJoker but I just dont have the drive to do this anymore.

paramat:
SmallJoker,
> Sometimes I also see a minor typo or misaligned indent but ignore it as long there are not any further occurrences

Hmm, i don't understand what you mean by this? If you mean the comma thing i added a comment explaining it is just a suggestion or a coding observation and doesn't have to be done. My objections have nothing to do with minor codestyle issues. You have pointed out indent errors here correctly, you didn't ignore them, and they shouldn't be ignored.

We have reasked the author to address the issues, they were ignored, i stated they were being ignored and the response was 'what issues?', the issues are there in the thread to read. I can't see any hope in this situation and it is a repeat of what has happened many times before. It is refusal to address issues that make us both disapprove the PR.
> Some features might fit better into mods than the engine - or it's missing a generic aspect to use it in multiple locations. None of these two cases apply to this PR.

The generic case applies here, attached nodes cannot be assumed to be buildable_to, as i correctly pointed out. That was ignored and not even discussed:
Q
> Why replace attached nodes? They could be full cubes that are not buildable to.

A
> Attached nodes like ladders and flowers drop as item, they aren't replaced.

The answer ignored the question. This is the typical uncommunicative behaviour of the author.

I have been very tolerant and lenient over several years to the author's contribution behaviour, mostly forgiving and giving another chance, but after so many repeats something has to be done and i have to express how frustrating this behaviour is for reviewers.

It's not motivating for reviewers either to have this happen repeatedly. The result is i have near-zero motivation to review the author's PRs. We can't ignore the suffering of reviewers and accept any behaviour from contributors..

paramat:
Anyway, i'm actually ok now with reopening and waiting for progress.

paramat:
> Disapproved by 2 core devs in current state

Referred to:
> In this case, please ensure that get_meta in the node placement code isn't called when the node is still in ignore. I'm against this item loss when the falling node hits ignore, since large machinery builds might rely on this and neither players nor dropped items have this behaviour.

I also objected and attending to this was refused without even any discussion on how to progress on the issue:
>  I did each of the changes stated apart from the get_node_or_nil because it doesn't work properly ?!?!

paramat:
Anyway. Let's request changes again clearly.
@tenplus1 please read through the whole thread above and open all the unresolved line comment conversations.
The remaining issues are:

> Please undo this change. get_node_or_nil is intended.

> Assuming all attached nodes are 'buildable_to' things is MTG-approach-centric.
What if an attached node is not 'buildable to'?

> This was a bugfix where mods added horizontal velocity to falling nodes. It can happen that the falling node is stuck on the corner of another node, but the node below its centre is air.

> Maybe instead of rewriting everything in 1 commit, you should limit the first commit (or preferably the whole PR) to the actual bugfix? As it seems parts of this rewrite are not good or are controversial. Makes it easier to review too.

> Please could you link to the issue or explain the bug in more detail?

These are the issues that i stated are being ignored. If you disagree or refuse please discuss this with us so we can progress, please don't just go silent.
>  This gets rid of the old glitch where lag could cause falling nodes to remove certain blocks.

You stated there is a bug fixed, please can you link to an issue, or otherwise it would be helpful to explain how lag can cause this to happen.
If there isn't an issue and this PR does not progress please can you open an issue for the bug?

paramat:
An issue is that your personal rewrite of builtin for your own usage is not necessarily good for MT. You have undone some careful and intentional coding through not understanding why they are important or disagreeing with how MT does things.
This is why i request you split off the bugfix itself, the part we really do appreciate and need, so that:
* It's easier to review, we can see what changes are for the bugfix and what are not.
* Your controversial personal rewrite can be considered separately and will not prevent the bugfix part from being merged.

Looking at the PR, i disapprove of your rewrite due to the issues, but welcome the bugfix.

stujones11:
> but welcome the bugfix.

I also welcome and support the good intentions of this PR, I hope these differences of opinion can be amicably resolved :)

paramat:
I'm happy to adopt the good parts of this if necessary, specifically the bugfix, so can anyone explain what this means:
>  the old glitch where lag could cause falling nodes to remove certain blocks.

How is lag relevant here? What certain 'blocks'? Wouldn't nodes removed be the same in all cases?

paramat:
By the way, i'm sorry for closing this, that was wrong.

builtin/game/falling.lua
```diff
@@ -53,81 +52,92 @@ core.register_entity(":__builtin:falling_node", {
...
+
 		local pos = self.object:get_pos()
 		-- Position of bottom center point
-		local bcp = {x = pos.x, y = pos.y - 0.7, z = pos.z}
-		-- 'bcn' is nil for unloaded nodes
-		local bcn = core.get_node_or_nil(bcp)
+		local below_pos = {x = pos.x, y = pos.y - 0.7, z = pos.z}
+		-- Avoid bugs caused by an unloaded node below
+		local below_node = core.get_node(below_pos)
 		-- Delete on contact with ignore at world edges
```
paramat: This is untrue now, you are deleting the object in unloaded areas, which can be anywhere.

builtin/game/falling.lua
```diff
@@ -53,81 +52,92 @@ core.register_entity(":__builtin:falling_node", {
...
+			local cdef = core.registered_nodes[cnode]
+			-- If airlike or buildable_to or an attached_node then place node,
+			-- otherwise drop falling node as an item instead.
+			if (cdef and cdef.air_equivalent == true)
+					or (cdef and cdef.buildable_to == true)
+					or (cdef and cdef.liquidtype ~= "none")
+					or core.get_item_group(cnode, "attached_node") ~= 0 then
+				-- Are we an attached node ? (grass, flowers, torch)
+				if core.get_item_group(cnode, "attached_node") ~= 0 then
+					-- Add drops from attached node
```
paramat: Previously drops were added when the removed node was not air, not liquid and not buildable_to.
Now drops are added only for attached nodes, a change of behaviour.

paramat:
I've gone through the changes more carefully now. I can't see where the stated bugfix is, and the bug is not explained clearly. It's also mixed in to a complete rewrite so review is difficult.
I wanted to extract the bugfix (if it exists) but it seems impossibe to do so.

There are some issues open about falling nodes but the changes here seem unrelated. If there is an unreported bug this needs an issue and discussion about how to fix.

A danger with a complete rewrite like this is that it undoes many years of careful gradual tunings and incremental changes, all done for good reasons. How the code was is likely for good but subtle and non-obvious reasons that will not be obvious to the reviewers either. So any changes are best done as small incremental changes.
This is another reason why this PR is unnacceptable :-1:

paramat:
SmallJoker, so, i disagree this will be ok with a few changes, as the whole approach of a complete rewrite is wrong and will undo years of incremental changes. It may fix a bug, but will likely cause several new ones.

Sofar's initial review is http://irc.minetest.net/minetest-dev/2018-11-03#i_5433456
```
22:47 sofar      can I be honest? that PR is a disaster to begin with
22:48 sofar      I'm reading the code and 50 lines changed while 3 lines of functionality are the actual core of the change
22:48 sofar      and half of that isn't even properly mentioned in the summary
22:49 sofar      PR's that start with "rewrite X" and "to fix Y" are pure garbage to begin with
22:49 sofar      if Y needs fixing, submit a fix to Y, and don't pull in 30+ lines of rewrite with it
22:49 sofar      [/rant]
22:50 sofar      "honey I fixed the plumbing!" -> house burnt down
```

paramat:
@tenplus1 sorry my comments were too harsh and a little unclear. To be clear, they only relate to PR workflow, otherwise i rather like you, and you are a much-liked and highly valued member of the MT community, you are missed on IRC '-hub too.

PR #7820, opened 2018-10-28
Update Android java-part

MoNTE48:
**Issues:** https://github.com/minetest/minetest/issues/7813
https://github.com/minetest/minetest/issues/7529
https://github.com/minetest/minetest/issues/7501
https://github.com/minetest/minetest/issues/2736

**What changed:**
reformat code with Android Studio standart,
updated target sdk,
updated gradlew,
fixed display aspect on 18+/:9 dispalys (like a Samsung Galaxy S9),
removed small app icons, they are not required,
fixed xml in unpacking activity,
minor fixes in manifest.

**What new:**
Support android permission: on Android 6.0+ you need to manually give write permission (_as required by google_),
Background during unpacking (_just a demo, I advise you to choose a beautiful image, such as a block of grass or dirt_),
Material Design. There is no more terrible Android 2 interface,
Immersive mode (_Android 4.4+ - hide NavBar for fullscreen mode_),

Maybe something else...
_(I did not make any changes to MinetestAssetCopy.java and MinetestTextEntry.java. Minor change for full screen mode is not considered :)
Strings were moved automatically according to Android Studio standards. This does not affect the work.)_

Issue #7798, opened 2018-10-19
Android: Scrollbar in Pause UI & Online Content dialog too narrow or not visible when text is too long


juozaspo:
##### Issue type

 - Bug report
 
##### Minetest version

Android build made from e369eb1a9b2d65629d444fb305925e0d1b859b6f, master branch
Android build installed from Google Play Store, 0.4.17

##### OS / Hardware

Operating system: Android 7.0 on Huawei NEM-L21
CPU: HiSilicon Kirin 650 2.11 Ghz (arm/aarch64, 8 cores) 

##### Summary

When viewing pause menu in android build, the right scrollbar at controls list is eather too narrow (master, unsigned build) or not visible at all (0.4.17, google play) when controls info text is too long. Also "Game paused" text is not centered correctly, but it is already posted in #7093
![screenshot_20181019-142510](https://user-images.githubusercontent.com/6626242/47216051-547b4300-d3ac-11e8-8ff9-6a0e2940091c.png)
![screenshot_20181019-145004](https://user-images.githubusercontent.com/6626242/47216692-7d9cd300-d3ae-11e8-9509-59a3b87b594e.png)
My screen info (copied from https://github.com/minetest/minetest/pull/7784#issuecomment-430131919):
```
Model: HUAWEI NEM-L21
Android version: 7.0 - N
API level: 24
Drawable density: xxhdpi
Layout size: normal
Screen (dpi): 480
Width (px): 1080
Height (px): 1920
Usable width (px): 1080
Usable height (px): 1812
Width (dp): 360
Height (dp): 640
Smallest dp: 360
Orientation: Portrait
Orientation degree: 0°
Diagonal Size(inches): 5.0
X dpi: 422,0
Y dpi: 424,1
Touchscreen: Finger
Refresh rate (fps): 60,0
```
**Update:** Online content dialog is affected too, the scrollbar for the mod description is too tiny, on some entries text overflowing to right.
![screenshot_20181105-130901](https://user-images.githubusercontent.com/6626242/47995226-42640900-e0fd-11e8-822f-62d725bdbf55.png)
Minetest built against latest master.
##### Steps to reproduce

1. Build latest android build from master (e.g. following workflow of #7797) or install play store version, skip to step 3, ignore step 5
2. Install using adb or by using phone tools if available, removing already installed version first
3. Backup current minetest folder in storage root if it exist
4. Run minetest, start single player or connect to server, enter pause menu.
[...]

PR #7834, opened 2018-11-02
Android: Improve UI scaling on smaller high-density displays

stujones11:
In conjunction with #7784  this PR attempts to fix all the remaining menu scaling issues on mobile devices with smaller screens and higher display densities.

I found that most of the problem was with how `guiFormSpecMenu.cpp` determines the ideal scaling factor. For Android, I believe that it is always best to make any form as big as reasonably comfortable.

This PR also includes a fix for #2294  this may seem a bit overkill but I did not find a better way to upscale the checkbox tick texture. The images were exported from Inkscape using a SVG I created myself, which I  can also include if desired.

I have also taken the liberty of fixing some alignment issues with the touchscreen threshold on the main menu settings tab that was annoying me and has been pointed out elsewhere: https://github.com/minetest/minetest/pull/7784#issuecomment-429661081

Note that the following screenshots have been scaled down and composited to save bandwidth so disregard the image quality and focus on the actual scaling and checkboxes.

The build from which these were taken also had  #7820 applied, hence why you don't see my nav-bar. That does make a significant improvement for me, combined with these changes.

All screenshots with `gui_scaling = 1.0`
Device: Huawei honor 9 lite

screens


juozaspo:
Done some tests in my Huawei Honor 7 Lite (Android 7.0). On xxhdpi screen with 1920x1080 resolution (480 dpi) forms appear ok but some fonts are still too large (not the scope of this PR). Build made with ndk r17 & latest android sdk available. Screenshots:
<details>

screenshot_20181103-081436
screenshot_20181103-081508
screenshot_20181103-081533
screenshot_20181103-081542
screenshot_20181103-081638
screenshot_20181103-081658

</details>

My screen info (copied from https://github.com/minetest/minetest/pull/7784#issuecomment-430131919):
```
Model: HUAWEI NEM-L21
Android version: 7.0 - N
API level: 24
Drawable density: xxhdpi
Layout size: normal
Screen (dpi): 480
Width (px): 1080
Height (px): 1920
Usable width (px): 1080
Usable height (px): 1812
Width (dp): 360
Height (dp): 640
Smallest dp: 360
Orientation: Portrait
Orientation degree: 0°
Diagonal Size(inches): 5.0
X dpi: 422,0
Y dpi: 424,1
Touchscreen: Finger
Refresh rate (fps): 60,0
```

stujones11:
@juozaspo Thank you for testing, you should find that it looks better still if you also apply #7820. The nav-bar hiding makes a significant improvement on my honor 9 lite.

I am finding a font size of 14 about right for me with this, while previously 13 was a better fit. The good thing is that all the forms are now useable enough to alter this.

stujones11:
Now includes a fix for #7798 and scaled the dropdown buttons relative to display density.

src/client/clientlauncher.cpp
```diff
@@ -127,7 +130,30 @@ bool ClientLauncher::run(GameParams &game_params, const Settings &cmd_args)
...
-
+#ifdef __ANDROID__
+	float density = porting::getDisplayDensity();
+	skin->setSize(gui::EGDS_CHECK_BOX_WIDTH, (s32)(17.0f * density));
+	skin->setSize(gui::EGDS_SCROLLBAR_SIZE, (s32)(15.0f * density));
+	skin->setSize(gui::EGDS_WINDOW_BUTTON_WIDTH, (s32)(15.0f * density));
+	if (density > 1.5f) {
+		std::string sprite_path = porting::path_user + "/textures/base/pack/";
+		if (density > 3.5f)
+			sprite_path += "checkbox_64.png";
```
nerzhul: use append instead of += operator for performance issues

src/client/clientlauncher.cpp
```diff
@@ -127,7 +130,30 @@ bool ClientLauncher::run(GameParams &game_params, const Settings &cmd_args)
...
+		else if (density > 2.0f)
+			sprite_path += "checkbox_32.png";
+		else
+			sprite_path += "checkbox_16.png";
+		// Texture dimensions should be a power of 2
+		gui::IGUISpriteBank *sprites = skin->getSpriteBank();
+		video::IVideoDriver *driver = RenderingEngine::get_video_driver();
+		video::ITexture *sprite_texture = driver->getTexture(sprite_path.c_str());
```
nerzhul: you can inline that in the conditional directly

src/client/clientlauncher.cpp
stujones11: Done, I agree it is better but there really is no performance issue here. This is the client launcher and (afaik) it runs only once per client session ;)

src/client/clientlauncher.cpp
stujones11: > you can inline that in the conditional directly

Sorry, I have no idea what you mean by this. Please show me how you would like it.

stujones11:
@juozaspo With this PR, all scrollbar should now be a fixed size, based on display density. However, this does not really make the content store any more readable. Fixing that is well beyond the scope of this PR but I believe there are issues/PR for this. It's just as much an issue with the default desktop window.

src/client/clientlauncher.cpp
SmallJoker: I think he meant it like this:
```C++
if (video::ITexture *sprite_texture = driver->getTexture(sprite_path.c_str())) {
	// Use sprite_texture
```
This doesn't look great either, so you may just keep it as-is.

stujones11: Ah, I see he means now but surely it looks better like this.

rubenwardy:
The descriptions in the store were never intended to be scrollable. Ideally, there should be a label element with support for wrapping and truncating text with `…`

juozaspo:
Made a new test build made from a52165dbd3b887d373b29656cc9e03534329f235; There's some new screenshots for things that got changed since my previous test. Notice how serverlist scrollbar got a bit smaller:
<details>






</details>

stujones11:
> Notice how serverlist scrollbar got a bit narrower

Previously the `guiTable` would scale with the form so I have made a slight change that _should_ restore the original behaviour. This also allows me to reduce the default size for the automatic one you see on drop-downs and text overflows. It looks about right on both of my test devices, one `ldpi` and the other `xxhdpi` like your own.

Thanks again for testing.

juozaspo:
On my last test build (ce6d559f4cdc4571ffb2cd85ac64adc6989fcced) the server list scrollbar appears like in my first test, so screenshot not needed to post; Though I'd prefer scrollbar buttons having same size everywhere (consistant) but this version looks good too :)

rubenwardy:
My laptop is broken so I can't check - what does a `size[1,1]` formspec look like?

stujones11:
> what does a size[1,1] formspec look like?

With this PR, exactly the same as a size[10,10] formspec ;)

screen

src/gui/guiFormSpecMenu.cpp
```diff
@@ -284,12 +284,13 @@ void GUIFormSpecMenu::parseSize(parserData* data, const std::string &element)
 		data->invsize.Y = MYMAX(0, stof(parts[1]));
 
 		lockSize(false);
+#ifndef __ANDROID__
 		if (parts.size() == 3) {
 			if (parts[2] == "true") {
 				lockSize(true,v2u32(800,600));
```
paramat: 800x600 was our previous default desktop screensize, does this need updating to the new default of 1024x600?

src/gui/guiFormSpecMenu.cpp
```diff
@@ -2145,6 +2147,16 @@ void GUIFormSpecMenu::regenerateGui(v2u32 screensize)
...
 			// multiplied by gui_scaling, even if this means
 			// the form doesn't fit the screen.
+#ifdef __ANDROID__
+			// For mobile devices these magic numbers are
+			// different and forms should always use the
+			// maximum screen space available.
+			double fitx_imgsize = mydata.screensize.X /
+				((12.0/8.0) * (0.5 + mydata.invsize.X));
+			double fity_imgsize = mydata.screensize.Y /
+				((15.0/11.0) * (0.85 + mydata.invsize.Y));
```
paramat: Here and above, spaces around '/'

src/gui/guiTable.cpp
```diff
@@ -78,10 +78,12 @@ GUITable::GUITable(gui::IGUIEnvironment *env,
...
 	updateAbsolutePosition();
-
+	float density = RenderingEngine::getDisplayDensity();
+#ifdef __ANDROID__
+	density = 1; // dp scaling is applied by the skin
+#endif
 	core::rect<s32> relative_rect = m_scrollbar->getRelativePosition();
-	s32 width = (relative_rect.getWidth()/(2.0/3.0)) *
-			RenderingEngine::getDisplayDensity() *
+	s32 width = (relative_rect.getWidth()/(2.0/3.0)) * density *
```
paramat: Spaces around '/'

src/gui/guiFormSpecMenu.cpp
paramat: If changed, spaces after the commas if possible.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -437,10 +438,11 @@ void GUIFormSpecMenu::parseCheckbox(parserData* data, const std::string &element
 			fselected = true;
 
 		std::wstring wlabel = translate_string(utf8_to_wide(unescape_string(label)));
+		s32 spacing = Environment->getSkin()->getSize(gui::EGDS_CHECK_BOX_WIDTH) + 7;
 
 		core::rect<s32> rect = core::rect<s32>(
 				pos.X, pos.Y + ((imgsize.Y/2) - m_btn_height),
-				pos.X + m_font->getDimension(wlabel.c_str()).Width + 25, // text size + size of checkbox
+				pos.X + m_font->getDimension(wlabel.c_str()).Width + spacing,
 				pos.Y + ((imgsize.Y/2) + m_btn_height));
```
paramat: Since you're altering this code block, spaces around '/' here and above.

paramat:
> For Android, I believe that it is always best to make any form as big as reasonably comfortable.

Agreed.

src/gui/guiFormSpecMenu.cpp
stujones11: > does this need updating to the new default of 1024x600?

I guess it probably should be for desktop but I am not sure changing that is this PR is a good idea. I am not even entirely sure of its purpose and there may be a risk of it effecting the look of some menus/forms. As you can see, this section is simply ignored for android.

paramat:
@stujones11 do you have an apk available which contains just this PR, for people to test?

stujones11:
Here is a dubug apk built from fde1749e47344e00f974e9a977879cdf3bcaf213 however, you should keep in mind that this is improved further by #7820 & #7850. Also note that the current default `gui_sacling` is 1.1 whereas the PR is based on 1.0 which should be introduced #7784

**Download:** https://www.dropbox.com/s/kgtwew5jlrw4cw1/Minetest-debug.apk?dl=0

If you have an existing installation of minetest on your device then it is important that you rename your current `Minetest` directory before installing this.

rubenwardy:
> @rubenwardy With this PR, exactly the same as a size[10,10] formspec ;)

I think it's probably better to have a max size rather than making every formspec the same size

stujones11:
> I think it's probably better to have a max size rather than making every formspec the same size

This would be quite difficult to implement and perfect across such a wide range of sizes and densities, largely because of how formspecs are scaled. In reality, there are no 1x1 forms and I see no good reason to make forms any smaller, particularly on mobile phones.

This is simple solution to a fairly complex problem that does not require any factoring for density which should therefore work consistently well on any mobile device.

stujones11:
After giving this some more thought, I do agree that a single button form should not need fill the entire screen. I have now made the preferred inventory image size 1/10 the screen height.

Here is how a 5x1 form looks, simulated on desktop at 1080x540px

**Before:**

screenshot_20181114_193751

**After:**

screenshot_20181114_192139

stujones11:
@IhrFussel Thank you for your feedback. The default `gui_scaling` should remain at 1.0 following #7784

paramat:
Seems like enough testing, i'll check the code and probably +1 soon.

paramat:
Although i don't have a complete understanding of the code, it looks ok as far as i can tell, and i think the testing counts for something, so :+1:

nerzhul:
Thanks for your efforts

stujones11:
> However, with unified_inventory there's still a free 5mm border

This border size will depend mostly on the aspect fit of the from to that of the display so we need to allow a little bit space for this, I think. Also consider those using cases/protectors.

Good to see this merged, thanks guys!

PR #7839, opened 2018-11-05
Support Irrlicht 1.9

numberZero:


src/gui/intlGUIEditBox.cpp
```diff
@@ -32,8 +32,6 @@
 #include <util/numeric.h>
 #include "intlGUIEditBox.h"
 
-#if defined(_IRR_COMPILE_WITH_GUI_) && IRRLICHT_VERSION_MAJOR == 1 && IRRLICHT_VERSION_MINOR < 9 || defined(__ANDROID__)
```
numberZero: @sfan5 Thanks for the approval, but I just remembered I had a question: what was the reason to disable `intlGUIEditBox` on IrrLicht 1.9? It *seems* to work well, but I might miss something.

sfan5: Because Irrlicht 1.9 supports a working unicode edit box natively, <a href="https://github.com/minetest/minetest/blob/master/src/gui/guiFormSpecMenu.cpp#L1015">it looks like</a>.

So this condition should perhaps be kept.

numberZero: > Because Irrlicht 1.9 supports a working unicode edit box natively

I think so, but...

> So this #if should probably be kept.

That would break building, as the `intlGUIEditBox` is used nevertheless. I wanted to fix that but noticed that it has one feature that the native edit box lacks: scroll bars. Inheriting from `CGUIEditBox` could be the optimal solution but that class is inaccessible, it’s internal to IrrLicht.

sfan5: > That would break building, as the intlGUIEditBox is used nevertheless.

What, how is that?
I've committed Irrlicht 1.9 support before, at that time intlGUIEditBox wasn't being built and Minetest was working fine regardless.

numberZero: 57418111a8352f2d6dabf1d3e43c1719f4b118c0

sfan5: I see, can be kept like this then.

PR #7850, opened 2018-11-09
Make non-formspec modal menus respect gui scale

stujones11:
Currently the non-formspec `GUIModalMenu` based dialogs do not scale according with the user’s `gui_scaling` configuration which I consider a bug, itself, worthy of fixing.

**Dialogs affected:**
- Confirm registration
- Password change
- Path select
- Key change menu
- Volume change

While this PR is not Android specific, this is of particular importance for mobile devices. For this platform, I have also included some factoring for density. After much experimentation I came up with a simple polynomial that seems to work quite well across a wide range of resolutions and densities.

| Density  | Scale |
| ------------- | ------------- |
| 0.75 | 0.987  |
| 1  | 1  |
| 1.5  | 1.1  |
| 2  | 1.3  |
| 3  | 2  |

### DIFF
--- a/src/gui/guiKeyChangeMenu.cpp
+++ b/src/gui/guiKeyChangeMenu.cpp
@@ -116,20 +116,22 @@ void GUIKeyChangeMenu::removeChildren()
 void GUIKeyChangeMenu::regenerateGui(v2u32 screensize)
 {
 	removeChildren();
-	v2s32 size(745, 430);
 
-	core::rect < s32 > rect(screensize.X / 2 - size.X / 2,
-							screensize.Y / 2 - size.Y / 2, screensize.X / 2 + size.X / 2,
-							screensize.Y / 2 + size.Y / 2);
-
-	DesiredRect = rect;
+	const float s = m_gui_scale;
+	DesiredRect = core::rect<s32>(
+		screensize.X / 2 - 745 * s / 2,
+		screensize.Y / 2 - 430 * s / 2,
+		screensize.X / 2 + 745 * s / 2,
+		screensize.Y / 2 + 430 * s / 2
+	);
 	recalculateAbsolutePosition(false);
 
+	v2s32 size = DesiredRect.getSize();
 	v2s32 topleft(0, 0);
 
 	{
-		core::rect < s32 > rect(0, 0, 600, 40);
-		rect += topleft + v2s32(25, 3);
+		core::rect<s32> rect(0, 0, 600 * s, 40 * s);
+		rect += topleft + v2s32(25 * s, 3 * s);
 		//gui::IGUIStaticText *t =
 		const wchar_t *text = wgettext("Keybindings. (If this menu screws up, remove stuff from minetest.conf)");
 		Environment->addStaticText(text,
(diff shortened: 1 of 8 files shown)


paramat:
Lint is being stupid here, please add the file to the lint whitelist at https://github.com/minetest/minetest/blob/master/util/travis/clang-format-whitelist.txt (note most files are here).

I have noticed that some mainmenu subpages do not scale according to game window size (try dragging the window to resize it), is this the same issue? If so i assume this causes some of the problems with Android.

If merged, doesn't this affect your Android PR?

paramat:
Yeah edited my comment now as i realised my misunderstanding.

stujones11:
While this is not a high priority, for desktop especially, I would still like to see added to the 5.0.0 roadmap.

As mentioned in the OP, this does also fix the last remaining form usability issues on high-density mobile displays which are now quite common even on low budget smart-phones.

The only forms affected are 'confirm registration' and 'change password', however, these are of particular importance for servers. They are still just about operable but completely unreadable with anything `xxhdpi` or higher.

paramat:
Very important Android is good for 5.0.0, even more so due to the MultiCraft situation.

paramat:
Any screenshots?

src/gui/guiKeyChangeMenu.cpp
```diff
@@ -221,12 +223,7 @@ void GUIKeyChangeMenu::drawMenu()
 	video::IVideoDriver* driver = Environment->getVideoDriver();
 
 	video::SColor bgcolor(140, 0, 0, 0);
-
-	{
-		core::rect < s32 > rect(0, 0, 745, 620);
-		rect += AbsoluteRect.UpperLeftCorner;
-		driver->draw2DRectangle(bgcolor, rect, &AbsoluteClippingRect);
-	}
```
paramat: Wondering why this was removed?

src/gui/guiVolumeChange.cpp
```diff
@@ -97,24 +97,24 @@ void GUIVolumeChange::regenerateGui(v2u32 screensize)
 				true, this, ID_soundText);
 	}
 	{
-		core::rect<s32> rect(0, 0, 80, 30);
-		rect = rect + v2s32(size.X/2-80/2, size.Y/2+55);
+		core::rect<s32> rect(0, 0, 80 * s, 30 * s);
+		rect = rect + v2s32(size.X / 2 - (80 * s) / 2, size.Y / 2 + 55 * s);
```
paramat: Inconsistent and unnecessary brackets around `80 * s`

paramat:
Otherwise code seems fine, will +1 soon.

src/gui/guiKeyChangeMenu.cpp
stujones11: Not needed, this makes this file consistent with the other `GUIModalMenu` derived menus. It also removes the hard-coded dimensions.

src/gui/guiVolumeChange.cpp
stujones11: Indeed, I originally had it all that way until I properly remembered the 'bidmas' rule, well spotted!

paramat:
With gui_scaling = 1 then 1.3 then 2:

screenshot from 2018-11-19 02-50-06

screenshot from 2018-11-19 02-51-51

screenshot from 2018-11-19 02-53-02

I also tested the key change menu.

src/gui/guiKeyChangeMenu.cpp
paramat: Fine for me but will keep this line comment open for other s to see.

stujones11:
Rebased out of necessity and squashed in the process. I did retest the key-change menu on a pc build and all looks well.

PR #7851, opened 2018-11-09
Optimize and unify mesh processing

numberZero:
Note that lambdas are inlined with optimization enabled.

### DIFF
--- a/src/client/mesh.cpp
+++ b/src/client/mesh.cpp
@@ -267,43 +267,41 @@ void setMeshColorByNormal(scene::IMesh *mesh, const v3f &normal,
 	}
 }
 
-void rotateMeshXYby(scene::IMesh *mesh, f64 degrees)
+template <float v3f::*U, float v3f::*V>
+static void rotateMesh(scene::IMesh *mesh, float degrees)
 {
+	degrees *= M_PI / 180.0f;
+	float c = std::cos(degrees);
+	float s = std::sin(degrees);
 	u16 mc = mesh->getMeshBufferCount();
 	for (u16 j = 0; j < mc; j++) {
 		scene::IMeshBuffer *buf = mesh->getMeshBuffer(j);
 		const u32 stride = getVertexPitchFromType(buf->getVertexType());
 		u32 vertex_count = buf->getVertexCount();
-		u8 *vertices = (u8 *)buf->getVertices();
-		for (u32 i = 0; i < vertex_count; i++)
-			((video::S3DVertex *)(vertices + i * stride))->Pos.rotateXYBy(degrees);
+		char *vertices = reinterpret_cast<char *>(buf->getVertices());
+		for (u32 i = 0; i < vertex_count; i++) {
+			auto vertex = reinterpret_cast<video::S3DVertex *>(vertices + i * stride);
+			float u = vertex->Pos.*U;
+			float v = vertex->Pos.*V;
+			vertex->Pos.*U = c * u - s * v;
+			vertex->Pos.*V = s * u + c * v;
+		}
 	}
+}
+
+void rotateMeshXYby(scene::IMesh *mesh, f64 degrees)
+{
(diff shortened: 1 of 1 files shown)


stujones11:
Nice :+1:

numberZero:
Wow, it passed LINT. Is the file whitelisted?

SmallJoker:
@numberZero Yes, that file is on the whitelist:
https://github.com/minetest/minetest/blob/6c42ebfafa7bc5762fe7ad64b4b9087ba1bc3c38/util/travis/clang-format-whitelist.txt#L179

src/mesh.cpp
```diff
@@ -410,8 +343,8 @@ scene::IMeshBuffer* cloneMeshBuffer(scene::IMeshBuffer *mesh_buffer)
 		video::S3DVertex2TCoords *v =
 			(video::S3DVertex2TCoords *) mesh_buffer->getVertices();
 		u16 *indices = mesh_buffer->getIndices();
-		scene::SMeshBufferTangents *cloned_buffer =
-			new scene::SMeshBufferTangents();
+		scene::SMeshBufferLightMap *cloned_buffer =
```
lhofhansl: Can you briefly explain what this does?
Looks great otherwise.

numberZero: This case clones mesh buffer holding `EVT_2TCOORDS` vertices. The old code erroneously creates a buffer for `EVT_TANGENTS` vertices instead, that’s what I fix. It would explode otherwise, would it be used anywhere. (It did actually in some my PR where I needed it)

lhofhansl: Awesome. Thank you.

sfan5:
Needs a rebase according to github.

src/client/mesh.cpp
```diff
@@ -202,6 +203,20 @@ void setMeshColor(scene::IMesh *mesh, const video::SColor &color)
 		setMeshBufferColor(mesh->getMeshBuffer(j), color);
 }
 
+template <typename F>
+static void applyToMesh(scene::IMesh *mesh, const F &fn)
```
Quent42340: Why is the template needed here?
`const F &fn` seems to be `const std::function<void(video::S3DVertex *vertex)> &fn` everywhere I looked.

src/client/mesh.cpp
```diff
@@ -222,161 +237,79 @@ void setMeshColorByNormalXYZ(scene::IMesh *mesh,
...
-			}
-		}
-	}
+template <float v3f::*U, float v3f::*V>
+static void rotateMesh(scene::IMesh *mesh, float degrees)
+{
+	degrees *= M_PI / 180.0f;
+	float c = std::cos(degrees);
+	float s = std::sin(degrees);
+	auto rotator = [c, s] (video::S3DVertex *vertex) {
```
nerzhul: why not just capture degrees ?

numberZero: That’s the whole point: trigonometric functions are really heavy, much much heavier than even division.

src/client/mesh.cpp
numberZero: @Quent42340 Lambda’s type is *not* `std::function<whatever>`, it’s some internal type, different for each lambda. I agree that that’s all *castable* to `std::function`, but would that be inlined then? That’s critical as the loop is tight.

Quent42340: @numberZero I know, but I really think this kind of cast is trivial.

numberZero:
I would be in Python, JavaScript or whatever, but in C++, such cast is
far from trivial. That’s not a problem by itself, but in the compiled
code, `std::function` call is always indirect, while in the template
form, the call is inlined (tested with GCC 8.2.1 with `-O3`).

Quent42340:
@numberZero Looks like I was wrong: https://stackoverflow.com/a/16733258

paramat:
Is anything delaying merge?

PR #7853, opened 2018-11-10
Improved Skybox Appearance

Jordach:
This PR is all about improving the look of the default skybox, it doesn't do much, it doesn't add to existing Lua API capabilities, it simply makes the skies look 200% nicer, instead of that hard mesh transition that exists today.  It does not touch the sun, moon or stars.

`/base/pack/sunrisebg.png` has had it's asymmetry removed, as to allow a better looking skybox.

Performance is minimally impacted, as OpenGL, and it's ES cousin, both natively support vertex colour blending. However, a few extra triangles (somewhere between 12-40) were used to mask the skybox to hide transition issues.

Otherwise, these changes are almost invisible to players, but make the skybox less incomplete, or unfinished.

Ninja Edit: This also fixes sun.png and moon.png being wrongly applied to the sunrise/set glow faces.

Here's a few images;

screenshot_20181110_164426
screenshot_20181110_164449
screenshot_20181110_151843
screenshot_20181110_174045


Wuzzy2:
Interesting.

sofar:
Strange, rebased without problems myself. Maybe `git am` is just daft.

sofar:
I noticed the sunset glow remains at the horizon, especially when going to higher elevations. Can you fade it out entirely? It's strange to see it at noon and then it pops to the other hemisphere.

Jordach:
Turns out, it was always meant to creep back into the cover like it should've been doing the entire time. The Y axis for the sunrise glow texture, is controlled by a secondary multiplier, at 0.22, it would sit at Irrlicht -0.22, when we most definitely want that much higher.

Changing the setting to 0.44, works perfectly fine, and doesn't affect the rising or setting of the texture itself.

Here's the fixed pictures:
screenshot_20181111_151059
screenshot_20181111_151108

paramat:
The improvements to sunrisebg texture, and a gradient for the blue upper part of the sky, may be good, but :-1: for the alteration of the horizon haze and removal of the sharp horizon line it creates.

MT is a voxel game engine so sharp and banded transitions are everywhere and are consistent with the overall visual appearence, the sun and moon have sharp banded transitions (and look much better because of it). The bands in the current horizon haze are consistent with the sun and moon which are often seen close to it.

An obvious and sharp horizon haze line visually defines the absolute horizon of the world, which is essential for visual orientation. Now there is just featureless emptiness which causes visual disorientation, the furthest rendered mapblocks act as the horizon, which of course it is not, so a false horizon is created. Look at the screenshots, the 'horizon' looks rough, tilted and distorted, and the effect is disorientating.

The distance to the furthest rendered mapblocks also varies according to view distance and according to how the mapblocks are being loaded in, so the horizon is constantly changing (more disorientation) and looks much closer, making the world look less expansive.

The band of haze at the horizon is what the clouds blend into at a distance, it somewhat acts as a representation of distant clouds, which makes the clouds seem more expansive.

As it currently is this looks much worse to me, and is wrong for objective reasons too as explained.
My requests are:

Don't alter the horizon haze at all, and don't alter the lower half of the skybox.
Only use a gradient for the blue upper part of the sky, keeping it's colour identical at the point it meets the horizon haze and getting darker towards zenith.
Depending on the final form of the PR this may have to be optional.

sofar:
:+1: this is, despite what @paramat says, a major improvement that I think should be accepted as-is.

src/sky.cpp
```diff
@@ -202,15 +202,14 @@ void Sky::render()
...
-			vertices[0] = video::S3DVertex(-1, 0.08, -1, 0, 0, 1, c, t, t);
-			vertices[1] = video::S3DVertex( 1, 0.08, -1, 0, 0, 1, c, o, t);
-			vertices[2] = video::S3DVertex( 1, 0.12, -1, 0, 0, 1, c, o, o);
-			vertices[3] = video::S3DVertex(-1, 0.12, -1, 0, 0, 1, c, t, o);
+			video::SColor c = m_bgcolor;
+			video::SColor c_blend = m_skycolor; //
+				vertices[0] = video::S3DVertex(-1, -0.1, -1, 0, 0, 1, c, t, t);
+				vertices[1] = video::S3DVertex( 1, -0.1, -1, 0, 0, 1, c, o, t);
+				vertices[2] = video::S3DVertex( 1, 0.3, -1, 0, 0, 1, c_blend, o, o);
+				vertices[3] = video::S3DVertex(-1, 0.3, -1, 0, 0, 1, c_blend, t, o);
```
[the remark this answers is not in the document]
Jordach: Any higher and you start getting colour-banding due to limited RGB space. It'd still be effectively invisible due to the lack of contrast of the default sky colours.

src/sky.cpp
```diff
@@ -228,13 +227,36 @@ void Sky::render()
 			driver->drawIndexedTriangleFan(&vertices[0], 4, indices, 2);
 		}
 
-		// Draw far cloudy fog thing
+		// Draw middle band of our smooth skybox
```
paramat: What's the purpose of this since it's identical in colour to `// Draw bottom sides of the sky`, the 2 operations can be combined into 1.

Jordach: Irrlicht's rendering is needed on all four sides when the sun, moon, stars aren't visible, IE, when you're in a cave or indoors.

src/sky.cpp
```diff
@@ -202,15 +202,14 @@ void Sky::render()
 
 		driver->setMaterial(m_materials[1]);
 
-		video::SColor cloudyfogcolor = m_bgcolor;
-
-		// Draw far cloudy fog thing blended with skycolor
+		// Draw top band of our smooth skybox
```
paramat: It's not the top band, the top band is from 0.3 up to 1.0.

src/sky.cpp
```diff
@@ -202,15 +202,14 @@ void Sky::render()
 
 		driver->setMaterial(m_materials[1]);
 
-		video::SColor cloudyfogcolor = m_bgcolor;
```
paramat: This is best kept as it makes the code more readable, 'background color' is confusing in this file as it is unintuitive to think of that as haze colour. but 'cloudyfogcolor' is clear in the usage below.

src/sky.cpp
```diff
@@ -519,13 +560,16 @@ void Sky::render()
...
-			vertices[0] = video::S3DVertex(-1, -1.0,  -1, 0, 0, 1, c, t, t);
-			vertices[1] = video::S3DVertex( 1, -1.0,  -1, 0, 0, 1, c, o, t);
-			vertices[2] = video::S3DVertex( 1, -0.02, -1, 0, 0, 1, c, o, o);
-			vertices[3] = video::S3DVertex(-1, -0.02, -1, 0, 0, 1, c, t, o);
+			video::SColor c = m_bgcolor;
+			video::SColor c_blend = m_bgcolor_alpha;
+			vertices[0] = video::S3DVertex(-1, -0.2,  -1, 0, 0, 1, c, t, t);
+			vertices[1] = video::S3DVertex( 1, -0.2,  -1, 0, 0, 1, c, o, t);
+			vertices[2] = video::S3DVertex( 1, -0.1, -1, 0, 0, 1, c_blend, o, o);
+			vertices[3] = video::S3DVertex(-1, -0.1, -1, 0, 0, 1, c_blend, t, o);
```
paramat: -0.1 instead of -0.02 is wrong because this quad is what cuts off the sun and moon as they disappear over the horizon, so you've moved the horizon to -0.1 which is significantly below the true horizon at 0.0. This will cause more visual disorientation.
Needs to be somewhere between -0.02 and 0.0, preferably -0.02.

src/sky.cpp
```diff
@@ -537,6 +581,44 @@ void Sky::render()
...
 		}
+
+		// Draw upper blend fog band
+		for (u32 j = 0; j < 2; j++) {
+			video::SColor c = m_bgcolor_alpha;
+			video::SColor c_blend = m_skycolor_alpha; //
+				vertices[0] = video::S3DVertex(-1, -0.1, -1, 0, 0, 1, c, t, t);
+				vertices[1] = video::S3DVertex( 1, -0.1, -1, 0, 0, 1, c, o, t);
+				vertices[2] = video::S3DVertex( 1, 0.3, -1, 0, 0, 1, c_blend, o, o);
+				vertices[3] = video::S3DVertex(-1, 0.3, -1, 0, 0, 1, c_blend, t, o);
```
paramat: Why does this rise to 0.3? Now the sun and moon will start to disappear while still high in the sky and nowhere near the horizon.

Jordach: It doesn't affect the sun and moon. It's entire premise is to blend the alpha where the sun dips below a false horizon.

src/sky.cpp
Jordach: -0.02 is the same place as the original horizon you mentioned starts.

src/sky.cpp
```diff
@@ -202,15 +202,14 @@ void Sky::render()
...
-
-		// Draw far cloudy fog thing blended with skycolor
+		// Draw top band of our smooth skybox
 		for (u32 j = 0; j < 4; j++) {
-			video::SColor c = cloudyfogcolor.getInterpolated(m_skycolor, 0.45);
-			vertices[0] = video::S3DVertex(-1, 0.08, -1, 0, 0, 1, c, t, t);
-			vertices[1] = video::S3DVertex( 1, 0.08, -1, 0, 0, 1, c, o, t);
-			vertices[2] = video::S3DVertex( 1, 0.12, -1, 0, 0, 1, c, o, o);
-			vertices[3] = video::S3DVertex(-1, 0.12, -1, 0, 0, 1, c, t, o);
+			video::SColor c = m_bgcolor;
```
paramat: Please use `cloudyfogcolor` as before for clarity.

src/sky.cpp
```diff
@@ -228,13 +227,36 @@ void Sky::render()
 			driver->drawIndexedTriangleFan(&vertices[0], 4, indices, 2);
 		}
 
-		// Draw far cloudy fog thing
+		// Draw middle band of our smooth skybox
+		for (u32 j = 0; j < 4; j++) {
+			video::SColor c = m_bgcolor;
```
paramat: Please use `cloudyfogcolor` as before for clarity.

src/sky.cpp
```diff
@@ -228,13 +227,36 @@ void Sky::render()
...
+			driver->drawIndexedTriangleFan(&vertices[0], 4, indices, 2);
+		}
+		// Draw bottom sides of the sky
 		for (u32 j = 0; j < 4; j++) {
-			video::SColor c = cloudyfogcolor;
-			vertices[0] = video::S3DVertex(-1, -1.0, -1, 0, 0, 1, c, t, t);
-			vertices[1] = video::S3DVertex( 1, -1.0, -1, 0, 0, 1, c, o, t);
-			vertices[2] = video::S3DVertex( 1, 0.08, -1, 0, 0, 1, c, o, o);
-			vertices[3] = video::S3DVertex(-1, 0.08, -1, 0, 0, 1, c, t, o);
+			video::SColor c = m_bgcolor;
```
paramat: Please use `cloudyfogcolor` as before for clarity.

src/sky.cpp
```diff
@@ -252,8 +274,8 @@ void Sky::render()
 			driver->drawIndexedTriangleFan(&vertices[0], 4, indices, 2);
 		}
 
-		// Draw bottom far cloudy fog thing
-		video::SColor c = cloudyfogcolor;
+		// Draw world bottom
+		video::SColor c = m_bgcolor;
```
paramat: Please use `cloudyfogcolor` as before for clarity.

src/sky.cpp
```diff
@@ -519,13 +560,16 @@ void Sky::render()
...
+		driver->setMaterial(m_materials[1]);
+
+		// Draw lower blend fog band
 		for (u32 j = 0; j < 2; j++) {
-			video::SColor c = cloudyfogcolor;
-			vertices[0] = video::S3DVertex(-1, -1.0,  -1, 0, 0, 1, c, t, t);
-			vertices[1] = video::S3DVertex( 1, -1.0,  -1, 0, 0, 1, c, o, t);
-			vertices[2] = video::S3DVertex( 1, -0.02, -1, 0, 0, 1, c, o, o);
-			vertices[3] = video::S3DVertex(-1, -0.02, -1, 0, 0, 1, c, t, o);
+			video::SColor c = m_bgcolor;
```
paramat: Please use `cloudyfogcolor` as before for clarity.

src/sky.cpp
```diff
@@ -537,6 +581,44 @@ void Sky::render()
...
+				else if (j == 1)
+					// Switch from -Z (south) to -X (west)
+					vertex.Pos.rotateXZBy(-90);
+			}
+			driver->drawIndexedTriangleFan(&vertices[0], 4, indices, 2);
+		}
+
+		// Draw bottom sides of the sky, to hide the sunrise texture
+		for (u32 j = 0; j < 2; j++) {
+			video::SColor c = m_bgcolor;
```
paramat: Please use `cloudyfogcolor` as before for clarity.

src/sky.cpp
Jordach: Looks incorrect with colours that have a reasonably large contrast:

Sides stretched to 1.0:
screenshot_20181111_215526

Sides kept at 0.3:
screenshot_20181111_215633

Sides lifted to 0.7:
screenshot_20181111_215733

src/sky.cpp
Jordach: Renaming `m_bgcolor` and `m_skycolor` to `m_sky_color_bottom` and `m_sky_color_top` would be better since `cloudyfogcolor` was used once to calculate the difference between the top half and the bottom half the sky.

src/sky.cpp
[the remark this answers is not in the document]
paramat: I don't mean the bgcolor/skycolor blend should reach higher, i mean the opposite.
For a blue gradient reaching from horizon to 1.0, the colours could be tuned to not create visual banding.

src/sky.cpp
paramat: I don't see the relevance of that.
These 4 quads seem identical in code to the next 4 quads, so surely can be combined?

src/sky.cpp
paramat: Yes, but i can't see the relevance of that because you've moved the effective horizon to around -0.15.

src/sky.cpp
paramat: Sorry my mistake, both colours have alpha 0.

src/sky.cpp
paramat: Yes using new and more suitable names might be a good idea.

src/sky.cpp
Jordach: Yes, it just makes sure that grey skybox looks the same as it would be outside (if the sky was a similar colour with the sun and moon visible, so to speak).

src/sky.cpp
Jordach: It's trivial to change blending colours on the fly, but, having the blending done with pre-defined colours  (or supplied by set_sky or other) would be better.

src/sky.cpp
Jordach: We have the false horizontal horizon lowered, so that if you were on top of a mountain or hill, the horizon still exists, but not looking flat at the camera. But standing at sea level, it's clear that the sun or moon sets as normal. It just masks the lack of an actual curvature or fixed Y level horizon.

paramat:
sofar apart from subjective taste, this has some actual problems to do with lack of horizon and shifting the effective horizon to 15% below the true horizon, which are both clearly wrong.

Contrary to the claims in the first post, this makes very significant changes which suggests this should be optional. Jordach is already preparing code for additional custom sky implementations so this could be one of those options, it could be another option within 'set sky' or maybe a clientside option.

src/sky.cpp
paramat: We could perhaps have skycolor gradient from horizon haze top to 0.3, then a darker blue from 0.3 to 1.0.

Jordach:
@paramat I'm wanting to redo set_sky (instead of a fat oversized PR) and do it in more human sized steps. Removing the need to abuse texture modifications in real time (and our limited VRAM on some machines) and use a smoothed skybox would save time, and add useful eye candy to a boring dated looking skybox. It's why my next pull after this would be allowing `set_sky` or another `player:set_?` should handle this.

src/sky.cpp
Jordach: That makes sense, however, doing so would interfere with rendering of fog, but would be easier. However, with the near cyanic colour of the current skies, darkening it would make it look horrendous. I'd say bump the height of the side to 0.65, then change actual sky colours later on to look cleaner and less comic-like.

paramat:
> It's why my next pull after this would be allowing set_sky or another player:set_? should handle this.

So this would become optional? if so that's very welcome.

My updated suggestions are:
Don't alter the horizon haze at all, and don't alter the lower half of the skybox. Or at the least, keep a sharp horizon haze top at 0.12 as before, to visualy define the absolute horizon.
Only use a gradient for the blue upper part of the sky, keeping it's colour identical at the point it meets the horizon haze and getting darker towards 0.3 or whatever looks best, then above that use a darker blue.

Jordach:
Clarification: I meant changing of the actual sky colour, not actually set the horizon modes. Reworking the skybox mechanism would be an excessive and futile way of doing it.

paramat:
Ok, i think this needs to be optional due to the significance of the change.

Jordach:
Then that makes this whole PR and better graphical updates fruitless.

paramat:
Why? Options are good, it could even be enabled by default.

Jordach:
I'd restore the old one with another pull, however, letting me finish this with room to expand later on would be the right thing to do.

paramat:
> I'd restore the old one with another pull, however, letting me finish this [...]

Not sure i understand but, if you're intending to make this optional eventually (are you?) it should be made optional in this PR if possible, since the optionality is of so much importance here. Core devs cannot ignore serious issues on the hope they will be fixed later :)

Jordach:
@paramat I'd personally prefer a minetest.conf switch over making set_sky bend over, since as we'd all know, it looks far better as a table with support for missing values, over the current (and inane) setup. I'll see what I can do in this pull.

Jordach:
I've gone and added the minetest.conf switch `use_old_skybox` as a way to denote the 0.4 era skybox. You can find the advanced settings option under Graphics -> In-Game -> Basic. By default, this is set to false.

Pre smooth_sky branch:
screenshot_20181111_234140

Post smooth_sky branch:
screenshot_20181111_234544

With `use_old_skybox = false`:
screenshot_20181111_234544

And the option in advanced settings:
screenshot_20181111_234337

Any questions?

paramat:
> I'd personally prefer a minetest.conf switch

Yeah that's fine, thanks for doing this :)

Jordach:
@paramat You're welcome, I've given the skybox a slight height boost to 0.5 I believe. Should be a tiny bit better than it was, and the horizon that the sun sets at is actually size of one sun lower than it should be.

paramat:
So anyway, let's discuss the geometry of a MT world.

Looking at the first screenshot in the first post, i'm guessing that you have moved the horizon down to -0.15 so that it is now located where the typical edge of the visible mapblocks is, to make the edge of the visible mapblocks the new horizon?

This location is not the true horizon of a MT world, as there are always more distant non-visible mapblocks beyond it that raise the horizon up higher. Imagine if the player was looking at a 1000 node distance of visible mapblocks, the horizon is now far higher up and very close to 0 degrees view pitch angle.

The true location of the horizon of a MT world is average terrain level (y = 0) at the world's edge, on average roughly 31000 nodes away.
The average altitude of a player is a little way above sea level, lets say y = 32 (we have to use an average as the horizon location on the skybox is not altered according to player altitude).

If you calculate the view pitch angle from player to the true horizon it is therefore negative but very very close to 0 degrees (completely level) because 31000 is very much larger than 32. (angle = -invtan(32 / 31000) = -0.059 degrees).
A view pitch of -0.059 degrees corresponds to vertex co-ordinate -0.001 on the sides of the skybox (almost the exact midpoint of the skybox side). (coord = tan(0.059) * 1)

Jordach:
Seeing as the skybox Y horizon never matches the world Y position, at least
from tall hills and mountains it looks just right.

Otherwise, on a flat area, it looks almost like the classic skybox. I was
accounting for higher vantage points, not just flat areas.

paramat:
Ok, let's say the player is at cloud level y = 128 as a typical time-averaged altitude that takes into account time spent in high terrain. The view pitch angle is now -0.24 degrees, still very close to 0, and the vertex co-ordinate is -0.004 on the skybox sides, still very close to the middle and even less than the current -0.02.

src/sky.cpp
paramat: Ok i probably misunderstand, is it because it needs to match the alpha colours added later to the East and West?

Jordach:
At worst it looks a little wrong, but at best, including rendering ranges,
it should look "roughly" right.

src/sky.cpp
paramat: Ok, this discussion isn't much of an issue anyway, marking as resolved :)

src/sky.cpp
Jordach: Eh, it's fine. Who cares other than the people demanding an *ultra realistic game with brown tinted shaders*. I mean, it looks fancy, and that's what I aimed for.

src/sky.cpp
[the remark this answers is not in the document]
Jordach: Banding can always add some charm to a boring bland skybox, almost as if it's a game using dithering. I obviously want to avoid it, but can be a charmer, sort of.

paramat:
It looks very obviously wrong to me, sorry.

Now about horizon haze.

IRL, as your view pitch angle approaches 0, you are exponentially looking through more and more atmosphere, and the whitish haze that occurs toward the horizon tends to get exponentially stonger. The result of this exponential gradient is that the location of the horizon is obvious even if you couldn't see land, sea or clouds. For example see the following photo, it is somewhat clear where the horizon is due to the rapid haze gradient over a small distance:

travel-photo-sicily-mount-etna-lone-figure-sits-on-volcanic-rock-gazing-out-at-distant-horizon-by-david-j-rodger

Obviously an exponential gradient is not possible within 1 quad so you could use 2 bands of differing gradients.
Note also that the sky is definitely blue only a little way above the horizon, i feel your code makes the whitish haze extend too far up the skybox side.

So it's possible to avoid any sharp transitions but to also clearly define the horizon, by creating a rapid gradient near the horizon, where the banding is currently (so i'm no longer requesting a sharp transition).
A clearly defined horizon is very important to keep the expansive appearence of the world, and for visual orientation because it defines a completely level line. The edge of the visible mapblocks is all over the place and not level, so those create a bad looking horizon.

Jordach:
If I'd used white as the bottom sky colour, you'd notice that the 0.3 height for blending fits as you described it, where the fade starts just under the horizon.

paramat:
See these 2 code blocks:
```
			// Draw middle band of our smooth skybox
			for (u32 j = 0; j < 4; j++) {
				video::SColor c = m_sky_color_bottom;
					vertices[0] = video::S3DVertex(-1, -0.2, -1, 0, 0, 1, c, t, t);
					vertices[1] = video::S3DVertex( 1, -0.2, -1, 0, 0, 1, c, o, t);
					vertices[2] = video::S3DVertex( 1, -0.1, -1, 0, 0, 1, c, o, o);
					vertices[3] = video::S3DVertex(-1, -0.1, -1, 0, 0, 1, c, t, o);
				for (video::S3DVertex &vertex : vertices) {
					if (j == 0)
						// Don't switch
						{}
					else if (j == 1)
						// Switch from -Z (south) to +X (east)
						vertex.Pos.rotateXZBy(90);
					else if (j == 2)
						// Switch from -Z (south) to -X (west)
						vertex.Pos.rotateXZBy(-90);
					else
						// Switch from -Z (south) to +Z (north)
						vertex.Pos.rotateXZBy(-180);
				}
				driver->drawIndexedTriangleFan(&vertices[0], 4, indices, 2);
			}

			// Draw bottom sides of the sky
			for (u32 j = 0; j < 4; j++) {
				video::SColor c = m_sky_color_bottom;
					vertices[0] = video::S3DVertex(-1, -0.2, -1, 0, 0, 1, c, t, t);
					vertices[1] = video::S3DVertex( 1, -0.2, -1, 0, 0, 1, c, o, t);
					vertices[2] = video::S3DVertex( 1, -1,   -1, 0, 0, 1, c, o, o);
					vertices[3] = video::S3DVertex(-1, -1,   -1, 0, 0, 1, c, t, o);
				for (video::S3DVertex &vertex : vertices) {
					if (j == 0)
						// Don't switch
						{}
					else if (j == 1)
						// Switch from -Z (south) to +X (east)
						vertex.Pos.rotateXZBy(90);
					else if (j == 2)
						// Switch from -Z (south) to -X (west)
						vertex.Pos.rotateXZBy(-90);
					else
						// Switch from -Z (south) to +Z (north)
						vertex.Pos.rotateXZBy(-180);
				}
				driver->drawIndexedTriangleFan(&vertices[0], 4, indices, 2);
```
I've looked through them a few times and as far as i can see they are completely identical apart from vertex co-ordinates, so this is unnecessarily drawing twice as many quads as needed, they can be combined to draw from -0.1 to -1.

src/sky.cpp
paramat: It has nothing to do with ultra realism, as you know i would be the last to insist on ultra realism :)
In an abstract MT world the horizon is still unarguably at a particular elevation, this is indpendent of realism.
You seem to now be arguing for sloppiness, we care because we care about MT and it's quality.
Fanciness alone is not enough, there also has to be quality and precision :)

paramat:
>  I've given the skybox a slight height boost to 0.5

Maybe there was a misunderstanding, but this is worse. The issue i pointed out was that the whitish haze extends too far up into the sky already when using 0.3.
I don't mind the use of gradients but they should roughly match the locations we had before, which is haze blending up to vertex co-ordinate 0.12.

paramat:
>  at least from tall hills and mountains it looks just right.
Otherwise, on a flat area, it looks almost like the classic skybox.

It looks very different even when the player is at a low altitude:

crop

This is a crop of your first screenshot. As well as being visible through water, the sun is now setting / rising behind the edge of a mapblock roughly 100 nodes away, so the world now looks tiny.
In MT master, the sun sets behind an almost infinitely distant horizon, creating an impression of extreme spaciousness:

screenshot_20181112_034158

The horizon is also completely level instead of rough, tilted and constantly changing according to the visible mapblocks.

Jordach:
I've tried a few new daytime colours at the moment, seems like the ones used in my game really do show off the fancy skybox drawing logic.

screenshot_20181112_134414
screenshot_20181112_134421

src/sky.cpp
paramat: See comments in the thread, the -0.02 used in master is well chosen to account for an average player altitude of y = 500. Please move the horizon back to -0.02, as in, the middle point of the gradient that hides the sun and moon should be at -0.02.

Jordach:
I've gone ahead with testing the height for the shading locally to -0.03.

I've been working on testing some other sky colours, either taken from my own mod for a deep sky blue, and some tweaking to make them closer to classic colours, but more contrasty.

However, you can re-enable the old colour scheme, with the old skybox mode, or `use_old_sky_color`.

Tweaked sky colours; as of like 4PM:
screenshot_20181112_182352
screenshot_20181112_182350
screenshot_20181112_182346

paramat:
Looking at the 2nd screenshot in https://github.com/minetest/minetest/pull/7853#issuecomment-437888007, those custom colours with higher contrast do make the horizon haze line more noticeable. A deeper blue for upper sky looks good.

So whatever colours are chosen will affect how rapid the gradient needs to be to create a clear horizon line.
In my own experiments, the MT master colours for cloudyfogcolor and skycolor are so similar the only way to get a clear horizon line is to have a very small gradient, for example from 0.08 to 0.12 which is only the width of the MT master transition band.
> I've gone ahead with testing the height for the shading locally to -0.03.

Please use -0.02, as in, the centre of the cutoff being at -0.02, for example the cutoff gradient could be from -0.01 to -0.03 or whatever.

Jordach:
> Looking at the 2nd screenshot in [#7853 (comment)](https://github.com/minetest/minetest/pull/7853#issuecomment-437888007), those custom colours with higher contrast do make the horizon haze line more noticeable. A deeper blue for upper sky looks good.
> 
> So whatever colours are chosen will affect how rapid the gradient needs to be to create a clear horizon line.
> In my own experiments, the MT master colours for cloudyfogcolor and skycolor are so similar the only way to get a clear horizon line is to have a very small gradient, for example from 0.08 to 0.12 which is only the width of the MT master transition band.
> 
> > I've gone ahead with testing the height for the shading locally to -0.03.
> 
> Please use -0.02, as in, the centre of the cutoff being at -0.02, for example the cutoff gradient could be from -0.01 to -0.03 or whatever.

The horizon fade in (for sun, moon) starts at -0.03, and ends at 0.07.

paramat:
There may be a bug, but how the night sky is has been carefully tuned in the presence of that bug, so fixing it does not necessarily result in a correct result. We need to fix it then retune night sky to somewhat preserve the appearence. 
I did notice when working on night sky that it seemed difficult to make it a dark blue instead of a dark grey-blue.
Please open a new issue or PR for this bug you have discovered. Fixing this bug should not be part of this PR. I've worked in night sky a lot so will attend to the bug.

src/sky.cpp
```diff
@@ -202,58 +203,132 @@ void Sky::render()
...
-				else if (j == 1)
-					// Switch from -Z (south) to +X (east)
-					vertex.Pos.rotateXZBy(90);
-				else if (j == 2)
-					// Switch from -Z (south) to -X (west)
-					vertex.Pos.rotateXZBy(-90);
-				else
-					// Switch from -Z (south) to +Z (north)
-					vertex.Pos.rotateXZBy(-180);
+			// Draw bottom sides of the sky
```
paramat: This code block can be combined with `// Draw middle band of our smooth skybox` above since the colour is identical.

src/sky.cpp
```diff
@@ -202,58 +203,132 @@ void Sky::render()
 
 		driver->setMaterial(m_materials[1]);
 
-		video::SColor cloudyfogcolor = m_bgcolor;
+		if (m_use_old_skybox) {
+			// Draw middle band of the sky, unsmoothed:
```
paramat: Please keep tjhe original comments for the classic skybox:
`// Draw far cloudy fog thing blended with skycolor`

src/sky.cpp
```diff
@@ -202,58 +203,132 @@ void Sky::render()
...
-				else if (j == 1)
-					// Switch from -Z (south) to +X (east)
-					vertex.Pos.rotateXZBy(90);
-				else if (j == 2)
-					// Switch from -Z (south) to -X (west)
-					vertex.Pos.rotateXZBy(-90);
-				else
-					// Switch from -Z (south) to +Z (north)
-					vertex.Pos.rotateXZBy(-180);
+			// Draw bottom half of the world, without smoothing
```
paramat: `// Draw far cloudy fog thing`

src/sky.cpp
```diff
@@ -202,58 +203,132 @@ void Sky::render()
...
+						// Switch from -Z (south) to +Z (north)
+						vertex.Pos.rotateXZBy(-180);
+				}
+				driver->drawIndexedTriangleFan(&vertices[0], 4, indices, 2);
+			}
+			
+			// Draw top band of our smooth skybox
+			for (u32 j = 0; j < 4; j++) {
+				video::SColor c = m_sky_color_bottom;
+				video::SColor c_blend = m_sky_color_top;
```
paramat: I think it's best to keep the original variable names, as either with or without a smooth gradient, the top is skycolor and the bottom is cloudyfogcolor, there's no need to change the varaiable names that the core devs are used to.

src/sky.cpp
```diff
@@ -202,58 +203,132 @@ void Sky::render()
...
+						vertex.Pos.rotateXZBy(-180);
+				}
+				driver->drawIndexedTriangleFan(&vertices[0], 4, indices, 2);
 			}
-			driver->drawIndexedTriangleFan(&vertices[0], 4, indices, 2);
 		}
 
-		// Draw bottom far cloudy fog thing
-		video::SColor c = cloudyfogcolor;
+		// Draw world bottom
```
paramat: // Draw bottom far cloudy fog thing

src/sky.cpp
```diff
@@ -519,23 +610,83 @@ void Sky::render()
...
-				if (j == 0)
-					// Switch from -Z (south) to +X (east)
-					vertex.Pos.rotateXZBy(90);
-				else
-					// Switch from -Z (south) to -X (west)
-					vertex.Pos.rotateXZBy(-90);
+		driver->setMaterial(m_materials[1]);
+
+		if (m_use_old_skybox) {
+			// Draw fog below east and west horizons
```
paramat: `// Draw far cloudy fog thing below east and west horizons`

src/sky.cpp
```diff
@@ -564,30 +715,53 @@ void Sky::update(float time_of_day, float time_brightness,
...
-	video::SColorf bgcolor_bright_normal_f(170. / 255, 200. / 255, 230. / 255, 1.0);
-	video::SColorf bgcolor_bright_dawn_f(0.666, 200. / 255 * 0.7, 230. / 255 * 0.5, 1.0);
-	video::SColorf bgcolor_bright_dawn_f(0.666, 0.549, 0.220, 1.0);
-	video::SColorf bgcolor_bright_dawn_f(0.666 * 1.2, 0.549 * 1.0, 0.220 * 1.0, 1.0);
-	video::SColorf bgcolor_bright_dawn_f(0.666 * 1.2, 0.549 * 1.0, 0.220 * 1.2, 1.0);
+	video::SColorf sky_bottom_bright_normal_f(170. / 255, 200. / 255, 230. / 255, 1.0);
+	video::SColorf sky_bottom_bright_dawn_f(0.666, 200. / 255 * 0.7, 230. / 255 * 0.5, 1.0);
+	video::SColorf sky_bottom_bright_dawn_f(0.666, 0.549, 0.220, 1.0);
+	video::SColorf sky_bottom_bright_dawn_f(0.666 * 1.2, 0.549 * 1.0, 0.220 * 1.0, 1.0);
+	video::SColorf sky_bottom_bright_dawn_f(0.666 * 1.2, 0.549 * 1.0, 0.220 * 1.2, 1.0);
```
paramat: Nuuuuu, please don't change these variable names, it's completely unnecessary to do so.

src/sky.cpp
```diff
@@ -564,30 +715,53 @@ void Sky::update(float time_of_day, float time_brightness,
...
+	video::SColorf sky_bottom_bright_indoor_f = video::SColor(255, 100, 100, 100);
+
+	// Ready our sky colours before we assign them based on skybox settings
+	video::SColorf sky_bottom_bright_normal_f;
+	video::SColorf sky_bottom_bright_dawn_f;
+	video::SColorf sky_bottom_bright_night_f;
+
+	video::SColorf sky_top_bright_normal_f;
+	video::SColorf sky_top_bright_dawn_f;
+	video::SColorf sky_top_bright_night_f;
```
paramat: Please don't change these variable names, it's completely unnecessary to do so.

src/sky.cpp
```diff
@@ -613,56 +787,73 @@ void Sky::update(float time_of_day, float time_brightness,
...
+				m_sky_color_top_bright_f = m_sky_color_top_bright_f.getInterpolated(
+					sky_top_bright_night_f, color_change_fraction);
 			} else {  // Day
-				m_bgcolor_bright_f = m_bgcolor_bright_f.getInterpolated(
-					bgcolor_bright_normal_f, color_change_fraction);
-				m_skycolor_bright_f = m_skycolor_bright_f.getInterpolated(
-					skycolor_bright_normal_f, color_change_fraction);
+				m_sky_color_bottom_bright_f = m_sky_color_bottom_bright_f.getInterpolated(
+					sky_bottom_bright_normal_f, color_change_fraction);
+
```
paramat: Here and above, unnecessary empty lines.

src/sky.cpp
```diff
@@ -613,56 +787,73 @@ void Sky::update(float time_of_day, float time_brightness,
...
-		m_skycolor_bright_f = m_skycolor_bright_f.getInterpolated(
-			bgcolor_bright_indoor_f, color_change_fraction);
+		m_sky_color_bottom_bright_f = m_sky_color_bottom_bright_f.getInterpolated(
+			sky_bottom_bright_indoor_f, color_change_fraction);
+		m_sky_color_top_bright_f = m_sky_color_top_bright_f.getInterpolated(
+			sky_bottom_bright_indoor_f, color_change_fraction);
 		m_cloudcolor_bright_f = m_cloudcolor_bright_f.getInterpolated(
 			cloudcolor_bright_normal_f, color_change_fraction);
 		m_clouds_visible = false;
 	}
```
paramat: The entire code block above has unnecessary changes due to an unnecessary changing of variable names.

src/sky.cpp
paramat: Essentially, unless there's good reason to, it's better to change as little as possible instead of as much as possible, to make the diff smaller and minimise disruption and confusion caused by changes.

src/sky.cpp
```diff
@@ -519,23 +610,83 @@ void Sky::render()
...
+			}
+		} else {
+			// Draw lower blend fog band
+			for (u32 j = 0; j < 2; j++) {
+				video::SColor c = m_sky_color_bottom;
+				video::SColor c_blend = m_sky_color_bottom_alpha;
+				vertices[0] = video::S3DVertex(-1, -0.03,  -1, 0, 0, 1, c, t, t);
+				vertices[1] = video::S3DVertex( 1, -0.03,  -1, 0, 0, 1, c, o, t);
+				vertices[2] = video::S3DVertex( 1,  0.02, -1, 0, 0, 1, c_blend, o, o);
+				vertices[3] = video::S3DVertex(-1,  0.02, -1, 0, 0, 1, c_blend, t, o);
```
paramat: Thanks, horizon is back to a good place now :)

paramat:
Please could you remove the generated files from this PR: conf.example and settings_translation_file. We update these independently, updating them in a PR increases the chance of conflicts.

paramat:
Still much to improve, but horizon is back to a good place so i've removed my -1. Thanks for your patience.

Please could you explain why the sunrise texture is drawn upside-down and underneath the upright version? Is it because it tends to be very slightly visible even when below the horizon so you are essentially removing it's sharp lower edge?

paramat:
I assume what you've done to the sunrise texture is fix the chopped right edge by taking the left half and mirroring it?

paramat:
Well, anything below the horizon is not seen, and the bottom of the existing texture never rose above the horizon?
I know you had some discussion in IRC about what you were doing, i'm requesting that this explanation is here in the thread too where it needs to be.

paramat:
Actually, now that you've moved the hard cutoff horizon to -0.03 the base of the upright texture is always below the horizon, the upside-down texture is now never seen currently.
With the upside-down texture removed there won't be any significant changes to sunrise texture, so it will then not need to be switched by the option.

paramat:
`float y = -(1.0 - a) * 0.3;`
This is a downwards movement to hide the texture behiind the horizon.
The texture top is at 0.2 and your complete horizon cutoff is now at -0.03 so this line now needs to be:
`float y = -(1.0 - a) * 0.23;`

paramat:
See https://github.com/minetest/minetest/pull/7750#issuecomment-438934177
"Ninja Edit: This also fixes sun.png and moon.png being wrongly applied to the sunrise/set glow faces." Please can you open a separate PR to fix this bug?

lhofhansl:
Good stuff!

MT can use a slight face lift :)
Very much in favor of this.

Patch does not apply for me, though. (not luck with git am)

paramat:
@lhofhansl which part of this PR code fixes https://github.com/minetest/minetest/pull/7750#issuecomment-438934177 ? I'll make the bugfix.

lhofhansl:
@paramat honestly I don't know. Does the patch apply for you (I can, of course, directly test on the fork)

paramat:
lhofhansl i get the impression Jordach is still experimenting and adjusting, there are also some code errors and many issues to sort out first, see my comments above. So rebase is not essential at this point.

src/sky.cpp
paramat: Actually, since in MT master we have
`video::SColor cloudyfogcolor = m_bgcolor;`
It seems clearest to change `bgcolor` to `cloudyfogcolor` everywhere. I agree with you that `bgcolor` is unclear and confusing. But please keep `skycolor` unchanged as it is accurate and familiar. So then the names are improved but changes are minimised.

src/sky.cpp
paramat: Actually, since in MT master we have
`video::SColor cloudyfogcolor = m_bgcolor;`
It seems clearest to change `bgcolor` to `cloudyfogcolor` everywhere. I agree with you that `bgcolor` is unclear and confusing. But please keep `skycolor` unchanged as it is accurate and familiar. So then the names are improved but changes are minimised.

src/sky.cpp
paramat: Actually, since in MT master we have
`video::SColor cloudyfogcolor = m_bgcolor;`
It seems clearest to change `bgcolor` to `cloudyfogcolor` everywhere. I agree with you that `bgcolor` is unclear and confusing. But please keep `skycolor` unchanged as it is accurate and familiar. So then the names are improved but changes are minimised.

src/sky.cpp
paramat: Actually, since in MT master we have
`video::SColor cloudyfogcolor = m_bgcolor;`
It seems clearest to change `bgcolor` to `cloudyfogcolor` everywhere. I agree with you that `bgcolor` is unclear and confusing. But please keep `skycolor` unchanged as it is accurate and familiar. So then the names are improved but changes are minimised.

src/sky.h
```diff
@@ -133,12 +133,16 @@ class Sky : public scene::ISceneNode
 	bool m_clouds_visible; // Whether clouds are disabled due to player underground
 	bool m_clouds_enabled = true; // Initialised to true, reset only by set_sky API
 	bool m_directional_colored_fog;
+	bool m_use_old_skybox; // Enable to display the classic MT skybox
+	bool m_use_old_colors; // Enable to use pre 5.0 sky colours
 	bool m_bodies_visible = true; // sun, moon, stars
-	video::SColorf m_bgcolor_bright_f = video::SColorf(1.0f, 1.0f, 1.0f, 1.0f);
```
paramat: Actually, since in MT master we have
`video::SColor cloudyfogcolor = m_bgcolor;`
It seems clearest to change `bgcolor` to `cloudyfogcolor` everywhere. I agree with you that `bgcolor` is unclear and confusing. But please keep `skycolor` unchanged as it is accurate and familiar. So then the names are improved but changes are minimised.

paramat:
Jordach, since in MT master we have
`video::SColor cloudyfogcolor = m_bgcolor;`
It seems clearest to change `bgcolor` to `cloudyfogcolor` everywhere. I agree with you that `bgcolor` is unclear and confusing. But please keep `skycolor` unchanged as it is accurate and familiar. So then the names are improved but changes are minimised.

paramat:
Note the author is currently rebasing and reworking this, i'm guessing it will be a new PR, so closing, but if not this should be reopened.

paramat:
@Jordach i know you are currently significantly reworking this feature, will it be a new PR or not? If so please close this PR.

happysmash27:
This sky is amazing! I feel like this is one of the things Minetest has been missing for ages now. Thank you so, so much! It makes the game actually look good!

tacotexmex:
No new PR for this feature?

TumeniNodes:
>No new PR for this feature?

>@Jordach i know you are currently significantly reworking this feature, will it be a new PR or not? If so please close this PR.

obviously, that is the answer. It means Jordach is most likely working on it.

lhofhansl:
@Jordach Are you still working on this?

PR #7873, opened 2018-11-17
Fix Android build

numberZero:
http://irc.minetest.net/minetest-dev/2018-11-16#i_5440265 (https://github.com/minetest/minetest/pull/7839#issuecomment-439547256)
Older IrrLicht revision used on Android is labelled as 1.9 but lacks `roundingError`, so I add the prototype.


srifqi:
It seems like this fixes that issue. +1

sfan5:
does this break anything for "desktop" irrlicht 1.9?

nerzhul:
Nice; it's trivial i merge

PR #7922, opened 2018-12-01
Improve world configure menu

pauloue:
Replace 'Subgame mods' with \<game name\> mods.
Display game description on left instead of dependencies.
Fix double DIR_DELIM in some paths.
Enable/disable MP --> modpack (fits nicely without increasing button size)

Does this require any changes to a translation file?

Before:
before
After:
after
Note: MTG could use a description in mod.conf.


pauloue:
Made the modpack button slightly wider, now it has as much padding as many other buttons. Also Aligned the Enable all button with the edge of the table (the math made sense before, not sure why extra .2 is needed).

PR #6537, opened 2017-10-16
Bind mouse buttons

samcaulfield:
These patches accomplish the following:

1. Create two new actions: dig and place, for digging and placing blocks. These actions can be bound to any key just like the other player actions e.g. sneak, jump. This involves removing hardcoded associations between the mouse and these actions from game.cpp.
2. Allow the mouse buttons to be bound to any action in the GUI just like any other button.

PR #7924, opened 2018-12-01
Allow binding dig, place actions to keys; remove LMB/RMB hardcoding

ClobberXD:
[Adopted from #6537]

- Remove hard-coded association of dig and place actions, with LMB and RMB respectively.
  - These actions are still bound to LMB and RMB by default.
- Allow binding actions to mouse buttons.
- `ObjectRef:get_player_control` now returns `dig` and `place` fields. Note: The old `LMB` and `RMB` fields still exist alongside `dig` and `place`, to ensure backwards compatibility.

Tested, works. This PR is ready for review.

#### Video demonstration

https://youtu.be/0Ca7nEMiq_U

#### How to test

- Suggested keymap to test this PR:
  ```
  keymap_place = KEY_KEY_L
  keymap_dig = KEY_KEY_X
  keymap_inventory = KEY_LBUTTON
  keymap_jump = KEY_RBUTTON
  ```
- Tweak `repeat_place_interval`, and test it by selecting a place-able item (e.g. a node), pointing at a node, and holding down the "place" keybind.

Due to code changes around keypress detection, it's advisable to also test various digging, placing, use-ing, and secondary_use-ing scenarios, and carefully observe that nothing has broken. Note that this PR doesn't modify any part of the actual digging and placing, even in a subtle manner. So if a change is noticeable, please do let me know.

### DIFF
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -110,9 +110,9 @@ doubletap_jump (Double tap jump for fly) bool false
 #    enabled.
 always_fly_fast (Always fly and fast) bool true
 
-#    The time in seconds it takes between repeated right clicks when holding the right
-#    mouse button.
-repeat_rightclick_time (Rightclick repetition interval) float 0.25 0.001
+#    The time in seconds it takes between repeated node placements when holding
+#    the place button.
+repeat_place_time (Place repetition interval) float 0.25 0.001
 
 #    Automatically jump up single-node obstacles.
 autojump (Automatic jumping) bool false
@@ -181,6 +181,14 @@ keymap_jump (Jump key) key KEY_SPACE
 #    Also used for climbing down and descending in water if aux1_descends is disabled.
 #    See http://irrlicht.sourceforge.net/docu/namespaceirr.html#a54da2a0e231901735e3da1b0edf72eb3
 keymap_sneak (Sneak key) key KEY_LSHIFT
+
+#    Key for digging.
+#    See http://irrlicht.sourceforge.net/docu/namespaceirr.html#a54da2a0e231901735e3da1b0edf72eb3
+keymap_dig (Dig key) key KEY_LBUTTON
+
+#    Key for placing.
+#    See http://irrlicht.sourceforge.net/docu/namespaceirr.html#a54da2a0e231901735e3da1b0edf72eb3
+keymap_place (Place key) key KEY_RBUTTON
 
 #    Key for opening the inventory.
 #    See http://irrlicht.sourceforge.net/docu/namespaceirr.html#a54da2a0e231901735e3da1b0edf72eb3
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -6105,15 +6105,23 @@ object you are working with still exists.
     * Only affects formspecs shown after this is called.
 * `get_formspec_prepend(formspec)`: returns a formspec string.
 * `get_player_control()`: returns table with player pressed keys
-    * The table consists of fields with boolean value representing the pressed
-      keys, the fields are jump, right, left, LMB, RMB, sneak, aux1, down, up, zoom.
-    * example: `{jump=false, right=true, left=false, LMB=false, RMB=false,
-      sneak=true, aux1=false, down=false, up=false, zoom=false}`
-    * The `zoom` field is available since 5.3
+    * The table consists of fields with the following boolean values
+      representing the pressed keys: `up`, `down`, `left`, `right`, `jump`,
+      `aux1`, `sneak`, `dig`, `place`, `LMB`, `RMB`, and `zoom`.
+    * The fields `LMB` and `RMB` are equal to `dig` and `place` respectively,
+      and exist only to preserve backwards compatibility.
 * `get_player_control_bits()`: returns integer with bit packed player pressed
-  keys.
-    * bit nr/meaning: 0/up, 1/down, 2/left, 3/right, 4/jump, 5/aux1, 6/sneak,
-      7/LMB, 8/RMB, 9/zoom (zoom available since 5.3)
+  keys. Bits:
+    * 0 - up
+    * 1 - down
+    * 2 - left
+    * 3 - right
+    * 4 - jump
+    * 5 - aux1
+    * 6 - sneak
+    * 7 - dig
+    * 8 - place
+    * 9 - zoom
 * `set_physics_override(override_table)`
     * `override_table` is a table with the following fields:
         * `speed`: multiplier to default walking speed value (default: `1`)
--- a/src/client/game.cpp
+++ b/src/client/game.cpp
@@ -602,7 +602,6 @@ public:
 #endif
 
 /****************************************************************************
-
  ****************************************************************************/
 
 const float object_hit_delay = 0.2;
@@ -623,15 +622,15 @@ struct GameRunData {
 	u16 new_playeritem;
 	PointedThing pointed_old;
 	bool digging;
-	bool ldown_for_dig;
+	bool punching;
+	bool btn_down_for_dig;
 	bool dig_instantly;
 	bool digging_blocked;
-	bool left_punch;
 	bool reset_jump_timer;
 	float nodig_delay_timer;
 	float dig_time;
 	float dig_time_complete;
-	float repeat_rightclick_timer;
+	float repeat_place_timer;
 	float object_hit_delay_timer;
 	float time_from_last_punch;
 	ClientActiveObject *selected_object;
@@ -796,6 +795,14 @@ protected:
 	{
 		return input->wasKeyDown(k);
 	}
+	inline bool wasKeyPressed(GameKeyType k)
+	{
+		return input->wasKeyPressed(k);
+	}
+	inline bool wasKeyReleased(GameKeyType k)
+	{
+		return input->wasKeyReleased(k);
+	}
 
 #ifdef __ANDROID__
 	void handleAndroidChatInput();
@@ -910,7 +917,7 @@ private:
 	bool m_cache_enable_free_move;
 	f32  m_cache_mouse_sensitivity;
 	f32  m_cache_joystick_frustum_sensitivity;
-	f32  m_repeat_right_click_time;
+	f32  m_repeat_place_time;
 	f32  m_cache_cam_smoothing;
 	f32  m_cache_fog_start;
 
@@ -944,7 +951,7 @@ Game::Game() :
 		&settingChangedCallback, this);
 	g_settings->registerChangedCallback("joystick_frustum_sensitivity",
 		&settingChangedCallback, this);
-	g_settings->registerChangedCallback("repeat_rightclick_time",
+	g_settings->registerChangedCallback("repeat_place_time",
 		&settingChangedCallback, this);
 	g_settings->registerChangedCallback("noclip",
 		&settingChangedCallback, this);
@@ -1002,7 +1009,7 @@ Game::~Game()
 		&settingChangedCallback, this);
 	g_settings->deregisterChangedCallback("mouse_sensitivity",
 		&settingChangedCallback, this);
-	g_settings->deregisterChangedCallback("repeat_rightclick_time",
+	g_settings->deregisterChangedCallback("repeat_place_time",
 		&settingChangedCallback, this);
 	g_settings->deregisterChangedCallback("noclip",
 		&settingChangedCallback, this);
@@ -2482,8 +2489,8 @@ void Game::updatePlayerControl(const CameraOrientation &cam)
 		isKeyDown(KeyType::SPECIAL1),
 		isKeyDown(KeyType::SNEAK),
 		isKeyDown(KeyType::ZOOM),
-		input->getLeftState(),
-		input->getRightState(),
+		isKeyDown(KeyType::DIG),
+		isKeyDown(KeyType::PLACE),
 		cam.camera_pitch,
 		cam.camera_yaw,
 		input->joystick.getAxisWithoutDead(JA_SIDEWARD_MOVE),
@@ -2498,8 +2505,8 @@ void Game::updatePlayerControl(const CameraOrientation &cam)
 			( (u32)(isKeyDown(KeyType::JUMP)                          & 0x1) << 4) |
 			( (u32)(isKeyDown(KeyType::SPECIAL1)                      & 0x1) << 5) |
 			( (u32)(isKeyDown(KeyType::SNEAK)                         & 0x1) << 6) |
-			( (u32)(input->getLeftState()                             & 0x1) << 7) |
-			( (u32)(input->getRightState()                            & 0x1) << 8) |
+			( (u32)(isKeyDown(KeyType::DIG)                           & 0x1) << 7) |
+			( (u32)(isKeyDown(KeyType::PLACE)                         & 0x1) << 8) |
 			( (u32)(isKeyDown(KeyType::ZOOM)                          & 0x1) << 9)
 		);
 
@@ -3081,7 +3088,7 @@ void Game::processPlayerInteraction(f32 dtime, bool show_hud, bool show_debug)
 
 	PointedThing pointed = updatePointedThing(shootline,
 			selected_def.liquids_pointable,
-			!runData.ldown_for_dig,
+			!runData.btn_down_for_dig,
 			camera_offset);
 
 	if (pointed != runData.pointed_old) {
@@ -3089,20 +3096,18 @@ void Game::processPlayerInteraction(f32 dtime, bool show_hud, bool show_debug)
 		hud->updateSelectionMesh(camera_offset);
 	}
 
-	if (runData.digging_blocked && !input->getLeftState()) {
-		// allow digging again if button is not pressed
+	// Allow digging again if button is not pressed
+	if (runData.digging_blocked && !isKeyDown(KeyType::DIG))
 		runData.digging_blocked = false;
-	}
 
 	/*
 		Stop digging when
-		- releasing left mouse button
+		- releasing dig button
 		- pointing away from node
 	*/
 	if (runData.digging) {
-		if (input->getLeftReleased()) {
-			infostream << "Left button released"
-					<< " (stopped digging)" << std::endl;
+		if (wasKeyReleased(KeyType::DIG)) {
+			infostream << "Dig button released (stopped digging)" << std::endl;
 			runData.digging = false;
 		} else if (pointed != runData.pointed_old) {
 			if (pointed.type == POINTEDTHING_NODE
@@ -3112,8 +3117,7 @@ void Game::processPlayerInteraction(f32 dtime, bool show_hud, bool show_debug)
 				// Still pointing to the same node, but a different face.
 				// Don't reset.
 			} else {
-				infostream << "Pointing away from node"
-						<< " (stopped digging)" << std::endl;
+				infostream << "Pointing away from node (stopped digging)" << std::endl;
 				runData.digging = false;
 				hud->updateSelectionMesh(camera_offset);
 			}
@@ -3124,55 +3128,57 @@ void Game::processPlayerInteraction(f32 dtime, bool show_hud, bool show_debug)
 			client->setCrack(-1, v3s16(0, 0, 0));
 			runData.dig_time = 0.0;
 		}
-	} else if (runData.dig_instantly && input->getLeftReleased()) {
-		// Remove e.g. torches faster when clicking instead of holding LMB
+	} else if (runData.dig_instantly && wasKeyReleased(KeyType::DIG)) {
+		// Remove e.g. torches faster when clicking instead of holding dig button
 		runData.nodig_delay_timer = 0;
 		runData.dig_instantly = false;
 	}
 
-	if (!runData.digging && runData.ldown_for_dig && !input->getLeftState()) {
-		runData.ldown_for_dig = false;
-	}
+	if (!runData.digging && runData.btn_down_for_dig && !isKeyDown(KeyType::DIG))
+		runData.btn_down_for_dig = false;
 
-	runData.left_punch = false;
+	runData.punching = false;
 
 	soundmaker->m_player_leftpunch_sound.name = "";
 
 	// Prepare for repeating, unless we're not supposed to
-	if (input->getRightState() && !g_settings->getBool("safe_dig_and_place"))
-		runData.repeat_rightclick_timer += dtime;
+	if (isKeyDown(KeyType::PLACE) && !g_settings->getBool("safe_dig_and_place"))
+		runData.repeat_place_timer += dtime;
 	else
-		runData.repeat_rightclick_timer = 0;
+		runData.repeat_place_timer = 0;
 
-	if (selected_def.usable && input->getLeftState()) {
-		if (input->getLeftClicked() && (!client->modsLoaded()
-				|| !client->getScript()->on_item_use(selected_item, pointed)))
+	if (selected_def.usable && isKeyDown(KeyType::DIG)) {
+		if (wasKeyPressed(KeyType::DIG) && (!client->modsLoaded() ||
+				!client->getScript()->on_item_use(selected_item, pointed)))
 			client->interact(INTERACT_USE, pointed);
 	} else if (pointed.type == POINTEDTHING_NODE) {
 		handlePointingAtNode(pointed, selected_item, hand_item, dtime);
 	} else if (pointed.type == POINTEDTHING_OBJECT) {
 		v3f player_position  = player->getPosition();
 		handlePointingAtObject(pointed, tool_item, player_position, show_debug);
-	} else if (input->getLeftState()) {
+	} else if (isKeyDown(KeyType::DIG)) {
 		// When button is held down in air, show continuous animation
-		runData.left_punch = true;
+		runData.punching = true;
 		// Run callback even though item is not usable
-		if (input->getLeftClicked() && client->modsLoaded())
+		if (wasKeyPressed(KeyType::DIG) && client->modsLoaded())
 			client->getScript()->on_item_use(selected_item, pointed);
-	} else if (input->getRightClicked()) {
+	} else if (wasKeyPressed(KeyType::PLACE)) {
 		handlePointingAtNothing(selected_item);
 	}
 
 	runData.pointed_old = pointed;
 
-	if (runData.left_punch || input->getLeftClicked())
-		camera->setDigging(0); // left click animation
+	if (runData.punching || wasKeyPressed(KeyType::DIG))
+		camera->setDigging(0); // dig animation
 
-	input->resetLeftClicked();
-	input->resetRightClicked();
+	input->clearWasKeyPressed();
+	input->clearWasKeyReleased();
 
-	input->resetLeftReleased();
-	input->resetRightReleased();
+	input->joystick.clearWasKeyDown(KeyType::MOUSE_L);
+	input->joystick.clearWasKeyDown(KeyType::MOUSE_R);
+
+	input->joystick.clearWasKeyReleased(KeyType::MOUSE_L);
+	input->joystick.clearWasKeyReleased(KeyType::MOUSE_R);
 }
 
 
@@ -3269,7 +3275,7 @@ PointedThing Game::updatePointedThing(
 
 void Game::handlePointingAtNothing(const ItemStack &playerItem)
 {
-	infostream << "Right Clicked in Air" << std::endl;
+	infostream << "Attempted to place item while pointing at nothing" << std::endl;
 	PointedThing fauxPointed;
 	fauxPointed.type = POINTEDTHING_NOTHING;
 	client->interact(INTERACT_ACTIVATE, fauxPointed);
@@ -3288,7 +3294,7 @@ void Game::handlePointingAtNode(const PointedThing &pointed,
 
 	ClientMap &map = client->getEnv().getClientMap();
 
-	if (runData.nodig_delay_timer <= 0.0 && input->getLeftState()
+	if (runData.nodig_delay_timer <= 0.0 && isKeyDown(KeyType::DIG)
 			&& !runData.digging_blocked
 			&& client->checkPrivilege("interact")) {
 		handleDigging(pointed, nodepos, selected_item, hand_item, dtime);
@@ -3309,13 +3315,14 @@ void Game::handlePointingAtNode(const PointedThing &pointed,
 		}
 	}
 
-	if ((input->getRightClicked() ||
-			runData.repeat_rightclick_timer >= m_repeat_right_click_time) &&
+	if ((wasKeyPressed(KeyType::PLACE) ||
+			runData.repeat_place_timer >= m_repeat_place_time) &&
 			client->checkPrivilege("interact")) {
-		runData.repeat_rightclick_timer = 0;
-		infostream << "Ground right-clicked" << std::endl;
+		runData.repeat_place_timer = 0;
+		infostream << "Place button pressed while looking at ground" << std::endl;
 
-		camera->setDigging(1);  // right click animation (always shown for feedback)
+		// Placing animation (always shown for feedback)
+		camera->setDigging(1);
 
 		soundmaker->m_player_rightpunch_sound = SimpleSoundSpec();
 
@@ -3381,8 +3388,7 @@ bool Game::nodePlacement(const ItemDefinition &selected_def,
 	}
 
 	verbosestream << "Node placement prediction for "
-		<< selected_def.name << " is "
-		<< prediction << std::endl;
+		<< selected_def.name << " is " << prediction << std::endl;
 	v3s16 p = neighbourpos;
 
 	// Place inside node itself if buildable_to
@@ -3541,7 +3547,7 @@ void Game::handlePointingAtObject(const PointedThing &pointed,
 
 	m_game_ui->setInfoText(infotext);
 
-	if (input->getLeftState()) {
+	if (isKeyDown(KeyType::DIG)) {
 		bool do_punch = false;
 		bool do_punch_damage = false;
 
@@ -3551,12 +3557,12 @@ void Game::handlePointingAtObject(const PointedThing &pointed,
 			runData.object_hit_delay_timer = object_hit_delay;
 		}
 
-		if (input->getLeftClicked())
+		if (wasKeyPressed(KeyType::DIG))
 			do_punch = true;
 
 		if (do_punch) {
-			infostream << "Left-clicked object" << std::endl;
-			runData.left_punch = true;
+			infostream << "Punched object" << std::endl;
+			runData.punching = true;
 		}
 
 		if (do_punch_damage) {
@@ -3571,8 +3577,8 @@ void Game::handlePointingAtObject(const PointedThing &pointed,
 			if (!disable_send)
 				client->interact(INTERACT_START_DIGGING, pointed);
 		}
-	} else if (input->getRightClicked()) {
-		infostream << "Right-clicked object" << std::endl;
+	} else if (wasKeyDown(KeyType::PLACE)) {
+		infostream << "Pressed place button while pointing at object" << std::endl;
 		client->interact(INTERACT_PLACE, pointed);  // place
 	}
 }
@@ -3618,7 +3624,7 @@ void Game::handleDigging(const PointedThing &pointed, const v3s16 &nodepos,
 			return;
 		client->interact(INTERACT_START_DIGGING, pointed);
 		runData.digging = true;
-		runData.ldown_for_dig = true;
+		runData.btn_down_for_dig = true;
 	}
 
 	if (!runData.dig_instantly) {
@@ -3712,7 +3718,7 @@ void Game::handleDigging(const PointedThing &pointed, const v3s16 &nodepos,
 		client->setCrack(-1, nodepos);
 	}
 
-	camera->setDigging(0);  // left click animation
+	camera->setDigging(0);  // Dig animation
 }
 
 
@@ -4051,7 +4057,7 @@ void Game::readSettings()
 	m_cache_enable_fog                   = g_settings->getBool("enable_fog");
 	m_cache_mouse_sensitivity            = g_settings->getFloat("mouse_sensitivity");
 	m_cache_joystick_frustum_sensitivity = g_settings->getFloat("joystick_frustum_sensitivity");
-	m_repeat_right_click_time            = g_settings->getFloat("repeat_rightclick_time");
+	m_repeat_place_time                  = g_settings->getFloat("repeat_place_time");
 
 	m_cache_enable_noclip                = g_settings->getBool("noclip");
 	m_cache_enable_free_move             = g_settings->getBool("free_move");
@@ -4143,30 +4149,32 @@ void Game::showPauseMenu()
 		"- %s: move backwards\n"
 		"- %s: move left\n"
 		"- %s: move right\n"
-		"- %s: jump/climb\n"
-		"- %s: sneak/go down\n"
+		"- %s: jump/climb up\n"
+		"- %s: dig/punch\n"
+		"- %s: place/use\n"
+		"- %s: sneak/climb down\n"
 		"- %s: drop item\n"
 		"- %s: inventory\n"
 		"- Mouse: turn/look\n"
-		"- Mouse left: dig/punch\n"
-		"- Mouse right: place/use\n"
 		"- Mouse wheel: select item\n"
 		"- %s: chat\n"
 	);
 
-	 char control_text_buf[600];
+	char control_text_buf[600];
 
-	 porting::mt_snprintf(control_text_buf, sizeof(control_text_buf), control_text_template.c_str(),
-			GET_KEY_NAME(keymap_forward),
-			GET_KEY_NAME(keymap_backward),
-			GET_KEY_NAME(keymap_left),
-			GET_KEY_NAME(keymap_right),
-			GET_KEY_NAME(keymap_jump),
-			GET_KEY_NAME(keymap_sneak),
-			GET_KEY_NAME(keymap_drop),
-			GET_KEY_NAME(keymap_inventory),
-			GET_KEY_NAME(keymap_chat)
-			);
+	porting::mt_snprintf(control_text_buf, sizeof(control_text_buf), control_text_template.c_str(),
+		GET_KEY_NAME(keymap_forward),
+		GET_KEY_NAME(keymap_backward),
+		GET_KEY_NAME(keymap_left),
+		GET_KEY_NAME(keymap_right),
+		GET_KEY_NAME(keymap_jump),
+		GET_KEY_NAME(keymap_dig),
+		GET_KEY_NAME(keymap_place),
+		GET_KEY_NAME(keymap_sneak),
+		GET_KEY_NAME(keymap_drop),
+		GET_KEY_NAME(keymap_inventory),
+		GET_KEY_NAME(keymap_chat)
+	);
 
 	std::string control_text = std::string(control_text_buf);
 	str_formspec_escape(control_text);
--- a/src/client/inputhandler.cpp
+++ b/src/client/inputhandler.cpp
@@ -37,6 +37,8 @@ void KeyCache::populate()
 	key[KeyType::JUMP] = getKeySetting("keymap_jump");
 	key[KeyType::SPECIAL1] = getKeySetting("keymap_special1");
 	key[KeyType::SNEAK] = getKeySetting("keymap_sneak");
+	key[KeyType::DIG] = getKeySetting("keymap_dig");
+	key[KeyType::PLACE] = getKeySetting("keymap_place");
 
 	key[KeyType::AUTOFORWARD] = getKeySetting("keymap_autoforward");
 
@@ -111,57 +113,81 @@ bool MyEventReceiver::OnEvent(const SEvent &event)
 	if (event.EventType == irr::EET_KEY_INPUT_EVENT) {
 		const KeyPress &keyCode = event.KeyInput;
 		if (keysListenedFor[keyCode]) {
+				// If the key is being held down then the OS may
+				// send a continuous stream of keydown events.
+				// In this case, we don't want to let this
+				// stream reach the application as it will cause
+				// certain actions to repeat constantly.
 			if (event.KeyInput.PressedDown) {
+				if (!IsKeyDown(keyCode)) {
+					keyWasDown.set(keyCode);
+					keyWasPressed.set(keyCode);
+				}
 				keyIsDown.set(keyCode);
-				keyWasDown.set(keyCode);
 			} else {
+				if (IsKeyDown(keyCode))
+					keyWasReleased.set(keyCode);
+
 				keyIsDown.unset(keyCode);
 			}
+
 			return true;
 		}
-	}
 
 #ifdef HAVE_TOUCHSCREENGUI
-	// case of touchscreengui we have to handle different events
-	if (m_touchscreengui && event.EventType == irr::EET_TOUCH_INPUT_EVENT) {
+	} else if (m_touchscreengui && event.EventType == irr::EET_TOUCH_INPUT_EVENT) {
+		// In case of touchscreengui, we have to handle different events
 		m_touchscreengui->translateEvent(event);
 		return true;
-	}
 #endif
 
-	if (event.EventType == irr::EET_JOYSTICK_INPUT_EVENT) {
+	} else if (event.EventType == irr::EET_JOYSTICK_INPUT_EVENT) {
 		/* TODO add a check like:
 		if (event.JoystickEvent != joystick_we_listen_for)
 			return false;
 		*/
 		return joystick->handleEvent(event.JoystickEvent);
-	}
-	// handle mouse events
-	if (event.EventType == irr::EET_MOUSE_INPUT_EVENT) {
-		if (isMenuActive()) {
-			left_active = false;
-			middle_active = false;
-			right_active = false;
-		} else {
-			left_active = event.MouseInput.isLeftPressed();
-			middle_active = event.MouseInput.isMiddlePressed();
-			right_active = event.MouseInput.isRightPressed();
-
-			if (event.MouseInput.Event == EMIE_LMOUSE_PRESSED_DOWN) {
-				leftclicked = true;
-			}
-			if (event.MouseInput.Event == EMIE_RMOUSE_PRESSED_DOWN) {
-				rightclicked = true;
-			}
-			if (event.MouseInput.Event == EMIE_LMOUSE_LEFT_UP) {
-				leftreleased = true;
-			}
-			if (event.MouseInput.Event == EMIE_RMOUSE_LEFT_UP) {
-				rightreleased = true;
-			}
-			if (event.MouseInput.Event == EMIE_MOUSE_WHEEL) {
-				mouse_wheel += event.MouseInput.Wheel;
-			}
+	} else if (event.EventType == irr::EET_MOUSE_INPUT_EVENT) {
+		// Handle mouse events
+		KeyPress key;
+		switch (event.MouseInput.Event) {
+		case EMIE_LMOUSE_PRESSED_DOWN:
+			key = "KEY_LBUTTON";
+			keyIsDown.set(key);
+			keyWasDown.set(key);
+			keyWasPressed.set(key);
+			break;
+		case EMIE_MMOUSE_PRESSED_DOWN:
+			key = "KEY_MBUTTON";
+			keyIsDown.set(key);
+			keyWasDown.set(key);
+			keyWasPressed.set(key);
+			break;
+		case EMIE_RMOUSE_PRESSED_DOWN:
+			key = "KEY_RBUTTON";
+			keyIsDown.set(key);
+			keyWasDown.set(key);
+			keyWasPressed.set(key);
+			break;
+		case EMIE_LMOUSE_LEFT_UP:
+			key = "KEY_LBUTTON";
+			keyIsDown.unset(key);
+			keyWasReleased.set(key);
+			break;
+		case EMIE_MMOUSE_LEFT_UP:
+			key = "KEY_MBUTTON";
+			keyIsDown.unset(key);
+			keyWasReleased.set(key);
+			break;
+		case EMIE_RMOUSE_LEFT_UP:
+			key = "KEY_RBUTTON";
+			keyIsDown.unset(key);
+			keyWasReleased.set(key);
+			break;
+		case EMIE_MOUSE_WHEEL:
+			mouse_wheel += event.MouseInput.Wheel;
+			break;
+		default: break;
 		}
 	} else if (event.EventType == irr::EET_LOG_TEXT_EVENT) {
 		static const LogLevel irr_loglev_conv[] = {
@@ -188,38 +214,28 @@ s32 RandomInputHandler::Rand(s32 min, s32 max)
 	return (myrand() % (max - min + 1)) + min;
 }
 
+struct RandomInputHandlerSimData {
+	std::string key;
+	float counter;
+	int time_max;
+};
+
 void RandomInputHandler::step(float dtime)
 {
-	{
-		static float counter1 = 0;
-		counter1 -= dtime;
-		if (counter1 < 0.0) {
-			counter1 = 0.1 * Rand(1, 40);
-			keydown.toggle(getKeySetting("keymap_jump"));
-		}
-	}
-	{
-		static float counter1 = 0;
-		counter1 -= dtime;
-		if (counter1 < 0.0) {
-			counter1 = 0.1 * Rand(1, 40);
-			keydown.toggle(getKeySetting("keymap_special1"));
-		}
-	}
-	{
-		static float counter1 = 0;
-		counter1 -= dtime;
-		if (counter1 < 0.0) {
-			counter1 = 0.1 * Rand(1, 40);
-			keydown.toggle(getKeySetting("keymap_forward"));
-		}
-	}
-	{
-		static float counter1 = 0;
-		counter1 -= dtime;
-		if (counter1 < 0.0) {
-			counter1 = 0.1 * Rand(1, 40);
-			keydown.toggle(getKeySetting("keymap_left"));
+	static RandomInputHandlerSimData rnd_data[] = {
+		{ "keymap_jump", 0.0f, 40 },
+		{ "keymap_special1", 0.0f, 40 },
+		{ "keymap_forward", 0.0f, 40 },
+		{ "keymap_left", 0.0f, 40 },
+		{ "keymap_dig", 0.0f, 30 },
+		{ "keymap_place", 0.0f, 15 }
+	};
+
+	for (auto &i : rnd_data) {
+		i.counter -= dtime;
+		if (i.counter < 0.0) {
+			i.counter = 0.1 * Rand(1, i.time_max);
+			keydown.toggle(getKeySetting(i.key.c_str()));
 		}
 	}
 	{
@@ -228,30 +244,6 @@ void RandomInputHandler::step(float dtime)
 		if (counter1 < 0.0) {
 			counter1 = 0.1 * Rand(1, 20);
 			mousespeed = v2s32(Rand(-20, 20), Rand(-15, 20));
-		}
-	}
-	{
-		static float counter1 = 0;
-		counter1 -= dtime;
-		if (counter1 < 0.0) {
-			counter1 = 0.1 * Rand(1, 30);
-			leftdown = !leftdown;
-			if (leftdown)
-				leftclicked = true;
-			if (!leftdown)
-				leftreleased = true;
-		}
-	}
-	{
-		static float counter1 = 0;
-		counter1 -= dtime;
-		if (counter1 < 0.0) {
-			counter1 = 0.1 * Rand(1, 15);
-			rightdown = !rightdown;
-			if (rightdown)
-				rightclicked = true;
-			if (!rightdown)
-				rightreleased = true;
 		}
 	}
 	mousepos += mousespeed;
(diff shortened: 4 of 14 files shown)


ClobberXD:
Rebased and fixed merge conflicts. But I'm not really happy with the choice of words used for both LMB (`DIG`) and RMB (`PLACE`) - LMB and RMB is also used for punching and misc. interacting with world respectively. `USE` and `SECONDARY_USE` would suit them better, IMHO. Thoughts?

jastevenson303:
Use reminds me of "aux1" and secondary_use reminds me of on_secondary_use
node def value.  But if that doesn't matter, USE and SECONDARY_USE sound
good to me.  Could also go with "add" and "subtract"; but this could turn
into a lot of thought over a somewhat inconsequential detail.

ClobberXD:
> Use reminds me of "aux1" and secondary_use reminds me of on_secondary_use node def value. But if that doesn't matter, USE and SECONDARY_USE sound good to me. Could also go with "add" and "subtract"; but this could turn into a lot of thought over a somewhat inconsequential detail.

Yes, but all references to that key now use `aux1` only, and I assume it's safe to name LMB `USE`. I'm not as convinced about `SECONDARY_USE` for RMB, but it's still much better than `PLACE` which is misleading.

SmallJoker:
> @SmallJoker could you please elaborate on this? What exactly needs to be done to fix this compat breakage?

Change
https://github.com/minetest/minetest/blob/2d05e09e6d07e4eb3a86dd13f1750c02fde9c844/src/script/lua_api/l_object.cpp#L1313-L1320

to

```C++
	lua_pushboolean(L, control.dig);
	lua_setfield(L, -2, "dig");
	lua_pushboolean(L, control.place);
	lua_setfield(L, -2, "place");
	// Legacy fields to ensure mod compatibility
	lua_pushboolean(L, control.LMB);
	lua_setfield(L, -2, "LMB");
	lua_pushboolean(L, control.RMB);
	lua_setfield(L, -2, "RMB");
```
Please also adjust the docs accordingly.

jastevenson303:
fair points clobberxd.  then how about use_secondary versus secondary_use?

ClobberXD:
`secondary_use` sounds better than `use_secondary`

jastevenson303:
When right clicking air, it's called "activating."  So how about `activate` instead of `secondary_use`?

ClobberXD:
> When right clicking air, it's called "activating." So how about `activate` instead of `secondary_use`?

Hmm... `Press the activate button to open/close this chest` - This doesn't sound as meaningful. The reason why I'm suggesting `secondary_use` is that it's much more generic and can apply to all the actions equally well. e.g. `Press the secondary use button to open the door.`

jastevenson303:
Yep, sounds good to me.  TBH I'm just excited and a little anxious so don't
mind me.

ClobberXD:
> Yep, sounds good to me. TBH I'm just excited and a little anxious so don't mind me.

I can understand. :)

I just wish I can understand why there're include errors popping up in a couple of files.

lhofhansl:
I'm very much in favor of the idea (haven't had a chance to look at the changes, though)

ClobberXD:
Squashed the commits for ease of manageability.

src/nodedef.cpp
```diff
@@ -366,7 +366,7 @@ void ContentFeatures::reset()
 	climbable = false;
 	buildable_to = false;
 	floodable = false;
-	rightclickable = true;
+	can_place_onto = true;
```
ClobberXD: This is misleading - `can_place_onto` is *not* the same as `rightclickable`. Please advise as to what this should be changed to.

ClobberXD:
Rebased. This PR should be ready for testing.

Wuzzy2:
Hmmm, I am not happy with the terms `USE` and `SECONDARY_USE` tbh.

Because leftclick's main (!) use is “dig/punch”, and rightclick's main use is “use/place”, where with “use” we mean “using the pointed node or thing”. At least this is defined that way in the Esc menu and README file. And it's mostly true. It's only that special items might subvert these definitions.

Calling the key that is primarily used for digging/punching “USE” is very confusing. And the name “SECONDARY_USE” does not make it any better.

ClobberXD:
> Hmmm, I am not happy with the terms USE and SECONDARY_USE tbh.
>
> Calling the key that is primarily used for digging/punching “USE” is very confusing. And the name “SECONDARY_USE” does not make it any better.

I agree that `USE` and `SECONDARY_USE` aren't the best names possible - while they're pretty generalised, they also do a bad job of actually describing what the buttons do.

> Because leftclick's main (!) use is “dig/punch”, and rightclick's main use is “use/place”, where with “use” we mean “using the pointed node or thing”. At least this is defined that way in the Esc menu and README file. And it's mostly true. It's only that special items might subvert these definitions.

So, do you suggest using "dig/punch" and "use/place" as the button names? They describe the function of these buttons quite well, but how would they be presented to the end-user the same way? I mean, isn't a single term better for this, rather than two terms with a `/` in the middle?

jastevenson303:
Tested.  Did not work.  I was able to jump with right mouse button, but couldn't place or dig regardless of where those two actions were bound.

jastevenson303:
Also can we please just settle on a name?  LOL

ClobberXD:
@jastevenson303 I haven't got around to testing this yet, so thanks for testing this. I'll look into why it doesn't work.

ClobberXD:
Can confirm - actions are bindable to mouse buttons and work properly, but dig and place don't work irrespective of their keybinds.

src/client/game.cpp
```diff
@@ -595,15 +594,16 @@ struct GameRunData {
 	u16 new_playeritem;
 	PointedThing pointed_old;
 	bool digging;
-	bool ldown_for_dig;
+	bool btn_down_for_dig;
 	bool dig_instantly;
 	bool digging_blocked;
-	bool left_punch;
+	bool punch;
+	bool update_wielded_item_trigger;
```
jastevenson303: It appears this line was not an addition in the original PR?  Just curious if it was taken out at some point.  Good luck, I hope you can get this to work again.  :)

ClobberXD: This line seems to have already existed in the code, before the original PR, but was removed in a later commit. This would've been added during rebase. I still kinda suck at rebasing in all but the simplest of situations :)

builtin/settingtypes.txt
```diff
@@ -110,9 +110,9 @@ doubletap_jump (Double tap jump for fly) bool false
 #    enabled.
 always_fly_fast (Always fly and fast) bool true
 
-#    The time in seconds it takes between repeated right clicks when holding the right
-#    mouse button.
-repeat_rightclick_time (Rightclick repetition interval) float 0.25
+#    The time in seconds it takes between repeated block placements when holding
+#    the place block button.
+repeat_place_time (Place repetition interval) float 0.25
```
ClobberXD: This would break existing clients in a non-violent way. i.e. `repeat_rightclick_interval` won't work anymore, but they could easily fix this by renaming it to `repeat_place_time`. Thoughts?

Wuzzy2:
> This would break existing clients in a non-violent way. i.e. repeat_rightclick_interval won't work anymore, but they could easily fix this by renaming it to repeat_place_time. Thoughts?

Does this work? E.g. has this been tested to actually *only* place things, and not trigger `on_rightclick` of pointed things? This is important.

src/client/game.cpp
```diff
@@ -3012,51 +3009,51 @@ void Game::processPlayerInteraction(f32 dtime, bool show_hud, bool show_debug)
...
 
-	input->resetLeftClicked();
-	input->resetRightClicked();
+	input->joystick.clearWasKeyDown(KeyType::MOUSE_L);
+	input->joystick.clearWasKeyDown(KeyType::MOUSE_R);
 
-	input->resetLeftReleased();
-	input->resetRightReleased();
+	input->joystick.clearWasKeyReleased(KeyType::MOUSE_L);
+	input->joystick.clearWasKeyReleased(KeyType::MOUSE_R);
```
ClobberXD: Does `KeyType::MOUSE_L` and `KeyType::MOUSE_R` need to be changed to `KeyType::DIG` and `KeyType::PLACE`?

ClobberXD:
> Does this work? E.g. has this been tested to actually only place things, and not trigger on_rightclick of pointed things? This is important.

`repeat_rightclick_interval` just limits the number of right-clicks simulated when RMB is held down. This PR just changes the name of the setting, but the behaviour and function of this setting is the exact same. i.e. if the pointed node defines `on_rightclick`, that is triggered, and the wielded node isn't placed. e.g. toggling doors. This is how it currently works too. Maybe I misunderstood what you said?

Wuzzy2:
But isn't the new name extremely misleading then? Triggering `on_rightclick` != placing a node.

Wuzzy2:
Note: I agree the name `rightclick_interval` is misleading, too, if mouse buttons are bindable. Need to think of a better name.

src/nodedef.h
```diff
@@ -354,7 +354,7 @@ struct ContentFeatures
 	// Player can build on these
 	bool buildable_to;
 	// Player cannot build to these (placement prediction disabled)
-	bool rightclickable;
+	bool can_place_onto;
```
ClobberXD: Another instance of the ironical nodedef property :)

> ```cpp
> // Player cannot build to these (placement prediction disabled)
> bool can_place_onto;
> ```

Note that this change also breaks existing mods that define nodes with the `rightclickable` property, and it's probably best to revert all changes to this nodedef property. Thoughts?

ClobberXD:
@Wuzzy2 From your previous comment:

> ... leftclick's main (!) use is “dig/punch”, and rightclick's main use is “use/place”, where with “use” we mean “using the pointed node or thing”. At least this is defined that way in the Esc menu and README file. And it's mostly true. It's only that special items might subvert these definitions.

`repeat_use_place_repetition_interval` :D

On a more serious note, convenience starts to become an issue with long names like this.

src/nodedef.h
DS-Minetest: Nodes can't be defined with the `rightclickable` property. This bool just says whether an `on_rightclick` field is set:
https://github.com/minetest/minetest/blob/4d7f2967172c5a358811ac1f24115b281eff2431/src/script/common/c_content.cpp#L507-L508

DS-Minetest:
I suggest using the names `luse` and `ruse`. (The problem here is that "ruse" is an english word.)

src/nodedef.h
ClobberXD: Oh, that clears things up, thanks.

Then this might just be fixed by inverting the variable. i.e.

```cpp
lua_getfield(L, index, "on_rightclick"); 
f.can_place_onto = !lua_isfunction(L, -1); 
```

Because, if a node has an `on_rightclick` function, it is **not** placeable onto (this can be bypassed by using Shift+RMB, but that's a different story).

ClobberXD:
> I suggest using the names `luse` and `ruse`

Well, the whole point of this PR is to get rid of the association between dig and place actions, and `L` and `R` respectively. ;)

src/nodedef.h
DS-Minetest: There might be other factors which disallow placement without sneaking, eg. formspec meta.
Furthermore, this field has to be sent to the client, too. Because of backwards compatibility you could not send the new bool to the client without inverting it. This will cause the client code to be different from the server code (at least the names).
All in all, it's much simpler to just use keep this as is, a bool field that tells the client whether there's an `on_rightclick` callback defined.

ClobberXD: Hm, how about changing the name to `is_not_placeable` to not require inversion?

DS-Minetest:
> > I suggest using the names `luse` and `ruse`
> 
> Well, the whole point of this PR is to get rid of the association between dig and place actions, and `L` and `R` respectively. ;)

Who says that the L and R have to mean left and right? ((Maybe luse is the labor key and ruse is the relevant key.))
Just give the keys very abstract names. But don't define an order, eg. not use1 and use2. Some examples: luse and ruse, iuse and ause, ause and zuse
`luse` and `ruse` would be good because of historical reasons.

ClobberXD:
How about `dig_punch` and `place_use`?

DS-Minetest:
> How about `dig_punch` and `place_use`?

Then you would have to document eg. that `on_use` is called when the `dig_punch` button is used and not when the `place_use` button is used (which has a "use" in it).
You can't give descriptive names to the buttons, they just do too many different things. (Well you could use endlessly long names like `place_and_secondary_use_and_open_meta_formspec_but_if_sneak_is_pressed_always_place`.)

Wuzzy2:
> (Well you could use endlessly long names like place_and_secondary_use_and_open_meta_formspec_but_if_sneak_is_pressed_always_place.)

:D

It's funny because it's true. Minetest's control scheme is, in fact, a little insane. A simple right-click has just so many meanings.

I still think `place_use` perfectly summarizes it.

> Then you would have to document eg. that on_use is called when the dig_punch button is used and not when the place_use button is used (which has a "use" in it).

That's true, but it just exposes just another insanity of Minetest IMO. It should have never been called `on_use` in the first place since it's clearly for the punch key (LMB). The word “use” in Minetest is used inconsistenly.

Note that the Lua API is problematic in other areas as well. There's a callback literally called `on_rightclick`. This naming will now bite us in the ass as well …

ClobberXD:
I agree with Wuzzy, `place_use` covers *most* of the use-cases of that button.

I'm afraid we cant touch the names of the existing stuff that interface with mods (aka the API :P), as that would break backwards compat. I doubt this would even happen for 6.0.0, as a breaking release needn't break all the existing mods unless \*really\* required.

****

I've finally fixed this PR, and I've tested that both

- Binding actions to mouse buttons
- Re-binding dig and place actions to other keys

work properly. Video coming up soon, once I fix my screen recorder.

jastevenson303:
Tested and works as expected.

ClobberXD:
Video demonstration: https://youtu.be/0Ca7nEMiq_U

src/client/game.cpp
```diff
@@ -3012,51 +3008,50 @@ void Game::processPlayerInteraction(f32 dtime, bool show_hud, bool show_debug)
 			client->setCrack(-1, v3s16(0, 0, 0));
 			runData.dig_time = 0.0;
 		}
-	} else if (runData.dig_instantly && input->getLeftReleased()) {
-		// Remove e.g. torches faster when clicking instead of holding LMB
+	} else if (runData.dig_instantly && !isKeyDown(KeyType::DIG)) {
```
DS-Minetest: `!isKeyDown` is perhaps not the same as `getLeftReleased`. The former just tells the state while the latter maybe returns whether the key was released once between the steps.
Or maybe they are the same, I don't know.

src/client/game.cpp
```diff
@@ -3012,51 +3008,50 @@ void Game::processPlayerInteraction(f32 dtime, bool show_hud, bool show_debug)
...
+		runData.punch = true;
+	} else if (wasKeyDown(KeyType::PLACE)) {
 		handlePointingAtNothing(selected_item);
 	}
 
 	runData.pointed_old = pointed;
 
-	if (runData.left_punch || input->getLeftClicked())
-		camera->setDigging(0); // left click animation
+	if (runData.punch || isKeyDown(KeyType::DIG))
```
DS-Minetest: Should `wasKeyDown` be used here?

src/client/game.cpp
```diff
@@ -3153,7 +3148,7 @@ PointedThing Game::updatePointedThing(
 
 void Game::handlePointingAtNothing(const ItemStack &playerItem)
 {
-	infostream << "Right Clicked in Air" << std::endl;
+	infostream << "Attempted to place block while pointing at nothing" << std::endl;
```
DS-Minetest:
```suggestion
	infostream << "Attempted to place node while pointing at nothing" << std::endl;
```
or even
```suggestion
	infostream << "Attempted to place item while pointing at nothing" << std::endl;
```

src/client/game.cpp
```diff
@@ -3193,18 +3188,17 @@ void Game::handlePointingAtNode(const PointedThing &pointed,
...
+		runData.repeat_place_timer = 0;
+		infostream << "Place button pressed while looking at ground" << std::endl;
 
 		if (meta && !meta->getString("formspec").empty() && !random_input
 				&& !isKeyDown(KeyType::SNEAK)) {
-			// Report right click to server
-			if (nodedef_manager->get(map.getNode(nodepos)).rightclickable) {
+			// Report place action to server
+			if (nodedef_manager->get(map.getNode(nodepos)).can_place_onto)
 				client->interact(INTERACT_PLACE, pointed);
```
DS-Minetest: The interact has to be sent to trigger the `on_rightclick` callback. This is not placement. (Well, the action is indeed called "place".)
Anyway, the `can_place_onto` is wrong. It has to be `rightclickable`, as in "has an `on_rightclick` callback".

src/client/game.cpp
```diff
@@ -3245,12 +3240,10 @@ void Game::handlePointingAtNode(const PointedThing &pointed,
 						SimpleSoundSpec();
 
 				if (def.node_placement_prediction.empty() ||
-						nodedef_manager->get(map.getNode(nodepos)).rightclickable) {
+						nodedef_manager->get(map.getNode(nodepos)).can_place_onto)
```
DS-Minetest: `rightclickable` is not the same as `can_place_onto`.

src/client/game.cpp
```diff
@@ -3269,8 +3262,8 @@ bool Game::nodePlacementPrediction(const ItemDefinition &selected_def,
 	if (!is_valid_position)
 		return false;
 
-	if (!prediction.empty() && !(nodedef->get(node).rightclickable &&
-			!isKeyDown(KeyType::SNEAK))) {
+	if (!(nodedef->get(node).can_place_onto && !isKeyDown(KeyType::SNEAK)) &&
```
DS-Minetest: Same as above.

src/client/game.cpp
```diff
@@ -3431,12 +3424,12 @@ void Game::handlePointingAtObject(const PointedThing &pointed,
 			runData.object_hit_delay_timer = object_hit_delay;
 		}
 
-		if (input->getLeftClicked())
+		if (isKeyDown(KeyType::DIG))
```
DS-Minetest: `wasKeyDown`?

src/client/game.cpp
```diff
@@ -3592,7 +3585,7 @@ void Game::handleDigging(const PointedThing &pointed, const v3s16 &nodepos,
 		client->setCrack(-1, nodepos);
 	}
 
-	camera->setDigging(0);  // left click animation
+	camera->setDigging(0);  // Dig animation
```
DS-Minetest: Another problem with the naming. Digging is also the name of the player model's hand movement.

src/client/inputhandler.cpp
```diff
@@ -112,56 +114,78 @@ bool MyEventReceiver::OnEvent(const SEvent &event)
...
+				// send a continuous stream of keydown events.
+				// In this case, we don't want to let this
+				// stream reach the application as it will cause
+				// certain actions to repeat constantly.
+				if (!IsKeyDown(keyCode))
+					keyWasDown.set(keyCode);
 				keyIsDown.set(keyCode);
-				keyWasDown.set(keyCode);
 			} else {
 				keyIsDown.unset(keyCode);
```
DS-Minetest: Maybe you need a new `keyWasUp`.
```cpp
if (IsKeyDown(keyCode))
	keyWasUp.set(keyCode);
keyIsDown.unset(keyCode);
```

src/client/inputhandler.cpp
```diff
@@ -112,56 +114,78 @@ bool MyEventReceiver::OnEvent(const SEvent &event)
...
-				mouse_wheel += event.MouseInput.Wheel;
-			}
+	} else if (event.EventType == irr::EET_MOUSE_INPUT_EVENT) {
+		// Handle mouse events
+		KeyPress key;
+		switch (event.MouseInput.Event) {
+		case EMIE_LMOUSE_PRESSED_DOWN:
+			key = "KEY_LBUTTON";
+			keyIsDown.set(key);
+			keyWasDown.set(key);
```
DS-Minetest: Not like above?
```cpp
if (!IsKeyDown(key))
	keyWasDown.set(key);
keyIsDown.set(key);
```

src/client/inputhandler.cpp
```diff
@@ -112,56 +114,78 @@ bool MyEventReceiver::OnEvent(const SEvent &event)
...
+
+		// Add cases for all other enum values to silence compiler warning
+		case EMIE_MOUSE_MOVED:
+		case EMIE_LMOUSE_DOUBLE_CLICK:
+		case EMIE_RMOUSE_DOUBLE_CLICK:
+		case EMIE_MMOUSE_DOUBLE_CLICK:
+		case EMIE_LMOUSE_TRIPLE_CLICK:
+		case EMIE_RMOUSE_TRIPLE_CLICK:
+		case EMIE_MMOUSE_TRIPLE_CLICK:
+		case EMIE_COUNT:
```
DS-Minetest: Would a `default:` not suffice?

src/client/inputhandler.cpp
```diff
@@ -188,38 +212,28 @@ s32 RandomInputHandler::Rand(s32 min, s32 max)
...
-		if (counter1 < 0.0) {
-			counter1 = 0.1 * Rand(1, 40);
-			keydown.toggle(getKeySetting("keymap_left"));
+	static RandomInputHandlerSimData rnd_data[] = {
+		{ "keymap_jump", 0.0f, 40 },
+		{ "keymap_special1", 0.0f, 40 },
+		{ "keymap_forward", 0.0f, 40 },
+		{ "keymap_left", 0.0f, 40 },
+		{ "keymap_dig", 0.0f, 30 },
+		{ "keymap_place", 0.0f, 15 }
```
DS-Minetest:
```suggestion
		{"keymap_place", 0.0f, 15}
```
https://www.kernel.org/doc/html/latest/process/coding-style.html#spaces
>Do not add spaces around (inside) parenthesized expressions.

src/gui/guiKeyChangeMenu.cpp
```diff
@@ -290,65 +291,119 @@ bool GUIKeyChangeMenu::resetMenu()
...
-			core::rect < s32 > rect(0, 0, 600, 40);
-			rect += v2s32(0, 0) + v2s32(25, 30);
-			const wchar_t *text = wgettext("Key already in use");
-			this->key_used_text = Environment->addStaticText(text,
-					rect, false, true, this, -1);
-			delete[] text;
-			//infostream << "Key already in use" << std::endl;
-		}
+	bool shift_went_down = false;
+	if(!shift_down &&
```
DS-Minetest:
```suggestion
	if (!shift_down &&
```

src/gui/guiKeyChangeMenu.cpp
```diff
@@ -290,65 +291,119 @@ bool GUIKeyChangeMenu::resetMenu()
...
-		{
-			key_setting *k = NULL;
-			for (key_setting *ks : key_settings) {
-				if (ks->id == activeKey) {
-					k = ks;
-					break;
-				}
+	// Remove Key already in use message
+	if(this->key_used_text)
+	{
```
DS-Minetest: `if (...) {`

src/gui/guiKeyChangeMenu.cpp
```diff
@@ -290,65 +291,119 @@ bool GUIKeyChangeMenu::resetMenu()
...
+		this->key_used_text->remove();
+		this->key_used_text = NULL;
+	}
+	// Display Key already in use message
+	if (std::find(this->key_used.begin(), this->key_used.end(), kp) != this->key_used.end())
+	{
```
DS-Minetest: Brace.

src/gui/guiKeyChangeMenu.cpp
```diff
@@ -290,65 +291,119 @@ bool GUIKeyChangeMenu::resetMenu()
...
-			this->key_used.push_back(kp);
+		this->key_used.push_back(kp);
 
-			// Allow characters made with shift
-			if(shift_went_down){
-				shift_down = true;
-				return false;
-			}
+		// Allow characters made with shift
+		if(shift_went_down){
```
DS-Minetest: Spaces.

src/gui/guiKeyChangeMenu.cpp
```diff
@@ -290,65 +291,119 @@ bool GUIKeyChangeMenu::resetMenu()
...
+		}
+
+		activeKey = -1;
+		return true;
+	}
+}
 
-			activeKey = -1;
-			return true;
+bool GUIKeyChangeMenu::OnEvent(const SEvent& event)
```
DS-Minetest:
```suggestion
bool GUIKeyChangeMenu::OnEvent(const SEvent &event)
```

src/gui/guiKeyChangeMenu.cpp
```diff
@@ -290,65 +291,119 @@ bool GUIKeyChangeMenu::resetMenu()
...
+{
+	// If an action is waiting for a keybind ("press key") and if that
+	// action's button is left-clicked then set the binding to LMB. This
+	// check has to be performed because a left-click on a GUI element is
+	// considered a GUI event and not a generic mouse input event like for
+	// MMB and RMB.
+	if ((activeKey >= 0) && (event.EventType == EET_GUI_EVENT) &&
+	    (event.GUIEvent.EventType == gui::EGET_BUTTON_CLICKED)) {
+
+		switch (event.GUIEvent.Caller->getID()) {
```
DS-Minetest:
```cpp
	if ((activeKey >= 0) && (event.EventType == EET_GUI_EVENT) &&
			(event.GUIEvent.EventType == gui::EGET_BUTTON_CLICKED)) {
		switch (event.GUIEvent.Caller->getID()) {
```

src/gui/guiKeyChangeMenu.cpp
```diff
@@ -290,65 +291,119 @@ bool GUIKeyChangeMenu::resetMenu()
...
+			case GUI_ID_BACK_BUTTON:
+				break;
```
DS-Minetest:
```cpp
	if ((activeKey >= 0) && (event.EventType == EET_GUI_EVENT) &&
			(event.GUIEvent.EventType == gui::EGET_BUTTON_CLICKED)) {
		switch (event.GUIEvent.Caller->getID()) {
		case GUI_ID_BACK_BUTTON:
			break;
...
```

src/nodedef.h
DS-Minetest: (The node itself is placeable, so not `is_not_placeable` but `is_not_placeable_to`.)
The name is a bit confusing as you _can_ still place to it. You just have to hold sneak.

src/script/common/c_content.cpp
```diff
@@ -505,7 +505,7 @@ ContentFeatures read_content_features(lua_State *L, int index)
 	lua_pop(L, 1);
 
 	lua_getfield(L, index, "on_rightclick");
-	f.rightclickable = lua_isfunction(L, -1);
+	f.can_place_onto = lua_isfunction(L, -1);
```
DS-Minetest: You can place onto the node even if it has no `on_rightclick`. :wink:

src/script/common/c_content.cpp
```diff
@@ -877,8 +877,8 @@ void push_content_features(lua_State *L, const ContentFeatures &c)
 	lua_setfield(L, -2, "climbable");
 	lua_pushboolean(L, c.buildable_to);
 	lua_setfield(L, -2, "buildable_to");
-	lua_pushboolean(L, c.rightclickable);
-	lua_setfield(L, -2, "rightclickable");
+	lua_pushboolean(L, c.can_place_onto);
+	lua_setfield(L, -2, "can_place_onto");
```
DS-Minetest: If you change this, you have to document it in `client_lua_api.txt`.

src/script/lua_api/l_object.cpp
```diff
@@ -1418,9 +1418,14 @@ int ObjectRef::l_get_player_control(lua_State *L)
 	lua_setfield(L, -2, "aux1");
 	lua_pushboolean(L, control.sneak);
 	lua_setfield(L, -2, "sneak");
-	lua_pushboolean(L, control.LMB);
+	lua_pushboolean(L, control.dig);
+	lua_setfield(L, -2, "dig");
+	lua_pushboolean(L, control.place);
+	lua_setfield(L, -2, "place");
+	// Legacy fields to ensure mod compatibility
```
DS-Minetest: Missing documentation.

src/client/inputhandler.cpp
ClobberXD:
```
(a + b * c / d)  // <--- This is a parenthesized expression

( )  // <--- This is a pair of parenthesis

{  }  // <--- This is a pair of braces

{a, b, c, d}  // <--- This is a... list-like thingy, as far as languages like C++ are concerned

{ a, b, c, d }  // <--- Adding spaces within braces is more readable (esp. with
                //      elements more complex than one-letter identifiers), IMO
```

src/client/game.cpp
ClobberXD: Agreed.

\*sigh\*, the naming issues are likely going to stall this PR big-time...

src/client/game.cpp
ClobberXD: Not really, because (hopefully I'm not getting this wrong):

```pseudocode
if (player is punching and they are currently holding down the DIG button)
    set appropriate digging animation
```

src/client/game.cpp
ClobberXD: I'll just rename it back to `on_rightclick`. I don't want to stall this PR with the ongoing naming contest.

src/client/game.cpp
ClobberXD: Okay...

src/client/game.cpp
ClobberXD: _Okay..._

src/client/game.cpp
DS-Minetest: But
https://github.com/minetest/minetest/blob/5f1cd555cd9d1c64426e173b30b5b792d211c835/src/client/camera.h#L123-L125
says that it starts the animation.
(Also there the buttons are still called left and right.)

src/gui/guiKeyChangeMenu.cpp
ClobberXD: ... for impact

src/client/inputhandler.cpp
ClobberXD: It should, I just didn't recompile after adding a `default`, which lead me to think that it didn't. Will fix it.

src/client/game.cpp
ClobberXD: Not sure if that's the way it's supposed to be, but I'll retest after changing this.

src/gui/guiKeyChangeMenu.cpp
ClobberXD:
```suggestion
		if (shift_went_down) {
```

src/script/lua_api/l_object.cpp
ClobberXD: Yes, I'm yet to add documentation.

jastevenson303:
Thank you @SmallJoker.  I am curious, how much support is there?  I understand changing a core fundamental such as this can be uncomfortable.

jastevenson303:
@SmallJoker what's the point of rebasing if NO CORE DEV supports this?
http://irc.minetest.net/minetest-hub/2019-09-22#i_5595007

ClobberXD:
@rubenwardy supports this, AFAIK

ClobberXD:
@rubenwardy _might_ be interested in this PR, as he self-requested a review in #6537 before it was closed.

jastevenson303:
There's a merge conflict now, and I'm not able to resolve it.

This happened before where I used the PR on my own, and I had to stop because I couldn't figure out how to resolve it.  @ClobberXD thank you so much for your contribution, it's really appreciated from here.

It's not really fair to ask so much of contributors.

jastevenson303:
@rubenwardy https://forum.minetest.net/viewtopic.php?p=315289#p315289 this could have gotten into 5.1.  It still could.

Regardless, can you help in supporting this?  Just imagine when it drops, and the on the "New Features" list it shows: * Can now bind mouse buttons.  I think there will be a fair amount of slow clapping.

Imagine all the people who tried Minetest, tried to bind their mouse buttons, and were surprised by what is rather unusual behavior.  Consider DOS games in the early 1990's that allowed binding mouse buttons.  Consider how odd this is, for people who might have a preference.

@sofar @calinou Can I ask for your support?  @ClobberXD has worked to resuscitate a neglected PR, and it stands a chance of happening again.  Minetest has a LOT of features, and a LOT of them are unused, undiscovered, or unimportant.

This is not one of those features.

Calinou:
I'm not a core developer, so my approval wouldn't be worth anything really :slightly_frowning_face:

rubenwardy:
I would like to see better input support, but I'm feeling too burnt out to review this. Even so, I don't think this should be in 5.1.0 as the deadline is too close now.

I don't see it as an essential feature like you claim. There are tons more features that are way more overdue, such as mobs and a working pathfinder, a main menu that doesn't stink, and particles which actually perform decently

ClobberXD:
@jastevenson303: Thanks for the loads of support you've shown towards this PR. I haven't seen the conflict yet, but I'm sure it's easily fixable. Also, I don't have the time to get this merged before 5.1.0, because remember, the PR itself works fine, but the naming problems is the elephant in the room, so to say. While I'd have loved to see this in 5.1.0, I agree with rubenwardy that this isn't *essential* to the release.

jastevenson303:
I was simply stating a possibility.  I would just like to see at least one
core dev support this PR is all.  That way there's at least hope it'll get
merged in 5.2-dev.

@clobberxd it didn't look simple to me!  haha :)

paramat:
I'm sure many core devs 'support' the concept of this PR, i do too, but 'support' is different to 'approval'.

jastevenson303:
Thank you @paramat.  Do you support mouse binds?  When this PR is rebased
and all issues are addressed, will you approve it?

ClobberXD:
Rebased. Needs testing to ensure that everything still works. I'm yet to find time to address the line comments.

I've also added another commit that fixes *most* of the code-style issues in `src/gui/guiKeyChangeMenu.cpp`

jastevenson303:
Tested and works.

Ferk:
About the naming of the actions, to me the right name for the right-button would be "interact".

When holding a The main difference between the "`on_use`" and the (ill-named) "`on_rightclick`" is the fact that `on_right_click` requires a pointed thing. At least the last time I checked (which was quite a while ago).

 - checks lua_api - 

Uhm.. sorry, I think I'm too late. The name was actually changed from `on_rightclick` to `on_secondary_use` already.
Was a change also made to have the action be triggered even when you are pointing at the sky or the nothing?

ClobberXD:
> About the naming of the actions, to me the right name for the right-button would be "interact".

*All* actions done by the player falls under "interact", incl. both LMB and RMB actions :)

And yes, `on_secondary_use` is triggered when a player right-clicks, while pointing at nothing. `on_place` is triggered when the player points at an object or a node.

Ferk:
> All actions done by the player falls under "interact", incl. both LMB and RMB actions :)

Not necessarily.
You can have "on_use" trigger a change in the item itself without interacting with the environment.
Like for example, when you eat an apple, you are not interacting with anything external.

> And yes, on_secondary_use is triggered when a player right-clicks, while pointing at nothing.

Ok, I'm glad that was changed then. In that case `on_rightclick` is no longer meant to be used for interacting with the world.

ClobberXD:
> You can have "on_use" trigger a change in the item itself without interacting with the environment.

Both LMB and RMB actions require the `interact` priv, and are hence considered interaction, at least by Minetest. That's what I meant.

paramat:
> When this PR is rebased and all issues are addressed, will you approve it?

Although i support the concept, approving depends on me reviewing and testing and the PR being fit for merge, so i cannot say in advance. I might review but it is not high priority for me.

src/client/game.cpp
paramat: ^ Needs consideration.

jastevenson303:
The word is 'activate.'

ClobberXD:
Rebased

ClobberXD:
The name issue is dragging this PR into oblivion, lol. I'd like to settle on the names "Dig" for LMB and "Place" for RMB, as those are the **primary** functions of the buttons. We'll have to get used to "Use the dig button to attack the monster" and "Use the place button to open/close the door"; we probably would... :P

_Inputs, suggestions, alternatives appreciated!_

Wuzzy2:
I suggest “Punch” for LMB as it's more general-purpose. You can punch nodes, objects and the air. But you can only dig nodes. Therefore, use “Punch”.

I'm fine with “Place” for RMB.

paramat:
> "Dig" for LMB and "Place" for RMB, as those are the primary functions of the buttons.

I agree with this, they form a nice pair of opposite actions, and digging and placing are on average the most used functions in MT.

Wuzzy2:
I have no strong feelings towards either alternative.

I prefer “Punch” for LMB and “Place” for RMB.
But I am not opposed to “Dig” for LMB and “Place” for RMB.

ClobberXD:
> I have no strong feelings towards either alternative.
> 
> I prefer “Punch” for LMB and “Place” for RMB.
> But I am not opposed to “Dig” for LMB and “Place” for RMB.

While Punch is a great option too, digging is the primary function of the LMB when it comes to Minetest (or so I'd like to think), which is the reason behind choosing that as the name of the button. That, and the fact that the original author had started using that name in a couple of places.

LoneWolfHT:
I don't care if it only describes a certain feature as long as I can tell what it does

src/client/game.cpp
ClobberXD: Good catch, thanks! It's indeed `wasKeyDown` that should be used here. `isKeyDown` corresponds to `getLeftState`, and `!isKeyDown` isn't the same as `wasKeyDown`.

ClobberXD:
I skimmed through the current code to figure out mouse button detection and fix a couple of issues with the PR. Unlike normal keys which have only `isKeyDown` and `wasKeyDown`, mouse buttons have `get(Left|Right)State` (`true` if pressed), `get(Left|Right)Clicked`, and `get(Left|Right)Released`. I'll note down a couple of observations regarding these, in case anyone would like to know.

- `get(Left|Right)State` is directly equivalent to `isKeyDown` for normal keys.
- `get(Left|Right)Clicked` is very similar to `get(Left|Right)State`, but its value is set to false one tick later, at the end of `Game::processPlayerInteraction`. This function is essentially used to check if the button was pressed in the current tick. Used for specifically detecting **clicks**, e.g. left-clicking on a torch to insta-dig it, which is a little faster than when holding down LMB.
- `get(Left|Right)Released` returns true when a mouse button has been released, but its value needs to be manually set to false at the end of `Game::processPlayerInteraction`, just like `get(Left|Right)Clicked`.
- So for most intents and purposes, `wasKeyDown` can be considered as the equivalent to `get(Left|Right)Clicked`.

****

Once I've fixed the existing issues with this PR, I plan on adding support for detecting MMB interactions.

src/client/game.cpp
ClobberXD: This should indeed be `wasKeyDown`. Changing, thanks for the heads-up.

src/client/game.cpp
ClobberXD: I understand that `!isKeyDown` isn't the same as `get(Left|Right)Released`. I'll implement an `isKeyReleased` instead.

src/client/game.cpp
ClobberXD: It is indeed `wasKeyDown`. Fixed.

src/client/game.cpp
ClobberXD: Ah, are you referring to the function name?

src/nodedef.h
ClobberXD: I'll revert back to `rightclickable` for now.

src/nodedef.cpp
ClobberXD: Never mind, reverted to `rightclickable` for now.

src/client/game.cpp
ClobberXD: Thoughts on renaming this to `Camera::setAnim`? Maybe we can also pass an Enum as the parameter, instead of a plain number.

src/client/game.cpp
ClobberXD: Actually, I'll replace this with the new `wasKeyPressed` method...

ClobberXD:
The merge conflict is scary... *shudders*

ClobberXD:
### Update

I'm yet to resolve the merge conflict. But I've added a new function `wasKeyPressed` (short for `wasKeyJustPressed`) which checks whether a key was just pressed. This will be set alongside `isKeyDown`, but will be unset right away in the subsequent iteration of `Game::processPlayerInteraction`. The idea, similar to how `get(Left|Right)Clicked` used to work, is to allow code to detect when the user has just started pressing a key, which is very different from detecting whether a key is down or not. e.g. to begin the wieldhand animation when holding down the dig button - only the initial press needs to be detected to trigger the animation. `isKeyDown` might help detect when a key is pressed, but would continue return true unless and until the key is released, which would result in a bunch of code unnecessarily running over and over again, when they'd have to run only once.

Apart from the merge conflict (I'll try to resolve this myself, but any help would be appreciated), I'd declare the PR ready for testing. Testing instructions in the first post.

LoneWolfHT:
If you fix the conflicts I think you'll find people who are willing to test this

ClobberXD:
@LoneWolfHT Thanks for the interest. It's the fixing-the-conflicts part that bothers me. There's a lot of conflicts in `guiKeyChangeMenu.cpp`, and unfortunately, I also don't understand much of the code involved. I've been putting this off as it'll take me some time to understand the code and what the original author sought to change here, but I suppose this needs to be done. I'll dig into this today (or within the next couple of days), if possible.

sfan5:
If the key change menu is an issue perhaps make a PR for the functionality first and another one for exposing it in the key change menu?

ClobberXD:
That's a very good idea, thanks. Also, this PR won't be incomplete as the keys could be set by editing minetest.conf anyway.

ClobberXD:
Removed all changes in `guiKeyChangeMenu.*` and rebased.

- Binding actions to mouse buttons works. :heavy_check_mark: 
- Re-binding dig and place actions to other keys works. :heavy_check_mark: 

I'm very surprised - *nothing* is broken after the rebase. :D

jastevenson303:
Works for me.  I imagine minetest.conf.example will be updated automatically?  Took me a minute to figure out the settings, but it was KEY_RBUTTON and KEY_SPACE.  Thank you!!

src/client/game.cpp
ClobberXD: I've added a `wasKeyPressed` which is equivalent to `get(Left|Right)Clicked`. This method is very similar to `wasKeyDown` but unlike the latter, it's state isn't reset immediately after calling but only at the end of `Game::processPlayerInteraction` - this means that this function will return the same value for all `wasKeyPressed` calls in `Game::processPlayerInteraction` in a given iteration.

(Thanks for reminding me to comment on this, @sfan5)

ClobberXD:
Documentation has been added. CSM method `player:get_control` now returns `dig` and `place` fields instead of `LMB` and `RMB`.

src/script/lua_api/l_localplayer.cpp
```diff
@@ -231,8 +231,8 @@ int LuaLocalPlayer::l_get_control(lua_State *L)
 	set("aux1", c.aux1);
 	set("sneak", c.sneak);
 	set("zoom", c.zoom);
-	set("LMB", c.LMB);
-	set("RMB", c.RMB);
+	set("dig", c.dig);
+	set("place", c.place);
```
ClobberXD, line 235: `LMB` and `RMB` are removed from the CSM method entirely. Backwards compat isn't a concern, as this method hasn't been in any release yet.

jastevenson303:
No sorry I was mistaken; I'm not able to eat apples (I did not change LMB/punch bind).

src/client/game.cpp
rubenwardy: joysticks largely don't work in Minetest :D

I suspect it may need to be DIG and PLACE though

src/client/game.cpp
sfan5: so is `!isKeyDown(KeyType::DIG)` the correct equivalent of `getLeftReleased()` then?

jastevenson303:
...

ClobberXD:
@jastevenson303 Try `make clean` and rebuild - worked for me.

ClobberXD:
@rubenwardy Thanks for testing!

****

I've re-added a "co-authored-by" attribution to the original author as they had done some much work on this before I adopted it. Please make sure the entire commit message makes it into `master` when merging.

src/client/game.cpp
ClobberXD: I expected that answer, but I don't want to screw it up even further. :)

Anyway, I'll see what I can do.

builtin/settingtypes.txt
ClobberXD: This needs attending to as well. I doubt this is a popular setting, but I'm pretty sure a good bunch of users have tweaked this to their requirements. IMO, the new name actually better describes the purpose of the setting, even if `place` is hardcoded to RMB.

SmallJoker: Backwards compatibility would be nice to have.
1) Read old setting value (if it exists)
2) Set new value
3) Remove old setting line

doc/lua_api.txt
```diff
@@ -6073,14 +6073,37 @@ object you are working with still exists.
     * Only affects formspecs shown after this is called.
 * `get_formspec_prepend(formspec)`: returns a formspec string.
 * `get_player_control()`: returns table with player pressed keys
-    * The table consists of fields with boolean value representing the pressed
-      keys, the fields are jump, right, left, LMB, RMB, sneak, aux1, down, up.
-    * example: `{jump=false, right=true, left=false, LMB=false, RMB=false,
-      sneak=true, aux1=false, down=false, up=false}`
+    * The table consists of fields with boolean values representing the pressed
+      keys. Example:
```
SmallJoker, line 6077: Is this example required? Just listing the table keys like before should be enough, isn't it?

src/client/game.cpp
```diff
@@ -623,15 +622,15 @@ struct GameRunData {
 	u16 new_playeritem;
 	PointedThing pointed_old;
 	bool digging;
-	bool ldown_for_dig;
+	bool btn_down_for_dig;
 	bool dig_instantly;
 	bool digging_blocked;
-	bool left_punch;
+	bool punch;
```
SmallJoker, line 628: `punching` would be consistent with `digging` - its counterpart.

doc/lua_api.txt
ClobberXD: Personally, I think listing the keys would suffice and the example is required at all. Shall I remove it?

jastevenson303:
Unusable until eating and using is fixed.

src/client/game.cpp
ClobberXD: `getLeftReleased` returns `true` only for one run of `Game::processPlayerInteraction` and is reset (https://github.com/minetest/minetest/pull/7924/files#diff-7f71a92de20504e8a5636b4317ea67b2L3169). So it's indeed different from `!isKeyDown()` which always evaluates to true if the key is up. *But*, the behaviour is the same in this case, as `dig_instantly` and `!isKeyDown` both need to be true for the inner code to run. Once the inner code runs, `dig_instantly` is set to false, and the code never runs again until the key is pressed again. Essentially, `dig_instantly` is already used to keep track of key release events, so `!down` is the same as `released` in this specific context.

src/client/game.cpp
ClobberXD: You're right, will do.

src/client/game.cpp
ClobberXD: I'll leave joystick unhardcoding to another PR as I'm completely unfamiliar with joystick code at the moment. Even if I try to work on it, this PR is going to be greatly delayed yet again.

src/client/game.cpp
ClobberXD: Done

doc/lua_api.txt

ClobberXD:
@jastevenson303 Thanks for reporting. I too notice some oddities incl. not being able to "use" items.

src/client/game.cpp
```diff
@@ -3119,55 +3119,56 @@ void Game::processPlayerInteraction(f32 dtime, bool show_hud, bool show_debug)
...
-		runData.repeat_rightclick_timer = 0;
+		runData.repeat_place_timer = 0;
 
-	if (selected_def.usable && input->getLeftState()) {
-		if (input->getLeftClicked() && (!client->modsLoaded()
-				|| !client->getScript()->on_item_use(selected_item, pointed)))
+	if (selected_def.usable && isKeyDown(KeyType::DIG)) {
+		if (wasKeyPressed(KeyType::DIG) && (!client->modsLoaded() ||
+				!client->getScript()->on_item_use(selected_item, pointed)))
 			client->interact(INTERACT_USE, pointed);
```
ClobberXD, line 3144: This part seems to be the issue with regards to items not being usable. `isKeyDown` returns true, but `wasKeyDown` seems to always return false, which means execution never reaches `client->interact`. `wasKeyPressed` checks if a key was pressed and released (sorta equivalent to a "tap", to make things clear). I think I might need to implement all four keypress states - up, down, pressed, released. :)

ClobberXD: Yeah, I just checked `getLeftClicked` again. It returns true only the instant LMB is pressed, and it's reset at the end of `processPlayerInteraction`. `wasKeyPressed` isn't supposed to replace both `get--Clicked` and `get--Released`...

src/client/game.cpp
ClobberXD: I've changed `wasKeyPressed` to evaluate to true only when the key was pressed down. So this is equivalent to `isKeyDown`, except that it's state would be reset after one step. I've now added a separate `wasKeyReleased` (which I should've done much earlier) which returns true for one step after a key has been released. I'm still in the process of testing this, but this should fix everything (TM).

src/client/game.cpp
```diff
@@ -3264,7 +3270,7 @@ PointedThing Game::updatePointedThing(
 
 void Game::handlePointingAtNothing(const ItemStack &playerItem)
 {
-	infostream << "Right Clicked in Air" << std::endl;
+	infostream << "Attempted to place item while pointing at nothing" << std::endl;
```
ClobberXD, line 3273: This is incorrect, this function handles both LMB and RMB while pointing at nodes.

src/client/game.cpp
ClobberXD: `wasKeyReleased` now exists. It's a drop-in replacement for `get(Left|Right)Released`.

ClobberXD:
Good news, I've fixed everything :tm:. Apart from items being un-use-able, there was also the issue of nodes being placed when the player was rightclicking air before pointing at a node (turned out to be a simple fix, though). Here's what I've tested:

- Digging - test by digging around.
- Item use (LMB) - test using apples or `testfood:*` items in devtest.
- Node placement - test by placing nodes that you dug.
- Item's secondary_use (RMB while pointing at air).
- Node inventory access (on_rightclick?) - test by opening chests in MTG (devtest's chest of everything isn't a node inv. AFAIK).

ClobberXD:
@jastevenson303 If you can, please test this PR again and let me know if there's anything else amiss.

jastevenson303:
LOL

testing

tested, FOR WHAT IT'S WORTH... so far so good, but I'll let you know if something odd happens, maybe.

jastevenson303:
Oh I SUPPOSE I should let you know about PLACE_TIME

hint: hold down place and walk around

ClobberXD:
@jastevenson303 I don't see any issues with `repeat_place_time` (if that's what you mean). The setting is obeyed correctly as well (I tested this by setting it to `0.1`). Would you be able to elaborate?

jastevenson303:
Nah its cool

src/defaultsettings.cpp
```diff
@@ -73,6 +73,8 @@ void set_default_settings(Settings *settings)
 	settings->setDefault("keymap_right", "KEY_KEY_D");
 	settings->setDefault("keymap_jump", "KEY_SPACE");
 	settings->setDefault("keymap_sneak", "KEY_LSHIFT");
+	settings->setDefault("keymap_dig", "KEY_LBUTTON");
+	settings->setDefault("keymap_place", "KEY_RBUTTON");
```
SmallJoker, line 77: Missing settingtypes.txt entries.

src/client/game.cpp
```diff
@@ -3119,55 +3123,57 @@ void Game::processPlayerInteraction(f32 dtime, bool show_hud, bool show_debug)
...
 
-	runData.left_punch = false;
+	runData.punching = false;
 
 	soundmaker->m_player_leftpunch_sound.name = "";
 
 	// Prepare for repeating, unless we're not supposed to
-	if (input->getRightState() && !g_settings->getBool("safe_dig_and_place"))
-		runData.repeat_rightclick_timer += dtime;
+	if (isKeyDown(KeyType::PLACE) && !g_settings->getBool("safe_dig_and_place"))
```
SmallJoker, line 3140: This does not seem to work properly. I've set the following in my configuration:
`keymap_place = KEY_KEY_H` (`H` is not in use anywhere else)

But the nodes are placed two times in a row. Always. No matter how short the key press is.

src/client/game.cpp
```diff
@@ -3119,55 +3123,57 @@ void Game::processPlayerInteraction(f32 dtime, bool show_hud, bool show_debug)
...
+		// Remove e.g. torches faster when clicking instead of holding dig button
 		runData.nodig_delay_timer = 0;
 		runData.dig_instantly = false;
 	}
 
-	if (!runData.digging && runData.ldown_for_dig && !input->getLeftState()) {
-		runData.ldown_for_dig = false;
-	}
+	if (!runData.digging && runData.btn_down_for_dig && !isKeyDown(KeyType::DIG))
+		runData.btn_down_for_dig = false;
```
SmallJoker, line 3133: Not working properly either. Released the dig key in between and the crack does not disappear and the progress is kept.

grafik

ClobberXD: Not sure if this is related, but I've come across this in master occasionally, especially when I'm trying to dig nodes right at the max range of the tool. I'll look into this, though. Thanks for reporting.

src/client/game.cpp
ClobberXD: Strange, I haven't come across this issue while testing...

src/defaultsettings.cpp
ClobberXD: Aye, will do. o7

src/client/game.cpp
ClobberXD: Hmm, this happens only when binding the place button to a key. Investigating...

src/client/game.cpp
ClobberXD: Good catch, again. This can only be reproduced when binding dig to a normal key. Investigating...

src/defaultsettings.cpp
ClobberXD: Done

src/client/inputhandler.cpp
```diff
@@ -112,56 +114,77 @@ bool MyEventReceiver::OnEvent(const SEvent &event)
 		const KeyPress &keyCode = event.KeyInput;
 		if (keysListenedFor[keyCode]) {
 			if (event.KeyInput.PressedDown) {
+				// If the key is being held down then the OS may
+				// send a continuous stream of keydown events.
+				// In this case, we don't want to let this
+				// stream reach the application as it will cause
+				// certain actions to repeat constantly.
+				if (!IsKeyDown(keyCode)) {
```
ClobberXD, line 122: I wonder if this check is redundant. Aren't events only sent if there are state changes? Pretty sure a key press event is sent only if the key was just pressed, and isn't continuously sent as long as the key is pressed.

ClobberXD: Apparently not - as long as the key is pressed, one event is sent, and then a stream of events are sent until the key is released; just as this helpful comment suggests. But... why? :thinking:

src/client/game.cpp
ClobberXD: Fixed.

src/client/game.cpp
ClobberXD: Fixed

ClobberXD:
@SmallJoker Turns out I hadn't even implemented key-release handlers for the normal keys. :D

Rebased. Fixed all the the things :tm:. PR is ready for review again.

jastevenson303:
Everything appears to be working correctly.  Thank you!

jastevenson303:
I had this really strange experience, as evidenced by the following video:
https://youtu.be/ihmASukVuK4

I have since been unable to reproduce it, and I'm not sure if it's related to setting a player's hand stack (https://github.com/jastevenson303/jastest/blob/master/mods/classes/init.lua#L106) or if it's even related to this PR.

I just wanted to show it just in case, and I'll update if I experience it again.

src/client/inputhandler.cpp
jastevenson303: Is this related to an operating system's key repeat settings?  Just curious.

LoneWolfHT:
Can also test the LMB code with this WE command:
```
//lua for i=0, 5, 1 do minetest.after(i, function() minetest.chat_send_all(dump(minetest.get_player_by_name("Lone_Wolf"):get_player_control().LMB))end) end
```

src/client/inputhandler.cpp
ClobberXD: It seems like it; *shouldn't* be, though...

ClobberXD:
Thanks for testing this PR, @LoneWolfHT. I'll take a look at the reported issue.

ClobberXD:
> When rightclicking with the hammer in my hand it instantly rightclicks the spawned entity instead of requiring a second click

Could you point me to the line of code in your mod that results in this behaviour? Please also provide a minimal reproducible example if possible - that'd be very helpful.

src/network/serverpackethandler.cpp
```diff
@@ -499,8 +499,7 @@ void Server::process_PlayerPos(RemotePlayer *player, PlayerSAO *playersao,
 	player->control.jump = (keyPressed & 16);
 	player->control.aux1 = (keyPressed & 32);
 	player->control.sneak = (keyPressed & 64);
-	player->control.LMB = (keyPressed & 128);
-	player->control.RMB = (keyPressed & 256);
+	player->control.place = (keyPressed & 256);
 	player->control.zoom = (keyPressed & 512);
```
ClobberXD, line 503: Aha, server isn't de-serializing the packet properly!

ClobberXD: Fixed

ClobberXD:
LMB detection has been fixed.

LoneWolfHT:
It's in an on_secondary_use https://github.com/LoneWolfHT/hammer_of_power/blob/master/init.lua#L258
Untested but this code should work. Just drop it in a `test` mod
```lua
minetest.register_node("test:testnode", {
	description = "testnode",
	tiles = {"air.png"},
	groups = {oddly_breakable_by_hand = 3, snappy = 1},
	on_secondary_use = function(_, user, _)
		local name = user:get_player_name()
		local pos = user:get_pos()
		pos.y = pos.y + 1.5

		local obj = minetest.add_entity(
			vector.add(pos, vector.multiply(user:get_look_dir(), 2.5)),
			"test:entity"
		)
	end,
})

minetest.register_entity("test:ent", {
	initial_properties = {
		physical = true,
		collide_with_objects = false,
		selectionbox = {-0.25, -0.25, -0.25, 0.25, 0.25, 0.25},
		pointable = true,
		visual = "cube",
		visual_size = {x = 0.5, y = 0.5, z = 0.5},
		textures = {},
		is_visible = true,
		automatic_rotate = 20,
		glow = 4,
		static_save = false,
	},
	on_punch = function(self, puncher, _, _, dir)
		self.object:remove()
	end,
	on_rightclick = function(self, clicker)
		self.object:remove()
	end,
})
```

doc/lua_api.txt
```diff
@@ -6105,15 +6105,23 @@ object you are working with still exists.
...
+    * 0 - up
+    * 1 - down
+    * 2 - left
+    * 3 - right
+    * 4 - jump
+    * 5 - aux1
+    * 6 - sneak
+    * 7 - dig
+    * 8 - place
+    * 9 - zoom
```
SmallJoker, line 6124: `(zoom available since 5.3)`  rebase error.

ClobberXD: Oops, sorry. Will fix.

ClobberXD:
I am AFK for a couple of days and I cant test the code you provided until I return. But could you please do me a favour by retesting the same mod after applying the latest updates to the PR?

****

@SmallJoker Thanks for testing. Good to know it works for you. :)

MoNTE48:
Guys, merge this PR after all.
I can’t release PR to improve Android management because I don’t want to make conflicts. This PR will greatly simplify the hack for touchscreen place / dig nodes.

celeron55:
This looks fine enough and the basic feature works for me. We can fix remaining problems with specific mods later, if there are any.

sfan5:
:+1: Not tested but briefly reviewed the code before.

LoneWolfHT:
This problem is still present https://github.com/minetest/minetest/pull/7924#pullrequestreview-433633357

jamesalexanderstevenson:
You mean the first item, right?

LoneWolfHT:
Oh, yeah the second item. It's like all entities are automatically rightclicked when pointed at

jamesalexanderstevenson:
Only once, and only after a prior place action is performed successfully on a node.

LoneWolfHT:
Someone else is having this issue

LoneWolfHT:
I can create an issue linking to the stuff here if wanted

sfan5:
Please create an issue for that, otherwise it will be forgotten.

PR #7931, opened 2018-12-02
Re-master and improve touchscreen gui icons

stujones11:
The current touchscreen gui icons are a mix of different styles from a number of different artists using a variety of different editors. This PR attempts to unify the style, while at the same time, increase the resolution of the smaller icons to look better on modern HD screens.

I was tempted to just make them all 256x256 px however, I am not sure that is really necessary at the moment. All the master SVGs have been included in the the android build directory to make any future scaling possible without any loss of quality.

These images were all made 100% from scratch using Inkscape. I am no artist by any means however, as a mechanical engineer, I do find my way around most CAD and drawing programs quite easily.

icon_grid

<details><summary>In-game screenshots</summary>
<p>

**Before:**

before
**After:**

after

Note that the 'after' shot has `gui_scale_filtering` enabled, otherwise the effect is not quire as good. For some reason this is disabled by default on android.

</details>
<p>

PS: The difference may appear subtle, especially on a low density monitor so you really need to zoom in on these images to fully appreciate it.


MoNTE48:
_Good work, but i have strange problems with the enabled gui_scale_filtering...
(Damn, I really do not remember. But you and I once talked about this)_

stujones11:
> Good work, but i have strange problems with the enabled gui_scale_filtering...

@MoNTE48 Thanks, I do see a slight glitch with the loading bar with this setting enabled.

screen

I will look into this again but would prefer to keep any code changes out of this PR.

paramat:
I'd like this merged for 5.0.0. Very good work on the icons. It seems they are all greyscale-alpha.
:+1:

stujones11:
> Very good work on the icons. It seems they are all greyscale-alpha.

Thank you for the +1. The greyscale was actually the work of `pngcrush` but I don't see any problem with that :)

paramat:
By the way, now that all the Android blocker PRs are merged perhaps you can provide an .apk for testing? Maybe after this PR is merged too.

PR #7939, opened 2018-12-04
Add an activeobject manager to hold active objects

nerzhul:
currently implement client only

It uses lambda to do a callback on the AO object

### DIFF
--- a/src/activeobjectmgr.h
+++ b/src/activeobjectmgr.h
@@ -1,0 +1,69 @@
+/*
+Minetest
+Copyright (C) 2010-2018 nerzhul, Loic BLOT <loic.blot@unix-experience.fr>
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation; either version 2.1 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License along
+with this program; if not, write to the Free Software Foundation, Inc.,
+51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+#pragma once
+
+#include <unordered_map>
+#include <client/clientobject.h>
+#include "irrlichttypes.h"
+
+namespace mt {
+
+template<typename T>
+class ActiveObjectMgr {
+public:
+	ActiveObjectMgr() = default;
+	~ActiveObjectMgr() = default;
+
+	virtual void step(float dtime, std::function<void(T *)> &f) = 0;
+	virtual void clear(bool force) = 0;
+	virtual bool registerObject(T *obj) = 0;
+	virtual void removeObject(u16 id) = 0;
+
+	T * getActiveObject(u16 id)
+	{
+		typename std::unordered_map<u16, T *>::const_iterator n = m_active_objects.find(id);
+		return (n != m_active_objects.end() ? n->second : NULL);
+	}
+
+	u16 getFreeId() const
+	{
+		// try to reuse id's as late as possible
+		static u16 last_used_id = 0;
+		u16 startid = last_used_id;
+		for(;;) {
+			last_used_id ++;
+			if (isFreeId(last_used_id))
+				return last_used_id;
+
+			if (last_used_id == startid)
+				return 0;
+		}
+	}
+
+	bool isFreeId(const u16 id) const
+	{
+		return id != 0 && m_active_objects.find(id) == m_active_objects.end();
+
+	}
+protected:
+	std::unordered_map<u16, T *> m_active_objects;
+};
+
+}
--- a/src/activeobjectmgr.h
+++ b/src/activeobjectmgr.h
@@ -23,15 +23,17 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include <client/clientobject.h>
 #include "irrlichttypes.h"
 
-namespace mt {
+class TestActiveObjectMgr;
 
 template<typename T>
 class ActiveObjectMgr {
 public:
+
+
 	ActiveObjectMgr() = default;
 	~ActiveObjectMgr() = default;
 
-	virtual void step(float dtime, std::function<void(T *)> &f) = 0;
+	virtual void step(float dtime, const std::function<void(T *)> &f) = 0;
 	virtual void clear(bool force) = 0;
 	virtual bool registerObject(T *obj) = 0;
 	virtual void removeObject(u16 id) = 0;
@@ -39,7 +41,7 @@ public:
 	T * getActiveObject(u16 id)
 	{
 		typename std::unordered_map<u16, T *>::const_iterator n = m_active_objects.find(id);
-		return (n != m_active_objects.end() ? n->second : NULL);
+		return (n != m_active_objects.end() ? n->second : nullptr);
 	}
 
 protected:
@@ -64,5 +66,3 @@ protected:
 
 	std::unordered_map<u16, T *> m_active_objects;
 };
-
-}
--- a/src/activeobjectmgr.h
+++ b/src/activeobjectmgr.h
@@ -22,10 +22,13 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include <unordered_map>
 #include "irrlichttypes.h"
 
-class TestActiveObjectMgr;
+class TestClientActiveObjectMgr;
+class TestServerActiveObjectMgr;
 
 template<typename T>
 class ActiveObjectMgr {
+	friend class ::TestClientActiveObjectMgr;
+	friend class ::TestServerActiveObjectMgr;
 public:
 
 	virtual void step(float dtime, const std::function<void(T *)> &f) = 0;
--- a/src/activeobjectmgr.h
+++ b/src/activeobjectmgr.h
@@ -48,14 +48,12 @@ protected:
 		// try to reuse id's as late as possible
 		static thread_local u16 last_used_id = 0;
 		u16 startid = last_used_id;
-		for(;;) {
-			last_used_id++;
-			if (isFreeId(last_used_id))
-				return last_used_id;
-
+		while (!isFreeId(++last_used_id)) {
 			if (last_used_id == startid)
 				return 0;
 		}
+
+		return last_used_id;
 	}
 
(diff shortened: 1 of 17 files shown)


src/activeobjectmgr.h
```diff
@@ -0,0 +1,70 @@
...
+		static thread_local u16 last_used_id = 0;
+		u16 startid = last_used_id;
+		for(;;) {
+			last_used_id++;
+			if (isFreeId(last_used_id))
+				return last_used_id;
+
+			if (last_used_id == startid)
+				return 0;
+		}
```
Quent42340: Wouldn't something like this be better ?

```cpp
while (!isFreeId(++last_used_id)) {
    if (last_used_id == startid)
        return 0;
}

return last_used_id;
```

src/activeobjectmgr.h
```diff
@@ -0,0 +1,70 @@
...
+
+	virtual void step(float dtime, std::function<void(T *)> &f) = 0;
+	virtual void clear(bool force) = 0;
+	virtual bool registerObject(T *obj) = 0;
+	virtual void removeObject(u16 id) = 0;
+
+	T * getActiveObject(u16 id)
+	{
+		typename std::unordered_map<u16, T *>::const_iterator n = m_active_objects.find(id);
+		return (n != m_active_objects.end() ? n->second : NULL);
```
Quent42340: Please use `nullptr` instead of `NULL` in new added code to remove dependency to `cstdlib`

src/client/activeobjectmgr.h
```diff
@@ -0,0 +1,44 @@
...
+namespace client
+{
+class ActiveObjectMgr : public  mt::ActiveObjectMgr<ClientActiveObject>
+{
+public:
+	ActiveObjectMgr() = default;
+	~ActiveObjectMgr() = default;
+
+	void clear(bool force) override;
+	void step(float dtime, std::function<void(ClientActiveObject *)> &f) override;
```
Quent42340: `const &` for `std::function` should be possible here

src/client/clientenvironment.cpp
```diff
@@ -262,29 +259,31 @@ void ClientEnvironment::step(float dtime)
...
 
-			obj->updateLight(light);
+			cao->updateLight(light);
 		}
-	}
+	};
+
+	// hold the callback state
+	std::function<void(ClientActiveObject *)> cb = std::ref(cb_state);
+	m_ao_manager.step(dtime, cb);
```
Quent42340: You should be able to pass directly cb_state (same signature than cb, `void(ClientActiveObject *)`)

nerzhul: i cannot it's why i added std::ref, compiler refused it due to lambda capture. Without capture it's allowed

Quent42340: If you can't, then remove `auto` and just do `std::function<void(ClientActiveObject *)> cb_state = ...`

paramat:
Is there a chance of causing subtle bugs? Just wondering if it should be left until after 5.0.0.

nerzhul:
@paramat it's basically a code move only, + enhancement of the code quality with unittests to ensure the behaviour is correct, it's possible to embed it in 5.0.0.

src/unittest/test_activeobjectmgr.cpp
```diff
@@ -0,0 +1,108 @@
...
+	UASSERT(caomgr.getFreeId() != aoid);
+
+	aoids.push_back(aoid);
+
+	// Register basic objects, ensure we never found
+	for (u8 i = 0; i < UINT8_MAX; i++) {
+		// Register an object
+		auto tcao = new TestClientActiveObject();
+		caomgr.registerObject(tcao);
+		aoid = tcao->getId();
```
SmallJoker: I think you forgot `aoids.push_back(aoid);` here.

src/serverenvironment.h
```diff
@@ -243,7 +244,10 @@ class ServerEnvironment : public Environment
 		-------------------------------------------
 	*/
 
-	ServerActiveObject* getActiveObject(u16 id);
+	ServerActiveObject* getActiveObject(u16 id)
```
SmallJoker: Needs style correction

src/serverenvironment.cpp
```diff
@@ -1542,43 +1481,11 @@ void ServerEnvironment::getAddedActiveObjects(PlayerSAO *playersao, s16 radius,
...
-
-		// Discard if already on current_objects
-		std::set<u16>::iterator n;
-		n = current_objects.find(id);
-		if(n != current_objects.end())
-			continue;
-		// Add to added_objects
-		added_objects.push(id);
-	}
+	m_ao_manager.getAddedActiveObjectsAroundPlayer(playersao->getBasePosition(), radius_f,
```
SmallJoker: IMO this function is too specific for the ActiveObject class. Suggestion: Add an iterator so that the (real active) active objects can read out quickly.

src/server/activeobjectmgr.cpp
```diff
@@ -0,0 +1,153 @@
...
+	g_profiler->avg("Server::ActiveObjectMgr: num of objects", m_active_objects.size());
+	for (auto &ao_it : m_active_objects) {
+		f(ao_it.second);
+	}
+}
+
+bool ActiveObjectMgr::registerObject(ServerActiveObject *obj)
+{
+	assert(obj); // Pre-condition
+	if (obj->getId() == 0) {
```
SmallJoker: Needs code style cleanups

src/unittest/test_activeobjectmgr.cpp
nerzhul: right !

src/serverenvironment.h
nerzhul: yes, code style will be fixed after functional review

src/serverenvironment.cpp
nerzhul: i wanted to prevent this to make the active object really owned by manager and protect them. it's also why there is a ServerAO Manager :) and i should rename to getAddedActiveObjectsAroundPosition because before we passed the playerSAO and now it's just a position.

SmallJoker: What's the profit of forwarding all functions to the manager? Just speaking of the code, it will be more independent - but also gets a complexer structure. The two classes cannot become independent anyway.

paramat:
> it's basically a code move only, + enhancement of the code quality with unittests

Thanks, seems fine then. I think i was over-concerned :)

src/activeobjectmgr.h
```diff
@@ -0,0 +1,67 @@
...
+		return last_used_id;
+	}
+
+	bool isFreeId(const u16 id) const
+	{
+		return id != 0 && m_active_objects.find(id) == m_active_objects.end();
+
+	}
+
+	std::unordered_map<u16, T *> m_active_objects;
```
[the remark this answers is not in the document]
nerzhul: yeah but not now as the pointer lifetime is not depending on the map storage but can be different. I don't touch this lifetime part now, 5.0.0 must be stable

nerzhul:
@SmallJoker , @Quent42340 PR is now ready

src/activeobjectmgr.h
```diff
@@ -0,0 +1,65 @@
...
+		u16 startid = last_used_id;
+		while (!isFreeId(++last_used_id)) {
+			if (last_used_id == startid)
+				return 0;
+		}
+
+		return last_used_id;
+	}
+
+	bool isFreeId(const u16 id) const
```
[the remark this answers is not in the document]
nerzhul: exact i kept it from the previous code :)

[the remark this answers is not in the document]
Quent42340: You're making a "Code quality" PR, so please take the time to remove useless stuff when you see it. :)

And by "useless" I mean **obvious** useless, like this `const`, or like a `const` in this code for example:
```cpp
const u8 getMember() const { return m_member; }
```

src/server/activeobjectmgr.cpp
```diff
@@ -0,0 +1,165 @@
...
+		u16 new_id = getFreeId();
+		if (new_id == 0) {
+			errorstream << "Server::ActiveObjectMgr::addActiveObjectRaw(): "
+				    << "no free id available" << std::endl;
```
SmallJoker: Please use tabs to indent. (same for the the entire file)

src/serverenvironment.cpp
```diff
@@ -1542,43 +1481,12 @@ void ServerEnvironment::getAddedActiveObjects(PlayerSAO *playersao, s16 radius,
...
-		std::set<u16>::iterator n;
-		n = current_objects.find(id);
-		if(n != current_objects.end())
-			continue;
-		// Add to added_objects
-		added_objects.push(id);
-	}
+	m_ao_manager.getAddedActiveObjectsAroundPos(playersao->getBasePosition(), radius_f,
+												player_radius_f, current_objects,
+												added_objects);
```
SmallJoker: Incorrect re-formatting.

src/serverenvironment.cpp
```diff
@@ -1976,53 +1832,46 @@ void ServerEnvironment::activateObjects(MapBlock *block, u32 dtime_s)
...
 		// is actually located in an active block, re-save to the block in
 		// which the object is actually located in.
-		if(!force_delete &&
-			obj->m_static_exists &&
-			!m_active_blocks.contains(obj->m_static_block) &&
-			m_active_blocks.contains(blockpos_o))
-		{
+		if (!force_delete && obj->m_static_exists &&
+		   !m_active_blocks.contains(obj->m_static_block) &&
+		   m_active_blocks.contains(blockpos_o)) {
```
SmallJoker: Incorrect re-formatting.

src/serverenvironment.cpp
```diff
@@ -2057,9 +1905,9 @@ void ServerEnvironment::deactivateFarObjects(bool _force_delete)
 							data_changed = false;
 					} else {
 						warningstream << "ServerEnvironment::deactivateFarObjects(): "
-								<< "id=" << id << " m_static_exists=true but "
-								<< "static data doesn't actually exist in "
-								<< PP(obj->m_static_block) << std::endl;
+									  << "id=" << id << " m_static_exists=true but "
+									  << "static data doesn't actually exist in "
+									  << PP(obj->m_static_block) << std::endl;
```
SmallJoker: Incorrect re-formatting.

src/server/activeobjectmgr.cpp
nerzhul: clang-format cheat :) i will fix that

src/serverenvironment.cpp
nerzhul: oh exact, mis configured clion installation on a new pc :)

PR #7941, opened 2018-12-05
Fix player rotations

CoderForTheBetter:
This fixes the rotation error introduced by my commit https://github.com/minetest/minetest/commit/faa358e797128ab07bb5644ce305a832d59e5596 and is related to this issue: https://github.com/minetest/minetest/issues/7938.

Here's a video that shows that this fixes the problem: [Video](https://streamable.com/rgpk8)

Thanks, everyone for being patient and testing!

### DIFF
--- a/src/content_sao.cpp
+++ b/src/content_sao.cpp
@@ -1091,7 +1091,7 @@ void PlayerSAO::step(float dtime, bool send_recommended)
 			pos,
 			v3f(0.0f, 0.0f, 0.0f),
 			v3f(0.0f, 0.0f, 0.0f),
-			v3f(0.0f, 0.0f, 0.0f),
+			m_rotation,
 			true,
 			false,
 			update_interval


jastevenson303:
Didn't work.  
screenshot from 2018-12-05 10-03-28

These players are looking at each other...

CoderForTheBetter:
Hmm, I see the players are wearing armor. I'm pretty sure that means they are attached to entities. Try detaching the players from entries and see if it works. There is still a bug, and I'm sure I have halfway fixed it.

CoderForTheBetter:
Hmm, I see the players are wearing armor. I'm pretty sure that means they are attached to entities. Try detaching the players from entries and see if it works. There is still a bug, and I'm sure I have halfway fixed it.

Edit: just saw your edit, will leave mine up just in case.

PR #7947, opened 2018-12-06
Update minetest.conf.example and run updatepo.sh

SmallJoker:
EDIT: And btw, is it intended to run the script like this?


nerzhul:
Merge when you want

PR #7950, opened 2018-12-08
Fix crash if display resolution is not set

martinxyz:
On my wayland / gnome3 setup DisplayHeightMM() returns 0. This resulted in a
misleading startup error suggesting to fix my font paths.


martinxyz:
I've also backported this to stable-0.4 in my fork. Should I make a separate PR?

SmallJoker:
@martinxyz Thanks for your efforts but once the commit is merged into master, it should be easy to just cherry-pick it into the backport branch. However, we're currently working towards 5.0, so once that one is out, there won't be any further 0.4.x release (i.e. useless backport).

PR #7957, opened 2018-12-09
Fix dissector script to work with newer versions of Wireshark (>2.6)

meme:
`ProtoField.bytes` are always displayed as hex, so the parameter asks for what symbol you'd like to separate them with. This PR replaces `base.DEC` with `base.NONE` for no separation.


ClobberXD:
I got the errors too, and I was wondering if the issue was on my end. :+1:

PR #7958, opened 2018-12-09
Make MapNode handle paramtype2≠leveled properly

### LINKED ISSUE #6913: Leveled nodeboxes work without paramtype2="leveled"
The snow mod uses a leveled nodebox for the snow node (https://github.com/Splizard/minetest-mod-snow/pull/31). It works although paramtype2 is not set to "leveled" in the nodedef.
l guess the difference between paramtype2 set or not is that other mods can easily find out whether nodes' param2 is used for leveled.
I think minetest should give a warning if a mod uses leveled nodeboxes while paramtype2 is not "leveled".

numberZero:
Fix #6913.
@HybridDog @paramat


nerzhul:
i think this getLevel function is a good candidate for unittests to ensure the result is the one we wanted.

Issue #7880, opened 2018-11-21
Missing strings for translation


Wuzzy2:
##### Issue type
- Translation issue
 
##### Minetest version
67b20ff15a4786edd76adcbdb921fcc4acac7113

##### Summary
A couple of strings are currently untranslatable or Minetest fails to show the translation:

* `builtin` strings enclosed in `core.gettext` (not collected by `updatepo.sh`)
* Strings in `game.cpp` when `showOverlayMessage` is called (e.g. “Shutting down...”)
* “Unknown node: [node name]” when you hover an unknown node

Please post in this issue if you have found more missing strings

SmallJoker:
Relevant code part - should be a quick task to add the functions mentioned above:
https://github.com/minetest/minetest/blob/67b20ff15a4786edd76adcbdb921fcc4acac7113/util/updatepo.sh#L51-L64

Wuzzy2:
#9513 will fix this:

> builtin strings enclosed in core.gettext (not collected by updatepo.sh)

Wuzzy2:
This one is already fixed:

> Strings in game.cpp when showOverlayMessage is called (e.g. “Shutting down...”)

And this one is no longer relevant, as it isn't displayed anymore:

> “Unknown node: [node name]” when you hover an unknown node

This means, as soon #9513 is merged, this issue is fixed.

Wuzzy2:
UPDATE: Ignore #9513 from now on, this issue has been superceded.

If #10693 is merged, this issue is fixed.

PR #7961, opened 2018-12-10
Make showOverlayMessage strings translatable

Wuzzy2:
This PR makes strings put into `showOverlayMessage` in `game.cpp`. translatable, i.e. these are the texts that show up in the loading screen on the loading bar.

Partial fix of #7880.


nerzhul:
there is a conflict, can your rebase ?

Wuzzy2:
As you can probably see, I completely screwed up this branch. XD
It's faster to start over than trying to repair the branch.

2nd try here: https://github.com/minetest/minetest/pull/7964

PR #7964, opened 2018-12-11
Make showOverlayMessage strings translatable (take 2)

Wuzzy2:
This PR makes strings put into `showOverlayMessage` in `game.cpp` translatable, i.e. these are the texts that show up in the loading screen on the loading bar.

Partial fix of #7880.

### DIFF
--- a/src/client/game.cpp
+++ b/src/client/game.cpp
@@ -1146,7 +1146,7 @@ void Game::shutdown()
 	if (current_formspec)
 		current_formspec->quitMenu();
 
-	showOverlayMessage("Shutting down...", 0, 0, false);
+	showOverlayMessage(N_("Shutting down..."), 0, 0, false);
 
 	if (clouds)
 		clouds->drop();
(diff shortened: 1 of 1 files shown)


nerzhul:
Why do you translate ? it's already done in the function itself:

https://github.com/minetest/minetest/blob/1d8f8f3f38b4e04cc87177f4492b7dfe167f2bf9/src/client/game.cpp#L3965

Wuzzy2:
False.

`_N` is a no-op that just returns the input string. It does *not* translate.

The sole purpose of `_N` is to make sure the string will be collected. This is neccessary because the actual translation happens in the function itself, but if the actual strings have no marker, they won't be collected.

This is pretty basic Gettext practice, please review your Gettext manual.
https://www.gnu.org/software/gettext/manual/html_node/Special-cases.html#Special-cases

nerzhul:
okay i see ty

PR #7968, opened 2018-12-11
Add a Dockerfile + gitlab CI build

nerzhul:
Permit to distribute minetest server as a docker file, embedded in the Gitlab CI + Gitlab registry permitting to provide minetest server daily builds.

It can also permit servers owners to use minetest server through docker easily, if it's a valid usecase do you want some documentation ?

Here is the demonstration build video connecting a local client on the minetest server inside the container

https://www.youtube.com/watch?v=fT4FBmUZPWI

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM debian:stretch
 USER root
 RUN apt-get update -y && \
 	apt-get -y install build-essential libirrlicht-dev cmake libbz2-dev libpng-dev libjpeg-dev \
-		libsqlite3-dev libcurl4-gnutls-dev zlib1g-dev libgmp-dev libjsoncpp-dev && \
+		libsqlite3-dev libcurl4-gnutls-dev zlib1g-dev libgmp-dev libjsoncpp-dev git && \
 		apt-get clean && rm -rf /var/cache/apt/archives/* && \
 		rm -rf /var/lib/apt/lists/*
 
@@ -17,7 +17,9 @@ RUN	mkdir -p /usr/src/minetest/cmakebuild && cd /usr/src/minetest/cmakebuild && 
 		-DENABLE_SYSTEM_JSONCPP=1 \
 		-DENABLE_SOUND=0 \
 		.. && \
-		make -j2 && make install
+		make -j2 && \
+		rm -Rf ../games/minetest_game && git clone https://github.com/minetest/minetest_game ../games/minetest_game && \
+		make install
 
 FROM debian:stretch
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,7 @@ RUN	mkdir /usr/src/minetest/cmakebuild && cd /usr/src/minetest/cmakebuild && \
 		-DENABLE_GETTEXT=TRUE \
 		-DBUILD_SERVER=TRUE \
 		-DBUILD_CLIENT=FALSE \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+COPY . /usr/src/minetest
+
+RUN	mkdir -p /usr/src/minetest/cmakebuild && cd /usr/src/minetest/cmakebuild && \
+    	cmake -DCMAKE_INSTALL_PREFIX=/usr/local -DCMAKE_BUILD_TYPE=Release -DRUN_IN_PLACE=FALSE \
+		-DENABLE_GETTEXT=TRUE \
+		-DBUILD_SERVER=TRUE \
+		-DBUILD_CLIENT=FALSE \
+		-DENABLE_SYSTEM_JSONCPP=1 \
+		-DENABLE_SOUND=0 \
```
sfan5: probably not needed since client is disabled anyway

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+	apt-get -y install build-essential libirrlicht-dev cmake libbz2-dev libpng-dev libjpeg-dev \
+		libsqlite3-dev libcurl4-gnutls-dev zlib1g-dev libgmp-dev libjsoncpp-dev git && \
+		apt-get clean && rm -rf /var/cache/apt/archives/* && \
+		rm -rf /var/lib/apt/lists/*
+
+COPY . /usr/src/minetest
+
+RUN	mkdir -p /usr/src/minetest/cmakebuild && cd /usr/src/minetest/cmakebuild && \
+    	cmake -DCMAKE_INSTALL_PREFIX=/usr/local -DCMAKE_BUILD_TYPE=Release -DRUN_IN_PLACE=FALSE \
+		-DENABLE_GETTEXT=TRUE \
```
sfan5: gettext for a server?

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+		rm -Rf ../games/minetest_game && git clone https://github.com/minetest/minetest_game ../games/minetest_game && \
+		make install
+
+FROM debian:stretch
+
+USER root
+RUN groupadd minetest && useradd -m -g minetest -d /var/lib/minetest minetest && \
+    apt-get update -y && \
+    apt-get -y install libcurl3-gnutls libjsoncpp1 liblua5.1-0 libluajit-5.1-2 libpq5 libsqlite3-0 \
+        libstdc++6 zlib1g libirrlicht1.8 libc6
```
sfan5: irrlicht is not actually needed at runtime for the server, so you may be able to omit it here

Dockerfile
nerzhul: oops i forgot this :)

Dockerfile
nerzhul: i try to launch without it, because we linked against it and uses some irrlicht headers at compilation time.

Dockerfile
nerzhul: BUILD_CLIENT is set to true by default. BUILD_SERVER is set to FALSE by default

nerzhul:
Validated by @sfan5 on IRC

http://irc.minetest.net/minetest-dev/2018-12-11#i_5456660

nerzhul:
And it works well ! 

capture d ecran de 2018-12-12 09-40-04

PR #7976, opened 2018-12-14
HUD: Font shadow added on item stacks for better readability. (Really trivial)

Quent42340:
What the title says.

***

<details><summary>Before</summary>
<p>

screenshot-20181214105149

</p>
</details>

***

<details><summary>After</summary>
<p>

screenshot-20181214105117

</p>
</details>

***

<details><summary>Offset fix</summary>
<p>

screenshot-20181214114602

</p>
</details>

***

### DIFF
--- a/src/client/hud.cpp
+++ b/src/client/hud.cpp
@@ -771,6 +771,9 @@ void drawItemStack(video::IVideoDriver *driver,
 		video::SColor bgcolor(128,0,0,0);
 		driver->draw2DRectangle(bgcolor, rect2, clip);
 
+		video::SColor shadow_color(0,0,0,255);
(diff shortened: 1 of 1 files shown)


Quent42340:
Well, currently the position isn't correct due to the shadow, and I noticed that a small shadow was already existing, so changes may occur here.

Quent42340:
The goal of this change is to make item counts more readable. I decided to use the same solution than in [my own C++ Minecraft-like game](https://youtu.be/Ub3nGgGvEBk?t=37) which is exactly the same solution Minecraft itself uses.

Minetest has a small shadow but it's barely visible given the size of the font on item counts.

However I'm open to any suggestion to replace this PR if it can't be accepted.

Quent42340:
Ok Minetest actually has a config option for this. This PR is useless then.

PR #7986, opened 2018-12-16
Inventory lists: Middle drag + Left drag (#6544 revived)

Unarelith:
Since #6544 was under `Adoption Needed` tag, I decided to rebase it and take it further.

I tested the PR, it seems to work perfectly, tell me if I need to change things in the code but looks good to me so far.


src/gui/guiFormSpecMenu.cpp
```diff
@@ -3396,6 +3421,47 @@ bool GUIFormSpecMenu::OnEvent(const SEvent& event)
...
 
 			m_selected_dragging = false;
+
+			// If the mouse was let go off once and we are in left drag mode
+			if (m_auto_place && m_left_dragging) {
+				int each = MYMIN(m_selected_amount_left_drag_copy/m_left_drag_to_stack.size(), m_left_drag_stack.getStackMax(m_client->idef()));
+
+				m_selected_amount = m_selected_amount_left_drag_copy;
+
+				for (std::set<ItemSpec*>::iterator it=m_left_drag_to_stack.begin(); it!=m_left_drag_to_stack.end(); ++it) {
```
Stenodyon:
```suggestion
				for (auto &s : m_left_drag_to_stack) {
```

In addition, I see a couple of lines above 80 characters.

Stenodyon:
I encountered a bug while testing:

1. Gave myself 99 wood
2. Left-click dragged in the crafting area to craft a chest
3. Put the remaining wood away
4. Took the chest from the crafting output
5. When trying to put the chest in an inventory slot, it would either not let me place it or would place 1 wood in the slot instead, and the game segfaulted after trying this repeatedly.

Quent42340:
Well, looks like this problem was inherited of the old PR...

Easy steps to reproduce:
- Drag a wood stack with right-click to craft a chest
- Try to place the chest in the inventory

It seems that something is wrong with the saved selected item, since sometimes when you try to put the chest, a wood node will be but instead. The problem seems to be reliability between inventories, since it only happens when moving to another inventory.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -3307,54 +3312,74 @@ bool GUIFormSpecMenu::OnEvent(const SEvent& event)
...
 						count = 1;
 					else  // left
 						count = s_count;
 
 					if (!event.MouseInput.Shift) {
 						// no shift: select item
+
+						// m_selected_item can be deleted by the black magic when it's empty
+						// As such we create our own copy of it for left drag handling
+						m_left_drag_item = new ItemSpec(s);
```
Stenodyon: This line is executed when taking an item stack from the inventory but is not executed when taking the item stack from the crafting output.

Quent42340: You're right, thanks for noticing. Though it's unclear why the author used `m_left_drag_item` and not `m_selected_item` only.

Unarelith:
This PR was ready for review I guess. I can rebase it when you're ready to review it, but it also can be rebased after #7970 if needed. ;)

SmallJoker:
The PR goals are good, and it would improve gameplay if at least middle-click dragging were a thing.
I don't know why this PR was never given real attention. After a quick look the code is all there and covers the wanted cases. However, these operations should be moved to one of the inherited `InventoryAction` classes (`clientApply`). `guiFormSpecMenu` should only handle inputs, and use inventory functions to perform the actions.

Unarelith:
@paramat You can close this as long as it keeps the "Adoption needed" tag, which should be removed from the original PR if it's still there because I made some fixes to it iirc

PR #6165, opened 2017-07-24
Time-of-day-dependent face shading, specular lighting

kaadmy:
Adds directional lighting and (very subtle) specular on everything, including models, if shaders are enabled.
Model shading requires the models to be properly exported with normals (well, obviously...)

Shading depends on the sun (or moon) angle and changes properly during the day/night and eases smoothly during a dawn/dusk transition.

Before/after screenshot:
Before and after

PR #7987, opened 2018-12-16
Time-of-day-dependent face shading (#6165 revived)

Unarelith:
Since #6165 is dead, I rebased it. ~However, even if all the changes are exactly the same than on the original PR, it currently doesn't work with current engine version.~

~I tried to investigate this issue but I didn't managed to fix it. I'll keep searching but your help is very welcome.~

**FINALLY WORKS**

***

<details><summary>Old lighting</summary>
<p>

screenshot-20181216125236

</p>
</details>

***

<details><summary>Dawn</summary>
<p>

screenshot-20181216124256

</p>
</details>

***

<details><summary>Midday</summary>
<p>

screenshot-20181216124304

</p>
</details>

***

<details><summary>Dusk</summary>
<p>

screenshot-20181216124315

</p>
</details>

***


niansa:
Looks good :thumb_up:

lhofhansl:
Nice. Lemme try.

lhofhansl:
Captures the mood of the specific day-time nicely. I like it, makes for more immersive game play.
Haven't looked closely at the code... I agree with the idea.

Now we need (IMHO) dynamic shadows and reflective water - these also do a lot for immersive game play. :)

lhofhansl:
I found a small problem with the wielded_light mod https://github.com/minetest-mods/wielded_light.git
In the morning hours the wielded light does not light anything (but it works when I place a torch)

paramat:
The original PR has some issues, causing strange lighting in some situations, so this needs testing.

Quent42340:
@paramat well, it works, it's an experimental feature and it has a config option, so if the code is good I'd say merge it and wait for skilled people to fix it.

paramat:
Edited my previous post, looking through the original something about the wielditem is mentioned, not sure what's wrong.
We merge PRs once the bugs are fixed not before :)

Quent42340:
@paramat just read @lhofhansl message, I think it's the same thing, and it's not really a bug, it's just a minor issue.

paramat:
Yes sorry, any bugs seem minor and considering the rebase issues i'd be fine with merging now as long as there's a comment in docs saying 'experimental' or something.

Quent42340:
@paramat Done. Tell me if it's ok like this please.

paramat:
Yes fine thanks.
/////////////

As far as i can see, the mix of time-dependent and conventional shading is linear and determined by: 
```
	// The alpha gives the ratio of sunlight in the incoming light.
	float outdoorsRatio = 1.0 - gl_Color.a;
```
Apparently alpha is the ratio of artificial light to day natural light.

I'm thinking there is potential for tuning this effect in future by altering how the mix depends on 'outdoorsRatio' and amount of daylight and making the relationships non-linear with both.

Consider that this effect is dependent on the direction of the sun and ideally is dependent on the nodes being exposed to sunlight, it follows that ideally the effect should only happen in high daylight light levels and fall off rapidly in darker places.
So a linear relationship seems possibly not ideal as it would seem to over-apply the effect in semi-dark areas.

screenshot_20170820_114316

For example, daylight comes in from the hole above and the whole room is lit by daylight, so the effect will happen on the walls here. Perhaps the effect should fall off rapidly for lower light levels, if possible.

client/shaders/nodes_shader/opengl_vertex.glsl
```diff
@@ -130,15 +156,26 @@ float disp_z;
...
 		nightRatio * artificialLight.rgb) * 2;
 	color.a = 1;
-	
+
+#if defined(ENABLE_DIRECTIONAL_SHADING) && !LIGHT_EMISSIVE
+	// Lighting color
+	vec3 resultLightColor = ((lightColor.rgb * gl_Color.a) + nightRatio);
+
+	// ((resultLightColor * ((max(dot(normal, lightDirection), -0.2) + 0.2) / 1.2)* 0.6)) + 0.4;
+	resultLightColor = (resultLightColor * ((max(dot(alwaysNormal, lightDirection), -0.2) + 0.2) * 0.5)) + 0.4;
```
paramat: Maybe this variable name is a little unclear? Should it refer to directional light somehow?

paramat:
Good thanks.

Specular lighting seems unrelated to this feature, it really shouldn't be part of this PR due to being a completely different effect. I think it should be moved out to a separate PR (although personally i don't think specular lighting is justified or worth adding).
As far as i can see specular lighting is easily separated from this PR.

src/client/game.cpp
```diff
@@ -539,14 +592,17 @@ class GameGlobalShaderConstantSetterFactory : public IShaderConstantSetterFactor
...
+	std::vector<GameGlobalShaderConstantSetter *> created_camera;
 public:
 	GameGlobalShaderConstantSetterFactory(bool *force_fog_off,
 			f32 *fog_range, Client *client) :
 		m_sky(NULL),
 		m_force_fog_off(force_fog_off),
 		m_fog_range(fog_range),
-		m_client(client)
+		m_client(client),
+		m_camera(NULL)
```
ClobberXD: `nullptr`?

ClobberXD:
Finally found the time to give this PR a try. The world looks much more alive and nicer, great. :+1: 

However, I did notice that the upper/lower faces aren't brightened at all. Even if at an angle, sunlight does fall on the upper face too, and it'd be nice if that was brightened as well. During sunrise and sunset, the difference in light between the upper face and the side facing the sun is very striking, and even borderline unpleasant to look at:

screenshot_20181217_165011
screenshot_20181217_171335

src/client/game.cpp
Quent42340: It was in the original PR, I plan to do a big `sed -i 's/NULL/nullptr/g'` in the code at some point so it's not a big deal. I'll fix it tho

Quent42340:
@ClobberXD yeah it's weird I agree, but it's either that or nothing at all, someone could still improve it in a future PR. Unfortunately I don't have enough skills on that topic to fix it. :(

src/client/game.cpp
ClobberXD: Ah, ok. Similar PR: #6759

Quent42340: I saw that one, I'll do another PR like this after 5.0.0

paramat:
The way each face is brightened probably needs tuning (could be done after merge), for example at midday the sides are perhaps a little too dark.
Light at sunrise / sunset looks somewhat ok to me.
///////////////////

For specular lighting, i'm :-1: for this PR if it's left in as it's implemented so badly. If we have specular lighting it needs to be applied to smooth, shiny, reflective surfaces only, by using a new nodedef and a new shader material. This would need to be a separate PR (specular lighting should be a separate PR anyway).

Quent42340:
@paramat I removed specular lighting, tell me if I forgot anything.

I also added an 'experimental' mention in the settings menu since I had more space, you may want some changes to it.

paramat:
Thanks will test soon.

lhofhansl:
Re: My wielded_light comment above. This seems to be related to how/when that mod decided to place the "fake" light sources. So it's not a lighting problem per se, and it is also not necessarily wrong. I still support this.

lhofhansl:
What happens with weather mods? If a rain mod "hides" the sun, we'd want to not do the day light dependent shading. I guess it's not possible to suit everyone or every scenario.

Quent42340:
If someone decides to use a weather mod that hides the sun, then he'll just disable this feature until someone adds support for this.

lhofhansl:
Let's get this in?

ClobberXD:
nerzhul has added this PR to the 5.1.0 milestone. Probably because we're approaching a feature freeze, and it's best to avoid any major changes to the code, apart from bugfixes. :)

nerzhul:
yeah it's an interesting but also a major feature, it's better to delay it

HybridDog:
Could you move directional light calculation to the fragment shader if bumpmapping is enabled?

If I understand the code correctly, bumpmapping currently assumes that the light comes from the camera and the normal vector from the normalmap lacks a rotation: bump.xyz does not incorporate the face direction. This code could be changed.
https://github.com/minetest/minetest/blob/a4677496f347a3c9bd66b39ca535f8346bac1708/client/shaders/nodes_shader/opengl_fragment.glsl#L186-L190
With bumpmapping, face shading can be done for each pixel. For example, a normalmap can be used to smooth the current hard transition between the top and side grass.
grass

ClobberXD:
This seems to fix both the issue of sides being dark at midday and top being dark at dawn and dusk, as this eases out the contrast between the two faces.

Quent42340:
@HybridDog I gave you write access to my fork, the only relevant branch is `shading`
Sorry, I don't have time to take care of this PR right now

benrob0329:
What's the status of this?

paramat:
The usual, waiting for review.

LoneWolfHT:
*bump*
There's a conflicting file

Unarelith:
@LoneWolfHT Sure there is conflicts, but this PR probably won't be reviewed soon so I won't bother rebasing this each time there's a conflict.

@paramat Please tell me when someone is interested to review this, I'll rebase it then.

paramat:
>  this PR probably won't be reviewed soon so I won't bother rebasing this each time there's a conflict.

That's wise, and we don't expect a rebase on each conflict.

SmallJoker:
Rebased to current HEAD for testing.

LoneWolfHT:
image

LoneWolfHT:
This doesn't look right either, not sure how it could be fixed though
image

LoneWolfHT:
image
image

LoneWolfHT:
I think the dark faces should be slightly brighter until the sun sets. It feels too dark in the evenings

HybridDog:
> I think the dark faces should be slightly brighter until the sun sets. It feels too dark in the evenings

You need to apply gamma correction because the colour values in the glsl shaders are sRGB values (for example, (0.5,0.5,0.5) is significantly darker than 50% linear brightness grey).

paramat:
https://github.com/minetest/minetest/pull/7987#issuecomment-447710228 needs addressing, but could be attended to after merge.

LoneWolfHT:
It feels like the sky is completely clouded over during midday
image

LoneWolfHT:
image

paramat:
Node tops are roughly the same brightness, node sides are darker and a little too dark.

LoneWolfHT:
> Node tops are roughly the same brightness, node sides are darker and a little too dark.

And the reverse is true for dawn/dusk

LoneWolfHT:
image
image
This may show the top face brightness difference better. Both screenshots were taken at the same time of day

benrob0329:
This can always be improved overtime with individual issues for addressing some smaller problems.

paramat:
I see, yes too dark. Certainly needs tuning.

LoneWolfHT:
[Directional Lighting.mp4.zip](https://github.com/minetest/minetest/files/3342165/Directional.Lighting.mp4.zip)

Been doing some tweaking. I think I've reduced the difference. And I don't think it makes everything else too dark anymore

LoneWolfHT:
image
I made nodes lighting slightly brighter on my client so I could go around with gamma = 1 and still be able to see at night. So they'll probably be about even in brightness without those changes

paramat:
That's still darker, should be equal.

benrob0329:
Actually I think it's slightly lighter when directional shading is turned on, which would make sense as the sun is illuminating the top face more the the others (while normally they are all lit evenly).

LoneWolfHT:
grenade
The grenade sprite isn't supposed to vanish like that. This only happens on my main client. Which I've added this PR to. The only other PR I've merged in is the CSM one: https://github.com/minetest/minetest/pull/8627. Which I doubt is the cause

paramat:
> Actually I think it's slightly lighter when directional shading is turned on, which would make sense as the sun is illuminating the top face more the the others (while normally they are all lit evenly).

Oops, my mistake.
And i agree.

benrob0329:
Can this be merged as experimental for 5.1?

LoneWolfHT:
> Can this be merged as experimental for 5.1?

There's still this bug https://github.com/minetest/minetest/pull/7987#issuecomment-507323610
Which happens with more than just the grenade entities.

client/shaders/nodes_shader/opengl_vertex.glsl
```diff
@@ -159,20 +171,22 @@ float disp_z;
...
-	directional_light *= 0.5;
-	resultLightColor = resultLightColor * to_sRGB(directional_light) + ambient_light;
+	directional_light *= (1.0 - ambient_light) / 1.2;
+	resultLightColor = resultLightColor * directional_light + ambient_light;
 
 	directional_light = dot(alwaysNormal, artificialLightDirection);
 	directional_light = max(directional_light + 0.5, 0.0);
-	directional_light *= 0.47;
-	float artificialLightShading = to_sRGB(directional_light + 0.3);
+	directional_light *= (1.0 - 0.3) / 1.5;
```
SmallJoker: How often is this part called? Multiplying by its inverse might give a slight performance improvement on weaker GPUs.

LoneWolfHT:
I think @HybridDog's changes have made the unlit sides look worse, I think I'm expecting shadows there
image

LoneWolfHT:
Still have the contrast problem
image

LoneWolfHT:
I've gone and randomly changed 3 `0.3` values to `0.5` hoping that the result would look better (I'm not sure what's going on in the code) and I think I like how it looks. The mornings aren't so dark either
image

Code:
(Selected parts were changed)
image

client/shaders/nodes_shader/opengl_vertex.glsl
HybridDog: I haven't added these changes to the other vertex shaders; this still needs testing.

HybridDog:
> […] I'm expecting stronger shadows there

When the angle is 90° or more, no parallel sunlight reaches the face, so with simple directional lighting and flat surfaces, the side and back should have the same darkness. This looks a bit bad, so currently 0.2 is added to the directional light.
For testing, you could change the `0.2` to `0.6` and `1.2` to `1.6`.
https://github.com/minetest/minetest/pull/7987/files#diff-13d708d8b99ea3c4d320fe0b2eccba9cR179

src/client/content_cao.cpp
```diff
@@ -520,6 +523,22 @@ void GenericCAO::removeFromScene(bool permanent)
 	}
 }
 
+void GenericCAO::setSceneNodeMaterial(scene::ISceneNode *node,
```
ClobberXD: Why is the scene node pointer passed as an argument? Wouldn't `GenericCAO::getSceneNode` work?

https://github.com/minetest/minetest/blob/4291961e09db7851fbc940eac58a1b919668d935/src/client/content_cao.cpp#L420-L438

Df458:
What's the status of this PR?

From a quick glance it looks to me like there isn't much work left to make this release-ready (a rendering bug and some tweaks to the light strength) but there hasn't been any activity since last summer. I'd hate to see a promising feature like this die, does this need adoption?

Unarelith:
@Df458 I just rebased this PR because I thought it could be cool to have this feature, but I never checked how it works and if there's some changes to make then I won't be able to make them.

So yep, if you want to adopt it, you can.

paramat:
Replaced by #9272

PR #7989, opened 2018-12-16
Fix entity rotation in existing worlds

SmallJoker:


pgimeno:
It fails to preserve the Z rotation, due to the change I noted in my review. It's saving Y twice.

https://github.com/SmallJoker/minetest/blob/47b55adff26c08fb17d15fb1d5296c20d8b5d0e7/src/content_sao.cpp#L607-L613

```c++
	writeF1000(os, m_rotation.Y);

	// version2. Increase this variable for new values
	writeU8(os, 1); // PROTOCOL_VERSION >= 37

	writeF1000(os, m_rotation.X);
	writeF1000(os, m_rotation.Y);
```

pgimeno:
Off topic, but, were my comments not visible until I pressed "Submit review"? That's not very intuitive. I thought I was adding comments to the code as always, that would be immediately visible. I didn't want to send a separate comment, so I didn't click "Submit review".

SmallJoker:
@pgimeno No, you need to press the "Submit review" button in order to publish your comments. Although, you can also comment one-by-one, so they're displayed instantly.

PR #8003, opened 2018-12-19
Android: Fix start-up crashes on version 4.x

stujones11:
Now that we are using a shared STL we must include and load the library. It would seem that `libc++_shared.so` is already included from around android 5 onwards, however, it is not present on earlier versions resulting in a start-up crash.
```
     Caused by: java.lang.UnsatisfiedLinkError: Cannot load library: soinfo_link_image(linker.cpp:1635): could not load library "libc++_shared.so" needed by "libiconv.so"; caused by load_library(linker.cpp:745): library "libc++_shared.so" not found
```

This PR now also includes a fix for the following error reported in #7994. I will let the developers decide whether to squash these commits or not.
```
E/AndroidRuntime(15502): Caused by: java.lang.NumberFormatException: Invalid int: "res/drawable/bg.xml"
```

Tested on Android 4.2.1 (Jelly Bean)

game

It's actually surprisingly playable on something that only cost me about 60 quid (bnib) 5 years ago :)

Test APK available here:  [Minetest-java-fix.apk](https://www.dropbox.com/s/8d8ghg38xx03oj0/Minetest-java-fix.apk?dl=0) @Fixer-007 @niansa @MoNTE48 

Edit: I just realized that the commit actually reads as Load standard template library library, nvm :D

### DIFF
--- a/build/android/src/main/java/net.minetest.minetest/MtNativeActivity.java
+++ b/build/android/src/main/java/net.minetest.minetest/MtNativeActivity.java
@@ -10,6 +10,7 @@ import android.view.WindowManager;
 public class MtNativeActivity extends NativeActivity {
 
 	static {
+		System.loadLibrary("c++_shared");
 		System.loadLibrary("openal");
 		System.loadLibrary("ogg");
 		System.loadLibrary("vorbis");
--- a/build/android/src/main/res/values-v21/styles.xml
+++ b/build/android/src/main/res/values-v21/styles.xml
@@ -4,7 +4,7 @@
 	<style name="AppTheme" parent="@android:style/android:Theme.Material.Light.NoActionBar.Fullscreen">
 		<item name="android:windowNoTitle">true</item>
 		<item name="android:windowAnimationStyle">@null</item>
-		<item name="android:background">@drawable/bg</item>
+		<item name="android:windowBackground">@drawable/bg</item>
 	</style>
 
 	<style name="Theme.Dialog" parent="@android:style/Theme.Material.Light.Dialog.NoActionBar"/>
--- a/build/android/src/main/res/values/styles.xml
+++ b/build/android/src/main/res/values/styles.xml
@@ -4,7 +4,7 @@
 	<style name="AppTheme" parent="@android:style/android:Theme.Holo.Light.NoActionBar.Fullscreen">
 		<item name="android:windowNoTitle">true</item>
 		<item name="android:windowAnimationStyle">@null</item>
-		<item name="android:background">@drawable/bg</item>
+		<item name="android:windowBackground">@drawable/bg</item>
 	</style>
 
 	<style name="Theme.Dialog" parent="@android:style/android:Theme.Holo.Light.Dialog.NoActionBar"/>


nerzhul:
I test it at home asap.

Fixer-007:
Still blue-screens for me on 4.4.2

Fixer-007:
It just shows blue everything and then closes with "This app stopped working" blah-blah. How can I get a backtrace?

Fixer-007:
It just shows blue everything for 1 sec and then closes with "This app stopped working" blah-blah. How can I get a backtrace?

Fixer-007:
Will try

stujones11:
> Will try

On second thoughts, just launching the app from the adb shell will likely yield more concise results. You will, of course, need to enable usb debugging on your device.

stujones11:
Now confirmed to be working on Android 4.4.2 (KitKat) https://github.com/minetest/minetest/issues/7994#issuecomment-449475300

@nerzhul To fix the 0.4.17.x release you should first backport f70f7875e2a2e49f80470832ae97e73f171d6460 then this, once merged ;)

niansa:
Oh I will try that out, wait...
It does not work, first black screen, then "About to give-up screen" in logcat. :(

stujones11:
@niansa There was an additional fix since the APK linked above. Please try the one I linked here https://github.com/minetest/minetest/issues/7994#issuecomment-449456606 instead.

niansa:
Okay wait...
Btw, it booted into recovery. ^^

nerzhul:
i think it can be due to the permission check which is different on 5.x and earlier

niansa:
Okay...

stujones11:
> Now it seems to take infinite time to install.

Indeed, I would not expect great performance on that device.

> But now it crashes at that screen...

My guess is that it ran out of memory. Supporting older android versions is one thing but older hardware is another.

niansa:
It didn't. I checked it via ADB.

niansa:
`shell@android:/ $ free
             total       used       free     shared    buffers     cached
Mem:        455744     359264      96480          0        492      12068`

stujones11:
> i think it can be due to the permission check which is different on 5.x and earlier

It is working fine for me on android 4.2.1 and for Fixer on 4.4.2

niansa:
I have android 4.1.2

nerzhul:
FYI i just published a release for the future 0.4.17.2 without this PR: https://github.com/minetest/minetest/releases/download/0.4.17.2/Minetest-0.4.17.2-debug.apk 

Can you confirm it works like without this PR ?

niansa:
That happens to the RAM if I start Minetest: https://youtu.be/EQAV7kDSpos

nerzhul:
512MB of memory is definitively too low to run android & minetest

stujones11:
> Can you confirm it works like without this PR ?

Working for me on Android 8, except I spawned right on the face of a massive cliff :(

screen

niansa:
Okay Version 0.4.17.2 just crashed at black screen.

niansa:
But there IS free memory.

stujones11:
> Okay Version 0.4.17.2 just crashed at black screen.

This is because it does not contain this fix.

niansa:
Yes. @nerzhul wanted to know what happens if I try that Version.

MoNTE48:
Minetest can start on 256MB RAM. And work on 512. 
@niansa you can copy full logcat on gist?

niansa:
@MoNTE48 Yes I can... Wait. (By the way, Minetest works on my old E-Reader with 512mb RAM...)

niansa:
Locat: https://gist.github.com/niansa/f1d7747788363d8a2e94eaef72219a70

nerzhul:
i just wanted to be sure. This fix will be backported to stable when ready

MoNTE48:
@niansa I do not see any problem or crash. This is bound to work... Stupid Samsung :)

stujones11:
> This fix will be backported to stable when ready

@nerzhul The PR is ready now, whatever the problem is for @niansa it not related to what this PR fixes. These are fairly trivial and obvious patches.

nerzhul:
Merging

nerzhul:
@niansa can you retry the newly updated https://github.com/minetest/minetest/releases/download/0.4.17.3/Minetest-0.4.17.2-debug.apk ?

niansa:
Okay wait...

niansa:
IT WORKS!!! Thank you so much! I have to wait until performance has stabilised before I start Minetest but it works at pretty stable 28 FPS!

niansa:
@nerzhul It works but it is VERY buggy:
1. To start the app I have to: Uninstall minetest, Install it again and tap the open button on the  installation successfull page. If I try to start Minetest from the launcher it crashes at preparing media.
2. The following thing happes after I write any mesage into the chat: I can see everything as usual but all touch buttons are hidden and I can not move in any way. I can not even leave the app using the home button.

niansa:
I found one issue more. Somehow the Minetest start screen totally breaks the picture of my camera if I film it. And that is definitely not an issue of my camera...

PR #8013, opened 2018-12-22
Fix various bugs (Anticheat, Lua helpers)

SmallJoker:
Anticheat: Use camera position instead of player position for shoot line calculations
 ^ was partially also my own fault
Lua helpers: Increase 'i' to not overwrite earlier added table values


src/network/serverpackethandler.cpp
```diff
@@ -969,8 +969,9 @@ bool Server::checkInteractDistance(RemotePlayer *player, const f32 d, const std:
 	else if (max_d < 0)
 		max_d = BS * 4.0f;
 
-	// cube diagonal: sqrt(3) = 1.732
-	if (d > max_d * 1.732) {
+	// Cube diagonal: sqrt(3) = 1.732
+	// Round up to avoid false-positives due to lag
```
sfan5: is this a theoretical concern or have there been complaints about this?

nerzhul:
I'm okay with all points, except the @sfan5 comment. If you can remove this part i will add my approval for other points

src/network/serverpackethandler.cpp
```diff
@@ -969,8 +969,8 @@ bool Server::checkInteractDistance(RemotePlayer *player, const f32 d, const std:
 	else if (max_d < 0)
 		max_d = BS * 4.0f;
 
-	// cube diagonal: sqrt(3) = 1.732
-	if (d > max_d * 1.732) {
+	// Cube diagonal: sqrt(3) = 1.732
+	if (d > max_d + 1.732f * BS) {
```
sfan5: I don't think this is correct.

The idea here is that the distance is up to `sqrt(3)` **times** longer when measuring through the cube diagonal, as opposed to a straight line (which has a length of `d`).

SmallJoker: If the node uses a larger selection box, then this check adds another half node tolerance in terms of allowed interact distance. `max_d` and `d` are both the radius of the allowed interact area.

sfan5: > another

where does the cube diagonal tolerance go then?

> half node

`sqrt(3)` is the diagonal of a full node

src/network/serverpackethandler.cpp
```diff
@@ -969,8 +969,9 @@ bool Server::checkInteractDistance(RemotePlayer *player, const f32 d, const std:
 	else if (max_d < 0)
 		max_d = BS * 4.0f;
 
-	// cube diagonal: sqrt(3) = 1.732
-	if (d > max_d * 1.732) {
+	// Cube diagonal * 1.5 for maximal supported node extents:
+	// sqrt(3) * 1.5 ≅ 2.6
+	if (d > max_d + 2.6f * BS) {
```
sfan5: And the `* 1.732`? Why is it gone?

ClobberXD: 2.6 =~= 1.732 * 1.5

sfan5: The `* 1.732` was there to account for a block diagonally `d` nodes away, making the distance `sqrt(3) * d`. The `+ 2.6` quite obviously won't be able to cover this.

Here's a (terrible) drawing illustrating the problem in the 2D space:
unbenannt

sfan5: After some testing it seems this isn't actually a problem in practice, though I'm sure it was added for this purpose.

SmallJoker: Demo

PR #8019, opened 2018-12-25
 Use true pitch/yaw/roll rotations without loss of precision by pgimeno

pauloue:
From https://gitlab.com/pgimeno/minetest/merge_requests/3/

> Store the rotation in the node as a 4x4 transformation matrix internally (through IDummyTransformationSceneNode), which allows more manipulations without losing precision or having gimbal lock issues.
>
> Network rotation is still transmitted as Eulers, though, not as matrix. But it will stay this way in 5.0.
>
> Fixes https://github.com/minetest/minetest/issues/7927

EDIT by paramat: Applies implementation discussed from https://github.com/minetest/minetest/issues/7927#issuecomment-445813318 onwards.
Details https://github.com/minetest/minetest/issues/7927#issuecomment-447430248

### DIFF
--- a/src/client/content_cao.cpp
+++ b/src/client/content_cao.cpp
@@ -647,8 +647,7 @@ void GenericCAO::addToScene(ITextureSource *tsrc)
 	} else if (m_prop.visual == "mesh") {
 		infostream<<"GenericCAO::addToScene(): mesh"<<std::endl;
 		scene::IAnimatedMesh *mesh = m_client->getMesh(m_prop.mesh, true);
-		if (mesh)
-		{
+		if (mesh) {
 			m_matrixnode = RenderingEngine::get_scene_manager()->
 				addDummyTransformationSceneNode(nullptr);
 			m_matrixnode->grab();
--- a/src/client/content_cao.cpp
+++ b/src/client/content_cao.cpp
@@ -507,6 +507,7 @@ void GenericCAO::removeFromScene(bool permanent)
 		m_spritenode->drop();
 		m_spritenode = nullptr;
 	}
+
 	if (m_matrixnode) {
 		m_matrixnode->remove();
 		m_matrixnode->drop();
--- a/src/client/content_cao.cpp
+++ b/src/client/content_cao.cpp
@@ -406,9 +406,6 @@ v3f GenericCAO::getPosition()
 		if (m_matrixnode)
 			return m_matrixnode->getAbsolutePosition();
 
-		if (m_spritenode)
-			return m_spritenode->getAbsolutePosition();
-
 		return m_position;
 	}
 	return pos_translator.val_current;
@@ -541,8 +538,11 @@ void GenericCAO::addToScene(ITextureSource *tsrc)
 
 	if (m_prop.visual == "sprite") {
 		infostream<<"GenericCAO::addToScene(): single_sprite"<<std::endl;
+		m_matrixnode = RenderingEngine::get_scene_manager()->
+				addDummyTransformationSceneNode();
+		m_matrixnode->grab();
 		m_spritenode = RenderingEngine::get_scene_manager()->addBillboardSceneNode(
-				NULL, v2f(1, 1), v3f(0,0,0), -1);
+				m_matrixnode, v2f(1, 1), v3f(0,0,0), -1);
 		m_spritenode->grab();
 		m_spritenode->setMaterialTexture(0,
 				tsrc->getTextureForMesh("unknown_node.png"));
@@ -785,12 +785,11 @@ void GenericCAO::updateNodePos()
 		v3s16 camera_offset = m_env->getCameraOffset();
 		v3f pos = pos_translator.val_current -
 				intToFloat(camera_offset, BS);
+		getPosRotMatrix().setTranslation(pos);
 		if (node != m_spritenode) { // rotate if not a sprite
 			v3f rot = m_is_local_player ? -m_rotation : -rot_translator.val_current;
 			setPitchYawRoll(getPosRotMatrix(), rot);
-			getPosRotMatrix().setTranslation(pos);
-		} else
-			node->setPosition(pos); // sprites have no m_matrixnode
+		}
 	}
 }
 
--- a/src/client/content_cao.cpp
+++ b/src/client/content_cao.cpp
@@ -365,7 +365,7 @@ void GenericCAO::processInitData(const std::string &data)
 		return;
 	}
 
-	// PROTOCOL_VERSION >= 37 
+	// PROTOCOL_VERSION >= 37
 	m_name = deSerializeString(is);
 	m_is_player = readU8(is);
 	m_id = readU16(is);
@@ -402,7 +402,7 @@ bool GenericCAO::getSelectionBox(aabb3f *toset) const
 
 v3f GenericCAO::getPosition()
 {
-	if (getParent() != NULL) {
+	if (getParent() != nullptr) {
 		if (m_matrixnode)
 			return m_matrixnode->getAbsolutePosition();
 
@@ -625,7 +625,7 @@ void GenericCAO::addToScene(ITextureSource *tsrc)
 		// Set it to use the materials of the meshbuffers directly.
 		// This is needed for changing the texture in the future
 		m_meshnode->setReadOnlyMaterials(true);
-	} else if(m_prop.visual == "cube") {
+	} else if (m_prop.visual == "cube") {
 		infostream<<"GenericCAO::addToScene(): cube"<<std::endl;
 		scene::IMesh *mesh = createCubeMesh(v3f(BS,BS,BS));
 		m_matrixnode = RenderingEngine::get_scene_manager()->
@@ -644,10 +644,10 @@ void GenericCAO::addToScene(ITextureSource *tsrc)
 		m_meshnode->setMaterialFlag(video::EMF_BILINEAR_FILTER, false);
 		m_meshnode->setMaterialType(material_type);
 		m_meshnode->setMaterialFlag(video::EMF_FOG_ENABLE, true);
-	} else if(m_prop.visual == "mesh") {
+	} else if (m_prop.visual == "mesh") {
 		infostream<<"GenericCAO::addToScene(): mesh"<<std::endl;
 		scene::IAnimatedMesh *mesh = m_client->getMesh(m_prop.mesh, true);
-		if(mesh)
+		if (mesh)
 		{
 			m_matrixnode = RenderingEngine::get_scene_manager()->
 				addDummyTransformationSceneNode(nullptr);
@@ -671,8 +671,7 @@ void GenericCAO::addToScene(ITextureSource *tsrc)
 			m_animated_meshnode->setMaterialFlag(video::EMF_FOG_ENABLE, true);
 			m_animated_meshnode->setMaterialFlag(video::EMF_BACK_FACE_CULLING,
 				m_prop.backface_culling);
-		}
-		else
+		} else
 			errorstream<<"GenericCAO::addToScene(): Could not load mesh "<<m_prop.mesh<<std::endl;
 	} else if (m_prop.visual == "wielditem" || m_prop.visual == "item") {
 		ItemStack item;
--- a/src/client/content_cao.cpp
+++ b/src/client/content_cao.cpp
@@ -23,7 +23,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include <IMeshManipulator.h>
 #include <IAnimatedMeshSceneNode.h>
 #include "content_cao.h"
-#include "util/numeric.h" // For IntervalLimiter
+#include "util/numeric.h" // For IntervalLimiter & setPitchYawRoll
 #include "util/serialize.h"
 #include "util/basic_macros.h"
 #include "client/sound.h"
@@ -403,9 +403,11 @@ bool GenericCAO::getSelectionBox(aabb3f *toset) const
 v3f GenericCAO::getPosition()
 {
 	if (getParent() != NULL) {
-		scene::ISceneNode *node = getSceneNode();
-		if (node)
-			return node->getAbsolutePosition();
+		if (m_matrixnode)
+			return m_matrixnode->getAbsolutePosition();
+
+		if (m_spritenode)
+			return m_spritenode->getAbsolutePosition();
 
 		return m_position;
 	}
@@ -486,7 +488,7 @@ void GenericCAO::removeFromScene(bool permanent)
 
 		LocalPlayer* player = m_env->getLocalPlayer();
 		if (this == player->parent) {
-			player->parent = NULL;
+			player->parent = nullptr;
 			player->isAttached = false;
 		}
 	}
@@ -494,24 +496,29 @@ void GenericCAO::removeFromScene(bool permanent)
 	if (m_meshnode) {
 		m_meshnode->remove();
 		m_meshnode->drop();
-		m_meshnode = NULL;
+		m_meshnode = nullptr;
 	} else if (m_animated_meshnode)	{
 		m_animated_meshnode->remove();
 		m_animated_meshnode->drop();
-		m_animated_meshnode = NULL;
+		m_animated_meshnode = nullptr;
 	} else if (m_wield_meshnode) {
 		m_wield_meshnode->remove();
 		m_wield_meshnode->drop();
-		m_wield_meshnode = NULL;
+		m_wield_meshnode = nullptr;
 	} else if (m_spritenode) {
 		m_spritenode->remove();
 		m_spritenode->drop();
-		m_spritenode = NULL;
+		m_spritenode = nullptr;
+	}
+	if (m_matrixnode) {
+		m_matrixnode->remove();
+		m_matrixnode->drop();
+		m_matrixnode = nullptr;
 	}
 
 	if (m_nametag) {
 		m_client->getCamera()->removeNametag(m_nametag);
-		m_nametag = NULL;
+		m_nametag = nullptr;
 	}
 }
 
@@ -608,17 +615,24 @@ void GenericCAO::addToScene(ITextureSource *tsrc)
 			mesh->addMeshBuffer(buf);
 			buf->drop();
 		}
-		m_meshnode = RenderingEngine::get_scene_manager()->addMeshSceneNode(mesh, NULL);
+		m_matrixnode = RenderingEngine::get_scene_manager()->
+			addDummyTransformationSceneNode();
+		m_matrixnode->grab();
+		m_meshnode = RenderingEngine::get_scene_manager()->
+			addMeshSceneNode(mesh, m_matrixnode);
 		m_meshnode->grab();
 		mesh->drop();
 		// Set it to use the materials of the meshbuffers directly.
 		// This is needed for changing the texture in the future
 		m_meshnode->setReadOnlyMaterials(true);
-	}
-	else if(m_prop.visual == "cube") {
+	} else if(m_prop.visual == "cube") {
 		infostream<<"GenericCAO::addToScene(): cube"<<std::endl;
 		scene::IMesh *mesh = createCubeMesh(v3f(BS,BS,BS));
-		m_meshnode = RenderingEngine::get_scene_manager()->addMeshSceneNode(mesh, NULL);
+		m_matrixnode = RenderingEngine::get_scene_manager()->
+			addDummyTransformationSceneNode(nullptr);
+		m_matrixnode->grab();
+		m_meshnode = RenderingEngine::get_scene_manager()->
+			addMeshSceneNode(mesh, m_matrixnode);
 		m_meshnode->grab();
 		mesh->drop();
 
@@ -630,14 +644,16 @@ void GenericCAO::addToScene(ITextureSource *tsrc)
 		m_meshnode->setMaterialFlag(video::EMF_BILINEAR_FILTER, false);
 		m_meshnode->setMaterialType(material_type);
 		m_meshnode->setMaterialFlag(video::EMF_FOG_ENABLE, true);
-	}
-	else if(m_prop.visual == "mesh") {
+	} else if(m_prop.visual == "mesh") {
 		infostream<<"GenericCAO::addToScene(): mesh"<<std::endl;
 		scene::IAnimatedMesh *mesh = m_client->getMesh(m_prop.mesh, true);
 		if(mesh)
 		{
+			m_matrixnode = RenderingEngine::get_scene_manager()->
+				addDummyTransformationSceneNode(nullptr);
+			m_matrixnode->grab();
 			m_animated_meshnode = RenderingEngine::get_scene_manager()->
-				addAnimatedMeshSceneNode(mesh, NULL);
+				addAnimatedMeshSceneNode(mesh, m_matrixnode);
 			m_animated_meshnode->grab();
 			mesh->drop(); // The scene node took hold of it
 			m_animated_meshnode->animateJoints(); // Needed for some animations
@@ -674,8 +690,12 @@ void GenericCAO::addToScene(ITextureSource *tsrc)
 			infostream << "serialized form: " << m_prop.wield_item << std::endl;
 			item.deSerialize(m_prop.wield_item, m_client->idef());
 		}
+		m_matrixnode = RenderingEngine::get_scene_manager()->
+			addDummyTransformationSceneNode(nullptr);
+		m_matrixnode->grab();
 		m_wield_meshnode = new WieldMeshSceneNode(
 			RenderingEngine::get_scene_manager(), -1);
+		m_wield_meshnode->setParent(m_matrixnode);
 		m_wield_meshnode->setItem(item, m_client,
 			(m_prop.visual == "wielditem"));
 
@@ -763,11 +783,14 @@ void GenericCAO::updateNodePos()
 
 	if (node) {
 		v3s16 camera_offset = m_env->getCameraOffset();
-		node->setPosition(pos_translator.val_current - intToFloat(camera_offset, BS));
+		v3f pos = pos_translator.val_current -
+				intToFloat(camera_offset, BS);
 		if (node != m_spritenode) { // rotate if not a sprite
 			v3f rot = m_is_local_player ? -m_rotation : -rot_translator.val_current;
-			node->setRotation(rot);
-		}
+			setPitchYawRoll(getPosRotMatrix(), rot);
+			getPosRotMatrix().setTranslation(pos);
+		} else
+			node->setPosition(pos); // sprites have no m_matrixnode
 	}
 }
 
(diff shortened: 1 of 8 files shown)


pgimeno:
Thanks for submitting it!

paramat:
Thanks.

stujones11:
This looks good to me and is pretty much the 'officially' endorsed solution to this common problem. While doing this is fine for your main game actor, my only minor concern would be the creation of an additional dummy scene node for every single game object.

I doubt that it will be a problem but we really should try to get some numbers on this.

stujones11:
>  I don't have a particularly slow computer

I would consider a better test to be spawn 1000+ objects, with and without this PR and compare memory usage and other vital stats.

paramat:
Tested with my original boats patch https://github.com/minetest/minetest/pull/7395#issuecomment-426861772
Rotation for the boat is now an intuitive 'intrinsic' yaw-pitch-roll.
Left-right for yaw, forward-back for pitch and jump-sneak for roll.

paramat:
As the large majority of objects in MT will only use yaw (dropped items, most mobs), might there be benefit (would it even be possible?) in dividing objects into yaw-only and yaw-pitch-roll versions? Adding a new object property to select between those.

stujones11:
> dividing objects into yaw-only and yaw-pitch-roll versions?

Although rather messy, we could simply have a boolean property to enable 6d rotation and only add the matrix node for those objects. Hopefully, this should not be necessary though.

pgimeno:
Yes, it would indeed get messy. A bunch of new methods would be necessary to abstract the operations that are currently performed sometimes on the object scene node and sometimes on the transformation node, so that they can work both with and without the latter transparently. That would make the patch larger and more invasive.

Considering that at some point, Irrlicht will fix this on their side (hopefully in not many years), keeping the patch as simple as possible will help reverting the changes when that happens.

As for memory and other vital stats, I'll defer testing to whoever knows what and how to measure.

src/util/numeric.cpp
```diff
@@ -174,3 +174,38 @@ s16 adjustDist(s16 dist, float zoom_fov)
 	return std::round(dist * std::cbrt((1.0f - std::cos(threshold_fov)) /
 		(1.0f - std::cos(zoom_fov / 2.0f))));
 }
+
+void setPitchYawRollRad(core::matrix4 &m, const v3f &rot)
```
nerzhul: please provide unittests for this function

src/util/numeric.cpp
```diff
@@ -174,3 +174,38 @@ s16 adjustDist(s16 dist, float zoom_fov)
...
+	M[4] = c1 * s2 * s3 - s1 * c3;
+	M[5] = c1 * c2;
+	M[6] = c1 * s2 * c3 + s1 * s3;
+
+	M[8] = c2 * s3;
+	M[9] = -s2;
+	M[10] = c2 * c3;
+}
+
+v3f getPitchYawRollRad(const core::matrix4 &m)
```
nerzhul: please provide unittests for this function

src/util/numeric.h
```diff
@@ -417,3 +418,17 @@ inline void wrappedApproachShortest(T &current, const T target, const T stepsize
 		current = target;
 	}
 }
+
+void setPitchYawRollRad(core::matrix4 &m, const v3f &rot);
+
+inline void setPitchYawRoll(core::matrix4 &m, const v3f &rot)
```
nerzhul: please provide unittests for this function

src/util/numeric.h
```diff
@@ -417,3 +418,17 @@ inline void wrappedApproachShortest(T &current, const T target, const T stepsize
...
+{
+	setPitchYawRollRad(m, rot * core::DEGTORAD64);
+}
+
+v3f getPitchYawRollRad(const core::matrix4 &m);
+
+inline v3f getPitchYawRoll(const core::matrix4 &m)
```
nerzhul: please provide unittests for this function

pgimeno:
Added unit tests. @pauloue could you fetch/push please?

pauloue:
There was a conflict in `networkprotocol.h`. That's the reason for the force push. If that's a problem you can do the rebase on your branch.

pgimeno:
Thanks, yes I had a rebased version ready locally, waiting for someone to ask me to rebase. I wanted to remove a line there actually, so I've force-pushed my gitlab branch. You may need to delete your local branch before fetching, not sure.

src/client/content_cao.cpp
```diff
@@ -630,14 +644,16 @@ void GenericCAO::addToScene(ITextureSource *tsrc)
 		m_meshnode->setMaterialFlag(video::EMF_BILINEAR_FILTER, false);
 		m_meshnode->setMaterialType(material_type);
 		m_meshnode->setMaterialFlag(video::EMF_FOG_ENABLE, true);
-	}
-	else if(m_prop.visual == "mesh") {
+	} else if(m_prop.visual == "mesh") {
```
nerzhul: space missing after if

src/client/content_cao.cpp
```diff
@@ -630,14 +644,16 @@ void GenericCAO::addToScene(ITextureSource *tsrc)
...
 		infostream<<"GenericCAO::addToScene(): mesh"<<std::endl;
 		scene::IAnimatedMesh *mesh = m_client->getMesh(m_prop.mesh, true);
 		if(mesh)
 		{
```
nerzhul: can you fix that conditional code style as you modified the 3 lines after ?

paramat:
One thing i need to test is using a "cube" or "item" visual entity and checking the correct face is treated as the 'front'.

stujones11:
I have done a little testing and have observed some unusual results from the matrix <-> Euler conversion methods. Notice that if I replace the the unittest test values with the following, the unittest then fails.

```
	v1 = v3f(M_PI/1.5f, 0.f, 0.f);
	v2 = v3f(120.0f, 0.f, 0.f);
```
Any idea why that would be?

pgimeno:
That's right; you can use any angles for `setPitchYawRoll`, but currently the pitch returned by `getPitchYawRoll` is always in the range [-90°, 90°] while yaw and roll have a [-180°, 180°] range (and similarly for `getPitchYawRollRad` but in radians). The unit test deliberately avoids setting a pitch greater than 90°. That is expected behaviour. Think of pitch as changing latitude (i.e. vertical movement) and yaw as changing longitude (horizontal movement) in an earth globe; that restriction means you can't go past the pole to the other side, but you can go to any longitude, so you can be in any point of the sphere.

It's possible to make a `getPitchYawRoll` whose pitch range is [-180°, 180°] and the yaw range is [-90°, 90°] (you can only move 90° left or right; if you want to go to the other side you have to cross through the pole), but that's less intuitive. It's not possible to make a `getPitchYawRoll` that returns all angles in [-180°, 180°] with the same values that were set, because some of these rotations are equivalent as you have noted.

The unit test in fact has a test with a greater angle, and a check that is commented out because it is not satisfied by the current code, due to the restriction applying to the pitch. I was going to check that it was not the same, but then I thought that if `getPitchYawRollRad` was modified to restrict an angle other than pitch, that would make the test fail without reason.

stujones11:
@pgimeno Thank you for the detailed explanation, I have worked on similar code in the past and it was my memory of such problems that prompted me to make these tests. I still don't fully understand why the Lua methods don't suffer the same problem, I can only guess that the getter method uses a different transposition of the matrix than the setter but all that really matters is that it works :)

Regarding the matrix to euler conversion, I did come across a much simpler version that does seemingly account for singularities at the poles. I am not sure how well it performs compared to your version but it seemed to return pretty much same results in the few tests I have made.

Source: https://stackoverflow.com/questions/1996957/conversion-euler-to-matrix-and-matrix-to-euler

Here it is transposed to your matrix format (pseudo code) 
```c++
	float x = asinf(-M[9]);
	float y = 0.f, z = 0.f;
	if (cosf(rot.x) > 0.0001) {
		y = atan2f(M[8], M[10]);
		z = atan2f(M[1], M[5]);
	}
	else {
		z = atan2f(M[4], M[0]);
	}
```

pgimeno:
@stujones11 IIRC Lua gets/sets a vector, and whatever it stores, it gets back. The values are interpreted as ZXY Eulers only when they get to the client. `getPitchYawRol[Rad]` is currently unused, but it will be necessary in order to implement relative rotations in the way suggested in https://github.com/minetest/minetest/issues/7927#issuecomment-444786523, so I thought I'd add an implementation.

As for the conversion function you found, on first sight it has two problems. One is the use of `asinf`. Near 90°, `asinf` has a precision of around 0.02° which is somewhat poor; you want to use `atan2f` where possible. The other is that when crossing the 89.994° (approx) frontier (the value of acos(0.0001)), the y angle jumps suddenly to zero, which is bad for interpolation; the formulation I gave does that only if atan2f gets a zero input vector, which can only happen exactly in the pole, and I think it can't actually happen (because as I already noted, 90° can't be represented exactly in radians).

Probably neither of these two problems will be noticeable when doing relative rotations for things like vehicles, though. But relative rotations scare me a bit precision wise for objects being rotated always on the same axis, like fans. In the long term, they will drift no matter the method used.

stujones11:
> but it will be necessary in order to implement relative rotations

Surely you would simply build a new matrix from the specified rotation and multiply that with your current rotation matrix? I fail to see how that function would be helpful anyway, especially given its limitations but I'll take your word for it, you seem to know what you are doing.

From my own humble attempt to create a semi-realistic flight simulation with this implementation, I am quickly realizing that to do it properly, having access to the matrix itself would be far more helpful than this weird vector of euler radians. Even a simple unit direction vector would be easier to work with and could have possibly saved a lot of trouble.

> for objects being rotated always on the same axis, like fans. In the long term, they will drift no matter the method used.

Objects that rotate constantly would be much better done using an animated model, for a fan, I would consider entity rotation a waste of resources.

pgimeno:
> Surely you would just build a new matrix from the specified rotation and multiply that with your current rotation matrix?

Which rotation matrix? :D

In the server, you need `setPitchYawRollRad` to actually get a rotation matrix from the Eulers. You can then do the matrix operations you want, and then you convert them back to Eulers with `getPitchYawRollRad` so the rotation is again in the internal storage format, which is also the network protocol format. That's why I implemented both.

Definitely, having the rotation in a math-friendly format (i.e. matrix or quaternion) would be a huge gain. But such patch would be more invasive, affecting the server as well as the client. Maybe I submit a patch for 6.0 when MT moves out of MS-GH ;)

Hmm... now that we're in feature freeze it's too late to add `get_matrix()` to the SAO's methods. But that can be done for 5.1.

paramat:
screenshot_20190110_022734

Here i am, sailing my furnace ..

Seems correct, as the holes of the furnace are on the 'back' of the node.
(When using 'visual = "wielditem"' the player is scaled very small for some reason.)

pgimeno:
@paramat 

> (When using 'visual = "wielditem"' the player is scaled very small for some reason.)

Is that a bug of the patch or does it happen also without it?

paramat:
Don't worry, also happens without your patch.

paramat:
screenshot_20190110_024519

More sillyness, but again tests correctly, rotations are aligned correctly with the 'front' texture of 'visual = "cube"'
```
	initial_properties = {
		physical = true,
		collisionbox = {-0.5, -0.35, -0.5, 0.5, 0.3, 0.5},
		visual = "cube",
		visual_size = {x = 1, y = 1},
		textures = { -- top base rightside leftside front back
			"mesecar_car1top.png",
			"mesecar_carbase.png",
			"mesecar_car1rightside.png",
			"mesecar_car1leftside.png",
			"mesecar_car1front.png",
			"mesecar_car1back.png",
		},
	},
```

stujones11:
I have done a little more testing and discovered that my model is not always rotating around the correct local axis. It seems to start out okay at first, however, after a series of different rotations thing seem to get out of kilter.

See the short video clip I made here: https://vimeo.com/311739154

Take note of the movement and rotation vectors dumped to the chat.

pgimeno:
I see nothing wrong in that video. Euler angles work that way, they define an absolute rotation. The only angle that appears to be local is roll (z angle).

Yaw acts like an always vertical pin (aligned with the Y axis) going through the centre of the object, no matter its orientation. That's "extrinsic" (absolute) rotation. Roll acts like a pin stuck in the object's nose. That's "intrinsic" (relative) rotation. Pitch is neither intrinsic nor extrinsic.

When pitch is nearly vertical (i.e. near ±90°), yaw and roll are almost the same, as seen in the video. When pitch is exactly vertical, yaw and roll do the same (that's what's called "gimbal lock").

Relative rotation for all axes is a planned feature, but not yet implemented; `getPitchYawRoll` was implemented to make it possible. See comment in https://github.com/minetest/minetest/issues/7927#issuecomment-444786523

stujones11:
> I see nothing wrong in that video. Euler angles work that way

I was under the impression that this PR was supposed to remove this limitation but I obviously misunderstood. Maybe this does fix does fix some issues but it's still completely unusable for my purposes :'(

pgimeno:
@stujones11 You can try this Lua version:

```lua
local function mul_mat3mat3(c, a, b)
  c[1], c[2], c[3],
  c[4], c[5], c[6],
  c[7], c[8], c[9]
    = a[1] * b[1] + a[2] * b[4] + a[3] * b[7]
    , a[1] * b[2] + a[2] * b[5] + a[3] * b[8]
    , a[1] * b[3] + a[2] * b[6] + a[3] * b[9]
    , a[4] * b[1] + a[5] * b[4] + a[6] * b[7]
    , a[4] * b[2] + a[5] * b[5] + a[6] * b[8]
    , a[4] * b[3] + a[5] * b[6] + a[6] * b[9]
    , a[7] * b[1] + a[8] * b[4] + a[9] * b[7]
    , a[7] * b[2] + a[8] * b[5] + a[9] * b[8]
    , a[7] * b[3] + a[8] * b[6] + a[9] * b[9]
end

local sin, cos, atan2, sqrt = math.sin, math.cos, math.atan2, math.sqrt

local function euler2matrix(rot)
  local a1, a2, a3 = rot.z, rot.x, rot.y
  local c1, s1 = cos(a1), sin(a1)
  local c2, s2 = cos(a2), sin(a2)
  local c3, s3 = cos(a3), sin(a3)

  return s1 * s2 * s3 + c1 * c3, s1 * c2, s1 * s2 * c3 - c1 * s3,
         c1 * s2 * s3 - s1 * c3, c1 * c2, c1 * s2 * c3 + s1 * s3,
         c2 * s3,               -s2,      c2 * c3
end

local function matrix2euler(m00, m01, m02, m10, m11, m12, m20, m21, m22)
  -- matrix to ZXY
  -- calculate new Tait-Bryan angles in the XYZ order expected by Irrlicht
  local a1 = atan2(m01, m11)
  local c2 = sqrt(m22*m22 + m20*m20)
  local a2 = atan2(-m21, c2)

  -- Shoemake's original
--  local a3 = atan2(m20, m22)

  -- Mike Day's replacement
  -- https://www.semanticscholar.org/paper/Extracting-Euler-Angles-from-a-Rotation-Matrix-Day/668137fa4b875d890f446e689eea1e334bcf6bf6
  local c1, s1 = cos(a1), sin(a1)
  local a3 = atan2(s1*m12 - c1*m02, c1*m00 - s1*m10)

  return {x = a2, y = a3, z = a1}
end


local mat1 = {0,0,0,0,0,0,0,0,0}
local mat2 = {0,0,0,0,0,0,0,0,0}

local rotate = function(obj, angles)
  local x, y, z = angles.x, angles.y, angles.z
  mat1[1], mat1[2], mat1[3],
  mat1[4], mat1[5], mat1[6],
  mat1[7], mat1[8], mat1[9] = euler2matrix(obj:get_rotation())
  local rotated = false

  if y ~= 0 then
    local c, s = cos(y), sin(y)
    mat2[1], mat2[2], mat2[3] = c,  0,  s
    mat2[4], mat2[5], mat2[6] = 0,  1,  0
    mat2[7], mat2[8], mat2[9] =-s,  0,  c
    mul_mat3mat3(mat1, mat2, mat1)
    rotated = true
  end

  if x ~= 0 then
    local c, s = cos(x), sin(x)
    mat2[1], mat2[2], mat2[3] = 1,  0,  0
    mat2[4], mat2[5], mat2[6] = 0,  c, -s
    mat2[7], mat2[8], mat2[9] = 0,  s,  c
    mul_mat3mat3(mat1, mat2, mat1)
    rotated = true
  end

  if z ~= 0 then
    local c, s = cos(z), sin(z)
    mat2[1], mat2[2], mat2[3] = c, -s,  0
    mat2[4], mat2[5], mat2[6] = s,  c,  0
    mat2[7], mat2[8], mat2[9] = 0,  0,  1
    mul_mat3mat3(mat1, mat2, mat1)
    rotated = true
  end

  if rotated then
    obj:set_rotation(matrix2euler(mat1[1], mat1[2], mat1[3],
                                  mat1[4], mat1[5], mat1[6],
                                  mat1[7], mat1[8], mat1[9]))
  end
end

return rotate
```

Pass it an object to rotate, and a vector with the X angle being the local pitch, the Y angle being the local yaw, and the Z angle being the local roll. Best rotating 1 axis at a time. That's what the planned function is going to do.

If you set the object's `rotate` to the returned `rotate`, you can even do `obj:rotate(myrot)`.

pgimeno:
@stujones11 By the way, there are several reasons for this patch, already discussed in #7927. To sum them up, it's to make `get_yaw` work correctly when the Z angle passed to `set_rotation()` is not zero, to make roll work intuitively, and to avoid sticking with that breakage forever.

The Eulers implemented in this PR are easy, and more useful for your plane than you think. They are what the instruments would read. Yaw is what the compass would indicate, i.e. the horizontal component of where the nose is pointing. Roll is the artificial horizon's horizontal angle, or the bank indicator's angle. Pitch is the elevation indicated in the artificial horizon; basically the angle between the horizontal plane and where the nose is pointing.

stujones11:
@pgimeno Thanks for the code, the matrix to euler (with correct notational order) has been very helpful. However, I am doing my rotations using quaternions as it results in less code/calculations but don't quote me on that. I am no expert as you can probably tell.

FWIW I have pretty much tested this PR to it's limits, I do not see any performance deficit from the dummy scene nodes, no matter how many I spawn.

 In conclusion, this is absolutely correct way to do rotations in MT :+1:

> Good looking aircraft model.

Thanks, you can see it with a nicer texture here: https://vimeo.com/313476186

paramat:
How would we profile the use of the extra scene nodes? I don't expect it to be a large impact but it would be good to check.
Maybe spawn 100 entities in an identical world and take note of FPS?

src/client/content_cao.cpp
paramat: Still seems uncorrected.
@pauloue

pgimeno: @paramat @pauloue Fix pushed. I thought nerzhul was talking about the space after the 'if', sorry. My personal style is Allman-like, so I easily miss these things because they look natural to me.

pgimeno:
@paramat

> Maybe spawn 100 entities in an identical world and take note of FPS?

I did that and the difference was under the noise level, i.e. unnoticeable.

pgimeno:
It would be interesting to compare an Android build with and without this patch.

As for the snippet I posted, I got the sign convention wrong once again. I've edited the post, and here's a fixed version installed in the boats mod, that allows testing relative rotation: https://gitlab.com/pgimeno/minetest_game/blob/rel-rot/mods/boats/init.lua

I know that pitch usually works the opposite way :) I respected paramat's sign convention, as it's based on his modification to the boats mod used to test rotations. It's trivial to correct, by swapping + and - in the up/down control code.

@stujones11 Thanks a lot for your testing!

stujones11:
> It would be interesting to compare an Android build with and without this patch.

No problem, I will make a build from the latest commit tomorrow (late for me now) I would be interested to see how well my wip aircraft mod performs there, too :)

PS. I had noticed the sign switching but thought it must be a retro compat thing so didn't complain. It is easy enough for me to adjust that in my mod, as you say.

Thank you again for your work on this.

stujones11:
@pgimeno I tried to make an android build, however, there must have been some problem around the time you made this comment that prevents the libs from loading so it would need to rebased against current master. Unfortunately, my git skills are not quite up to doing that locally :-/

pgimeno:
Gah, I already rebased in GitLab before seeing your update :)

I don't have a build without it, so I can't compare. This is CPU-side processing, therefore the slower the CPU, the more noticeable the performance difference. That's why I thought of Android for testing.

stujones11:
> Gah, I already rebased in GitLab before seeing your update :)

I am sorry to waste your time, I truly doubt this will be a problem on any platform and Irrlicht is pretty lightweight to start with.

I will do some tests on android tomorrow, it is late again here for me now. However, I am fairly confident there will be no noticeable difference there either.

stujones11:
My tests on android were pretty much as I expected, no measurable difference

With PR:
screen_pr

Without PR:
screen

Note that the apk has been updated and now includes the test mod as well as MTG! sorry to anyone who tried the last one but it _was_ getting late when I made that,I also fixed a silly mistake in the test code =)

PS. You don't need to enable the test mod since I had to include it inside MTG, however, that's another issue.

paramat:
Thanks for the tests, i'm therefore not concerned about performance impact then.

pgimeno:
I've got the SD card I was waiting for, and spent most of my spare time today setting up and compiling both master and this branch in a RPi 3b, then preparing and performing a rigorous test using stujones11's mod.

#### Test procedure:

The tests were performed with my desktop as a server, so that the server side processing didn't affect measurements. I teleported to 2000m height to stay away from terrain and clouds. I created a cobble node using worldedit, to use as a reference for up/down. I looked around for a bit, for everything to be generated server-side.

I pointed away from the entities created, because the matrix calculations are performed even when not looking (the renderer can't know whether to exclude an entity until it knows where it's placed). The hotbar was empty (that affects FPS significantly, to my surprise), and I waited for the connection text to vanish. I only had in the screen the F5 debug text, the empty hotbar, the empty hand and the sky. time_speed was set to 0 in order for lighting changes to not interfere either. The sun was out of the field of view. The mouse was not moved at all between entering and pressing ESC to exit.

I raised the timer in your mod to 50 s, because entities took 5-15 s to appear and they often disappeared right after spawning, or not at all. While no entities were spawned, my FPS were at 60 (I couldn't disable vsync). During the tests, I assumed that a sudden drop indicated the spawning, and waited for it to stabilize.

#### Test results:

In the master branch I got a pretty steady 46 FPS. In this branch I got 41 FPS, occasionally 42. I assume that the 46 FPS result must be near the midpoint between 46 and 47, so I used 46.5 for the calculations, and the 41 FPS result must be near 42, so I used 41.8.

Based on these numbers, 256 entities add ~2.5 ms per frame, therefore each matrix node adds around 10 µs time per frame per entity.

#### Conclusions:

Such low-end CPUs are of course not expected to be equipped with a high-end GPU, therefore graphics rendering overwhelmingly dominates frame time, which matches your results as well. The above test procedure was the only way I found of being able to observe a difference. It's worth mentioning that using software rendering, I got 1 FPS or less while moving around normally, and with GPU, ~20 FPS or less. So, if one gets ~20 FPS in the presence of 256 entities, which I doubt, the extra 256 transformations would be taking ~20% of the frame time. If it's performed in parallel with the GPU's rendering, it most likely fits within the drawing time, and if it's a CPU renderer, the extra transformation's time is going to be negligible. I think that this conclusion is extendable to any CPU-GPU combo.

To sum up, yes, it slows down computing (no one expected otherwise), but the slowdown is negligible in comparison to the rendering time, so it's not going to be noticeable. Note the effort I had to make in order to make CPU time dominate.

----

@stujones11 

> I am very sorry to waste your time

I was more concerned about pauloue's time actually :) But I've reverted the GitLab branch to the same commit this one was at, so there's no need to bother him.

Thanks for your testing.

stujones11:
> setting up and compiling both master and this branch in a RPi 3b

Off-topic but just out of interest, did you manage to get that working with the ES drivers are are you using the 'experimental' openGL one? I have only ever managed the latter.

pgimeno:
'Experimental'. I haven't tried anything else. I just wanted a slow computer to test on :)

stujones11:
> Based on these numbers, 256 entities add ~2.5 ms per frame

I am not sure if changed the code but there should have been 1023 entities (slightly less in the original). I just thought I would point that out since you did not seem to mention it.

pgimeno:
Eep, you're right, so my numbers are off by a factor of almost 4. It's 1023 entities, not 256. I didn't pay enough attention, sorry :)

So, the actual results are: 1023 entities add ~2.5 ms per frame, therefore each matrix node adds around 2.4 µs time per frame per entity. And when I said "So, if one gets ~20 FPS in the presence of 256 entities, which I doubt, the extra 256 transformations..." I should have said: "So, if one gets ~20 FPS in the presence of 1023 entities, which I doubt, the extra 1023 transformations..."

Edited the post to match. I'll look into performing the other test you're suggesting.

pgimeno:
I've tested the dynamic updates under the same conditions. Without limiting the number of entities, the results were as follows: 42-44 FPS on master, 38-40 FPS on this branch. The numbers oscillated, which is expected due to interference between the frequencies of the server ticks and the client frames. As a consequence, this result is a bit less accurate. The extra calculation time is in this case 1/39 minus 1/44, approx. 2.9 ms for 1023 entities. Given the bigger measurement error, I think that sounds about the same. That matches my expectations.

The FPS I expected was actually worse; my calculation was 41.8/(41.8+11.1) = 33 FPS (all numbers are approximate). Perhaps the server isn't sending all updates to the client due not all entities being in front of the player. The camera wasn't in the same position as the previous test, even though it was in the same position during this testing; maybe that affected the results somehow too.

I'm satisfied with these results, so I'll consider it tested ant call it a day.

paramat:
Performance seems good, it tests well, what this implements is good and seems a widely accepted solution. Lot's of attention which i apreciate, so once i check codestyle i think all this counts for something and i will probably +1.
However, i do not *completely* understand the changes, so i will request that the core dev who gives the 2nd +1 does.

pgimeno:
In case it helps understanding the PR, here's a diagram of what the patch implements:

----
Without patch:
scenegraph-old

----
With patch:
scenegraph-new

paramat:
Looked through code :+1:

pgimeno:
@ClobberXD There seems to be a confusion. The code is designed to be able to work with MT's axis conventions. The convention you point out is that of the native Irrlicht `ISceneNode` rotations, which  caused #7989 (see there for details) and the whole problem that this PR is aimed to work around. Note how the `ISceneNode` rotation is kept at 0 with the patch, so it doesn't interfere with the different, Minetest-friendly convention introduced by `setPitchYawRoll()`.

I may have made a mistake, though. If so, could you please let me know which specific part of the code uses Z for up?

paramat:
> I notice that the rotations follow a different coordinate system (X: forward-backward, Y: left-right, Z: up-down)

That's a confusing way to refer to rotations, how can a rotation be 'forward-backward' and what does that mean?
Can you specify a part of this code you are concerned about?

ClobberXD:
@pgimeno Looks like I was indeed confused, and I misinterpreted stuff. Sorry for the trouble. Good work :+1: :)

src/client/content_cao.cpp
```diff
@@ -486,32 +485,37 @@ void GenericCAO::removeFromScene(bool permanent)
...
 		m_wield_meshnode->drop();
-		m_wield_meshnode = NULL;
+		m_wield_meshnode = nullptr;
 	} else if (m_spritenode) {
 		m_spritenode->remove();
 		m_spritenode->drop();
-		m_spritenode = NULL;
+		m_spritenode = nullptr;
+	}
+	if (m_matrixnode) {
```
nerzhul: please add an empty line before

paramat:
I'll merge this once that's done.

pgimeno:
@pauloue could you update it please?

paramat:
I might make that change on merge.

pauloue:
Updated

pgimeno:
Big thanks to @pauloue for hosting this PR.

PR #8034, opened 2018-12-30
Disable auto-setting server as favorite on connect; display "Add Favorite" button

ClobberXD:
Currently, a server is automatically added to one's favourites when they connect to it (or even attempt to connect to it). I think this isn't a good idea, as players would be trying out new servers, and might not ever visit them again, and yet, those servers will unnecessarily be added to their favourites.

This PR disables "auto-favouriting" all servers the player joins, and provides an "Add Favorite" button. Depending on whether a server is marked as favourite, there will be a "Del. Favorite" button or an "Add Favorite" button. A server will be marked as favorite only when a player explicitly adds it by pressing the "Add Favorite" button.

### DIFF
--- a/builtin/mainmenu/tab_online.lua
+++ b/builtin/mainmenu/tab_online.lua
@@ -61,7 +61,11 @@ local function get_formspec(tabview, name, tabdata)
 		if gamedata.fav then
 			retval = retval .. "button[7.73,4.9;2.3,1;btn_delete_favorite;" ..
 				fgettext("Del. Favorite") .. "]"
+		else
+			retval = retval .. "button[7.73,4.9;2.3,1;btn_add_favorite;" ..
+				fgettext("Add Favorite") .. "]"
 		end
+
 		if fav_selected.description then
 			retval = retval .. "textarea[8.1,2.3;4.23,2.9;;;" ..
 				core.formspec_escape((gamedata.serverdescription or ""), true) .. "]"
@@ -228,6 +232,17 @@ local function main_button_handler(tabview, fields, name, tabdata)
 
 		tabdata.fav_selected = fav_idx
 		return true
+	end
+
+	if fields.btn_add_favorite then
+		core.add_favorite({
+			name        = gamedata.servername,
+			address     = gamedata.address,
+			port        = gamedata.port,
+			description = gamedata.description
+		})
+		asyncOnlineFavourites()
+		tabdata.fav_selected = core.get_table_index("favourites")
 	end
 
 	if fields.btn_delete_favorite then
--- a/doc/menu_lua_api.txt
+++ b/doc/menu_lua_api.txt
@@ -17,12 +17,12 @@ Gamedata
 --------
 The "gamedata" table is read when calling core.start(). It should contain:
 {
-	playername     = <name>,
-	password       = <password>,
-	address        = <IP/adress>,
-	port           = <port>,
-	selected_world = <index>, -- 0 for client mode
-	singleplayer   = <true/false>,
+    playername     = <name>,
+    password       = <password>,
+    address        = <IP/adress>,
+    port           = <port>,
+    selected_world = <index>, -- 0 for client mode
+    singleplayer   = <true/false>,
 }
 
 Functions
@@ -102,12 +102,12 @@ core.show_path_select_dialog(formname, caption, is_file_select)
 ^ returns nil or selected file/folder
 core.get_screen_info()
 ^ returns {
-	density         = <screen density 0.75,1.0,2.0,3.0 ... (dpi)>,
-	display_width   = <width of display>,
-	display_height  = <height of display>,
-	window_width    = <current window width>,
-	window_height   = <current window height>
-	}
+    density         = <screen density 0.75,1.0,2.0,3.0 ... (dpi)>,
+    display_width   = <width of display>,
+    display_height  = <height of display>,
+    window_width    = <current window width>,
+    window_height   = <current window height>
+    }
 
 ### Content and Packages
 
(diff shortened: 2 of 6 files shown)


ClobberXD:
Updated first post with the remaining issues. PR will remain WIP until these are taken care of, but I could always use some help in fixing these. :)

paramat:
PR concept makes good sense, current behaviour doesn't.

Fixer-007:
I was frustrated with this for a long time, but was too lazy to create an issue, thank you for doing this PR, +1

ClobberXD:
I think I got the hang of this now - seems to work. fine now. Ready for testing and review. :)

pgimeno:
I sort of like the current behaviour. Could this be made optional? At least until #8318 is implemented.

jastevenson303:
The current is more like "recently visited" and less like "favorited."
Ideally, I think, both would be present.

ClobberXD:
Instead of making this optional for the time-being, putting this PR on hold seems to be an easier approach to me :)

doc/menu_lua_api.txt
```diff
@@ -17,12 +17,12 @@ Gamedata
...
-	address        = <IP/adress>,
-	port           = <port>,
-	selected_world = <index>, -- 0 for client mode
-	singleplayer   = <true/false>,
+    playername     = <name>,
+    password       = <password>,
+    address        = <IP/adress>,
+    port           = <port>,
+    selected_world = <index>, -- 0 for client mode
+    singleplayer   = <true/false>,
```
pgimeno: Is this a pure text file, or a .md file renamed to .txt in the style of lua_api.txt? If the latter, then the formatting is wrong all over the place, and this change is incomplete. See https://notabug.org/pgimeno/Gists/src/minetest-clobberxd---menu_lua_api/menu_lua_api.md

I suggest to either leave the formatting alone, or add a separate cleanup commit or even a new PR to clean it up.

DS-Minetest:
You might want to have a server in the serverlist (=favorite list) that is not in the public server list (=online favorites). This feature is important and was there before we got the online favorites. It should not be simply removed.

I think, the whole change should be done in one PR.
(Rename favorites to visited and add real favorites or something.)

pgimeno:
I'm working on a patch based on this one.

doc/menu_lua_api.txt
```diff
@@ -98,74 +98,86 @@ core.file_open_dialog(formname,caption)
...
 }
-core.delete_favorite(id, location) -> success
+
+- core.add_favorite(server)
+  - Adds a server to the local favorites list
+  - `server`: {
+        `name` (server name)
+        `address`
+        `port`
+        `description`
```
DS-Minetest: This is no real table and will cause confusion.
One might think that it's a list (eg. `{"mycoolserver", "bla.net",30000,"come and join"}` but the commas are missing.

It should be something like:
```
{
  name    = <server name>,
  address = <address of the server>,
  ...
}
```

pgimeno:
@ClobberXD Mind if I take your code but removing most changes to menu_lua_api.txt, as a basis for a new patch?

DS-Minetest:
There are many changes that don't have too much to do with favorites not being marked on join (eg. trying to delete a server from the public serverlist (which does nothing) is here removed). I think, these changes should be made in separate commits.

pgimeno:
@DS-Minetest I believe the list is only of favourite servers, therefore removing an element from the list effectively removes the favourite.

doc/menu_lua_api.txt
```diff
@@ -98,74 +98,86 @@ core.file_open_dialog(formname,caption)
...
+    }
+
+- core.delete_favorite(idx)
+  - Removes a server from the local favorites list. Returns `true` on success.
```
DS-Minetest: I think, the ``Returns `true` on success.`` is wrong.

DS-Minetest:
>eg. trying to delete a server from the public serverlist (which does nothing) is here removed

> @DS-Minetest I believe the list is only of favourite servers, therefore removing an element from the list effectively removes the favourite.

Yes, but removing an element from the online favorite list (= public server list) doesn't really work. [I've tested it.](https://github.com/DS-Minetest/minetest/tree/mainmenu_online_bla)
This PR changes `delete_favorite` to only delete servers from the local list. This is good but should be done in a separate commit.
I'm sorry for expressing myself so unclear. >_<

builtin/mainmenu/tab_online.lua
```diff
@@ -229,6 +233,17 @@ local function main_button_handler(tabview, fields, name, tabdata)
 		return true
 	end
 
+	if fields.btn_add_favorite then
+		core.add_favorite({
+			name        = gamedata.servername,
+			address     = gamedata.address,
+			port        = gamedata.port,
+			description = gamedata.description
```
pgimeno: This is causing a crash. `gamedata` doesn't seem to hold this information. It is in `serverlist[tabdata.fav_selected]` (if `tabdata.fav_selected` is not nil or 0).

src/serverlist.cpp
```diff
@@ -115,9 +115,7 @@ bool deleteEntry(const ServerListSpec &server)
 	std::string path = ServerList::getFilePath();
 	std::ostringstream ss(std::ios_base::binary);
 	ss << ServerList::serialize(serverlist);
-	if (!fs::safeWriteToFile(path, ss.str()))
-		return false;
-	return true;
+	return !fs::safeWriteToFile(path, ss.str()) ? true : false;
```
pgimeno: This is the wrong way around. I suggest to revert this change.

src/serverlist.cpp
```diff
@@ -134,10 +132,7 @@ bool insert(const ServerListSpec &server)
 	std::string path = ServerList::getFilePath();
 	std::ostringstream ss(std::ios_base::binary);
 	ss << ServerList::serialize(serverlist);
-	if (!fs::safeWriteToFile(path, ss.str()))
-		return false;
-
-	return true;
+	return !fs::safeWriteToFile(path, ss.str()) ? true : false;
```
pgimeno: This is the wrong way around. I suggest to revert this change. There's no point in touching this file just for rewriting the style of the condition.

builtin/mainmenu/tab_online.lua
DS-Minetest: There's a `gamedata.serverdescription`.

src/serverlist.cpp
DS-Minetest: Can't this simply be `return fs::safeWriteToFile(path, ss.str())`?

builtin/mainmenu/tab_online.lua
pgimeno: It's empty at that point. Also, I mean all four fields, but I didn't want to comment on all four lines :)

Have you tried the patch? It crashes for me due to this data being empty.

src/serverlist.cpp
pgimeno: Yes, but not point in changing the file just for that, which is why I suggested a reversal.

src/serverlist.cpp
ClobberXD: Will revert both these changes.

ClobberXD:
@pgimeno Feel free to use this PR as a base for another patch - I have no issues with it. :)

ClobberXD:
Is someone working on #8318? If no one's working on that PR, what'd happen to this PR?

ClobberXD:
\*bump\*

Core dev inputs needed. This PR was put on hold because pgimeno was planning to implement #8318 first, but has now stopped working on it (see prev. post). What's the fate of this PR now?

This PR is still not perfect. But this PR is still a significant step in the right direction, even without #8318.

paramat:
I closed #8318 as it is abandoned. So just continue as if it does not exist.

ClobberXD:
Thanks, paramat. Will continue with this.

paramat:
To clarify, i think this change is still desirable even without the 'recently visited' ordering that #8318 implemented.

Fixer-007:
Any news on this?

ClobberXD:
I'd say this might take a week or two to attend to. But fortunately for me, there's only a small amount of cleaning up to do, and the PR would be fresh as a daisy - there aren't any mammoth merge conflicts I have to deal with. :)

sfan5:
If this is added, how would I add a server this is not on the server list to my favorites?

ClobberXD:
> If this is added, how would I add a server this is not on the server list to my favorites?

This PR removes the ability to do this; this is a very valid concern.

I'm thinking of re-adding the auto-favourite-upon-connect functionality, but only if a server isn't on the server list. Thoughts?

sfan5:
> I'm thinking of re-adding the auto-favourite-upon-connect functionality, but only if a server isn't on the server list. Thoughts?

Sounds fine to me.

SmallJoker:
Non-listed servers should be added to the favourites automatically. So yes, that suggestion sounds good.

> I sort of like the current behaviour. 

> though I still prefer the overall "favorite-on-connect" feature to stay as is.

Me as well. It's helpful to keep track of the servers I usually play on. Thus I'm not really in favour of this PR. A setting could be added for sure, but should then also be integrated directly into the mainmenu (checkbox, perhaps).

ClobberXD:
Then the *correct* solution would be to have two categories "visited" and "favourites". While there's support for this PR, I do also see some opposition - I'd prefer to not merge this PR until we arrive upon  an 'ideal' solution.

PR #8041, opened 2019-01-02
delete repeated text

maxis27:


SmallJoker:
This file is generated automatically. Thus, this PR won't change anything because it will be overwritten the next time anyway. 
Please remove the line from here: https://github.com/minetest/minetest/blob/master/builtin/settingtypes.txt#L118

PR #8043, opened 2019-01-03
Make sqlite3 default auth & player backends for new worlds

nerzhul:
We forgot this important change for 5.0.0

Also notify about auth backend depreciation

### DIFF
--- a/src/serverenvironment.cpp
+++ b/src/serverenvironment.cpp
@@ -412,6 +412,7 @@ ServerEnvironment::ServerEnvironment(ServerMap *map,
 		if (!player_backend_exists) {
 			// fall back to files
 			conf.set("player_backend", "files");
+			player_backend_name = "files";
 
 			if (!conf.updateConfigFile(conf_path.c_str())) {
 				errorstream << "ServerEnvironment::ServerEnvironment(): "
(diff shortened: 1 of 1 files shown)


sofar:
:+1: on the concept, I didn't test or check the code though, but yes this was planned a long time back.

nerzhul:
@rubenwardy you said it was due to the player attribute backend but playermetaref can mitigate the issue, but no update, can you provide it ?. Auth should not be a problem it's called at auth only.

nerzhul:
@rubenwardy performance penalty fixed by c1d7dbfc38e8f48b42d957afad2696a8e56b740b (pushed to master). There are many incoherences in our save resulting in bad performancein SQL backends and some non write ensurance, i make all homogeneous refactoring the write flagging part

src/serverenvironment.cpp
```diff
@@ -396,36 +396,60 @@ ServerEnvironment::ServerEnvironment(ServerMap *map,
...
+	// If we open world.mt read the backend configurations.
+	if (succeeded) {
+		// Read those values before setting defaults
+		bool player_backend_exists = conf.exists("player_backend");
+		bool auth_backend_exists = conf.exists("auth_backend");
+
+		// player backend is not set, assume it's legacy file backend.
+		if (!player_backend_exists) {
+			// fall back to files
+			conf.set("player_backend", "files");
```
SmallJoker: Also update `player_backend_name`

src/serverenvironment.cpp
```diff
@@ -396,36 +396,60 @@ ServerEnvironment::ServerEnvironment(ServerMap *map,
...
+			conf.getNoEx("player_backend", player_backend_name);
 		}
-	}
 
-	std::string name;
-	conf.getNoEx("player_backend", name);
-	m_player_database = openPlayerDatabase(name, path_world, conf);
+		// auth backend is not set, assume it's legacy file backend.
+		if (!auth_backend_exists) {
+			conf.set("auth_backend", "files");
```
SmallJoker: Also update `auth_backend_name`

PR #8050, opened 2019-01-05
Increase default active_object_send_range_blocks

lhofhansl:
See #8048.

Increase default active_object_send_range_blocks to match max_block_send_distance.

### DIFF
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -1069,7 +1069,7 @@ ask_reconnect_on_crash (Ask to reconnect after crash) bool false
 #    Setting this larger than active_block_range will also cause the server
 #    to maintain active objects up to this distance in the direction the
 #    player is looking. (This can avoid mobs suddenly disappearing from view)
-active_object_send_range_blocks (Active object send range) int 3
+active_object_send_range_blocks (Active object send range) int 9
 
 #    The radius of the volume of blocks around every player that is subject to the
 #    active block stuff, stated in mapblocks (16 nodes).
--- a/src/defaultsettings.cpp
+++ b/src/defaultsettings.cpp
@@ -339,7 +339,7 @@ void set_default_settings(Settings *settings)
 	settings->setDefault("ask_reconnect_on_crash", "false");
 
 	settings->setDefault("profiler_print_interval", "0");
-	settings->setDefault("active_object_send_range_blocks", "3");
+	settings->setDefault("active_object_send_range_blocks", "9");
 	settings->setDefault("active_block_range", "3");
 	//settings->setDefault("max_simultaneous_block_sends_per_client", "1");
 	// This causes frametime jitter on client side, or does it?


paramat:
I'm ok with increasing it a little, but such a large increase may cause performance issues in some situations. This makes me prefer leaving this to after 5.0.0 to have long term testing first.
If this was just increased a little to 4 i would be ok with this for 5.0.0.
So, as it is currently, not for 5.0.0.

lhofhansl:
I doubt it's having any impact. Default view range on Android is 50 on "Desktop" it's 100.

So that translates to just 3 blocks on Android and 6 on "Desktop". This just increases the potential *if* the the client requested a larger view range.

Happy to delay it - though I don't think it's necessary.

paramat:
Seeing terrain is more important than seeing objects, so in a way it makes sense to have active object range smaller than block send distance.
As a player i would be irritated by having to reduce my view range because objects are active at extreme distances, seeing terrain is more important.
So, i think this value is too large, before or after 5.0.0.
:-1: for 9 mapblocks. Somewhere between active block range and send distance.

The default view range is somewhat irrelevant, players change view range all the time in-game, and tend to increase it as much as possible.

I'm still a little concerned about objects over 3 mapblocks away only being active when you're looking at them, it's optimal but still seems a little weird.

lhofhansl:
I'm not saying this right. You'll never see object past your client's viewing range.

I'm getting a bit tired of this, lemme just close this.

paramat:
> You'll never see object past your client's viewing range.

Yes i understand that :)
What i mean is:
In a situation where there is heavy object usage (quite common on servers) and i increase my view range to see 256 nodes of terrain, i don't want to also see objects 256 nodes away that then significantly reduces client performance. Because at large distances terrain is more important to see than objects.
I'm fine with a smaller distance and i agree it should be larger than the current 3 mapblocks, so i prefer this PR stays open. Currently a larger object limit in view direction is not being used at all so an increase is probably a good idea.

jastevenson303:
That may be your preference, but mine is the inverse.  If I extend view
range, I expect it to include objects.

Objects should be visible where there are nodes, and vise versa.

paramat:
I'm now thinking about server load for the increased number of objects that will be running their 'on step' functions.

If we consider 1 player on a server who is using a view range of at least 9 blocks = 144 nodes:
With default FOV the horizontal FOV is roughly 90 degrees.
Active disc around player has area PI * (3 * 16) ^ 2 = 7238 square nodes.
Area of viewed quadrant from radius 48 to 144 = (PI * (9 * 16) ^ 2) / 4 - (PI * (3 * 16) ^ 2) / 4 = 14476 square nodes.
Total = 21714 square nodes.

So the PR would triple the number of active objects, 9 is far too high.

jastevenson303 ideally yes but probably not practical.
This is about: At what distance should objects become visible to players?
Most objects are less than 1.5 stacked nodes in size (mob size), at 144 nodes an object that size looks like (obsidian cube plus slab):

screenshot_20190106_072729

Which considering intensity of 3 times more active objects is not worth rendering. Consider that many objects are smaller.

I do support increasing active object range but 64 nodes seems a good balance between distance and server load, an increase of 1 mapblock:

screenshot_20190106_073100

Area of quadrant from 48 to 64 = 1407 square nodes. This would be a 19% increase in area from the current 48 node disc around the player.
So i'm ok with the setting increased to 4.

lhofhansl:
My point remains that it wouldn't unless you increase your viewing_range, or unless you zoom (in which case the FOV is smaller). So on Android by default nothing would change and on the desktop version it would double to 6. 9 can only be reached if/when a client increases the viewing_range.

I agree if/when client do increase the viewing_range there's more server load. At the same time I also still prefer defaults that look good and behave well rather than (perhaps) premature optimizations.

By the same argument {{max_block_send_distance}} is also too large, since you cannot really see the small blocks.

In any case a server owner can increase this, so it's critical to increase the default... But I assume most don't know about this.

(I cringe every time I see a Minetest video on Youtube where the person moves around and suddenly objects - cars, mobs, etc - pop into existence and then vanish again when the distance increases a bit. Looks bad and unprofessional.)

Can we at least default it to 6, so that objects appear and disappear smoothly with the default viewing_range.

Another option is to document this better. Is there a wiki page for server setting where I could add an explanation for these settings?

sfan5:
> I'm now thinking about server load for the increased number of objects that will be running their 'on step' functions.

huh? `active_block_range` is not being increased here.

> Can we at least default it to 6, so that objects appear and disappear smoothly with the default viewing_range.

I'd much prefer 9, but if paramat is not okay with that increasing it to 6 is alright too.

paramat:
> My point remains that it wouldn't unless you increase your viewing_range,

Most desktop players use a viewing range of at least 144 nodes, so the tripling of object load will occur most of the time, for the sake of rendering objects (and running their on_step) code that are insignificantly visually small.
Desktop users don't use the default viewing range, they change it in-game all the time, and since 100 is unpleasantly close they increase it as far as they can.

I won't insist on 4 but more is risky. I do agree it should be increased. But keep in mind if you do this before 5.0.0 you risk adding significant load to servers and server network traffic, this is why i'm being cautious.
I think an increase to at least 4 is needed for 5.0.0, but also, a larger increase is risky to do now for 5.0.0 and before long term testing. This is why i'm suggesting a small increase for 5.0.0 then possibly more increase later.

sfan5, it seems that objects will be kept active by the server out to this distance from a player when within the FOV of that player, otherwise i wouldn't be so concerned. mob on_step functions are quite intensive:
```
#    From how far clients know about objects, stated in mapblocks (16 nodes).
#
#    Setting this larger than active_block_range will also cause the server
#    to maintain active objects up to this distance in the direction the
#    player is looking. (This can avoid mobs suddenly disappearing from view)
active_object_send_range_blocks (Active object send range) int 3

#    The radius of the volume of blocks around every player that is subject to the
#    active block stuff, stated in mapblocks (16 nodes).
#    In active blocks objects are loaded and ABMs run.
#    This is also the minimum range in which active objects (mobs) are maintained.
#    This should be configured together with active_object_range.
active_block_range (Active block range) int 3
```
Knowing this perhaps you can see the how much more intensive this will be on a server?

So i now also realise that active objects will be sent to the player within a complete disk of the setting radius.
3 -> 9 means disk area is multiplied by 9 ^ 2 / 3 ^ 2 = 9, that's 9 times more object network traffic which could be a problem.

So, even 6 might be too much, but i'm unsure.
Would objects appearing at great distance be worth it if server performance is significantly affected? We have to be careful about balancing this and i'm not sure we have time before 5.0.0 to balance it correctly.

Anyway, for 5.0.0 i'm :+1: for 4, unsure about 5 or 6, :-1: for anything over 6.

sfan5:
> sfan5, it seems that objects will be kept active by the server out to this distance from a player when within the FOV of that player, otherwise i wouldn't be so concerned. mob on_step functions are quite intensive

I assumed "maintain" just meant "keep loaded in memory" (without calling on_step), I checked the implementation and that's not the case
yes, 4-6 might indeed be a better choice than 9

PR #8070, opened 2019-01-07
Fix a crash on Android with Align2Npot2

### LINKED ISSUE #8069: [Android] Minetest 0.4.17.3 crash
##### Issue type

 - Bug report
 
##### Minetest version

```
0.4.17.3
```

##### OS / Hardware

Operating system: Android 6.0
CPU: LGE LG X Style (k6b)

##### Summary

New crash on android

```
backtrace:
   #00  pc 0000000000018108  /system/lib/libc.so (strlen+71)
   #01  pc 0000000000459c8d  /data/app/net.minetest.minetest-2/lib/arm/libminetest.so (_Z11Align2Npot2PN3irr5video6IImageEPNS0_12IVideoDriverE+64)
   #02  pc 000000000045af3d  /data/app/net.minetest.minetest-2/lib/arm/libminetest.so (_ZN13TextureSource24rebuildImagesAndTexturesEv+168)
   #03  pc 0000000000245d61  /data/app/net.minetest.minetest-2/lib/arm/libminetest.so (_ZN6Client20afterContentReceivedEPN3irr14IrrlichtDeviceE+296)
   #04  pc 00000000002b41d7  /data/app/net.minetest.minetest-2/lib/arm/libminetest.so (_ZN4Game12createClientERKNSt6__ndk112basic_stringIcNS0_11char_traitsIcEENS0_9allocatorIcEEEES8_PS6_t+206)
   #05  pc 00000000002b3fc3  /data/app/net.minetest.minetest-2/lib/arm/libminetest.so (_ZN4Game7startupEPbbP12InputHandlerPN3irr14IrrlichtDeviceERKNSt6__ndk112basic_stringIcNS6_11char_traitsIcEENS6_9allocatorIcEEEESE_SE_PSC_tRSC_S0_P11ChatBackendRK11SubgameSpecb+374)
   #06  pc 00000000002c523f  /data/app/net.minetest.minetest-2/lib/arm/libminetest.so (_Z8the_gamePbbP12InputHandlerPN3irr14IrrlichtDeviceERKNSt6__ndk112basic_stringIcNS5_11char_traitsIcEENS5_9allocatorIcEEEESD_SD_SD_tRSB_R11ChatBackendS_RK11SubgameSpecb+198)
   #07  pc 0000000000452ac1  /data/app/net.minetest.minetest-2/lib/arm/libminetest.so (_ZN14ClientLauncher3runER10GameParamsRK8Settings+2416)
   #08  pc 0000000000320923  /data/app/net.minetest.minetest-2/lib/arm/libminetest.so (main+8186)
   #09  pc 00000000003850ab  /data/app/net.minetest.minetest-2/lib/arm/libminetest.so (android_main+90)
   #10  pc 000000000083fe21  /data/app/net.minetest.minetest-2/lib/arm/libminetest.so
   #11  pc 0000000000041a03  /system/lib/libc.so (_ZL15__pthread_startPv+30)
   #12  pc 00000000000193a5  /system/lib/libc.so (__start_thread+6)
```

nerzhul:
glGetString can be NULL. If stored in a string it triggers a SIGSEGV.
Instead do a basic strstr and verify the pointer

Fixed #8069

### DIFF
--- a/src/client/tile.cpp
+++ b/src/client/tile.cpp
@@ -1012,13 +1012,14 @@ video::IImage * Align2Npot2(video::IImage * image,
 
 	core::dimension2d<u32> dim = image->getDimension();
 
-	const char *extensions = glGetString(GL_EXTENSIONS);
-
 	// Only GLES2 is trusted to correctly report npot support
-	if (get_GL_major_version() > 1 && extensions != NULL &&
-			strstr(extensions, "GL_OES_texture_npot") != NULL) {
+	// Note: we cache the boolean result. GL context will never change on Android.
+	static const bool hasNPotSupport = get_GL_major_version() > 1 &&
+		glGetString(GL_EXTENSIONS) && 
+		strstr(glGetString(GL_EXTENSIONS), "GL_OES_texture_npot");
+	
+	if (hasNPotSupport)
 		return image;
-	}
 
 	unsigned int height = npot2(dim.Height);
 	unsigned int width  = npot2(dim.Width);
--- a/src/client/tile.cpp
+++ b/src/client/tile.cpp
@@ -1012,11 +1012,11 @@ video::IImage * Align2Npot2(video::IImage * image,
 
 	core::dimension2d<u32> dim = image->getDimension();
 
-	std::string extensions = (char*) glGetString(GL_EXTENSIONS);
+	const char *extensions = glGetString(GL_EXTENSIONS);
 
 	// Only GLES2 is trusted to correctly report npot support
-	if (get_GL_major_version() > 1 &&
-			extensions.find("GL_OES_texture_npot") != std::string::npos) {
+	if (get_GL_major_version() > 1 && extensions != NULL &&
+			strstr(extensions, "GL_OES_texture_npot") != NULL) {
 		return image;
 	}
 


src/client/tile.cpp
```diff
@@ -1012,11 +1012,11 @@ video::IImage * Align2Npot2(video::IImage * image,
...
 	core::dimension2d<u32> dim = image->getDimension();
 
-	std::string extensions = (char*) glGetString(GL_EXTENSIONS);
+	const char *extensions = glGetString(GL_EXTENSIONS);
 
 	// Only GLES2 is trusted to correctly report npot support
-	if (get_GL_major_version() > 1 &&
-			extensions.find("GL_OES_texture_npot") != std::string::npos) {
+	if (get_GL_major_version() > 1 && extensions != NULL &&
+			strstr(extensions, "GL_OES_texture_npot") != NULL) {
```
numberZero: ` != NULL` is superfluous, pointer-to-bool conversion is well-defined. If you want such syntax anyway, use `nullptr` at least.

src/client/tile.cpp
```diff
@@ -1012,11 +1012,11 @@ video::IImage * Align2Npot2(video::IImage * image,
 
 	core::dimension2d<u32> dim = image->getDimension();
 
-	std::string extensions = (char*) glGetString(GL_EXTENSIONS);
```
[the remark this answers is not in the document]
nerzhul: exact but i prefer to keep C primitives if not used outside a C function and called in the same stack :)

src/client/tile.cpp
nerzhul: will nullptr work with a C function call which doesn't know nullptr_t ?

nerzhul:
For performance i will change that to static. The GL context will not change on android

nerzhul:
@numberZero i updated it it will be better and remove the performance penalty on the gl verification

src/client/tile.cpp
numberZero: That doesn’t matter. Null pointer is a null pointer, regardless of how you name it. The reason to use `nullptr` is that it is only compatible with pointers, while `NULL` may (and is, in glibc at least) be defined as literal `0`.

Actually, the only allowed definitions for `NULL` are `nullptr` and `0` in C++. `((void*)0)` may only work in C as C++ forbids most implicit pointer conversions.

P.S. Sorry for OT.

nerzhul: No problem, is the new version okay to you ?

PR #8081, opened 2019-01-08
Valleys Mapgen: Optionally use 3D noise for rivers

Gael-de-Sailly:
This follows the changes I've made on my Lua version of Valleys Mapgen some time ago (see [here](https://forum.minetest.net/viewtopic.php?p=330950#p330950)).

The idea is to use a 3D noise instead of 2D for `noise_rivers`. This has an effect on the shape of valleys, allowing steeper and more irregular slopes, sometimes even overhanging the river. The deepest the valleys are, the most noticeable it is. This is why this feature has been named Canyons.

The algorithm remains the same, except that the river noise value now depends on elevation, so I moved all calculations involving it from the 2D loop to the 3D one (including `terrainLevelFromNoise(&tn)` that I had to remove from the function `calculateNoise()` and put it in `generateTerrain()`).

It is optional, enabled by the `canyon` spflag, disabled by default for backward compatibility.
Since it needs to handle both mapgen methods, I've done my best to make the code as flexible as possible regarding river noise dimension.

Notes:
- The 2D version uses the X-Z spread values for the river noise, instead of X-Y, so that adjusting the vertical spread for 3D noise won't stretch the rivers horizontally when switching back to 2D.
- For the 3D version, spawnplayer function doesn't avoid rivers.
- I'm wondering whether doing `if (spflags & MGVALLEYS_CANYONS)` results in a significant loss of time in the 3D loop, in which case I will convert it into a boolean outside the loop. What do you think?

Some screenshots (using my custom mapgen parameters)
Screenshot
Screenshots
screenshot_20190108_224531
screenshot_20190106_210041

### DIFF
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -1834,7 +1834,9 @@ mgfractal_np_dungeons (Dungeon noise) noise_params_3d 0.9, 0.5, (500, 500, 500),
 #    'vary_river_depth': If enabled, low humidity and high heat causes rivers
 #    to become shallower and occasionally dry.
 #    'altitude_dry': Reduces humidity with altitude.
-mgvalleys_spflags (Mapgen Valleys specific flags) flags altitude_chill,humid_rivers,vary_river_depth,altitude_dry,canyons altitude_chill,noaltitude_chill,humid_rivers,nohumid_rivers,vary_river_depth,novary_river_depth,altitude_dry,noaltitude_dry,canyons,nocanyons
+#    'canyons': Use 3D noise for valleys, resulting in more cliffs and steep
+#    structures around rivers
+mgvalleys_spflags (Mapgen Valleys specific flags) flags altitude_chill,humid_rivers,vary_river_depth,altitude_dry altitude_chill,noaltitude_chill,humid_rivers,nohumid_rivers,vary_river_depth,novary_river_depth,altitude_dry,noaltitude_dry,canyons,nocanyons
 
 #    The vertical distance over which heat drops by 20 if 'altitude_chill' is
 #    enabled. Also the vertical distance over which humidity drops by 10 if
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -1834,7 +1834,7 @@ mgfractal_np_dungeons (Dungeon noise) noise_params_3d 0.9, 0.5, (500, 500, 500),
 #    'vary_river_depth': If enabled, low humidity and high heat causes rivers
 #    to become shallower and occasionally dry.
 #    'altitude_dry': Reduces humidity with altitude.
-mgvalleys_spflags (Mapgen Valleys specific flags) flags altitude_chill,humid_rivers,vary_river_depth,altitude_dry altitude_chill,noaltitude_chill,humid_rivers,nohumid_rivers,vary_river_depth,novary_river_depth,altitude_dry,noaltitude_dry
+mgvalleys_spflags (Mapgen Valleys specific flags) flags altitude_chill,humid_rivers,vary_river_depth,altitude_dry,canyons altitude_chill,noaltitude_chill,humid_rivers,nohumid_rivers,vary_river_depth,novary_river_depth,altitude_dry,noaltitude_dry,canyons,nocanyons
 
 #    The vertical distance over which heat drops by 20 if 'altitude_chill' is
 #    enabled. Also the vertical distance over which humidity drops by 10 if
@@ -1886,7 +1886,9 @@ mgvalleys_np_filler_depth (Filler depth) noise_params_2d 0, 1.2, (256, 256, 256)
 mgvalleys_np_cavern (Cavern noise) noise_params_3d 0, 1, (768, 256, 768), 59033, 6, 0.63, 2.0
 
 #    Defines large-scale river channel structure.
-mgvalleys_np_rivers (River noise) noise_params_2d 0, 1, (256, 256, 256), -6050, 5, 0.6, 2.0, eased
+#    Used as 3D noise if 'canyons' spflag is enabled,
+#    otherwise used as 2D noise with Y spread ignored.
+mgvalleys_np_rivers (River noise) noise_params_3d 0, 1, (256, 256, 256), -6050, 5, 0.6, 2.0, eased
 
 #    Base terrain height.
 mgvalleys_np_terrain_height (Terrain height) noise_params_2d -10, 50, (1024, 1024, 1024), 5202, 6, 0.4, 2.0, eased
--- a/src/mapgen/mapgen_valleys.cpp
+++ b/src/mapgen/mapgen_valleys.cpp
@@ -83,10 +83,10 @@ MapgenValleys::MapgenValleys(MapgenValleysParams *params, EmergeManager *emerge)
 	noise_valley_profile     = new Noise(&params->np_valley_profile,     seed, csize.X, csize.Z);
 
 	//// River noise, 2D or 3D
-	if (spflags & MGVALLEYS_CANYONS)
+	if (spflags & MGVALLEYS_CANYONS) {
 		noise_rivers     = new Noise(&params->np_rivers, seed,
 				csize.X, csize.Y + 2, csize.Z);
-	else {
+	} else {
 		v3f *spread = &params->np_rivers.spread;
 		spread->Y = spread->Z;
 		noise_rivers     = new Noise(&params->np_rivers,             seed, csize.X, csize.Z);
--- a/src/mapgen/mapgen_valleys.cpp
+++ b/src/mapgen/mapgen_valleys.cpp
@@ -284,12 +284,6 @@ void MapgenValleys::makeChunk(BlockMakeData *data)
 
 int MapgenValleys::getSpawnLevelAtPoint(v2s16 p)
 {
-	// Check if in a river channel
-	float n_rivers = NoisePerlin2D(&noise_rivers->np, p.X, p.Y, seed);
-	if (std::fabs(n_rivers) <= river_size_factor)
-		// Unsuitable spawn point
-		return MAX_MAP_GENERATION_LIMIT;
-
 	float n_slope          = NoisePerlin2D(&noise_inter_valley_slope->np, p.X, p.Y, seed);
 	float n_terrain_height = NoisePerlin2D(&noise_terrain_height->np, p.X, p.Y, seed);
 	float n_valley         = NoisePerlin2D(&noise_valley_depth->np, p.X, p.Y, seed);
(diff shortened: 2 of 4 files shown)


paramat:
I'll look at this after 5.0.0 release. It'll be a headache because valleys mapgen is coded in a strange and confusing way that i can't stand. I still can't get my head around how it works, and oddly it seems no-one understands it, including duane who actually coded it. It was our, and especially my, big mistake to allow this mapgen to be merged with code in such a bizarre and messy state.

It's possible i will have to rewrite the code (without breaking worlds) to make it understandable and maintainable. I get the feeling it has been made far more complex and confusing than it needs to be.
I'm unsure about adding to the mess before cleaning up the mess.

Also i'm unsure this is worth adding as it makes the valleys look like mgv7 canyons, people appreciate mgvalleys not looking like mgv7. I thought the idea of mgvalleys was to have subtle use of 3D noise, only appearing at the summits of some mountains. A concern i have is that the entire terrain is composed of valley sides, so if you add 3D noise to valley sides the entire terrain will have 3D noise applied, which will be good around rivers but ugly elsewhere.

I'll test this sometime to get a better feel of what it actually does.

Gael-de-Sailly:
I see. I have to admit that understanding that code has not been easy for me either, I had to compare carefully with my Lua to understand what was equivalent to what.
I think duane had coded something functionally equivalent to my Lua without actually understanding it, so all the issue seems to come from my original code, that no one is understanding either.

I'm very little experienced in C++ for now, but if one day I have enough skills (i'm very motivated to code more on the core), rewriting Valleys Mapgen should be for myself.

About what the 3D river noise does, its effect will only be noticeable on riverbeds and valleys edges.
It's not completely true that the entire terrain is made of valleys sides. Their shape is a Gaussian function of `noise_rivers` and `noise_valley_profile`, it quickly becomes constant as we move away from the river.

One thing is worth being noted: for those who find the new option too extreme, it's possible to have subtle variations between the original and the new canyons feature, by using `canyons` with an increased value for the Y spread of `noise_rivers` (because a 2D noise is equivalent to a 3D noise with infinite Y spread).

paramat:
> About what the 3D river noise does, its effect will only be noticeable on riverbeds and valleys edges.

Good to know.

paramat:
The code being so confusing, and me not understanding it, is a big problem. So i intend to work through the code and clean it up, remove the horrible confusing stuff, and in the process understand it
I can't let the code become even more complex and confusing before i work on it, i can barely face this code as it is, which is partly why i keep procrastinating.

So due to that, unfortunately :-1: for this PR but the feature can probably be added after i clean up the code, hopefully during 5.1.0.

paramat:
I was in the mood so i just spent a few hours starting the job of cleaning up the mgvalleys code, making good progress.

Gael-de-Sailly:
I wrote a new (WIP) version rebased on #8101:
https://github.com/Gael-de-Sailly/minetest/tree/vmg_canyon_2

paramat:
Valleys mapgen code rewrite has been merged now so feel free to rebase on that, then i'll test.

Gael-de-Sailly:
Ok that's done. WIP again, since I had to rewrite it from scratch, some features are not implemented yet, like mainmenu settings and documentation.

Gael-de-Sailly:
Oops... did I forgot I left it WIP? Back on this today.

paramat:
It was functional, so i could have tested it by now, will do so as soon as i can. I will probably +1.

Gael-de-Sailly:
I was actually working on your cleaned version since January, when it was not merged, then I did the rebasing in a hurry in April, but now that I get into it again, it works better than I expected. I will double-check the spawnplayer function, but for the mapgen itself I don't see anything to change, so it's ready for testing.

For the settings I just have one question: I'm using one noise for `mgvalleys_np_rivers` either as 2D or as 3D, depending on the `spflags`. I made the 2D version use the XZ parameters for `spread` instead of XY, to allow setting a different value for Y spread in 3D version without altering the 2D version.

Is that OK or do you prefer two different settings? (I prefer one setting because, 2D or 3D, they have pretty much the same role, and this avoids adding one more setting in an already long list).

paramat:
Fine for me if it works.
By the way, you shouldn't make changes to conf.example, that is regularly auto-updated. Only change settingtypes.txt.

builtin/settingtypes.txt
```diff
@@ -1819,7 +1819,7 @@ mgfractal_np_cave2 (Cave2 noise) noise_params_3d 0, 12, (67, 67, 67), 10325, 3,
 #    'vary_river_depth': If enabled, low humidity and high heat causes rivers
 #    to become shallower and occasionally dry.
 #    'altitude_dry': Reduces humidity with altitude.
```
paramat: Needs a comment explaining that 'canyons' uses 3D noise for valley structure.

src/mapgen/mapgen_valleys.cpp
```diff
@@ -77,11 +78,20 @@ MapgenValleys::MapgenValleys(MapgenValleysParams *params, EmergeManager *emerge)
 	//// 2D Terrain noise
 	noise_filler_depth       = new Noise(&params->np_filler_depth,       seed, csize.X, csize.Z);
 	noise_inter_valley_slope = new Noise(&params->np_inter_valley_slope, seed, csize.X, csize.Z);
-	noise_rivers             = new Noise(&params->np_rivers,             seed, csize.X, csize.Z);
 	noise_terrain_height     = new Noise(&params->np_terrain_height,     seed, csize.X, csize.Z);
 	noise_valley_depth       = new Noise(&params->np_valley_depth,       seed, csize.X, csize.Z);
 	noise_valley_profile     = new Noise(&params->np_valley_profile,     seed, csize.X, csize.Z);
 
+	//// River noise, 2D or 3D
+	if (spflags & MGVALLEYS_CANYONS)
```
paramat: Needs a '{'
Don't mix use of braces and no use of braces in a conditional.

src/mapgen/mapgen_valleys.cpp
```diff
@@ -77,11 +78,20 @@ MapgenValleys::MapgenValleys(MapgenValleysParams *params, EmergeManager *emerge)
...
+		noise_rivers     = new Noise(&params->np_rivers, seed,
+				csize.X, csize.Y + 2, csize.Z);
+	else {
```
paramat: Needs a '}'

src/mapgen/mapgen_valleys.cpp
```diff
@@ -109,7 +119,7 @@ MapgenValleysParams::MapgenValleysParams():
 	np_filler_depth       (0.0,   1.2,  v3f(256,  256,  256),  1605,  3, 0.5,  2.0),
 	np_inter_valley_fill  (0.0,   1.0,  v3f(256,  512,  256),  1993,  6, 0.8,  2.0),
 	np_inter_valley_slope (0.5,   0.5,  v3f(128,  128,  128),  746,   1, 1.0,  2.0),
-	np_rivers             (0.0,   1.0,  v3f(256,  256,  256),  -6050, 5, 0.6,  2.0),
+	np_rivers             (0.0,   1.0,  v3f(256,  256,  256),  -6050, 5, 0.6,  2.0, NOISE_FLAG_EASED),
```
paramat: Eased 3D noise is intensive compared to normal 3D noise, just so you know. However i'm not insisting on non-eased noise. But with 'canyons' enabled valleys mapgen will become more intensive than v7.

src/mapgen/mapgen_valleys.cpp
```diff
@@ -270,25 +280,31 @@ void MapgenValleys::makeChunk(BlockMakeData *data)
...
-	float tv = std::fmax(river / n_valley_profile, 0.0f);
-	float valley_h = valley_d * (1.0f - std::exp(-tv * tv));
-	float surface_y = base + valley_h;
-	float slope = n_slope * valley_h;
+	float surface_y = base;
 	float river_y = base - 1.0f;
+	float slope = 0.0f;
+
+	if (spflags & MGVALLEYS_CANYONS) {
+		float n_rivers = NoisePerlin3D(&noise_rivers->np, p.X, std::floor(base), p.Y, seed);
```
paramat: I'm not sure this is ideal. Even with 'canyons' enabled, this is meant to keep spawn away from the width of the river channel as if it were 2D noise.
I would prefer this uses the existing code for both cases for simplicity. In the 3D case it's fine for it to just roughly keep spawn away from canyon areas.

paramat:
> I'm wondering whether doing if (spflags & MGVALLEYS_CANYONS) results in a significant loss of time in the 3D loop, in which case I will convert it into a boolean outside the loop. What do you think?

Don't worry about that, hmmmm told me that flags check is extremely lightweight.

builtin/settingtypes.txt
```diff
@@ -1819,7 +1819,7 @@ mgfractal_np_cave2 (Cave2 noise) noise_params_3d 0, 12, (67, 67, 67), 10325, 3,
...
-mgvalleys_spflags (Mapgen Valleys specific flags) flags altitude_chill,humid_rivers,vary_river_depth,altitude_dry altitude_chill,noaltitude_chill,humid_rivers,nohumid_rivers,vary_river_depth,novary_river_depth,altitude_dry,noaltitude_dry
+mgvalleys_spflags (Mapgen Valleys specific flags) flags altitude_chill,humid_rivers,vary_river_depth,altitude_dry,canyons altitude_chill,noaltitude_chill,humid_rivers,nohumid_rivers,vary_river_depth,novary_river_depth,altitude_dry,noaltitude_dry,canyons,nocanyons
```
paramat: Enabled by default? Not sure about this, the mapgen is very popular as it is.

paramat:
First test:
There are a lot of extreme overhangs in the river bed in gentle terrain. The riverbeds often become narrow too. This is often ugly, is excessive and unacceptable to me.

screenshot_20190602_223501

screenshot_20190602_224045

screenshot_20190602_223705

However the valley sides have some nice structures:

screenshot_20190602_223832

Is it possible to leave riverbeds as they are and only have the 3D noise affect the valley sides?

builtin/settingtypes.txt
Gael-de-Sailly: I perfectly agree. If it is enabled by default, it's a mistake.

Gael-de-Sailly:
I'm aware of this, but it's hard to change the shape of riversides without affecting the riverbeds. I don't find these riverbeds so ugly but that's a matter of taste.

You may get better results by increasing `spread.Y` in `mgvalleys_np_rivers`. Would generate potentially larger, but rarer cliffs at valley sides (what I mean by "canyons") and less extreme riverbeds. We could discuss about what setting should be the default but I'm afraid a different setting for `spread.Y` would cause compatibility issues because of the way `spread` is interpreted (see my previous post).

Gael-de-Sailly:
I figured out how to give the riverbeds their original shapes in my Lua, it's a bit tricky and can't be done cleanly using the noisemap (I had to rely on `minetest.get_perlin`). Will see how I could implement this in C++ but it will be for another day.

src/mapgen/mapgen_valleys.cpp
Gael-de-Sailly: Non-eased 3D noise doesn't look nice for what I've tested. And this makes it more consistent with the 2D version that is eased by default.

Gael-de-Sailly:
The riverbed thing is done. However this has the side effect of removing the deep water ponds from the 3D version.

paramat:
Thanks will retest.

paramat:
Sorry for delay, i flew for a few 1000 nodes to test the terrain.
In the mountainous terrain the river valleys look like the following screenshots:

screenshot_20190825_211553

screenshot_20190825_211658

screenshot_20190825_211829

screenshot_20190825_211904

screenshot_20190825_212117

screenshot_20190825_212405

Taken in isolation this is mildly interesting, weird and alien. But it is quite extreme and very out of place in Mapgen Valleys which is liked so much because of the somewhat realistic valleys, the lack of crazy 3D noise and floating rocks.
Also, this weird terrain is everywhere in the mountainous terrain, so the standard mountain valleys are much missed.

I think this is an interesting experiment, but is out of character and unsuitable for the mapgen, which surprises me since you are the author of the original.
It really belongs in the Lua version, which is freer and can be more experimental.
:-1:

Gael-de-Sailly:
Thank you for your reviewing, I understand your criticisms.

Note that I used to test it with my custom parameters and it wasn't that extreme. Increasing the Y spread of `mgvalleys_np_rivers` should already give more realistic patterns. The issue is that this noise is at `(256, 256, 256)` by default and it will still appear like your screenshots for most users. I could have duplicated the noise, but I avoided it because mapgen params are already complicated enough without this.

I think it's worth testing with `mgvalleys_np_rivers.spread = (256, 512, 256)`, still not sure you will like it. If not, I won't insist to get it merged, as most of players are already happy with the current VMG.

PR #8097, opened 2019-01-12
Speed up the craft definition handling

osjc:
The craft definition handling code that collects the names of
the craftable nodes suffers from vector reallocation
performance hits, slowing down instances with lots of
crafting recipes (VanessaE's DreamBuilder and most public
server some to my mind when thinking about this). As in each
instance the size of the resulting vector is already known,
add a reserve() call before the offending loops to allocate
the needed chunk of memory within the result vector in one
go, getting rid of the overhead.


osjc:
Didn't notice this until SmallJoker pointed this out.

osjc:
Clarified the commit message a bit.

PR #8098, opened 2019-01-12
Optimize string handling in path search

osjc:
Use "+=" to construct the various game paths. Use a temporary
to extract and reuse a result of a few common subexpressions
to further reduce the overhead.

### DIFF
--- a/src/content/subgames.cpp
+++ b/src/content/subgames.cpp
@@ -67,15 +67,19 @@ SubgameSpec findSubgame(const std::string &id)
 	std::vector<GameFindPath> find_paths;
 	while (!search_paths.at_end()) {
 		std::string path = search_paths.next(PATH_DELIM);
-		find_paths.emplace_back(path + DIR_DELIM + id, false);
-		find_paths.emplace_back(path + DIR_DELIM + id + "_game", false);
+		path.append(DIR_DELIM).append(id);
+		find_paths.emplace_back(path, false);
+		path.append("_game");
+		find_paths.emplace_back(path, false);
 	}
-	find_paths.emplace_back(
-			user + DIR_DELIM + "games" + DIR_DELIM + id + "_game", true);
-	find_paths.emplace_back(user + DIR_DELIM + "games" + DIR_DELIM + id, true);
-	find_paths.emplace_back(
-			share + DIR_DELIM + "games" + DIR_DELIM + id + "_game", false);
-	find_paths.emplace_back(share + DIR_DELIM + "games" + DIR_DELIM + id, false);
+
+	std::string game_base = DIR_DELIM;
+	game_base = game_base.append("games").append(DIR_DELIM).append(id);
+	std::string game_suffixed = game_base + "_game";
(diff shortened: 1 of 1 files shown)


osjc:
The new code is pretty assemblyish but that is what happens when you try to optimize a bunch of expressions that mix char[some-random-constant] with std:string.

src/content/subgames.cpp
```diff
@@ -67,15 +67,22 @@ SubgameSpec findSubgame(const std::string &id)
 	std::vector<GameFindPath> find_paths;
 	while (!search_paths.at_end()) {
 		std::string path = search_paths.next(PATH_DELIM);
-		find_paths.emplace_back(path + DIR_DELIM + id, false);
-		find_paths.emplace_back(path + DIR_DELIM + id + "_game", false);
+		path += DIR_DELIM;
```
nerzhul: I prefer path.append(DIR_DELIM).append(id) to prevent multiple useless lines

src/content/subgames.cpp
```diff
@@ -67,15 +67,22 @@ SubgameSpec findSubgame(const std::string &id)
...
 	}
-	find_paths.emplace_back(
-			user + DIR_DELIM + "games" + DIR_DELIM + id + "_game", true);
-	find_paths.emplace_back(user + DIR_DELIM + "games" + DIR_DELIM + id, true);
-	find_paths.emplace_back(
-			share + DIR_DELIM + "games" + DIR_DELIM + id + "_game", false);
-	find_paths.emplace_back(share + DIR_DELIM + "games" + DIR_DELIM + id, false);
+
+	std::string game_base = DIR_DELIM;
+	game_base += "games";
```
nerzhul: same here

src/content/subgames.cpp
osjc: Tested, works, so here you are.

osjc:
A less assemblyish version suggested by nerzhul. It works too so let it be.

src/content/subgames.cpp
```diff
@@ -67,15 +67,19 @@ SubgameSpec findSubgame(const std::string &id)
 	std::vector<GameFindPath> find_paths;
 	while (!search_paths.at_end()) {
 		std::string path = search_paths.next(PATH_DELIM);
-		find_paths.emplace_back(path + DIR_DELIM + id, false);
-		find_paths.emplace_back(path + DIR_DELIM + id + "_game", false);
+		path = path.append(DIR_DELIM).append(id);
+		find_paths.emplace_back(path, false);
+		path = path.append("_game");
```
nerzhul: you don't need to affect the reference. Append modify the path object itself

http://www.cplusplus.com/reference/string/string/append/

osjc: At least SOMETHING behaves normally in C++. As a matter of fact, I disliked the "path = path.blabla" part of the new lines. Good to see that I can get rid of them.

osjc: Ok. Got rid of the pesky "path = path. ...".

PR #8123, opened 2019-01-21
Fix thin strips at node edges

### LINKED ISSUE #6860: White stripes when fsaa is enabled
When using fsaa (here 2x), there's no proper edge between two polygons of nodebox nodes:
screenshot_20171230_150038

Without fsaa there's no white stripe:
screenshot_20171230_150123

As numberZero wrote, this happens because the neighbouring texel is used by fsaa. If it is transparent, the edge becomes half-transparent after blending.

numberZero:
Fix #6860. This uses different approaches for nodeboxes and solid nodes: for the latter, disabling tiling should be enough, while for the former, the texture has to be stretched as face edge is not the texture edge. I hope such tiny offset is enough as the effect was at sub-pixel level; can’t test, however, as FSAA is not supported on my current GPU.
@HybridDog

### DIFF
--- a/src/client/content_mapblock.cpp
+++ b/src/client/content_mapblock.cpp
@@ -67,7 +67,7 @@ MapblockMeshGenerator::MapblockMeshGenerator(MeshMakeData *input, MeshCollector 
 	nodedef   = data->m_client->ndef();
 	meshmanip = RenderingEngine::get_scene_manager()->getMeshManipulator();
 
-	std::string clean_nodebox_textures = g_settings->get("clean_nodebox_textures");
+	const std::string &clean_nodebox_textures = g_settings->get("clean_nodebox_textures");
 	if (clean_nodebox_textures == "enable") {
 		stretch_nodebox_textures = true;
 	} else if (clean_nodebox_textures == "disable") {
--- a/src/client/content_mapblock.cpp
+++ b/src/client/content_mapblock.cpp
@@ -227,7 +227,7 @@ void MapblockMeshGenerator::drawCuboid(const aabb3f &box,
 		2, 6, 4, 0
 	};
 
-	static const f32 texture_offset = 1.f / 8192.f;
+	static constexpr f32 texture_offset = 1.f / 8192.f;
 
 	for (int face = 0; face < 6; face++) {
 		int tileindex = MYMIN(face, tilecount - 1);
--- a/src/client/content_mapblock.cpp
+++ b/src/client/content_mapblock.cpp
@@ -67,12 +67,19 @@ MapblockMeshGenerator::MapblockMeshGenerator(MeshMakeData *input, MeshCollector 
 	nodedef   = data->m_client->ndef();
 	meshmanip = RenderingEngine::get_scene_manager()->getMeshManipulator();
 
-	bool fsaa = g_settings->getU16("fsaa") > 1;
-	bool filtering = g_settings->getBool("bilinear_filter") || g_settings->getBool("trilinear_filter");
+	std::string clean_nodebox_textures = g_settings->get("clean_nodebox_textures");
+	if (clean_nodebox_textures == "enable") {
+		stretch_nodebox_textures = true;
+	} else if (clean_nodebox_textures == "disable") {
+		stretch_nodebox_textures = false;
+	} else {
+		bool fsaa = g_settings->getU16("fsaa") > 1;
+		bool filtering = g_settings->getBool("bilinear_filter") || g_settings->getBool("trilinear_filter");
+		stretch_nodebox_textures = fsaa && !filtering;
+	}
 
 	enable_mesh_cache = g_settings->getBool("enable_mesh_cache") &&
 		!data->m_smooth_lighting; // Mesh cache is not supported with smooth lighting
-	stretch_nodebox_textures = fsaa && !filtering;
 
 	blockpos_nodes = data->m_blockpos * MAP_BLOCKSIZE;
 }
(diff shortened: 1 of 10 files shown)


paramat:
Have you profiled the performance impact by disabling fast faces? This might have to have a setting, if not insignificant i would personally prefer a few bright pixels and the higher performance.

Is it possible to visually see the stretched textures of a nodebox crossing at edges if close enough?

numberZero:
> Have you profiled the performance impact by disabling fast faces?

Ugh... thanks for asking that. I forgot to disable fast faces, while that’s mandatory for the fix to work on solid nodes. Certainly that needs to be optional as that increases face count. But should also fix #4912 =)

> and stretching adds calculations to it

`drawCuboid()` profile is low, despite the fact it uses heavy math already (`v2f::rotateBy`), much heavier than my addition.

paramat:
Sorry, i was actually confused and thought this was a fix for #4912 only, will read up on #6860

paramat:
The texture stretching only happens with fsaa enabled, fsaa is 'experimental', intensive and off by default, so i'm not concerned with a very small additional performance impact.

paramat:
Just one thought: With slightly stretched nodebox faces, faces of nodeboxes next to each other will overlap slightly and might z-fight. If the textures are different colours (like black and white) this might cause visible z-fighting along nodebox edges? Which might look worse than the bug fixed.

src/client/content_mapblock.cpp
```diff
@@ -216,6 +220,8 @@ void MapblockMeshGenerator::drawCuboid(const aabb3f &box,
 		2, 6, 4, 0
 	};
 
+	static const f32 texture_offset = 1.f / 8192.f;
```
nerzhul: static const constexpr (to calculate at compilation time)

numberZero:
@paramat The faces aren’t stretched. The textures are. So, if there 
*were* overlapping faces, they may start to z-fight due to different 
stretching, but not otherwise. (That doesn’t affect overlays as they are 
created later)

 > fsaa is <...> intensive <...> so i'm not concerned with a very small 
additional performance impact.

Note that while FSAA affects rendering time, this patch (the nodebox 
part) affects mesh generation time only.

paramat:
> The faces aren’t stretched. The textures are.

I see, thanks, that's no problem then.

src/client/content_mapblock.cpp
```diff
@@ -67,6 +67,17 @@ MapblockMeshGenerator::MapblockMeshGenerator(MeshMakeData *input, MeshCollector
 	nodedef   = data->m_client->ndef();
 	meshmanip = RenderingEngine::get_scene_manager()->getMeshManipulator();
 
+	std::string clean_nodebox_textures = g_settings->get("clean_nodebox_textures");
```
nerzhul: const std::string &

src/client/content_mapblock.cpp
numberZero: `constexpr` implies `const`

src/client/content_mapblock.cpp
numberZero: Done.
By the way, just noticed that `Settings` is not entirely thread-safe: concurrent access and change of the same entry is unsafe.

HybridDog:
I've tested it with nodeboxes and there are still stripes, but it looks better than before:
screenshot_20190125_074936
screenshot_20190125_074948

The offset I implemented there worked for me (`1.0f / (16.0f * 8)`) but `1.0f / (16.0f * 16)` did not work.
https://github.com/HybridDog/minetest/commit/63b70243b084a7b1cd805678a5a6f9998343ff2b#diff-15a21807c0c1462b310d02ab1a16763bR264

numberZero:
1/64 is too much:
screenshot_20190126_004634
(view image at full resolution)
Try tuning the offset at [content_mapblock.cpp:230](https://github.com/minetest/minetest/pull/8123/files#diff-15a21807c0c1462b310d02ab1a16763bR230)
Also note that some hardware (or maybe IrrLicht itself?) is buggy and may add a constant offset, for example:
screenshot_20190126_013116
That’s *not* stretching, but actually an offset, equal for all vertices.

HybridDog:
With 1/512, the lines are not visible when standing next to the node. This may be a suitable compromise:
2019-01-26-105737_1920x1056_scrot
2019-01-26-105802_1920x1056_scrot
The stripes are easier to notice from far a way, when having the camera next to the edge, the stripe disappears. Maybe it's possible to use the vertex shader to change the offset during runtime, but this could result in too much code for a workaround.
Maybe the window resolution also has an impact on when the stripes appear.

I'm not sure if the current clean_nodebox_textures is useful. Without FSAA you don't need texture stretching and with FSAA you usually wouldn't want stripes.
You could change this setting to a float where the player can manually set/change texture_offset. -1 means disabled and it's only enabled when stripes can appear (FSAA and no filtering).

numberZero:
> and it's only enabled when stripes can appear (FSAA and no filtering).

To clarify: you still need to fix (pad) your nodebox textures. Because otherwise filtering will break them, in one way or another. And stretching may break filtering (making edges mismatch), so can’t be relied upon.

This fix is more for textures that can’t be fixed themselves, like full node textures used on normal nodes, stairs, slabs etc.

paramat:
@numberZero i added 'WIP' but not sure if it is anymore.
Also, have you stopped working on your PRs? If so please let us know so we can mark your PRs as 'adoption needed'.

paramat:
No progress for 10 months, marking as 'adoption needed' as is a bugfix that may be desired.

HybridDog:
I have found a (in my opinion) much better fix: Use the `centroid` glsl keyword.
https://github.com/HybridDog/minetest/commit/e15436f8abcc50d416c4af92fd70a7a8b72f987b?w=1
It only works when shaders are enabled; this is not a problem because people with slow old hardware probably do not use MSAA.
centroid

Issue #8118, opened 2019-01-21
Send `Referer: server_address` to remote media Curl requests


sofar:
The Remote Media server protocol currently doesn't send a `Referer:` header. I would like it to do this, because this will allow the PRMS to distinguish between various uses and abuses:

- distinguish between authorized and non-authorized Minetest servers on the HTTP/mtmediasrv side
- possibly ban clients from origins such as non-minetest servers or advertisement supported servers
- get usage statistics to improve service

If you're worried about privacy: 
- including the referer header would send the server address, not the player name, to the remote media server
- we can consider a minetest.conf option to disable this (but clients that don't sent a referer can still be on the remote media server side)
 
I can easily find how to insert a header myself in the code, but as always, I have no friggen idea how to get the std::string address from the client startup code to the httpfetch function. So, if @nerzhul or someone else knows this, please drop a hint so I can write a functional patch:

```
diff --git a/src/client/clientmedia.cpp b/src/client/clientmedia.cpp
index e3ad92db..7fcc8882 100644
--- a/src/client/clientmedia.cpp
+++ b/src/client/clientmedia.cpp
@@ -254,6 +254,8 @@ void ClientMediaDownloader::initialStep(Client *client)
                        fetch_request.post_data = required_hash_set;
                        fetch_request.extra_headers.emplace_back(
                                "Content-Type: application/octet-stream");
+                       fetch_request.extra_headers.emplace_back(
+                               std::string("Referer: " + server_address); // how to get a handle on the client address here?
                        httpfetch_async(fetch_request);
 
                        m_httpfetch_active++;
```

nerzhul:
I'm fine with it, but you should also include the server port to be consistent :)

sofar:
@nerzhul, yes but, how can I get a reference to it at this point of code?

srifqi:
Hint: `client->getServerAddress()`

PR #8135, opened 2019-01-26
Add referer to remote media requests.

sofar:
This sends the following header to a remote media server:

    Referer: minetest://<server_name>:port

This was verified with CTF and the Minetest Public Remote Media
server.

### DIFF
--- a/src/client/clientmedia.cpp
+++ b/src/client/clientmedia.cpp
@@ -254,6 +254,16 @@ void ClientMediaDownloader::initialStep(Client *client)
 			fetch_request.post_data = required_hash_set;
 			fetch_request.extra_headers.emplace_back(
 				"Content-Type: application/octet-stream");
+
+			// Encapsulate possible IPv6 plain address in []
+			std::string addr = client->getAddressName();
+			if (addr.find(":", 0) != std::string::npos)
+				addr = '[' + addr + ']';
+			fetch_request.extra_headers.emplace_back(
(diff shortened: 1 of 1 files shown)


sofar:
Implements #8118

src/client/clientmedia.cpp
```diff
@@ -254,6 +254,10 @@ void ClientMediaDownloader::initialStep(Client *client)
 			fetch_request.post_data = required_hash_set;
 			fetch_request.extra_headers.emplace_back(
 				"Content-Type: application/octet-stream");
+			fetch_request.extra_headers.emplace_back(
+				std::string("Referer: minetest://") +
+				client->getAddressName() + ":" +
```
sfan5: this looks like it wouldn't account for a raw IPv6 which needs to be enclosed in `[]` to form a valid URI

sofar: do people do that? lol

I suppose I can use the methods to determine if it is ipv6, but testing if it's raw would, I think, mean parsing the actual address string if I'm not mistaken.

I'll look at the Address class methods...

sofar: eh probably testing for `:` would be sufficient... I'll see if i can cook that up.

sofar: Please check this, I think it'll be ok.

nerzhul:
it will be merged after release, thanks

PR #8190, opened 2019-02-07
Fix grayed-out modpacks in world configuration dialog

random-geek:
Previously, enabled modpacks would not show as green in the world configuration screen. This was due to some Lua weirdness causing a `true` value to become `"true"` and causing unexpected behaviour.

Before:
worldconfig_before
After:
worldconfig_after

The first modpack was intentionally left disabled.

### DIFF
--- a/builtin/mainmenu/pkgmgr.lua
+++ b/builtin/mainmenu/pkgmgr.lua
@@ -603,7 +603,11 @@ function pkgmgr.preparemodlist(data)
 				end
 			end
 			if element ~= nil then
-				element.enabled = value ~= "false" and value ~= "nil" and value
+				if value ~= "false" and value ~= "nil" and value then
+					element.enabled = true
+				else
+					element.enabled = false
+				end
 			else
 				core.log("info", "Mod: " .. key .. " " .. dump(value) .. " but not found")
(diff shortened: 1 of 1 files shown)


Fixer-007:
Noticed that problem too..

builtin/mainmenu/pkgmgr.lua
```diff
@@ -603,7 +603,11 @@ function pkgmgr.preparemodlist(data)
 				end
 			end
 			if element ~= nil then
-				element.enabled = value ~= "false" and value ~= "nil" and value
+				if value ~= "false" and value ~= "nil" and value then
```
SmallJoker: Lua CLI:
```
> return "a" and 9 and true
true
```
->
```Lua
element.enabled = value ~= "false" and value ~= "nil" and value and true
```

SmallJoker:
Alternative fix:
https://github.com/minetest/minetest/blob/b7b7791cd199b920edf0db122293660bb2eaa219/builtin/mainmenu/pkgmgr.lua#L308
-> `not rawlist[j].enabled`

SmallJoker:
a809f73

PR #8194, opened 2019-02-08
Replace for loop with call to standard library function

blindley:
This loop makes multiple passes over m_stack (type std::list) in order to remove all elements with a specified value. Replacing the loop with a call to std::list::remove does the same job, but in only one pass.


nerzhul:
strange historical code :)

nerzhul:
we will merge after release, in case of we miss something strange here

PR #8244, opened 2019-02-16
Fix content store crash

pauloue:
Crash occurred when clicking `pback` or `pend` when there were no search results. Reported here: https://github.com/minetest/minetest/issues/8165#issuecomment-460861862.

Also the formspec was getting updated twice (fstk handles this).

@rubenwardy


builtin/mainmenu/dlg_contentstore.lua
```diff
@@ -515,12 +512,11 @@ function store.get_formspec()
 	return table.concat(formspec, "")
 end
 
-function store.handle_submit(this, fields, tabname, tabdata)
+function store.handle_submit(this, fields)
```
pauloue: store is a dialog

Issue #1965, opened 2014-12-14
Copyright Years


srifqi:
I see that _almost all_ files still use **2013** for the **copyright years**.
**Last commit** that change the copyright years: 6d0ea26c2d62c3774ff384cf1bfc2a3372b49a3b by @sfan5 and @PilzAdam 
Should we update it as the release date for **0.4.11** is _December 24th_ **2014**?

PR #8263, opened 2019-02-19
Update copyright notice

archfan7411:
I doubt a more trivial PR is possible, but here it is. I changed the copyright notice from 2018 to 2019.
If this would be best combined into an existing PR, please feel free to close this one. Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Minetest
 
 An InfiniMiner/Minecraft inspired game.
 
-Copyright (C) 2010-2018 Perttu Ahola <celeron55@gmail.com>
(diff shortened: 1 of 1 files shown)


srifqi:
Old discussion about copyright year: #1965
Another PR (closed) about copyright year: #2768

It seems like we haven't made decision about updating copyright year. We have options:
1. Update those every year (same for all files).
2. Update those every year based on commit time range of each file.
3. Remove end year for all files.

archfan7411:
Ah. After reading those discussions I agree with the general opinion that removing the end date would be best...
I suppose I can close this, then. Thanks for the info!

srifqi:
I think we should wait for other devs. Maybe, they have another opinion.

PR #8264, opened 2019-02-20
PlayerSAO: Range-limit values passed to setHP and setBreath

### LINKED ISSUE #8212: Players's breath can be greater than breath_max
##### Issue type
 - Bug report
 
##### Minetest version
```
a809f7361e8a7232b3c9ec241be5adcae96b3b2d
```

##### Summary
When a player's max. breath is reduced below their current breath, the current breath is not capped at the max. breath value. This is a bug because the documentation says:

```
 * `set_breath(value)`: sets players breath
     * values:
         * `0`: player is drowning
         * max: bubbles bar is not shown
         * See [Object properties] for more information
```
So mods that only check if `breath` is equal to `max` (rather than equal or greater) are screwed.

This is also inconsistent with HP. When max. HP are reduced below your current HP, your current HP get reduced as well.

##### Steps to reproduce

Step 1: Use the test mod from #8211 to make your current and max hp/breath easily visible
Step 2: Use `luacmd`
Step 3: `/lua me:set_properties({breath_max=100})`
Step 4: `/lua me:set_breath(100)`
Step 5: `/lua me:set_properties({breath_max=20})`

You now have 100 breath, which is more than your max. breath.

Tested in singleplayer only.

ClobberXD:
Fixes #8212. Tested; works.

### DIFF
--- a/src/content_sao.cpp
+++ b/src/content_sao.cpp
@@ -1334,6 +1334,8 @@ void PlayerSAO::setHP(s32 hp, const PlayerHPChangeReason &reason)
 {
 	s32 oldhp = m_hp;
 
+	hp = rangelim(hp, 0, m_prop.hp_max);
+
 	s32 hp_change = m_env->getScriptIface()->on_player_hpchange(this, hp - oldhp, reason);
 	if (hp_change == 0)
 		return;
@@ -1355,7 +1357,7 @@ void PlayerSAO::setBreath(const u16 breath, bool send)
 	if (m_player && breath != m_breath)
 		m_player->setDirty(true);
 
-	m_breath = MYMIN(breath, m_prop.breath_max);
+	m_breath = rangelim(breath, 0, m_prop.breath_max);
 
 	if (send)
 		m_env->getGameDef()->SendPlayerBreath(this);


ClobberXD:
Ok sounds good. Will do.

ClobberXD:
Now, setting health/breath higher than their respective maximums throw a warning, and limit the value to its max. @nerzhul

ClobberXD:
Good idea. The message would be the following when a mod tries to set an HP value of, say 50, when `hp_max` is 20:

`WARNING[Server]: Attempting to set HP (50) higher than hp_max (20).`

nerzhul:
Okay for me. the lint failure is in another file, not important there

SmallJoker:
Why exactly is a warning message needed here? Mods can read the maximal values easily from the object properties; invalid values should be clamped automatically (with this PR now also for breath).

ClobberXD:
True, that. Should I remove the numbers from the warnings, or remove the warnings altogether?

SmallJoker:
@ClobberXD I would prefer to remove them entirely. It's not helpful at all for players and server admins. Modders will understand why the breath cannot be higher than the maximal value.

ClobberXD:
Done.

src/script/lua_api/l_object.cpp
```diff
@@ -252,6 +252,11 @@ int ObjectRef::l_set_hp(lua_State *L)
 		reason.lua_reference = luaL_ref(L, LUA_REGISTRYINDEX);
 	}
 
+	// Abort operation if attempting to set HP higher than hp_max
+	int hp_max = co->accessObjectProperties()->hp_max;
```
SmallJoker: Please extend `PlayerSAO::setHP` with this check, preferably before this Lua callback is executed:
https://github.com/minetest/minetest/blob/ae1caba6aaa6f0a13e70f195f371fd4e7a378491/src/content_sao.cpp#L1325

src/script/lua_api/l_object.cpp
```diff
@@ -729,16 +734,25 @@ int ObjectRef::l_set_properties(lua_State *L)
...
 		PlayerHPChangeReason reason(PlayerHPChangeReason::SET_HP);
 		co->setHP(prop->hp_max, reason);
 		if (co->getType() == ACTIVEOBJECT_TYPE_PLAYER)
 			getServer(L)->SendPlayerHPOrDie((PlayerSAO *)co, reason);
 	}
+
+	if (co->getType() == ACTIVEOBJECT_TYPE_PLAYER) {
+		PlayerSAO *playersao = getplayersao(ref);
+		if (prop->breath_max < playersao->getBreath())
+			playersao->setBreath(prop->breath_max);
```
SmallJoker: Move this code to `PlayerSAO::setBreath`; range limit `breath` there.
https://github.com/minetest/minetest/blob/ae1caba6aaa6f0a13e70f195f371fd4e7a378491/src/content_sao.cpp#L1343

src/script/lua_api/l_object.cpp
ClobberXD: Dunno, I thought it's easier to implement in l_object.cpp

Is there a reason to add this to PlayerSAO::setup instead? :) 

Also note that this will then only apply to players, and entities would be able to bypass *_max properties.

SmallJoker: @ClobberXD That's the idea of it. `hp_max` is only for players. For entities the mods control the maximal health (if it ever regains).

ClobberXD: `hp_max` is a part of object properties; shouldn't it apply to all ObjectRefs?

SmallJoker: Question is whether it would make sense. Mobs don't gain health by the engine - their HP is entirely controlled by mods, so I don't really see a practical use for that.

ClobberXD: Fair point... will move the checks to content_sao.cpp

src/script/lua_api/l_object.cpp
ClobberXD: Done

src/script/lua_api/l_object.cpp
ClobberXD: Done

ClobberXD:
@SmallJoker The changes you requested have been completed. This PR is now super-trivial with only two lines changed.

PR #8271, opened 2019-02-22
Fix breath statbar scaling; defer breath_bar hiding by one second

### LINKED ISSUE #8210: Breath statbar might show 10.5 bubbles for high breath values
##### Issue type
 - Bug report
 
##### Minetest version
```
f5bdc04ab5b47bd094ba0ced5f086e8fe6b12298
```

##### Description
If you have a high breath value (>30), the breath statbar might show 10.5 bubbles.

##### How to repro

1. Obtain `luacmd`
2. Set your max. breath to 100 or another high value with /lua me:set_properties({breath_max=100})
3. Dive in water until the bubbles appear

There will be 10.5 bubbles shown.

ClobberXD:
`PLAYER_MAX_BREATH_DEFAULT` is set to 11 instead of 10, probably so that all 10 bubbles are shown before the breath bar disappears. See https://github.com/minetest/minetest/issues/8210#issuecomment-465919487 for more details.

This PR sets `PLAYER_MAX_BREATH_DEFAULT` to 10, and hides the breath bar after a delay of one second, so that all 10 bubbles are visible.

Fixes #8210. Tested.

To test:

- Start losing breath in a drown-able liquid
- Regain breath
- Observe that all 10 bubbles are shown, before the breath bar disappears a second later

### DIFF
--- a/builtin/game/statbars.lua
+++ b/builtin/game/statbars.lua
@@ -3,22 +3,22 @@ local enable_damage = core.settings:get_bool("enable_damage")
 
 local health_bar_definition = {
 	hud_elem_type = "statbar",
-	position = { x=0.5, y=1 },
+	position = {x = 0.5, y = 1},
 	text = "heart.png",
 	number = core.PLAYER_MAX_HP_DEFAULT,
 	direction = 0,
-	size = { x=24, y=24 },
-	offset = { x=(-10*24)-25, y=-(48+24+16)},
+	size = {x = 24, y = 24},
+	offset = {x = (-10 * 24) - 25, y = -(48 + 24 + 16)},
 }
 
 local breath_bar_definition = {
 	hud_elem_type = "statbar",
-	position = { x=0.5, y=1 },
+	position = {x = 0.5, y = 1},
 	text = "bubble.png",
 	number = core.PLAYER_MAX_BREATH_DEFAULT,
 	direction = 0,
-	size = { x=24, y=24 },
-	offset = {x=25,y=-(48+24+16)},
+	size = {x = 24, y = 24},
+	offset = {x = 25, y= -(48 + 24 + 16)},
 }
 
 local hud_ids = {}
@@ -26,7 +26,7 @@ local hud_ids = {}
 local function scaleToDefault(player, field)
 	-- Scale "hp" or "breath" to the default dimensions
 	local current = player["get_" .. field](player)
-	local nominal = core["PLAYER_MAX_".. field:upper() .. "_DEFAULT"]
+	local nominal = core["PLAYER_MAX_" .. field:upper() .. "_DEFAULT"]
 	local max_display = math.max(nominal,
 		math.max(player:get_properties()[field .. "_max"], current))
 	return current / max_display * nominal
@@ -49,6 +49,7 @@ local function update_builtin_statbars(player)
 	local hud = hud_ids[name]
 
 	local immortal = player:get_armor_groups().immortal == 1
+
 	if flags.healthbar and enable_damage and not immortal then
 		local number = scaleToDefault(player, "hp")
 		if hud.id_healthbar == nil then
@@ -63,19 +64,28 @@ local function update_builtin_statbars(player)
 		hud.id_healthbar = nil
 	end
 
+	local show_breathbar = flags.breathbar and enable_damage and not immortal
+
+	local breath     = player:get_breath()
 	local breath_max = player:get_properties().breath_max
-	if flags.breathbar and enable_damage and not immortal and
-			player:get_breath() < breath_max then
+	if show_breathbar and breath <= breath_max then
 		local number = 2 * scaleToDefault(player, "breath")
-		if hud.id_breathbar == nil then
+		if not hud.id_breathbar and breath < breath_max then
 			local hud_def = table.copy(breath_bar_definition)
 			hud_def.number = number
 			hud.id_breathbar = player:hud_add(hud_def)
-		else
+		elseif hud.id_breathbar then
 			player:hud_change(hud.id_breathbar, "number", number)
 		end
-	elseif hud.id_breathbar then
-		player:hud_remove(hud.id_breathbar)
+	end
+
+	if hud.id_breathbar and (not show_breathbar or breath == breath_max) then
+		minetest.after(1, function(player_name, breath_bar)
+			local player = minetest.get_player_by_name(player_name)
+			if player then
+				player:hud_remove(breath_bar)
+			end
+		end, name, hud.id_breathbar)
 		hud.id_breathbar = nil
 	end
 end
--- a/builtin/game/constants.lua
+++ b/builtin/game/constants.lua
@@ -24,7 +24,7 @@ core.MAP_BLOCKSIZE = 16
 -- Default maximal HP of a player
 core.PLAYER_MAX_HP_DEFAULT = 20
 -- Default maximal breath of a player
-core.PLAYER_MAX_BREATH_DEFAULT = 11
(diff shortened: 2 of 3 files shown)


ClobberXD:
At the moment, the breath bar does disappear once full. Should I make it permanently visible?

nerzhul:
breath bar always disappear when full :)

ClobberXD:
> breath bar always disappear when full :)

I know, but doesn't that seem inconsistent with the health bar? If this behaviour is correct, shouldn't the health bar disappear when the player has full health as well?

nerzhul:
i don't think. Health should be shown every time like in many RPG, and breath is only shown when consumed generally, the current behaviour is fine and natural for me

paramat:
Breath bar disappearing is ok for me, less visual obstruction.

jastevenson303:
it's humorous to me that we are going from showing one half too many
bubbles to showing one half too few.  an improvement nonetheless.

SmallJoker:
Usually you don't care about breath, thus hiding it frees some space for the relevant game screen. I'd like to keep the old behaviour; internally the maximal value could be set to 1 higher so that the HUD may disappear correctly.

ClobberXD:
Makes sense. What do you suggest I do now?

builtin/game/statbars.lua
```diff
@@ -30,8 +30,8 @@ local function scaleToDefault(player, field)
 	local current = player["get_" .. field](player)
 	local nominal = core["PLAYER_MAX_".. field:upper() .. "_DEFAULT"]
```
SmallJoker: How about:
```suggestion
	local nominal = core["PLAYER_MAX_".. field:upper() .. "_DEFAULT"]
	if field == "breath" then
		-- Show up/scale to 10 bubbles. Default 11 makes the bar disappear.
		nominal = nominal - 1
	end
```

paramat:
I don't think 10 bubbles should ever appear, if 10 is full then 10 should be invisible. No bar means 10. This 11 thing is silly and messy.

jastevenson303:
I disagree.  Having the bubbles statbar disappear when only at nine bubbles
looks awkward.  I suggest you try it and see for yourself.

paramat:
I don't need to try it i can imagine from current behaviour.
Currently, when at 10 bubbles there is another breath regain cycle as the 11th is gained, then it disappears at 11. I can see why it was done but it was a bad and messy decision because 11 is a weird and inconsistent number.

jastevenson303:
it disappeared at ten, you never saw eleven.

paramat:
I know, that's what i wrote.

jastevenson303:
You wrote that ten shouldn't be visible, which would be a change in current behavior.

ClobberXD:
SmallJoker's suggestion of reducing `nominal` by 1 seems like hard-coding, and there's still quite a bit of wizardry behind the scenes. Maybe we can add support for statbars to automatically disappear when they are full, in 5.1.0. Support in the sense, there won't be the need to use magical numbers to make sure 10 bubbles appear, and the engine would take care of the hiding part without any intervention by us. :)

paramat:
Yes i was proposing 10 not being visible, a change of behaviour.

Yes a more careful implementation is probably needed.
This PR won't be merged for 5.0.0 as it's a low-priority visual-only issue.

ClobberXD:
Any thoughts or suggestions as to what's the best way to do this?

ClobberXD:
On printing a bunch of values, I noticed that the player's breath_max is 11, in spite of having changed `PLAYER_MAX_BREATH_DEFAULT` to 10. MTG doesn't seem to set `breath_max` to any value either. This can't be due to old modifications I made when testing this PR, as it's 11 in a new world as well. I'll see if this is initialized somewhere in the engine...

ClobberXD:
It seems like I just didn't re-compile after setting `PLAYER_MAX_BREATH_DEFAULT` to 10. It works now. All 10 bubbles are displayed, and if the bar is full, it is hidden one second later.

This PR now works, and is ready for review. :)

@SmallJoker

builtin/game/statbars.lua
```diff
@@ -65,19 +67,26 @@ local function update_builtin_statbars(player)
...
+		elseif hud.id_breathbar then
 			player:hud_change(hud.id_breathbar, "number", number)
 		end
-	elseif hud.id_breathbar then
-		player:hud_remove(hud.id_breathbar)
+	end
+
+	-- Hide the breath bar if breathbar HUD flag is set to false,
+	-- player is immortal, or breath == breath_max
+	if hud.id_breathbar and (not flags.breathbar or immortal or breath == breath_max) then
```
SmallJoker: Could you please use a new variable to check whether the player can have a breath bar? This check duplication and inversion feels very easy to forget in future changes.

builtin/game/statbars.lua
```diff
@@ -65,19 +67,26 @@ local function update_builtin_statbars(player)
...
+		minetest.after(1, function(breath_bar)
+			player:hud_remove(breath_bar)
```
SmallJoker: If the player leaves within this second, the reference becomes invalid. Either you'll trigger a warning or crash the server (probably former). The sane solution here would be to check whether the player is still online.

builtin/game/statbars.lua
```diff
@@ -65,19 +67,26 @@ local function update_builtin_statbars(player)
...
 
+	local breath     = player:get_breath()
 	local breath_max = player:get_properties().breath_max
 	if flags.breathbar and enable_damage and not immortal and
-			player:get_breath() < breath_max then
+			breath <= breath_max then
 		local number = 2 * scaleToDefault(player, "breath")
-		if hud.id_breathbar == nil then
- 			local hud_def = table.copy(breath_bar_definition)
+		if not hud.id_breathbar and breath < breath_max then
```
SmallJoker: Why `breath < breath_max`? Even if it's the same level I think the breath bar should be shown for a second.

ClobberXD: The breath bar would be flashing in and out of existence every time this function runs.

builtin/game/statbars.lua
ClobberXD: Will do.

builtin/game/statbars.lua
ClobberXD: Will do

builtin/game/statbars.lua
ClobberXD: This seems to be tricky - half of the conditions are actually unique, and I was only able to abstract 3 out of 5 conditions into a single variable.

builtin/game/statbars.lua
ClobberXD: Done

ClobberXD:
PR's ready for review (again). Tested by toggling `breathbar` HUD flag and `immortal`, and the statbar disappeared as expected. However, I did notice a delay of more than a second before the statbar disappeared due to the aforementioned actions. This is _probably_ because there's no straight way to detect whether HUD flags or armor groups change, but I just wanted to put it out here.

ClobberXD:
Rebased.

ClobberXD:
Bumpity-bump

ClobberXD:
***Bump***

Added testing instructions to the first post.

ClobberXD:
@SmallJoker Merge?

paramat:
Not really trivial enough for a 1-approval merge.

ClobberXD:
Delays may cause even more delays due to the need for re-testing. Better to merge this IMHO, instead of letting it stagnate, lest we have more PRs modifying the same areas of code and breaking this PR in the process.

ClobberXD:
Thanks for the review!

PR #8272, opened 2019-02-22
Fix typo in automatic forward setting

ClobberXD:
This PR removes an extra `s` from `Automatic forwards`.

I'm still not quite familiar with translations, so do let me know if I did something wrong.

Tested - works in English.

### DIFF
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -244,7 +244,7 @@ keymap_decrease_volume (Dec. volume key) key
 
 #    Key for toggling autoforward.
 #    See http://irrlicht.sourceforge.net/docu/namespaceirr.html#a54da2a0e231901735e3da1b0edf72eb3
-keymap_autoforward (Automatic forwards key) key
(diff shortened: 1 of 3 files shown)


paramat:
'forwards' is good English, but 'forward' is also ok, cleaner and probably easier to translate? So :+1:

paramat:
The 's' version seems to be more a British English thing, and MT uses American English:
"When these words are used as adverbs, they can be used with or without the –s ending. While the forms with –s are more common in British English, the forms without –s are usually used in American English."
:+1:

jastevenson303:
"good English" haha

i think you mean "well" English.

i'll see myself out.

paramat:
Yes writing 'good English' was a mistake, changed that to 'ok English'. For this usage i think no 's' may actually be better English, not sure. Anyway, it's cleaner, shorter, less awkward and more consistent with code and keymap name.

Strange spelling in MT has often been my fault, as i am British but MT uses USA English spelling.

PR #8282, opened 2019-02-23
Key settings: Clear with escape

### LINKED ISSUE #8279: Ability to clear key-binds from within the "Change Keys" dialog
##### Issue type

 - Feature request

##### Summary

One way to do this is to keep track of the focused field, and clear the corresponding key-bind when a clear button is pressed.

SmallJoker:
Closes #8279 

**How to test**
1) Open the key change menu
2) Assign keys
3) Remove assignment by pressing ~backspace or~ delete
4) Key is now set to ""

### DIFF
--- a/src/gui/guiKeyChangeMenu.cpp
+++ b/src/gui/guiKeyChangeMenu.cpp
@@ -246,7 +246,13 @@ void GUIKeyChangeMenu::drawMenu()
 bool GUIKeyChangeMenu::acceptInput()
 {
 	for (key_setting *k : key_settings) {
-		g_settings->set(k->setting_name, k->key.sym());
+		std::string default_key;
+		g_settings->getDefaultNoEx(k->setting_name, default_key);
+
+		if (k->key.sym() != default_key)
+			g_settings->set(k->setting_name, k->key.sym());
+		else
+			g_settings->remove(k->setting_name);
 	}
 
 	{
--- a/src/gui/guiKeyChangeMenu.cpp
+++ b/src/gui/guiKeyChangeMenu.cpp
@@ -86,8 +86,6 @@ GUIKeyChangeMenu::GUIKeyChangeMenu(gui::IGUIEnvironment* env,
 GUIModalMenu(env, parent, id, menumgr)
 {
 	init_keys();
-	for (key_setting *ks : key_settings)
-		key_used.push_back(ks->key);
 }
 
 GUIKeyChangeMenu::~GUIKeyChangeMenu()
@@ -112,6 +110,7 @@ void GUIKeyChangeMenu::removeChildren()
 	for (gui::IGUIElement *i : children_copy) {
 		i->remove();
 	}
+	key_used_text = nullptr;
 }
 
 void GUIKeyChangeMenu::regenerateGui(v2u32 screensize)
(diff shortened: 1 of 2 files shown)


ClobberXD:
Works - the keybind is cleared, but it leaves an incomplete line in `minetest.conf` like so:
`keymap_autoforward = `

Note that all key-binds are written to minetest.conf, even if only one of them is changed, and the empty keybinds end up creating incomplete lines. So I don't think this issue is strictly related to this PR, but it'd be nice if you could fix them here. :)

SmallJoker:
@ClobberXD I extended that request to remove all default values from the config file. Thanks for the input.

ClobberXD:
That's great, thanks. Here's how I tested this the second time:

- I removed all `keymap_*` settings from minetest.conf
- Ran MT, set Cinematic to `F8`.
- Opened minetest.conf, and saw a new line `keymap_cinematic = KEY_F8`. :heavy_check_mark: 
- Ran MT again, cleared Cinematic key-bind.
- No trace left behind in minetest.conf :heavy_check_mark: 

However I also noticed three other keybinds that were added to minetest.conf after setting Cinematic keybind:

```
keymap_cmd_local = KEY_PERIOD
keymap_decrease_viewing_range_min = KEY_MINUS
keymap_increase_viewing_range_min = KEY_PLUS
```

AFAIK, these are the default key-binds, and shouldn't have crept into minetest.conf. This could even be my fault (although I don't realize how), but everything else works perfectly fine :+1:

DS-Minetest:
There might be users who want to bind a key to delete or to backspace.

ClobberXD:
True - I was thinking of a separate "Clear" button for clearing the selected keybind. But this is probably a simpler solution - we could just document that the backspace and delete keys can't be assigned to keybinds.

DS-Minetest:
Or what about allowing backspace and delete only if the key has currently its default binding?

SmallJoker:
@ClobberXD A button would overcomplicate the dialogue a lot; since there's barely any space to add another button easily. I'd prefer to keep the delete/backspace behaviour.
@DS-Minetest For me it would be quite unintuitive when the delete/backspace keys have different functionalities, depending on the state of the currently selected setting. It's surely doable but I'd like to hear more player opinions about this.

Fixer-007:
>There might be users who want to bind a key to delete or to backspace.

there is like 100 spare keys, are you sure it is intuitive to bind to those? I guess we can use just backspace for deletion.

DS-Minetest:
> > There might be users who want to bind a key to delete or to backspace.
> 
> there is like 100 spare keys, are you sure it is intuitive to bind to those?

Yes I am. The delete and backspace keys are some of the most used keys, ergo some players might be used to use them.
Furthermore the backspace key is one of the largest keys on many keyboard layouts.

Actually I've counted 68 `keymap_*` settings in `minetest.conf.example`, so 100 keys is not that much as it might seem at first glance. Not to forget that some keyboards don't have a numpad.

I don't think that delete/backspace keys having different functionalities, depending on the state of the currently selected setting, would be too unintuitive. Humans are able to learn and to adapt. The users will learn how this functionality works.

But, as Krock already wrote, I too would
>like to hear more player opinions about this.

NathanSalapat:
Most programs I've used will let you clear keybinds using Esc. I think Delete or Backspace are fine too, I can't really see anybody wanting to use either of those keys for anything in game.

ClobberXD:
> Most programs I've used will let you clear keybinds using Esc.

I agree with this. Most programs, if not all, make use of Esc key for clearing existing keybinds because it can't be a part of a keybind itself - Esc is used to pause the game, or to bring up the pause menu.

Fixer-007:
Hmm, using Esc maybe very good idea actually

SmallJoker:
After some inactivity I now added the ESC key to those which erase the registered key. Does someone really want to use backspace? It seems more intuitive to use that key for erasing as well.

DS-Minetest:
I'm still for having the ability to bind to backspace or delete.

Having more than one key to reset the setting might be as confusing as https://github.com/minetest/minetest/pull/8282#issuecomment-466810429.

DS-Minetest:
:+1: PR looks good.

ClobberXD:
Tested; works :+1:

SmallJoker:
Well so.. I adjusted the PR according to the requests but I'm not happy about the current state.
ESC is meant as "go back, undo local changes" in formspecs, but I'd rather clear by backspace and cancel the key change callback using ESC.

SmallJoker:
Update
1) `activeKey` ID replaced with `key_setting *`. The available keys do not change, either do their pointers
2) Segfault fix when resizing a window while `key_used_text` was set
3) Key ESCAPE -> Cancel change
4) Key DELETE -> Remove entry

PR #8295, opened 2019-02-27
Make FindCURL ready for vcpkg

adrido:
This allows to find and use curl with the vcpkg toolchain file.

There will be another PR regarding vcpkg on Windows, but this is kind of urgent: Never ever copy a file out of the Windows directory and redistribute them!
That might not work from technical aspects (x32 vs x64), it is also very questionable in terms of license.

Last but not least: This also allows to build Minetest on windows with a static linked version of curl.

### DIFF
--- a/cmake/Modules/FindCURL.cmake
+++ b/cmake/Modules/FindCURL.cmake
@@ -1,19 +1,16 @@
 mark_as_advanced(CURL_LIBRARY CURL_INCLUDE_DIR)
 
-find_library(CURL_LIBRARY NAMES curl)
+find_library(CURL_LIBRARY NAMES curl libcurl)
 find_path(CURL_INCLUDE_DIR NAMES curl/curl.h)
 
-set(CURL_REQUIRED_VARS CURL_LIBRARY CURL_INCLUDE_DIR)
-
 if(WIN32)
-	find_file(CURL_DLL NAMES libcurl-4.dll
-		PATHS
-		"C:/Windows/System32"
-		DOC "Path to the cURL DLL (for installation)")
-	mark_as_advanced(CURL_DLL)
-	set(CURL_REQUIRED_VARS ${CURL_REQUIRED_VARS} CURL_DLL)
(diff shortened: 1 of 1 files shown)


cmake/Modules/FindCURL.cmake
```diff
@@ -1,19 +1,16 @@
...
-		PATHS
-		"C:/Windows/System32"
-		DOC "Path to the cURL DLL (for installation)")
-	mark_as_advanced(CURL_DLL)
-	set(CURL_REQUIRED_VARS ${CURL_REQUIRED_VARS} CURL_DLL)
+	# If VCPKG_APPLOCAL_DEPS is ON, dll's are automatically handled by VCPKG
+	if(NOT VCPKG_APPLOCAL_DEPS)
+		find_file(CURL_DLL NAMES libcurl-4.dll libcurl.dll
+			DOC "Path to the cURL DLL (for installation)")
+		mark_as_advanced(CURL_DLL)
```
SmallJoker: Also supply this value to `find_package_handle_standard_args` to ensure CURL is loaded correctly without vcpkg.

adrido: This would prevent using a static build of curl. Do you want that? 
Also, for consistency, the other FindXXX.cmake modules does not fail if the dll is not found.

SmallJoker: Oh, I did not see that. Thanks for pointing out.
For the consistency's sake it's fine this way.

sfan5:
:+1:

MoNTE48:
0 changed files...
https://github.com/minetest/minetest/commit/af728769214b8a0aa9ec0f46494db4a2cf76b7d9

SmallJoker:
Included in bd6f1cc @adrido.
Interestingly this PR did not show up rebase conflicts due to the identical code changes.

adrido:
Sorry, my fault. Should have closed this PR.

PR #8298, opened 2019-02-28
Add math.factorial

HybridDog:
A simple helper function, such as math.sign.


DS-Minetest:
I think, it's called "factorial" in english. (https://en.wikipedia.org/wiki/Factorial)

HybridDog:
> I think, it's called "factorial" in english. (https://en.wikipedia.org/wiki/Factorial)

fixed

PR #8299, opened 2019-02-28
Add table.shuffle

HybridDog:
https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle
Example use case: Randomly shuffle items in a chest inventory

### DIFF
--- a/builtin/common/misc_helpers.lua
+++ b/builtin/common/misc_helpers.lua
@@ -576,6 +576,20 @@ function table.key_value_swap(t)
 end
 
 
+function table.shuffle(t, from, to, random)
+	from = from or 1
+	to = to or #t
+	random = random or math.random
+	local n = to - from + 1
+	while n > 1 do
+		local r = from + n-1
+		local l = from + random(0, n-1)
+		t[l], t[r] = t[r], t[l]
+		n = n-1
+	end
+end
+
+
 --------------------------------------------------------------------------------
 -- mainmenu only functions
 --------------------------------------------------------------------------------
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2930,7 +2930,8 @@ Helper functions
     * `from` defaults to `1`
     * `to` defaults to `#table`
     * `random_func` defaults to `math.random`. This function receives two
-      integers as arguments and should return a random integer between them.
+      integers as arguments and should return a random integer inclusively
+      between them.
 * `minetest.pointed_thing_to_face_pos(placer, pointed_thing)`: returns a
   position.
     * returns the exact position on the surface of a pointed node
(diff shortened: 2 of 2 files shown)


builtin/common/misc_helpers.lua
```diff
@@ -561,6 +561,21 @@ function table.insert_all(t, other)
 end
 
 
+function table.shuffle(t, from, to, rank)
+	from = from or 1
+	to = to or #t
+	local n = to - from + 1
+	rank = rank or math.random(0, math.faculty(n)-1)
```
sfan5: when shuffling a list of 40 things, `40!` would be larger than the largest possible float (3.4E+38)
this function won't work correct in that case

is there any reason you didn't just implement a "plain" shuffle?

HybridDog: When putting items in a chest in random order during mapgen, the order should be the same if the chest is generated again at the same position with the same seed. I didn't consider the rank's size.

doc/lua_api.txt
```diff
@@ -2424,6 +2424,11 @@ Helper functions
 * `table.insert_all(table, other_table)`:
     * Appends all values in `other_table` to `table` - uses `#table + 1` to
       find new indices.
+* `table.shuffle(table, [from], [to], [random_func])`:
+    * Shuffles elements `from` to `to` in `table` in place
+    * `from` defaults to `1`
+    * `to` defaults to `#table`
+    * `random_func` defaults to `math.random`
```
raymoo: I think the documentation should describe what the function parameter should do. it doesn't look like it's required to provide the full functionality of math.random

doc/lua_api.txt
```diff
@@ -2424,6 +2424,12 @@ Helper functions
 * `table.insert_all(table, other_table)`:
     * Appends all values in `other_table` to `table` - uses `#table + 1` to
       find new indices.
+* `table.shuffle(table, [from], [to], [random_func])`:
+    * Shuffles elements `from` to `to` in `table` in place
+    * `from` defaults to `1`
+    * `to` defaults to `#table`
+    * `random_func` defaults to `math.random`. This function receives two
+      integers as arguments and should return a random integer between them.
```
raymoo: I think it should specify that it's inclusive

HybridDog: Done. I thought that people can infer this from how math.random, the default function, works, but it's probably inconvenient for inexperienced persons to look up the documentation.

doc/lua_api.txt
raymoo: You marked this resolved but you didn't make any related change or respond with why it's unnecessary.

HybridDog:
rebased.

PR #8301, opened 2019-03-01
prometheus skeleton

thomasrudin:
PR draft for #8004 

# TODO
* Code style/conventions
* Lua implementation (minetest_engine_* and minetest_mod_* namespace)
* Documentation

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -235,6 +235,15 @@ if(ENABLE_SPATIAL)
 endif(ENABLE_SPATIAL)
 
 
+# prometheus monitoring
+find_package(prometheus-cpp QUIET)
+if(prometheus-cpp_FOUND)
+	set(USE_PROMETHEUS TRUE)
+	add_compile_definitions(ENABLE_PROMETHEUS TRUE)
+	message(STATUS "prometheus-cpp enabled")
+endif()
+
+
 if(NOT MSVC)
 	set(USE_GPROF FALSE CACHE BOOL "Use -pg flag for g++")
 endif()
@@ -412,6 +421,7 @@ set(common_SRCS
 	mapsector.cpp
 	metadata.cpp
 	modchannels.cpp
+	monitoring.cpp
 	nameidmapping.cpp
 	nodedef.cpp
 	nodemetadata.cpp
@@ -554,6 +564,12 @@ if(BUILD_CLIENT)
 		${PLATFORM_LIBS}
 		${CLIENT_PLATFORM_LIBS}
 	)
+
+	# prometheus monitoring
+	if(USE_PROMETHEUS)
+		target_link_libraries(${PROJECT_NAME} prometheus-cpp::core prometheus-cpp::pull)
+	endif()
+
 	if(APPLE)
 		target_link_libraries(
 			${client_LIBS}
@@ -613,6 +629,12 @@ if(BUILD_SERVER)
 		${GMP_LIBRARY}
 		${PLATFORM_LIBS}
 	)
+
+	# prometheus monitoring
+	if(USE_PROMETHEUS)
+		target_link_libraries(${PROJECT_NAME}server prometheus-cpp::core prometheus-cpp::pull)
+	endif()
+
 	set_target_properties(${PROJECT_NAME}server PROPERTIES
 			COMPILE_DEFINITIONS "SERVER")
 	if (USE_CURSES)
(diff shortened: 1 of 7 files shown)


src/server.cpp
```diff
@@ -468,11 +481,14 @@ void Server::step(float dtime)
 		}
 		throw ServerError("AsyncErr: " + async_err);
 	}
+
```
ClobberXD: Please remove this newline if not created on purpose.

src/server.h
```diff
@@ -661,6 +662,12 @@ class Server : public con::PeerHandler, public MapEventReceiver,
 
 	// ModChannel manager
 	std::unique_ptr<ModChannelMgr> m_modchannel_mgr;
+
+	/*
+	Metrics
+	*/
+	Counter* m_tick_count_metric;
+	Counter* m_async_run_count_metric;
```
ClobberXD:
```cpp
Counter *m_tick_count_metric;
Counter *m_async_run_count_metric;
```

src/monitoring.h
```diff
@@ -0,0 +1,50 @@
...
+	virtual void observe(const double value) = 0;
+};
+
+class Monitoring {
+public:
+	virtual void start() = 0;
+	virtual void stop() = 0;
+	virtual Gauge* createGauge(const std::string &name, const std::string &help) = 0;
+	virtual Counter* createCounter(const std::string &name, const std::string &help) = 0;
+	virtual Histogram* createHistogram(const std::string &name, const std::string &help, const std::vector<double> buckets) = 0;
```
ClobberXD: `Gauge *createGauge( ...`

Needs to be corrected in the next two lines as well.

src/server.cpp
```diff
@@ -405,6 +405,17 @@ void Server::init()
 	m_max_chatmessage_length = g_settings->getU16("chat_message_max_size");
 	m_csm_restriction_flags = g_settings->getU64("csm_restriction_flags");
 	m_csm_restriction_noderange = g_settings->getU32("csm_restriction_noderange");
+
+	// Register metrics
+	m_tick_count_metric = g_monitoring->createCounter(
```
nerzhul: useless metric but interesting example

src/CMakeLists.txt
```diff
@@ -412,6 +421,7 @@ set(common_SRCS
 	mapsector.cpp
 	metadata.cpp
 	modchannels.cpp
+	monitoring.cpp
```
nerzhul: compile that only if USE_PROMETHEUS = true

src/monitoring.cpp
```diff
@@ -0,0 +1,162 @@
...
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License along
+with this program; if not, write to the Free Software Foundation, Inc.,
+51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+#if ENABLE_PROMETHEUS
```
nerzhul: unneeded, this should be done at cmake configuration time

thomasrudin:
I'm not done yet. Don't worry... ;)

src/server.cpp
thomasrudin: Yeah, that was a simple test-metric :)

src/CMakeLists.txt
thomasrudin: That won't work: there is a stub-implementation in there for the monitoring if the prometheus library is not found. It is either that or on every metric an `#if PROMETHEUS #endif` clause

thomasrudin:
Can someone please validate my stub-implementation in `prometheus.cpp` and check if thats the right approach?

src/CMakeLists.txt
nerzhul: you should have instead an interface for this implementation and create a stub implementation when not enabled and a normal implementation when enabled.

nerzhul: and also, replace monitoring -> metrology it's a better wording.

nerzhul: inheritance: 
```
class IMetrology { // abstract }
class PrometheusMetrology : public IMetrology {}
class StubMetrology : public IMetrology {}
 ```

thomasrudin: I won't rename that to metrology, sounds too odd... :/

nerzhul:
why close this ?

thomasrudin:
I realized discussing position of braces and naming of things is just not my thing..

Keep up the good work though :+1:

nerzhul:
code style can be fixed with clang-format if it's new files :)

PR #8304, opened 2019-03-02
getS16NoEx() returns true unless syntactical error in conf.

sofar:
The getS16NoEx() handler will return true unless there is a
`[num_emerge_threads]` line in the `minetest.conf` at which
point the excption handler part is reached. Due to the fact that
`defaultsettings.cpp` has a default value set for this setting,
that never will happen.

Because of this, the code will never check the number of threads on
the system, and keep `nthreads = 0`. If that happens, the value is
changed to `1` and only 1 emerge thread will be used.

The default should be set to `1` instead, due to the potential unsafe
consequences for the standard sqlite map files, but that should be a
separate commit that also adds documentation for that setting. This
commit focuses on removing this `hiding` bug instead.

### DIFF
--- a/src/emerge.cpp
+++ b/src/emerge.cpp
@@ -130,11 +130,13 @@ EmergeManager::EmergeManager(Server *server)
 
 	// If unspecified, leave a proc for the main thread and one for
 	// some other misc thread
-	s16 nthreads = 0;
-	if (!g_settings->getS16NoEx("num_emerge_threads", nthreads))
+	s16 nthreads;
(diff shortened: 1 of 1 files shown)


sofar:
BTW, I suspect that *many* settings bugs like this exist. The entire class of getXXNoEx() settings getters are all affected, I think.

SmallJoker:
`get*NoEx()` is most likely though for
1) Legacy settings (removed defaults)
2) Settings without defaults (mapgen)
2) Incomplete group settings (noise params)

For regular minetest.conf settings, `get*()` can be used without any problems as there's always a fallback.

PR #8305, opened 2019-03-02
Don't include and link to gettext if gettext is not found

adrido:
If gettext is enabled (by default) but gettext is not found, cmake fails because it tries to include and link to it.


adrido:
Thank you.

PR #8330, opened 2019-03-06
Fix definitively string serialization coding problems

nerzhul:
NetworkPacket strings are serialized in the same way for short and long, we use long method.

This has side effects the serialization code, short has been updated to use long too.


nerzhul:
putting that to 5.1.0. Also needs to add protocol bump

SmallJoker:
:warning: This PR will break the network compatibility again.
How are you going to make it 5.0 compatible, if even?

nerzhul:
i will think about it in a future version as now we are not 1 day after release. Closing.

PR #8342, opened 2019-03-08
Correctly set visibility of local player's attached objects

PiezoU005F:
This fixes the issue of objects becoming temporarily invisible when attached to the local player while the client isn't in first person view.

I proposed this change before, but I initially had no clue what I was doing, and figuring out the best approach resulted in a few pointless commits, and eventually, the pull request became so far behind the main branch that it might have broken something if it were merged, so I decided it was better to delete it.

This time, I've kept it simple, and based the changes off of the latest version as well.

### DIFF
--- a/src/client/content_cao.cpp
+++ b/src/client/content_cao.cpp
@@ -1531,8 +1531,11 @@ void GenericCAO::processMessage(const std::string &data)
 		setAttachment(parent_id, bone, position, rotation);
 
 		// localplayer itself can't be attached to localplayer
-		if (!m_is_local_player)
-			m_is_visible = !m_attached_to_local;
+		if (!m_is_local_player) {
+			// Objects attached to the local player should be hidden in first person
+			m_is_visible = (!m_attached_to_local) ||
+					(m_client->getCamera()->getCameraMode() != CAMERA_MODE_FIRST);
+		}
 	} else if (cmd == GENERIC_CMD_PUNCHED) {
 		u16 result_hp = readU16(is);
 


ClobberXD:
(Continued from #8046)

Yukitty:
👍 Looks good to me.
I would have preferred `!= CAMERA_MODE_FIRST` over `> CAMERA_MODE_FIRST` personally, but there is no functional difference in the present circumstance afaik.

src/client/content_cao.cpp
```diff
@@ -1502,8 +1502,9 @@ void GenericCAO::processMessage(const std::string &data)
 		// localplayer itself can't be attached to localplayer
 		if (!m_is_local_player) {
 			m_attached_to_local = getParent() != NULL && getParent()->isLocalPlayer();
-			// Objects attached to the local player should be hidden by default
-			m_is_visible = !m_attached_to_local;
+			// Objects attached to the local player should be hidden in first person
+			m_is_visible = (!m_attached_to_local) 
+				|| (m_client->getCamera()->getCameraMode() > CAMERA_MODE_FIRST);
```
nerzhul: why > ? it's error prone if we add more things in the enum fields. I prefer != if possible

PiezoU005F: I suppose it could have been useful if more camera modes were added where the entities shouldn't be shown (in which case, they would be set below CAMERA_MODE_FIRST), but that seems unlikely, in retrospect, so I've changed it.

ClobberXD:
Tested, this PR does fix the child entities disappearing after a little bit in the third-person modes. :+1:

ClobberXD:
Something seems wrong after rebase:

screenshot_20190823_200814

Only the nametag entity is visible, and I can't see the wieldentity or the health gauge now. These entities do indeed exist, as I can see these entities attached to other players.

ClobberXD:
This PR probably doesn't work as expected due to 4aa9a669cb184b77213e8df82eb20eda5aad9004 modifying client-side attachment code. I hope this can be fixed easily.

paramat:
Author seems absent, marked as 'adoption needed' as this is a bugfix that may be desired.

LoneWolfHT:
So this would fix the CTF nametags showing up in third-person?

rubenwardy:
Are you sure that the noted bug isn't due to another reason?

LoneWolfHT:
hah, I realize what this does now. I was confusing it with first-person showing attached objects

LoneWolfHT:
Guess I could look into doing a PR for this now that I know what this fixes. Seems like a pretty simple fix is possible...

SmallJoker:
Author absent. PR needs changes.

cronvel:
I just came across this bug while modding. I lost hours searching for a bug on my side, and just figured out that it was an engine bug... 

FYI, I am working on a mod for managing entities AND players, with skins/models/animation/(more) capabilities. The mod is able to manage *gears* (like wielded item, head-gear, and so on). But it failed to display any wielded items or *gears* once I change the model. I need to move in the 1st person view and return to 3rd person for this to work.

In the famous *wield3d* mod, it happens when wielding the torch, because the torch is attached with a different rotation than tools, so it needs a `set_attach()` call. 

So meanwhile, is there a workaround in Lua? I have already tried to remove all attached entities multiple times to force a sort of *reflow* but without any luck.

PR #8347, opened 2019-03-09
Fix set_hp(0, reason) Lua error

### LINKED ISSUE #8227: set_hp(0, {type="fall"}) causes Lua error
##### Issue type
 - Bug report
 
##### Minetest version
```
Minetest 5.0.0-dev-ffb17f1c (Linux)
```

##### Summary
When set_hp is called to kill the player with a hpchange reasons that already exists in the engine (?), the mod crashes with a Lua error that is missing both description and backtrace.
```
ERROR[Main]: ServerError: AsyncErr: ServerThread::run Lua: Runtime error from mod 'test2' in callback on_joinplayer(): <no description>
```
The call finishes correctly if hp > 0 or the reason is omitted.

##### Steps to reproduce

simple example code:
```lua
	player:set_hp(0, {type="fall"})
end)
```

pauloue:
Fixes #8227.

The fix is this line: https://github.com/minetest/minetest/pull/8347/files#diff-2d01836e6c0329cf2b516efb9b407661R390
That addition seems to necessitate the removal of `const` wherever `PlayerHPChangeReason` is used.

The crash occurred because the index of the reference to the Lua `PlayerHPChangeReason` table was not reset when the reference was deleted. This caused a problem when `pushPlayerHPChangeReason` was called twice with the same `PlayerHPChangeReason`, as is done when a player dies (for `on_player_hpchange` and `on_dieplayer` callbacks).

### DIFF
--- a/src/content_sao.cpp
+++ b/src/content_sao.cpp
@@ -646,8 +646,8 @@ int LuaEntitySAO::punch(v3f dir,
 
 	if (!damage_handled) {
 		if (result.did_punch) {
-			setHP((s32)getHP() - result.damage,
-				PlayerHPChangeReason(PlayerHPChangeReason::SET_HP));
+			PlayerHPChangeReason reason(PlayerHPChangeReason::SET_HP);
+			setHP((s32)getHP() - result.damage, reason);
 
 			if (result.damage > 0) {
 				std::string punchername = puncher ? puncher->getDescription() : "nil";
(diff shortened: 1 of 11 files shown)


pauloue:
I realize now why the Lua reason table is saved in the registry: it's for custom fields. But this doesn't work if the same `PlayerHPChangeReason` is pushed twice. Not sure why it doesn't just allow a custom type field?

@rubenwardy your code

SmallJoker:
The code changes look good, but wouldn't it be possible to keep the inline struct initialization instead of creating a new variable? Just wondering.

pauloue:
I think not because the parameter is a non-const reference. Otherwise I get this error: `cannot bind non-const lvalue reference of type ‘PlayerHPChangeReason&’ to an rvalue of type ‘PlayerHPChangeReason’`

PR #8368, opened 2019-03-12
Add node field to PlayerHPChangeReason table

### LINKED ISSUE #8345: Add node in damage reason table
##### Issue type
 - Feature request

##### Minetest version
```
5.0.0
```

##### Summary
5.0.0 supports damage reasons. One of the damage reasons is node damage. But the `reason` table is missing the node that dealt the damage. Mods still have to guesstimate a bit to actually find the damaging node.

I suggest to add the field `node` to the `reason` table. If `reason.type` equals `"node_damage"`, `node` shall contain the nodename of the node that dealt the damage (e.g. `"default:lava_source"`).

##### Special considerations
After a quick test, I figured out that if a player is in multiple different nodes with different values of `damage_per_second` at the same time, Minetest will deal pick the highest damage and only deal that one. E.g. if the player is in lava (8 dmg) and fire (2 dmg) at the same time, the damge will be 8.

So obviously, the `node` field of `reason` should be the node that was actually used to deal the damage.

pauloue:
Indicates the node that dealt the damage when the reason type is `node_damage`.

Closes #8345.

To test:
```lua
minetest.register_on_player_hpchange(function(player, hp_change, reason)
	print(dump(reason))
end)
```
and walk into some fire.

Output:
```
{
	type = "node_damage",
	from = "engine",
	node = "fire:basic_flame"
}
```

### DIFF
--- a/src/content_sao.cpp
+++ b/src/content_sao.cpp
@@ -1022,6 +1022,7 @@ void PlayerSAO::step(float dtime, bool send_recommended)
 
 	if (m_node_hurt_interval.step(dtime, 1.0f)) {
 		u32 damage_per_second = 0;
+		std::string nodename;
 		// Lowest and highest damage points are 0.1 within collisionbox
 		float dam_top = m_prop.collisionbox.MaxEdge.Y - 0.1f;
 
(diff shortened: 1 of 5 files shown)


src/content_sao.cpp
```diff
@@ -1294,8 +1302,9 @@ int PlayerSAO::punch(v3f dir,
 				hitparams.hp);
 
 	if (!damage_handled) {
-		setHP((s32)getHP() - (s32)hitparams.hp,
-				PlayerHPChangeReason(PlayerHPChangeReason::PLAYER_PUNCH, puncher));
+		PlayerHPChangeReason reason(PlayerHPChangeReason::PLAYER_PUNCH);
```
nerzhul: why not use the correct constructor like in previous version ?

src/content_sao.h
```diff
@@ -450,7 +454,7 @@ struct PlayerHPChangeReason {
 		}
 	}
 
-	PlayerHPChangeReason(Type type, ServerActiveObject *object=NULL):
```
nerzhul: i see, don't drop it, add the new contructor only

src/script/lua_api/l_object.cpp
```diff
@@ -170,8 +170,8 @@ int ObjectRef::l_punch(lua_State *L)
 	ObjectRef *puncher_ref = checkobject(L, 2);
 	ServerActiveObject *co = getobject(ref);
 	ServerActiveObject *puncher = getobject(puncher_ref);
-	if (co == NULL) return 0;
-	if (puncher == NULL) return 0;
+	if (co == NULL || puncher == NULL)
```
nerzhul: if (!co || !puncher)

pauloue:
@nerzhul updated

src/content_sao.h
```diff
@@ -400,11 +400,15 @@ struct PlayerHPChangeReason {
 		RESPAWN
 	};
 
-	Type type = SET_HP;
```
nerzhul: why remove the default value ?

pauloue: is it needed?

pauloue: oh I guess it's good to have, re-added.

PR #8408, opened 2019-03-20
Add .vscode/ to .gitignore

ClobberXD:
(+ some more trivial changes)

Is the `test_config.h` line alright, or should I add `*`s in the beginning/end?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@
 *.swp
 *.bak*
 *.orig
+# VSCode
+.vscode/*
 # Vim
 *.vim
 # Kate
@@ -66,10 +68,6 @@ doc/doxygen_*
 ## Build files
 CMakeFiles
 Makefile
-!build/android/Makefile
-build/android/path.cfg
-build/android/*.apk
-build/android/.externalNativeBuild
 cmake_install.cmake
 CMakeCache.txt
 CPackConfig.cmake
@@ -100,6 +98,10 @@ CMakeDoxy*
 compile_commands.json
 
 ## Android build files
+!build/android/Makefile
+build/android/path.cfg
+build/android/*.apk
+build/android/.externalNativeBuild
 build/android/src/main/assets
 build/android/build
 build/android/deps


Calinou:
I think editor-specific ignores should go to your global gitignore file (`~/config/git/ignore` by default) :smiley:

pauloue:
I agree with Calinou. Here's what the manpage says:

> Which file to place a pattern in depends on how the pattern is meant to be used.
> 
>    * Patterns which should be version-controlled and distributed to other repositories via clone (i.e., files that all developers will want to ignore) should go into a .gitignore file.
> 
>  * Patterns which are specific to a particular repository but which do not need to be shared with other related repositories (e.g., auxiliary files that live inside the repository but are specific to one user’s workflow) should go into the $GIT_DIR/info/exclude file.
> 
>   *  Patterns which a user wants Git to ignore in all situations (e.g., backup or temporary files generated by the user’s editor of choice) generally go into a file specified by core.excludesFile in the user’s ~/.gitconfig. Its default value is $XDG_CONFIG_HOME/git/ignore. If $XDG_CONFIG_HOME is either not set or empty, $HOME/.config/git/ignore is used instead.

ClobberXD:
Interesting. Since we already had an "Editor / Development environment" section, I thought I'd add VSCode's workspace folder as well.

I think we should also consider whether we want to trust the users with the task of adding their IDE by-products to their global gitignore. New contributors might not even realize that they've staged and committed files belonging to their IDE. Shouldn't we protect ourselves from that instead of relying upon the contributors to ignore their IDE files?

Calinou:
> New contributors might not even realize that they've staged and committed files belonging to their IDE. Shouldn't we protect ourselves from that instead of relying upon the contributors to ignore their IDE files?

I believe it's easy enough to notice in a pull request by checking the list of changed files.

ClobberXD:
Forgot about this PR :P

Closing now. We should probably consider removing the other editor-specific patterns from `.gitignore` as well.

PR #8409, opened 2019-03-20
Rewrite camera modes; implement per-player camera modes

ClobberXD:
#### Use-cases

- 3rd-person mode generally offers players unfair advantages over players using 1st-person mode, like peripheral vision (looking around walls), noclip glitch (seeing through nodes in narrow passageways), etc.
- Gun mods work with the assumption that the player is in 1st-person mode, and if the player is in 3rd-person mode, most calculations will be borked and incorrect.

#### Changes

- enum `CameraMode` moved to class `Player`.
- New `EnumString` array `es_CameraModes` to store all camera mode enums and their string representation pairs.
- New attribute `std::set<CameraMode> Player::m_camera_modes` to store the enabled camera modes (both client-side and server-side) along with getter `Player::getCameraModes` and setter `Player::setCameraModes`.
- New Server->client packet (`TOCLIENT_CAMERA_MODES`).
- New Lua API methods `ObjectRef:get/set_camera_modes`.
- New method `Server::SendCameraModes` for serializing `Player::m_camera_modes` into a `u16` and sending it over the network.
- New client-side packet handler `Client::handleCommand_CameraModes` to deserialize the received `u16` back to `std::set<CameraMode>` and store it client-side.
- Documentation!

(Note: This PR also includes a couple of code-style fixes around the touched code)

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -6224,6 +6224,19 @@ object you are working with still exists.
     * Returns `false` if failed.
     * Resource intensive - use sparsely
     * To get blockpos, integer divide pos by 16
+* `set_camera_modes(modes)`: Sets the allowed camera modes for the player.
+    * `modes` is a table containing the following fields:
+      ```lua
+      {
+          "first" = true,
+          "third" = true,
+          "third_front" = true
+      }
+      ```
+    * Calling `set_camera_modes` without any table is shorthand for enabling
+      all camera modes (default).
+* `get_camera_modes()`: Returns a table of camera modes usable by the player
+  as in `set_camera_modes`.
 
 `PcgRandom`
 -----------
--- a/src/client/camera.cpp
+++ b/src/client/camera.cpp
@@ -454,7 +454,7 @@ void Camera::update(LocalPlayer* player, f32 frametime, f32 busytime, f32 tool_r
 	// Reposition the camera for third person view
 	if (m_camera_mode > CAMERA_MODE_FIRST)
 	{
-		if (m_camera_mode == CAMERA_MODE_THIRD_FRONT)
+		if (m_camera_mode == CAMERA_MODE_FRONT)
 			m_camera_direction *= -1;
 
 		my_cp.Y += 2;
@@ -667,7 +667,7 @@ void Camera::drawWieldedTool(irr::core::matrix4* translation)
 	// Draw the wielded node (in a separate scene manager)
 	scene::ICameraSceneNode* cam = m_wieldmgr->getActiveCamera();
 	cam->setAspectRatio(m_cameranode->getAspectRatio());
-	cam->setFOV(72.0*M_PI/180.0);
+	cam->setFOV(72.0*M_PI / 180.0);
 	cam->setNearValue(10);
 	cam->setFarValue(1000);
 	if (translation != NULL)
@@ -683,6 +683,34 @@ void Camera::drawWieldedTool(irr::core::matrix4* translation)
 		cam->setTarget(focusPoint);
 	}
 	m_wieldmgr->drawAll();
+}
+
+void Camera::toggleCameraMode()
+{
+	LocalPlayer *player = m_client->getEnv().getLocalPlayer();
+	std::set<CameraMode> modes = player->getCameraModes();
+	std::set<CameraMode>::iterator it;
+
+	// If server sends no camera modes, revert to default set of camera modes
+	if (modes.empty()) {
+		warningstream << "Camera::toggleCameraMode() | No server-sent camera modes. Reverting to default..." << std::endl;
+
+		// Use first available camera mode if current camera
+		// mode isn't available in m_default_camera_modes
+		it = m_default_camera_modes.find(m_camera_mode);
+		if (it == m_default_camera_modes.end())
+			it = m_default_camera_modes.begin();
+	} else {
+		warningstream << "Camera::toggleCameraMode() | Server-sent camera modes available. Cycling to next available mode..." << std::endl;
+
+		// If current camera mode not found, or is the last
+		// camera mode, set to first available camera mode
+		auto it = modes.find(m_camera_mode);
+		if (it == modes.end() || ++it == modes.end())
+			it = modes.begin();
+	}
+
+	m_camera_mode = *it;
 }
 
 void Camera::drawNametags()
--- a/src/client/camera.h
+++ b/src/client/camera.h
@@ -22,6 +22,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include "irrlichttypes_extrabloated.h"
 #include "inventory.h"
 #include "client/tile.h"
+#include "player.h"
 #include <ICameraSceneNode.h>
 #include <ISceneNode.h>
 #include <list>
@@ -47,8 +48,6 @@ struct Nametag {
 	video::SColor nametag_color;
 	v3f nametag_pos;
 };
-
-enum CameraMode {CAMERA_MODE_FIRST, CAMERA_MODE_THIRD, CAMERA_MODE_THIRD_FRONT};
 
 /*
 	Client camera class, manages the player and camera scene nodes, the viewing distance
@@ -142,14 +141,7 @@ public:
 	void drawWieldedTool(irr::core::matrix4* translation=NULL);
 
 	// Toggle the current camera mode
-	void toggleCameraMode() {
-		if (m_camera_mode == CAMERA_MODE_FIRST)
-			m_camera_mode = CAMERA_MODE_THIRD;
-		else if (m_camera_mode == CAMERA_MODE_THIRD)
-			m_camera_mode = CAMERA_MODE_THIRD_FRONT;
-		else
-			m_camera_mode = CAMERA_MODE_FIRST;
-	}
+	void toggleCameraMode();
 
 	// Set the current camera mode
 	inline void setCameraMode(CameraMode mode)
@@ -185,7 +177,7 @@ private:
 	WieldMeshSceneNode *m_wieldnode = nullptr;
 
 	// draw control
-	MapDrawControl& m_draw_control;
+	MapDrawControl &m_draw_control;
 
 	Client *m_client;
 
@@ -241,6 +233,11 @@ private:
 	ItemStack m_wield_item_next;
 
 	CameraMode m_camera_mode = CAMERA_MODE_FIRST;
+	const std::set<CameraMode> m_default_camera_modes = {
+		CAMERA_MODE_FIRST,
+		CAMERA_MODE_FIRST,
+		CAMERA_MODE_THIRD
+	};
 
 	f32 m_cache_fall_bobbing_amount;
 	f32 m_cache_view_bobbing_amount;
--- a/src/network/clientopcodes.cpp
+++ b/src/network/clientopcodes.cpp
@@ -68,7 +68,7 @@ const ToClientCommandHandler toClientCommandTable[TOCLIENT_NUM_MSG_TYPES] =
 	{ "TOCLIENT_TIME_OF_DAY",              TOCLIENT_STATE_CONNECTED, &Client::handleCommand_TimeOfDay }, // 0x29
 	{ "TOCLIENT_CSM_RESTRICTION_FLAGS",    TOCLIENT_STATE_CONNECTED, &Client::handleCommand_CSMRestrictionFlags }, // 0x2A
 	{ "TOCLIENT_PLAYER_SPEED",             TOCLIENT_STATE_CONNECTED, &Client::handleCommand_PlayerSpeed }, // 0x2B
-	null_command_handler,
+	{ "TOCLIENT_CAMERA_MODES",             TOCLIENT_STATE_CONNECTED, &Client::handleCommand_CameraModes }, // 0x2C
 	null_command_handler,
 	null_command_handler,
 	{ "TOCLIENT_CHAT_MESSAGE",             TOCLIENT_STATE_CONNECTED, &Client::handleCommand_ChatMessage }, // 0x2F
--- a/src/network/clientpackethandler.cpp
+++ b/src/network/clientpackethandler.cpp
@@ -1615,3 +1615,24 @@ void Client::handleCommand_ModChannelSignal(NetworkPacket *pkt)
 	if (valid_signal)
 		m_script->on_modchannel_signal(channel, signal);
 }
+
+void Client::handleCommand_CameraModes(NetworkPacket *pkt)
+{
+	u16 serialized_modes;
+	*pkt >> serialized_modes;
+	std::set<CameraMode> modes;
+
+	errorstream << "\thandleCommand_CameraModes | " << serialized_modes << std::endl;
+
+	LocalPlayer *player = m_env.getLocalPlayer();
+	if (!player)
+		return;
+
+	// De-serialise bit flags
+	for (u16 i = 0; es_CameraModes[i].num != CAMERA_MODE_NULL; ++i) {
+		if (serialized_modes & (1 << es_CameraModes[i].num))
+			modes.emplace(static_cast<CameraMode>(es_CameraModes[i].num));
+	}
+
+	player->setCameraModes(modes);
+}
--- a/src/network/serveropcodes.cpp
+++ b/src/network/serveropcodes.cpp
@@ -167,7 +167,7 @@ const ClientCommandFactory clientCommandFactoryTable[TOCLIENT_NUM_MSG_TYPES] =
 	{ "TOCLIENT_TIME_OF_DAY",              0, true }, // 0x29
 	{ "TOCLIENT_CSM_RESTRICTION_FLAGS",    0, true }, // 0x2A
 	{ "TOCLIENT_PLAYER_SPEED",             0, true }, // 0x2B
-	null_command_factory, // 0x2C
+	{ "TOCLIENT_CAMERA_MODES",             0, true }, // 0x2C
 	null_command_factory, // 0x2D
 	null_command_factory, // 0x2E
 	{ "TOCLIENT_CHAT_MESSAGE",             0, true }, // 0x2F
--- a/src/player.h
+++ b/src/player.h
@@ -22,6 +22,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include "irrlichttypes_bloated.h"
 #include "inventory.h"
 #include "constants.h"
+#include "script/common/c_types.h"
 #include "network/networkprotocol.h"
 #include "util/basic_macros.h"
 #include <list>
@@ -114,6 +115,18 @@ struct PlayerSettings
 	void readGlobalSettings();
 };
 
+enum CameraMode {
+	CAMERA_MODE_NULL = -1, CAMERA_MODE_FIRST, CAMERA_MODE_THIRD, CAMERA_MODE_FRONT
+};
+
+const EnumString es_CameraModes[] = {
+	{ CAMERA_MODE_FIRST,       "first" },
+	{ CAMERA_MODE_THIRD,       "third" },
+	{ CAMERA_MODE_FRONT,       "front" },
+	{ CAMERA_MODE_NULL,        "" }
+	// FIXME: A class or struct with member functions might probably be a better solution
+};
+
 class Map;
 struct CollisionInfo;
 struct HudElement;
@@ -200,11 +213,14 @@ public:
 		return m_fov_override_spec;
 	}
 
+	void setCameraModes(std::set<CameraMode> &modes) { m_camera_modes = modes; }
+	const std::set<CameraMode> &getCameraModes() const { return m_camera_modes; }
+
 	u32 keyPressed = 0;
 
-	HudElement* getHud(u32 id);
+	HudElement *getHud(u32 id);
 	u32         addHud(HudElement* hud);
-	HudElement* removeHud(u32 id);
+	HudElement *removeHud(u32 id);
 	void        clearHud();
 
 	u32 hud_flags;
@@ -217,6 +233,7 @@ protected:
 	PlayerFovSpec m_fov_override_spec = { 0.0f, false, 0.0f };
 
 	std::vector<HudElement *> hud;
+	std::set<CameraMode> m_camera_modes;
 private:
 	// Protect some critical areas
 	// hud for example can be modified by EmergeThread
--- a/src/script/lua_api/l_object.cpp
+++ b/src/script/lua_api/l_object.cpp
@@ -2200,6 +2200,59 @@ int ObjectRef::l_get_day_night_ratio(lua_State *L)
 	return 1;
 }
 
+int ObjectRef::l_set_camera_modes(lua_State *L)
+{
+	NO_MAP_LOCK_REQUIRED;
+	ObjectRef *ref = checkobject(L, 1);
+	RemotePlayer *player = getplayer(ref);
+	if (!player)
+		return 0;
+
+	if (!lua_istable(L, 2))
+		return 0;
+
+	errorstream << "\tl_set_camera_modes" << std::endl;
+
+	std::set<CameraMode> modes;
+
+	lua_pushnil(L);
+	while (lua_next(L, 2) != 0) {
+		int enum_val;
+		if (string_to_enum(es_CameraModes, enum_val, lua_tostring(L, -2))) {
+			if (readParam<bool>(L, -1, true))
+				modes.emplace(static_cast<CameraMode>(enum_val));
+		}
+		lua_pop(L, 1);
+	}
+
+	// If mod passed empty table, add CAMERA_MODE_FIRST by default
+	if (modes.size() == 0)
+		modes.emplace(CAMERA_MODE_FIRST);
+
+	player->setCameraModes(modes);
+	getServer(L)->SendCameraModes(player, modes);
+	return 0;
+}
+
+int ObjectRef::l_get_camera_modes(lua_State *L)
+{
+	NO_MAP_LOCK_REQUIRED;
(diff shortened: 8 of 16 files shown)


src/network/clientopcodes.cpp
```diff
@@ -56,72 +56,72 @@ const ToClientCommandHandler toClientCommandTable[TOCLIENT_NUM_MSG_TYPES] =
...
+	{ "TOCLIENT_MODCHANNEL_SIGNAL",             TOCLIENT_STATE_CONNECTED, &Client::handleCommand_ModChannelSignal }, // 0x58
+	{ "TOCLIENT_NODEMETA_CHANGED",              TOCLIENT_STATE_CONNECTED, &Client::handleCommand_NodemetaChanged }, // 0x59
+	null_command_handler,
+	null_command_handler,
+	null_command_handler,
+	null_command_handler,
+	null_command_handler,
+	null_command_handler,
+	{ "TOCLIENT_SRP_BYTES_S_B",                 TOCLIENT_STATE_NOT_CONNECTED, &Client::handleCommand_SrpBytesSandB }, // 0x60
+	{ "TOCLIENT_FORMSPEC_PREPEND",              TOCLIENT_STATE_CONNECTED, &Client::handleCommand_FormspecPrepend }, // 0x61,
```
sofar: Can't see your changes - too many unrelated whitespace changes.

src/network/serveropcodes.cpp
```diff
@@ -145,70 +145,70 @@ const ClientCommandFactory clientCommandFactoryTable[TOCLIENT_NUM_MSG_TYPES] =
...
 	null_command_factory,
 	null_command_factory,
-	{ "TOCLIENT_INVENTORY",                0, true }, // 0x27
+	{ "TOCLIENT_INVENTORY",                     0, true }, // 0x27
 	null_command_factory,
-	{ "TOCLIENT_TIME_OF_DAY",              0, true }, // 0x29
-	{ "TOCLIENT_CSM_RESTRICTION_FLAGS",    0, true }, // 0x2A
+	{ "TOCLIENT_TIME_OF_DAY",                   0, true }, // 0x29
+	{ "TOCLIENT_CSM_RESTRICTION_FLAGS",         0, true }, // 0x2A
+	{ "TOCLIENT_CAMERA_MODE_RESTRICTION_FLAGS", 0, true }, // 0x2B
```
sofar: `_MODE_` could be removed, maybe just make it TOCLIENT_CAMERA_RESTRICTION_FLAGS ? Not that important though.

sofar:
neat idea, though, I like it.

src/network/serveropcodes.cpp
ClobberXD: I felt `mode` made it clear that this restricts camera modes, and not the camera itself (i.e. camera movement, rotation, etc.)

src/network/clientopcodes.cpp
ClobberXD: Sorry about that, I indented all the other lines to make up for the extra large length of the network packet in question. :)

src/network/serveropcodes.cpp
ClobberXD: Should I omit `_MODE_` and if yes, should I omit it everywhere (incl. packet name, getter/setter names, Lua API method name, etc.)?

src/client/camera.cpp
```diff
@@ -602,6 +602,28 @@ void Camera::drawWieldedTool(irr::core::matrix4* translation)
 	m_wieldmgr->drawAll();
 }
 
+void Camera::toggleCameraMode()
+{
+	if (m_camera_mode == CAMERA_MODE_FIRST) {
```
nerzhul: switch case please

src/client/camera.h
```diff
@@ -48,6 +48,10 @@ struct Nametag {
 	v3f nametag_pos;
 };
 
+struct CameraModeRestrictions {
+	bool first, third, third_front;
```
nerzhul: initialize values in the definition please

src/client/camera.h
```diff
@@ -133,13 +137,16 @@ class Camera
...
-	void toggleCameraMode() {
-		if (m_camera_mode == CAMERA_MODE_FIRST)
-			m_camera_mode = CAMERA_MODE_THIRD;
-		else if (m_camera_mode == CAMERA_MODE_THIRD)
-			m_camera_mode = CAMERA_MODE_THIRD_FRONT;
-		else
-			m_camera_mode = CAMERA_MODE_FIRST;
+	void toggleCameraMode();
+
+	inline void setCameraModeRestrictions(CameraModeRestrictions &restrictions)
```
nerzhul: constness

src/client/camera.h
```diff
@@ -133,13 +137,16 @@ class Camera
...
+	{
+		m_camera_mode_restrictions = restrictions;
+	}
+
+	inline CameraModeRestrictions getCameraModeRestrictions() const
```
nerzhul: missing ref

src/network/clientpackethandler.cpp
```diff
@@ -1474,3 +1475,15 @@ void Client::handleCommand_ModChannelSignal(NetworkPacket *pkt)
 	if (valid_signal)
 		m_script->on_modchannel_signal(channel, signal);
 }
+
+void Client::handleCommand_CameraModeRestrictionFlags(NetworkPacket *pkt)
+{
+	u16 byteflags;
+	*pkt >> byteflags;
+
+	CameraModeRestrictionFlags *flags = new CameraModeRestrictionFlags;
```
nerzhul: parenthesis please

src/client/camera.cpp
ClobberXD: Done.

src/client/camera.h
ClobberXD: Done. I hope that's what you asked for. :)

src/network/clientpackethandler.cpp
ClobberXD: Done. Also added a ctor for `CameraModeRestrictions`

src/client/camera.h
ClobberXD: Done.

src/client/camera.h
ClobberXD: Done; also added a ctor.

src/client/camera.cpp
```diff
@@ -60,6 +60,8 @@ Camera::Camera(MapDrawControl &draw_control, Client *client):
 	m_wieldnode->setItem(ItemStack(), m_client);
 	m_wieldnode->drop(); // m_wieldmgr grabbed it
 
+	m_camera_mode_restrictions = new CameraModeRestrictions();
```
nerzhul: memleak, this value is never delete. I suggest to use stack instead of pointer here, due to the usecase.

src/client/camera.h
```diff
@@ -48,6 +48,15 @@ struct Nametag {
 	v3f nametag_pos;
 };
 
+struct CameraModeRestrictions {
+	bool first, third, third_front;
+
+	CameraModeRestrictions()
+	{
+		first = third = third_front = false;
```
nerzhul: no, this is C++98 and lesser. This must be done on variable declaration.

```
bool first = false, third = false ...
```

src/client/camera.h
```diff
@@ -133,13 +142,18 @@ class Camera
...
-		if (m_camera_mode == CAMERA_MODE_FIRST)
-			m_camera_mode = CAMERA_MODE_THIRD;
-		else if (m_camera_mode == CAMERA_MODE_THIRD)
-			m_camera_mode = CAMERA_MODE_THIRD_FRONT;
-		else
-			m_camera_mode = CAMERA_MODE_FIRST;
+	void toggleCameraMode();
+
+	// Set camera mode restrictions
+	inline void setCameraModeRestrictions(const CameraModeRestrictions restrictions)
```
nerzhul: okay it's const, but where is the ref now ? :)

src/client/camera.h
```diff
@@ -133,13 +142,18 @@ class Camera
...
+	{
+		m_camera_mode_restrictions = restrictions;
+	}
+
+	// Get camera mode restrictions
+	inline CameraModeRestrictions &getCameraModeRestrictions() const
```
nerzhul: return const ref please

src/player.h
```diff
@@ -180,6 +180,7 @@ class Player
 	HudElement* removeHud(u32 id);
 	void        clearHud();
 
+	u16 camera_mode_restrictions;
```
nerzhul: initialize 0x01 | 0x02 | 0x04 here instead of the constructor function

src/client/camera.h
```diff
@@ -176,10 +190,12 @@ class Camera
 	WieldMeshSceneNode *m_wieldnode = nullptr;
 
 	// draw control
-	MapDrawControl& m_draw_control;
+	MapDrawControl &m_draw_control;
 
 	Client *m_client;
 
+	CameraModeRestrictions *m_camera_mode_restrictions;
```
nerzhul: this attribute must be private, we already have a setter and getter

src/script/lua_api/l_object.cpp
```diff
@@ -1806,6 +1806,36 @@ int ObjectRef::l_get_day_night_ratio(lua_State *L)
...
+int ObjectRef::l_get_camera_mode_restriction_flags(lua_State *L)
+{
+	NO_MAP_LOCK_REQUIRED;
+	ObjectRef *ref = checkobject(L, 1);
+	RemotePlayer *player = get_player(ref);
+	if (!player)
+		return 0;
+
+	lua_newtable(L);
+	lua_pushboolean(L, player->camera_mode_restrictions & 0x01);
```
nerzhul: use the getter instead of the raw attribute please

ClobberXD: I think the getter/setter methods are only accessible by the clients, and not servers - are files in `src/client/` compiled when building the server?

src/client/camera.h
ClobberXD: It is private, as far as I can tell :)

src/client/camera.h
ClobberXD: Done

src/client/camera.h
ClobberXD: Done.

src/client/camera.h
ClobberXD: Done

src/client/camera.cpp
ClobberXD: Done what I think you suggested. :)

src/player.h
ClobberXD: Done.

src/script/lua_api/l_object.cpp
ClobberXD: Nvm, added getter/setter methods for `RemotePlayer::m_camera_mode_restrictions`

ClobberXD:
Everything has been implemented, and this PR is ready for review. :tada: 

(This by no means indicates that the PR is functional, btw :P)

ClobberXD:
Trying to tackle this error:

```
/src/client/camera.cpp:43:1: error: uninitialized reference member in ‘struct CameraModeRestrictions&’
```

Let me know what's the best way to fix this. Should I not use references here? @nerzhul wanted its setter and getter to take and return references respectively. Shall I de-reference the passed reference param and store it in `m_camera_mode_restrictions`?

src/client/camera.h
```diff
@@ -176,10 +187,12 @@ class Camera
 	WieldMeshSceneNode *m_wieldnode = nullptr;
 
 	// draw control
-	MapDrawControl& m_draw_control;
+	MapDrawControl &m_draw_control;
 
 	Client *m_client;
 
+	CameraModeRestrictions &m_camera_mode_restrictions;
```
nerzhul: remove the reference here please it's just an owned object on the stack

src/network/clientpackethandler.cpp
```diff
@@ -1474,3 +1475,15 @@ void Client::handleCommand_ModChannelSignal(NetworkPacket *pkt)
 	if (valid_signal)
 		m_script->on_modchannel_signal(channel, signal);
 }
+
+void Client::handleCommand_CameraModeRestrictionFlags(NetworkPacket *pkt)
+{
+	u16 byteflags;
+	*pkt >> byteflags;
+
+	CameraModeRestrictions *flags = new CameraModeRestrictions();
```
nerzhul: don't use heap, just use stack

src/script/lua_api/l_object.cpp
```diff
@@ -1806,6 +1806,54 @@ int ObjectRef::l_get_day_night_ratio(lua_State *L)
...
+	NO_MAP_LOCK_REQUIRED;
+	ObjectRef *ref = checkobject(L, 1);
+	RemotePlayer *player = get_player(ref);
+	if (!player)
+		return 0;
+
+	u16 restrictions = player->getCameraModeRestrictions();
+
+	lua_newtable(L);
+	lua_pushboolean(L, restrictions & 0x01);
```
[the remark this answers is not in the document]
ClobberXD: `camera.h` already contains the `CameraModes` enum, and I didn't want to duplicate the exact same enum elsewhere in the code. Should I move the existing enum to a common (i.e. accessible by both server and client) place, say, `player.h`?

src/network/clientpackethandler.cpp
ClobberXD: Done.

src/client/camera.h
ClobberXD: Done

ClobberXD:
This PR finally compiles without errors, yay!

Tested the Lua API methods using the `luacmd` mod - the camera mode restrictions work flawlessly.

src/client/camera.cpp
```diff
@@ -602,6 +602,32 @@ void Camera::drawWieldedTool(irr::core::matrix4* translation)
...
+			else if (!m_camera_mode_restrictions.first)
+				m_camera_mode = CAMERA_MODE_FIRST;
+			break;
+		}
+		case CAMERA_MODE_THIRD_FRONT: {
+			if (!m_camera_mode_restrictions.first)
+				m_camera_mode = CAMERA_MODE_FIRST;
+			else if (!m_camera_mode_restrictions.third)
+				m_camera_mode = CAMERA_MODE_THIRD;
+		}
```
DS-Minetest: A break is missing here. (https://dev.minetest.net/Code_style_guidelines#Miscellaneous)

Also there are many curly brackets. I don't know whether it looks weird to make one block per case.

And everything is indented too much (see https://www.kernel.org/doc/html/latest/process/coding-style.html#indentation).

src/network/serveropcodes.cpp
DS-Minetest: What about `TOCLIENT_CAMERA_MODE_FLAGS`? Imho that the flags restrict seems kinda obvious.

src/client/camera.cpp
ClobberXD: Thanks for pointing that out. Fixed. :+1:

src/script/lua_api/l_object.cpp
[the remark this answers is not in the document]
SmallJoker: I'd prefer to see something like this using `string_to_enum` and a `set` of the available camera modes.
```C++
const struct EnumString es_CameraModes[] =
{
	{CAMERA_MODE_FIRST,       "first"},
	{CAMERA_MODE_THIRD,       "third"}
	{CAMERA_MODE_THIRD_FRONT, "front"},
	{0, NULL},
};

std::set<CameraMode> m_camera_modes;

int ObjectRef::l_set_camera_modes(lua_State *L)
{
	// ...

	if (!lua_istable(L, 1))
		return 0; // must be like {}

	lua_pushnil(L);
	// Possibly 1, maybe some other index
	while (lua_next(L, 1) != 0) {
		int enum_val;
		if (string_to_enum(es_CameraModes, enum_val, lua_tostring(L, -2))) {
			if (readParam<bool>(L, -1, true))
				m_camera_modes.emplace((CameraMode)enum_val);
			else
				m_camera_modes.erase((CameraMode)enum_val);
		}
		lua_pop(L, 1);
	}

	// Needs at least one
	if (m_camera_modes.count() == 0)
		m_camera_modes.emplace(CAMERA_MODE_FIRST);
}

void Camera::toggleCameraMode()
{
	auto it = m_camera_modes.find(m_camera_mode);
	if (it == m_camera_modes.end()) {
		m_camera_mode = *m_camera_modes.begin();
		return;
	}

	if (++it == m_camera_modes.end())
		it = m_camera_modes.begin();
	m_camera_mode = *it;
}
```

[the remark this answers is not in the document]
ClobberXD: Setting the camera modes themselves is much better than setting "restrictions" for the camera modes. :+1: 

Will do.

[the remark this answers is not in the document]
ClobberXD: Please also suggest the best place to declare `std::set<CameraMode> m_camera_modes`. ;)

From your code above, I think it could be declared in `l_object.cpp`, but the problem is that `Camera::toggleCameraMode` won't be able to access it directly, as above.

How about in the `Player` class? Both the server and the client would store the camera modes in `RemotePlayer` and `LocalPlayer` respectively.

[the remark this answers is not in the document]
SmallJoker: `Player` is probably the best choice. I'd use a getter function (non-const reference to the set) in this case to expose it to the Lua part.

src/network/serveropcodes.cpp
SmallJoker: `TOCLIENT_CAMERA_MODES` based on my yesterday's suggested design idea.

src/script/lua_api/l_object.cpp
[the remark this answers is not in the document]
ClobberXD: Sounds good.

src/network/serveropcodes.cpp
ClobberXD: Yes, `TOCLIENT_CAMERA_MODES` is better suited for the name, due to SmallJoker's idea.

ClobberXD: I thought of `TOCLIENT_CAMERA_MODES` as well, and that is also better suited for the packet name, due to [SmallJoker's idea](https://github.com/minetest/minetest/pull/8409#discussion_r270634673).

ClobberXD:
Can't take it anymore; squashing... :)

src/script/lua_api/l_object.cpp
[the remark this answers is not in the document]
ClobberXD: @SmallJoker How would these be sent over the network? Currently, these modes are encoded and sent as a `u16` over to the client, where the value is decoded again.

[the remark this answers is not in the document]
ClobberXD: As I'm not familiar with `EnumString`, would you also be able to help me understand how the getter API method would work?

src/network/serveropcodes.cpp
ClobberXD: Shortened the name across the PR to camera modes. Packet renamed to `TOCLIENT_CAMERA_MODES` (yay!)

src/script/lua_api/l_object.cpp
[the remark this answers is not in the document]
SmallJoker: @ClobberXD You can either continue to send the entire vector (starting with length=`u16`) as `u16` (enum values) or `std::string` (Lua names). Both options would be OK.
Regarding the EnumString functions: 
 * Get enum pendant for a string value: `string_to_enum(enumstring_array, out_as_int, input_string)`
 * Get string pendant for an enum value: [TODO] `enum_to_string(enumstring_array, out_as_string, input_enum_as_int)` 

The idea of the code above would be that modders can indicate the changes of the available camera modes. If one mode is not specified in the setter function, it won't be updated.
The getter function would then push a table containing all camera modes: (iterate though all available modes until CAMERA_MODE_MAX (TODO).
```Lua
{
	first = true, -- contained in m_camera_modes
	third = false, -- not in m_camera_modes
	front = true
}
```

I see this is getting quite a long story for such a small feature. I'm sorry if this is tiring.
FYI: `luamethod(ObjectRef, set_camera_mode_restriction_flags),` needs updating in your source as well.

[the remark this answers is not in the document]
ClobberXD: > You can either continue to send the entire vector (starting with length=u16) as u16 (enum values) or std::string (Lua names).

Alright, will look into it.

> Get string pendant for an enum value: **[TODO]** 
> `enum_to_string(enumstring_array, out_as_string, input_enum_as_int)`

Seeing `string_to_enum`, I already tried looking for `enum_to_string`, but couldn't find it.

> I see this is getting quite a long story for such a small feature. I'm sorry if this is tiring.

This is indeed getting too draggy, but I wholeheartedly support your approach as it's much more straightforward. The end result would be very robust and easy to work with, and that makes the wasted efforts worthwhile. Thanks for all your help. :)

>  luamethod(ObjectRef, set_camera_mode_restriction_flags), needs updating in your source as well.

Will do, thanks.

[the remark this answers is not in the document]
ClobberXD: > The getter function would then push a table containing all camera modes: (iterate though all available modes until CAMERA_MODE_MAX (TODO)

I was wondering if there's a better way to iterate through all available modes. Maybe convert the set into a map (`std::map`)? Each player will own a copy of this map, and each key can be assigned `true` or `false`.

ClobberXD:
@SmallJoker I think this PR is ready. I've added documentation for the new Lua API methods. I've also implemented `enum_to_string`, but now I don't know where I should use it, lol :)

I still haven't tested this yet (I haven't even compiled it) - I'll do it when I get some time later today.

rubenwardy:
APIs like this make me feel uncomfortable. I would quite like to have a unified way to control the current state a user is in in and which states they can change to. Unified in Lua API, packets, and handling code. The problem with hud flags is that they specifically are for HUD, and they are also readonly by the client. Ie: it's a flag, not a command to go into a particular state.

Argh, ramble. But maybe something like this would be better:

```c++
player:set_flags({
	camera_mode_first = true,
	minimap = true,
})

player:set_camera_mode("first")
player:se
-- all these methods send the same packets
```

obviously changing anything over than camera mode related things are out of scope, but it would be nice to have a plan for a unified API and packet handling

ClobberXD:
I really agree that we need a more unified API for all these features. We already have too many packets going to and fro, and there are separate getters and setters for each and every feature, making the API cluttered. While it's impractical to implement the unified API, I'm open to suggestions for incorporating a more unified approach into this feature PR.

ClobberXD:
@nerzhul Would you be able to update (or dismiss) your review?

paramat:
I dismissed nerzhul's review, i assume it was all attended to.

@nerzhul i have a request, please could you dismiss your reviews when the issues are resolved? You never seem to and i have had clean those up very many times.
If you don't want the bother of having to dismiss them (i can understand this) then you can use the 'add single comment' option instead of the review option, when leaving line comments.

Also , when you +1 a PR please add the appropriate 'approval' label.

ClobberXD:
Rebased; will test again. But this PR is complete, and only requires reviews.

paramat:
Code still broken?

ClobberXD:
@paramat Yes, I wasn't able to fix it within the evening. It seems fixable though, nothing to worry. :)

paramat:
No problem, just needed to know to re-add WIP label.

ClobberXD:
Added `cout`s everywhere. I noticed that the server completely freezes (client still works fine, even in singleplayer) when executing `Server::SendCameraModes`. This is the function that serialises the camera mode enums and sends them over the network. It's a very simple and straightforward function, and I don't see anything wrong with it. Will dig deeper...

ClobberXD:
What should be the default camera modes of players, if the server doesn't immediately send camera modes upon join? The most restrictive, with only first-person mode? Or the least restrictive, with all three modes? I personally prefer the most restrictive option, as I believe that the server should have control over this aspect, but the least restrictive is the backwards-compatible approach.

doc/lua_api.txt
```diff
@@ -5881,6 +5881,18 @@ This is basically a reference to a C++ `ServerActiveObject`
     * Returns `false` if failed.
     * Resource intensive - use sparsely
     * To get blockpos, integer divide pos by 16
+* `set_camera_modes(modes)`: Sets the camera modes for the player.
```
paramat: "Sets the allowed camera modes ..."

paramat:
Hmm good question, difficult to decide.
Allowing all 3 modes is harmless in most situations, and certainly harmless for creative servers.
If the default is first only, considering that most servers will just leave the default in effect, this would then be very unpopular and unnecessarily restrictive.
So in this case i think i prefer allowing all modes by default.

ClobberXD:
> If the default is first only, considering that most servers will just leave the default in effect, this would then be very unpopular and unnecessarily restrictive.

Good point, but this is more of a game/mod thing - server owners can make custom mods if they want, but it's mostly upto games like MTG to decide upon this (this would be set by the player api mod, I think?). But then again, games/mods would be *forced* to implement this, and even if they do end up adding the necessary code, it'd mean that all servers would have to update to the latest versions.

So I think the least restrictive option is the way to go (though I personally prefer the latter), for the sake of backwards-compat.

ClobberXD:
Rebased. I'll continue to debug the server freeze now.

esanhueza:
Hi, new here. I took the work made by @ClobberXD, added some corrections and it's working just fine over the master branch. But I need to solve two things.
1. If the player is in first person mode and a Mod restrict the camera modes to allow only third person mode, should I change the camera mode of the player immediately? or should I let that work for the Mod?
2. I don't know how to share my changes. As I understand, I have to make a pull request, but this feature already have one. How should I proceed?

PR #8425, opened 2019-03-25
Cleanups around the mapgen init

nerzhul:
* Cleanup emergeManager::initMapgens function (error handling)
* cleanup Mapgen::createMapgen (mapgen enum ids mustn't be set to a misc value in constructor for childs
* initialize serverthread object in server, not in init phase (init phase is called after constructor, it's identical)

note: **don't squash commits**

### DIFF
--- a/src/emerge.cpp
+++ b/src/emerge.cpp
@@ -186,12 +186,12 @@ EmergeManager::~EmergeManager()
 
 void EmergeManager::initMapgens(MapgenParams *params)
 {
-	FATAL_ERROR_IF(!m_mapgens.empty(), "mapgen already inited.");
+	FATAL_ERROR_IF(!m_mapgens.empty(), "Mapgen already initialised.");
 
 	mgparams = params;
 
 	for (u32 i = 0; i != m_threads.size(); i++)
-		m_mapgens.push_back(Mapgen::createMapgen(params->mgtype, i, params, this));
+		m_mapgens.push_back(Mapgen::createMapgen(params->mgtype, params, this));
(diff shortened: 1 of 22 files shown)


nerzhul:
@SmallJoker @sfan5 can you take time to look at this PR please ?

src/emerge.cpp
```diff
@@ -184,33 +184,28 @@ EmergeManager::~EmergeManager()
...
 
 
-bool EmergeManager::initMapgens(MapgenParams *params)
+void EmergeManager::initMapgens(MapgenParams *params)
 {
-	if (!m_mapgens.empty())
-		return false;
-
-	this->mgparams = params;
+	FATAL_ERROR_IF(!m_mapgens.empty(), "mapgen already inited.");
```
ClobberXD:
```suggestion
	FATAL_ERROR_IF(!m_mapgens.empty(), "Mapgen already initialised.");
```

src/mapgen/mapgen.cpp
```diff
@@ -153,23 +153,23 @@ Mapgen *Mapgen::createMapgen(MapgenType mgtype, int mgid,
 {
 	switch (mgtype) {
 	case MAPGEN_CARPATHIAN:
-		return new MapgenCarpathian(mgid, (MapgenCarpathianParams *)params, emerge);
```
SmallJoker: `mgid` is now no longer used.

nerzhul: exact

PR #8435, opened 2019-03-31
Find PostgreSQL correctly

adrido:
@nerzhul , You implemented the original part, please review. Thanks.
https://github.com/minetest/minetest/commit/ce42ff9cf74ebb8d4b68bc78c95e90ea3db02b78

As there might be questions:
See https://cmake.org/cmake/help/v3.14/module/FindPostgreSQL.html

The bulk to get an `PostgreSQL_ADDITIONAL_SEARCH_PATHS` did not work correctly the whole years, so I consider it as not needed.

This makes PostgereSQL usable on Windows too


adrido:
Just fixed a typo in the commit message.

Issue #2881, opened 2015-07-04
Prioritize shaped recipes over shapeless ones.


beyondlimits:
Currently shapeless recipes may get in conflict with shaped, overriding them. Since shaped recipes have narrower scope, prioritizing them over shapeless could bring more flexibility to the API and allow modders to write less code to avoid conflicts.

Example: https://github.com/Calinou/moreblocks/blob/master/stairsplus/slabs.lua#L69 cannot be declared shapeless, because it may conflict with https://github.com/Calinou/moreblocks/blob/master/stairsplus/stairs.lua#L206.

nanepiwo:
+1

beyondlimits:
Bump! Anyone here?

beyondlimits:
Nah, closing just to get rid of dangling issue.

PR #8445, opened 2019-04-03
Prioritise craft recipes

### LINKED ISSUE #3972: Select crafting recipes without groups first
There could be following situation:
default adds this recipe:

```
minetest.register_craft({
    output = 'default:stick 4',
    recipe = {
        {'group:wood'},
    }
})
```

A mod called smd adds its own wood node and its own stick; the craft recipe to get the stick from the wood doesn't work:

```
minetest.register_craft({
    output = 'smd:stick 4',
    recipe = {
        {'smd:wood'},
    }
})
```

smd:wood should still be in the wood group (for crafting default chests etc.) but smd:stick should be the output when crafting sticks from it, not default:stick.

l think crafting should be changed, it should get a table of all possible recipes and then it should select the recipe which fits best (which has the lowest amount of "group"s).

Edit: group recipes are already prioritised due to the craft hash type selection. The problem happens for shapeless recipes.

HybridDog:
Fixes #3972 and  #2881

I've also simplified getHashType (the first commit). 

### How to test the changes

Add this recipe to the **end** of default/crafting.lua:
```lua
minetest.register_craft({
	type = "shapeless",
	output = 'default:mese',
	recipe = {'default:coal_lump', 'group:stick'}
})
```
Then start Minetest and try to craft a Torch.

With this PR applied, the Torch recipe is prioritised over the shapeless one.

### DIFF
--- a/src/craftdef.cpp
+++ b/src/craftdef.cpp
@@ -407,22 +407,6 @@ void CraftDefinitionShaped::decrementInput(CraftInput &input, std::vector<ItemSt
 	craftDecrementOrReplaceInput(input, output_replacements, replacements, gamedef);
 }
 
-CraftHashType CraftDefinitionShaped::getHashType() const
-{
-	assert(hash_inited); // Pre-condition
-	bool has_group = false;
-	for (const auto &recipe_name : recipe_names) {
-		if (isGroupRecipeStr(recipe_name)) {
-			has_group = true;
-			break;
-		}
-	}
-	if (has_group)
-		return CRAFT_HASH_TYPE_COUNT;
-
-	return CRAFT_HASH_TYPE_ITEM_NAMES;
-}
-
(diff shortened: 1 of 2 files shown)


beyondlimits:
Does this by chance fix #2881?

ClobberXD:
Looks like it. See https://github.com/minetest/minetest/pull/8445/files#diff-7352ab73138ecb6fe1bcb3d74a6b52cbR135

Shaped recipes have the highest priority, while tool repair and shapeless recipes have the least priorities.

SmallJoker:
I like this PR for its simple implementation, although I'm not familiar with this code, which means that I need to test it (earliest in a few weeks) to ensure there are no unwanted side-effects.

SmallJoker:
1604b94
12e3d3d

PR #8447, opened 2019-04-03
Allow custom values for item wear bar via meta

DS-Minetest:
Allow to set the wear bar of an item with item metadata.
The field "wear_bar" is used. Values range from `1`(empty) to `0`(full).

I've used this to test:
```lua
minetest.register_chatcommand("bla", {
	params = "",
	description = "",
	privs = {},
	func = function(name, param)
		local item = player:get_wielded_item()
		local meta = item:get_meta()
		meta:set_string("wear_bar", param)
		player:set_wielded_item(item)
		return true, "bla"
	end,
})
```
Screenshot:
screenshot_20190403_194125

Todo: test

Fixes: #6448

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -1849,6 +1849,7 @@ Some of the values in the key-value store are handled specially:
 * `color`: A `ColorString`, which sets the stack's color.
 * `palette_index`: If the item has a palette, this is used to get the
   current color from the palette.
+* `wear_bar`: A float number, which is used to display a wear bar.
 
 Example:
 
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -1849,7 +1849,8 @@ Some of the values in the key-value store are handled specially:
 * `color`: A `ColorString`, which sets the stack's color.
 * `palette_index`: If the item has a palette, this is used to get the
   current color from the palette.
-* `wear_bar`: A float number, which is used to display a wear bar.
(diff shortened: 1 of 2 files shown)


rubenwardy:
This feels hacky to me

stujones11:
> This feels hacky to me. The distinction between tools and items should probably be removed instead

Ideally, I would like to see #3551 implemented, however, this is far from trivial. This simple patch would offer pretty much the same possibilities without the disruption. IMO there _should_ be a distinction between tools and items, however, some things don't quite fit into either category, like armor or guns. This would allow modders to bridge that gap if need be.

However, adding this would not prevent such refactor in future version, It just offers more flexibility. I also think that this is highly unlikely to happen anytime before version 6.0.0 (assuming anyone even takes it on) I have looked at sapier's original PR and it's absolutely massive!

DS-Minetest:
Documentation is added.

stujones11:
I have done some testing and this works perfectly for my armor requirements :+1: 

screenshot_20190404_220518

DS-Minetest:
Please review, approve or merge.

SmallJoker:
Whereas this PR works es intended, I don't think it's the best solution. As @rubenwardy mentioned, the distinction should be clear. A saner solution is to allow meshes for items and tools.

paramat:
sfan5 thumbed-up rubenwardy's comment, so that's 3 core dev disapprovals.

Issue #2690, opened 2015-05-11
positions of particle_spawner output not entirely accurate


VanessaE:
Picture says it all:

![image](https://cloud.githubusercontent.com/assets/1733537/7557278/572fd99c-f75c-11e4-987f-c48baa021fac.png)

There's only one particle spawner in this picture.  If I turn off that one, all three of these streams will disappear.  And, those extraneous particles respond to/collide with stuff that's placed in their way.

Engine is at commit 5a1975a4e38caa1385cfefac299ce17018172714

est31:
Works for me:
![screenshot_20150511_164840](https://cloud.githubusercontent.com/assets/8872119/7567511/af30aee0-f7fd-11e4-8170-9500e8a6e415.png)

kilbith:
@est31 Try with a sink instead...
![screenshot_20150511_170952](https://cloud.githubusercontent.com/assets/7883281/7568021/bb851d86-f800-11e4-9dab-79dcb2f6b841.png)

Or a faucet...
![screenshot_20150511_171215](https://cloud.githubusercontent.com/assets/7883281/7568073/f2802ce0-f800-11e4-838c-659fba4ca6ae.png)

est31:
My guess this is a glitch in collision detection.

est31:
Clearly there is a physics issue:
![screenshot_20150511_173633](https://cloud.githubusercontent.com/assets/8872119/7568678/4f457f5e-f804-11e4-8335-dc0e6b10985f.png)
I wonder however whether the original Image @VanessaE shared can be explained by that too.

obneq:
Problem seems to go away if you disable handling of negative nearest_dtime in collision.cpp, line 446.
so something is off with the checks done for bool step_up in the same file, line 431

Issue #1414, opened 2014-06-24
Particles are very slow


ShadowNinja:
Minetest uses a custom particle(spawner) implementation using entities, this is very slow.
Irrlicht's built-in particles are much faster, as you can see [here](https://www.youtube.com/watch?v=5WnmQYLJzlE).

Irrlicht's particles don't support collision by default, but this can be done with particle affectors (see [here](https://www.youtube.com/watch?v=-brLc8uUeQk)).

EDIT: Done (but with issues) by #2587.

neoascetic:
+1

neoascetic:
If collision detection implementation is hard, it would be cool to have Irrilicht's implementation at least for particles with turned off collision detection, and fallback to current implementation otherwise.

Wuzzy2:
The first thing I noticed is that the rendering distance for particles is infinite. It seems that once a client has "found" a particlespawner, it will handle and render it even if the player is very far away. So my first suggestion to boost performance is to limit the rendering distance of particles. Particles and particles that appear more than X nodes away from the player are ignored. You don't need to simulate particles of a block of fire over 1000 nodes away. ;-)

https://satoshiupload.com/images/UMZYItK8y5.png

Maybe the modder can be permitted to set a flag in the particle definition to overwrite the distance limit regardless (useful for large particles).

kay27:
Particles spawners have such options as speed, acceleration and distance (position), so maybe the logic is that the particles can fly very far and finally reach you after 1000 nodes... So the easiest way of fixing it for myself is to add just one option: maximum rendering distance from spawner position. For anarchy-type servers it should be not so big value, I think not more than 100 nodes.

ExeVirus:
So this was discussed in now closed duplicates of this issue, but the best idea(s) put forward are:

Simple version: max distance clipping
Complex version: FOV clipping. In this case, we render only particles/particle spawners that are directly in front of the player's vision up to a max distance (basically a rectangular prism). This could use an octree to subdivide the particles and spawners by block, and just specify if a given block is visible in the cone or not. 

If this complex version works for particles, it was suggested that something similar be done for all entity rendering, especially animated entities.

kay27:
@ExeVirus I'm not sure that Complex version as you described here is correct at whole. The simple example: water drops aren't in front of the player but they MUST reach them. So spawned-specific distance setup seems would be still preferred, for myself

AFCMS:
[...]

PR #8461, opened 2019-04-08
Reimplement particles using the Irrlicht particle system and improve collision handling

DTA7:
Update to pull request #6540.
Addresses #6523, #1414, #2690.

I've included your suggestions and implemented the special collision handling for points, as well as addressed the previous TODOs.
-The scaling is correct now, the velocities for the Irrlicht particles are given per millisecond. @numberZero
-The collision detection is much faster, about ~5 times when I tested it (by doing the calculation
with both methods by uncommenting the `collisionMoveSimple` code in "particles.cpp" of 46dc879cf143fb838b6db3c9e330974762dbe0fe and comparing the Profiler info). This speedup comes from the much lower amount of nodes considered by the treatment as points.
-I added the `bounce_fraction` and the `bounce_threshold` properties to specify the retained fraction of the velocity component in the direction of the bounce, and the threshold for the ratio of this component to the total speed, below which the particle stops completely. This causes the "digging particles" not to slide anymore as before.
-The `vertical` property was removed.
-To prevent particles partially hidden in nodes (which also causes wrong lighting), the colliding nodeboxes are treated as enlarged by the particle size. This causes problems on the edges of nodes for large particles, but to change this more nodes have to be considered. For the most part, this is not really noticable.

The changed properties require a change in the network protocol, so the change in 9c57ce0122b54fc26f57fd8ab77d7de5e972ec73 could be removed in this pull request.

[Updated mod for testing](https://pastebin.com/sBKE13jv)

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -6831,14 +6831,14 @@ Used by `minetest.add_particle`.
         -- `physical = true,` and `collide_with_objects = true,`.
         -- Requires collisiondetection = true to have any effect.
 
-		bounce_fraction = 1,
-		-- Fraction of the velocity component that is kept after bouncing
-		-- in the corresponding direction.
+        bounce_fraction = 1,
+        -- Fraction of the velocity component that is kept after bouncing
+        -- in the corresponding direction.
 
-		bounce_threshold = 0,
-		-- If the absolute value of the velocity component in the direction
-		-- of a bounce is smaller than bounce_threshold * speed,
-		-- the velocity is set to, i.e. prevents "gliding".
+        bounce_threshold = 0,
+        -- If the absolute value of the velocity component in the direction
+        -- of a bounce is smaller than bounce_threshold * speed,
+        -- the velocity is set to, i.e. prevents "gliding".
 
         texture = "image.png",
 
@@ -6874,7 +6874,7 @@ Used by `minetest.add_particlespawner`.
         maxvel = {x=0, y=0, z=0},
         minacc = {x=0, y=0, z=0},
         maxacc = {x=0, y=0, z=0},
-		-- ^ Is cut off to 0 if the lengths of both minacc and maxacc are smaller than 0.001
+        -- ^ Is cut off to 0 if the lengths of both minacc and maxacc are smaller than 0.001
         minexptime = 1,
         maxexptime = 1,
         minsize = 1,
@@ -6896,14 +6896,14 @@ Used by `minetest.add_particlespawner`.
         -- `physical = true,` and `collide_with_objects = true,`.
         -- Requires collisiondetection = true to have any effect.
 
-		bounce_fraction = 1,
-		-- Fraction of the velocity component that is kept after bouncing
-		-- in the corresponding direction.
+        bounce_fraction = 1,
+        -- Fraction of the velocity component that is kept after bouncing
+        -- in the corresponding direction.
 
-		bounce_threshold = 0,
-		-- If the absolute value of the velocity component in the direction
-		-- of a bounce is smaller than bounce_threshold * speed,
-		-- the velocity is set to, i.e. prevents "gliding".
+        bounce_threshold = 0,
+        -- If the absolute value of the velocity component in the direction
+        -- of a bounce is smaller than bounce_threshold * speed,
+        -- the velocity is set to, i.e. prevents "gliding".
 
         attached = ObjectRef,
         -- If defined, particle positions, velocities and accelerations are
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -6831,6 +6831,9 @@ Used by `minetest.add_particle`.
         -- `physical = true,` and `collide_with_objects = true,`.
         -- Requires collisiondetection = true to have any effect.
 
+	-- deprecated: `vertical = false`
+	-- Particles now always face player.
+
         bounce_fraction = 1,
         -- Fraction of the velocity component that is kept after bouncing
         -- in the corresponding direction.
@@ -6908,6 +6911,9 @@ Used by `minetest.add_particlespawner`.
         attached = ObjectRef,
         -- If defined, particle positions, velocities and accelerations are
         -- relative to this object's position and yaw
+
+	-- deprecated: `vertical = false`
+	-- Particles now always face player.
 
         texture = "image.png",
 
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -6868,6 +6868,7 @@ Used by `minetest.add_particlespawner`.
         maxvel = {x=0, y=0, z=0},
         minacc = {x=0, y=0, z=0},
         maxacc = {x=0, y=0, z=0},
+		-- ^ Is cut off to 0 if the lengths of both minacc and maxacc are smaller than 0.001
         minexptime = 1,
         maxexptime = 1,
         minsize = 1,
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -6831,8 +6831,8 @@ Used by `minetest.add_particle`.
         -- `physical = true,` and `collide_with_objects = true,`.
         -- Requires collisiondetection = true to have any effect.
 
-	-- deprecated: `vertical = false`
-	-- Particles now always face player.
+        -- deprecated: `vertical = false`
+        -- Particles now always face player.
 
         bounce_fraction = 1,
         -- Fraction of the velocity component that is kept after bouncing
@@ -6912,8 +6912,8 @@ Used by `minetest.add_particlespawner`.
         -- If defined, particle positions, velocities and accelerations are
         -- relative to this object's position and yaw
 
-	-- deprecated: `vertical = false`
-	-- Particles now always face player.
+        -- deprecated: `vertical = false`
+        -- Particles now always face player.
 
         texture = "image.png",
 
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -6831,8 +6831,14 @@ Used by `minetest.add_particle`.
         -- `physical = true,` and `collide_with_objects = true,`.
         -- Requires collisiondetection = true to have any effect.
 
-        vertical = false,
-        -- If true faces player using y axis only
+		bounce_fraction = 1,
+		-- Fraction of the velocity component that is kept after bouncing
+		-- in the corresponding direction.
+
+		bounce_threshold = 0,
+		-- If the absolute value of the velocity component in the direction
+		-- of a bounce is smaller than bounce_threshold * speed,
+		-- the velocity is set to, i.e. prevents "gliding".
 
         texture = "image.png",
 
@@ -6890,12 +6896,18 @@ Used by `minetest.add_particlespawner`.
         -- `physical = true,` and `collide_with_objects = true,`.
         -- Requires collisiondetection = true to have any effect.
 
+		bounce_fraction = 1,
+		-- Fraction of the velocity component that is kept after bouncing
+		-- in the corresponding direction.
+
+		bounce_threshold = 0,
+		-- If the absolute value of the velocity component in the direction
+		-- of a bounce is smaller than bounce_threshold * speed,
+		-- the velocity is set to, i.e. prevents "gliding".
+
         attached = ObjectRef,
         -- If defined, particle positions, velocities and accelerations are
         -- relative to this object's position and yaw
-
-        vertical = false,
-        -- If true face player using y axis only
 
         texture = "image.png",
 
--- a/src/client/particles.cpp
+++ b/src/client/particles.cpp
@@ -219,13 +219,6 @@ public:
 			v3f position = p.pos + camera_offset - dtime * p.vector * 1e3;;
 			v3f velocity = p.vector * 1e3f;
 
-			/*
-			aabb3f box {-half_width,-half_width,-half_width,half_width,half_width,half_width};
-			collisionMoveResult r_old = collisionMoveSimple(env, env->getGameDef(),
-					BS * 0.5, box, 0, dtime, &position, &velocity,
-					v3f {0.f, 0.f, 0.f}, nullptr, object_collision);
-			*/
-
 			collisionMoveResult r = collisionMovePoint(env, env->getGameDef(),
 					half_width, dtime, &position, &velocity, v3f {0.f, 0.f, 0.f},
 					nullptr, bounce_fraction, bounce_threshold, object_collision);
@@ -644,8 +637,6 @@ void ParticleManager::handleParticleEvent(ClientEvent *event, Client *client, Lo
 		ps->addAffector(lighting_affector);
 		lighting_affector->drop();
 
-		ps->setDebugDataVisible(irr::scene::EDS_BBOX);
-
 		// Delete allocated content of event
 		delete event->add_particlespawner.minpos;
 		delete event->add_particlespawner.maxpos;
@@ -714,7 +705,6 @@ void ParticleManager::handleParticleEvent(ClientEvent *event, Client *client, Lo
 			new LightingAffector(m_env, client, 0);
 		ps->addAffector(lighting_affector);
 		lighting_affector->drop();
-
 
 		ps->setMaterialFlag(video::EMF_LIGHTING, false);
 		ps->setMaterialFlag(video::EMF_BACK_FACE_CULLING, false);
--- a/src/client/particles.cpp
+++ b/src/client/particles.cpp
@@ -59,15 +59,15 @@ static irr::core::matrix4 bottomUpTextureMatrix(float scale_x, float scale_y,
 class CameraOffsetAffector : public irr::scene::IParticleAffector
 {
 public:
-	CameraOffsetAffector(const ClientEnvironment &env,
+	CameraOffsetAffector(const ClientEnvironment *env,
 		irr::scene::IParticleSystemSceneNode *ps, const v3f &position) :
-		env(env), ps(ps), position(position)
+		m_env(env), ps(ps), position(position)
 	{
 	}
 
 	void affect(u32 now, irr::scene::SParticle *particlearray, u32 count) override
 	{
-		v3s16 camera_offset = env.getCameraOffset();
+		v3s16 camera_offset = m_env->getCameraOffset();
 		ps->setPosition(position - intToFloat(camera_offset, BS));
 	}
 
@@ -76,8 +76,8 @@ public:
 		return irr::scene::EPAT_NONE;
 	}
 private:
-	const ClientEnvironment &env;
-	irr::scene::IParticleSystemSceneNode *const ps;
+	const ClientEnvironment *m_env;
+	irr::scene::IParticleSystemSceneNode *ps;
 	v3f position;
 };
 
@@ -86,7 +86,7 @@ class LightingAffector : public irr::scene::IParticleAffector
 public:
 	LightingAffector(ClientEnvironment *env, IGameDef *gamedef,
 		u8 glow, const video::SColor &base_color = video::SColor(0xFFFFFFFF))
-		: env(env), gamedef(gamedef), glow(glow),
+		: m_env(env), m_gamedef(gamedef), glow(glow),
 		base_color(base_color)
 	{
 	}
@@ -95,7 +95,7 @@ public:
 	{
 		for (u32 i = 0; i < count; ++i) {
 			v3f pos = particlearray[i].pos +
-				intToFloat(env->getCameraOffset(), BS);
+				intToFloat(m_env->getCameraOffset(), BS);
 			color = getParticleLightColor(pos / BS);
 			particlearray[i].color = color;
 		}
@@ -106,11 +106,10 @@ public:
 		return irr::scene::EPAT_NONE;
 	}
 private:
-	ClientEnvironment *env;
-	IGameDef *gamedef;
+	ClientEnvironment *m_env;
+	IGameDef *m_gamedef;
 	u8 glow;
 	video::SColor base_color;
-
 	video::SColor color;
 
 	video::SColor getParticleLightColor(const v3f &position)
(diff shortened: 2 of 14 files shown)


rubenwardy:
This is something that is high on my wish list! The current particle implementation is awful. Is this complete?

src/client/particles.cpp
```diff
@@ -31,635 +29,808 @@ with this program; if not, write to the Free Software Foundation, Inc.,
...
+private:
+	v3f acc;
+	bool disabled;
+	u32 last_time;
+};
+class CustomEmitter : public irr::scene::IParticleEmitter
+{
+public:
+	// unused pure virtual methods
+	virtual void setDirection(const core::vector3df& /* newDirection */) {}
```
rubenwardy: Haven't checked, but if these are from IParticleEmitter then please use override instead like this instead:

```c++
void setDirection(const core::vector3df& /* newDirection */) override {}
```

src/client/particles.cpp
```diff
@@ -31,635 +29,808 @@ with this program; if not, write to the Free Software Foundation, Inc.,
...
+	~SingleEmitter()
+	{
+		--single_particles_count;
+	}
 
-	if (is_attached) {
-		// Apply attachment yaw
-		vel.rotateXZBy(attached_yaw);
-		acc.rotateXZBy(attached_yaw);
+	virtual s32 emitt(u32 now, u32 timeSinceLastCall, irr::scene::SParticle *&outArray)
```
rubenwardy: Why the two `t`s?

src/client/particles.cpp
```diff
@@ -31,635 +29,808 @@ with this program; if not, write to the Free Software Foundation, Inc.,
...
+					rand() % 150 / 50.f,
+					rand() % 150 / 50.f - 1.5f
+				};
+				size = rand() % 8 / 64.f * BS;
+			}
+			p.pos.set(pos);
+			p.startVector = p.vector = vel * BS * 1e-3f;
+
+			p.startTime = now;
+			//TEST
```
rubenwardy: test?

src/client/particles.cpp
```diff
@@ -31,635 +29,808 @@ with this program; if not, write to the Free Software Foundation, Inc.,
...
+		acceleration_affector->drop();
+
+
+		scene::IParticleAffector *lighting_affector =
+			new LightingAffector(m_env, client, 0);
+		ps->addAffector(lighting_affector);
+		lighting_affector->drop();
+
+		// Delete allocated content of event
+		delete event->add_particlespawner.minpos;
```
rubenwardy: I don't like this, but it was there before I guess

src/collision.cpp
```diff
@@ -178,6 +178,138 @@ int axisAlignedCollision(
...
+	// Possible to move backwards in time if inside collision box
+	v3f d;
+	if (inside)
+		d = {0.5f * xsize, 0.5f * ysize, 0.5f * zsize};
+	else
+		d = {0.f, 0.f, 0.f};
+
+	v3f relpos = pos - staticbox.MinEdge;
+
+	if(speed.X > 0) // Check for collision with X- plane
```
rubenwardy: messed up code style

rubenwardy (commented):
This needs to go through a lot of regression testing to make sure that things aren't lost. That, and performance testing too.

I didn't check the collision detection bit. Also, the code style is messed up there

Some modders that may be interested in this: @orwell96 @stujones11 @Fixer-007

src/client/particles.cpp
DTA7: That's a misspelling in the Irrlicht API...
http://irrlicht.sourceforge.net/docu/classirr_1_1scene_1_1_i_particle_emitter.html

src/client/particles.cpp
DTA7: For this, the ClientEvent would have to be restructured.

DTA7:
This is complete as described apart from the code style in the collision part and of course testing.
-Are the properties `bounce_fraction` and `bounce_threshold` appropriate?
-Is the behavior near edges due to `surface_dist` a problem?
-Is there too much code duplication between `collisionMoveSimple` and `collisionMovePoint`? The overall structure is almost the same, but the node finding and speed calculation are not.

src/client/particles.cpp
nerzhul: our clientevent model using a union is just painful, it should use inheritance

doc/lua_api.txt
```diff
@@ -6831,8 +6831,14 @@ Used by `minetest.add_particle`.
...
-        vertical = false,
-        -- If true faces player using y axis only
+		bounce_fraction = 1,
+		-- Fraction of the velocity component that is kept after bouncing
+		-- in the corresponding direction.
+
+		bounce_threshold = 0,
+		-- If the absolute value of the velocity component in the direction
+		-- of a bounce is smaller than bounce_threshold * speed,
+		-- the velocity is set to, i.e. prevents "gliding".
```
SmallJoker: For some reason the docs use spaces to indent. Please use them here as well.

src/client/particles.cpp
```diff
@@ -31,635 +29,801 @@ with this program; if not, write to the Free Software Foundation, Inc.,
...
 	}
 
-	// Update lighting
-	updateLight();
+	irr::scene::E_PARTICLE_AFFECTOR_TYPE getType() const override
+	{
+		return irr::scene::EPAT_NONE;
+	}
+private:
+	ClientEnvironment *env;
```
SmallJoker: Private members ought to have the prefix `m_`

src/collision.cpp
```diff
@@ -60,112 +60,95 @@ struct NearbyCollisionInfo {
...
+		movingbox.MinEdge.Y - staticbox.MinEdge.Y,
+		movingbox.MinEdge.Z - staticbox.MinEdge.Z,
+		movingbox.MaxEdge.X - staticbox.MinEdge.X,
+		movingbox.MaxEdge.Y - staticbox.MinEdge.Y,
+		movingbox.MaxEdge.Z - staticbox.MinEdge.Z
 	);
 
-	if(speed.X > 0) // Check for collision with X- plane
-	{
+	if(speed.X > 0) { // Check for collision with X- plane
```
SmallJoker: Unrelated changes. Please undo or add spaces before `(`.

src/collision.cpp
```diff
@@ -178,24 +161,131 @@ int axisAlignedCollision(
...
+{
+	f32 dtime = 0.f;
+
+	// Keep distance from collision box
+	staticbox.MinEdge.X -= surface_dist;
+	staticbox.MinEdge.Y -= surface_dist;
+	staticbox.MinEdge.Z -= surface_dist;
+	staticbox.MaxEdge.X += surface_dist;
+	staticbox.MaxEdge.Y += surface_dist;
+	staticbox.MaxEdge.Z += surface_dist;
```
SmallJoker: Shorter:
`staticbox.MinEdge += surface_dist;`
`staticbox.MaxEdge += surface_dist;`

http://irrlicht.sourceforge.net/docu/vector3d_8h_source.html#l00043

SmallJoker:
> -Is there too much code duplication between collisionMoveSimple and collisionMovePoint? The overall structure is almost the same, but the node finding and speed calculation are not.

It would be great if they could be combined in some way. Perhaps try to split up common code into parts/functions which logically make sense.

src/client/particles.cpp
numberZero: > ...or m_ for members are allowed.

Allowed, not required. https://dev.minetest.net/Code_style_guidelines

numberZero:
There are some nearly identical parts, like the `if(collideWithObjects)` blocks or adding a node to `cinfo` (and that code, with all its `getNeighborConnectingFace` calls, looks very familiar; `MapNode::getNeighbors` seems to do the same). These should be merged.

src/server.cpp
```diff
@@ -1631,14 +1631,15 @@ void Server::SendSpawnParticle(session_t peer_id, u16 protocol_version,
 	pkt << pos << velocity << acceleration << expirationtime
 			<< size << collisiondetection;
 	pkt.putLongString(texture);
-	pkt << vertical;
```
sfan5: backwards compatibility required

src/server.cpp
```diff
@@ -1676,7 +1677,7 @@ void Server::SendAddParticleSpawner(session_t peer_id, u16 protocol_version,
 
 	pkt.putLongString(texture);
 
-	pkt << id << vertical;
+	pkt << id;
```
sfan5: same here

sfan5:
Something I noticed is that dig particles look weird with half-height nodes (or anything that isn't a full node):
out

The previous behavior looked better.

src/client/particles.cpp
ClobberXD: If it helps readability, then why not? :)

DTA7:
@sfan5 Thanks, I forgot a `* BS`.

src/server.cpp
DTA7: How can I do that? Is there an example for backwards compatible additional data in a packet?

DTA7:
Regarding the `attached` property, I tried it with the player as attached object. The position works correctly, but `attached->getPosition()` always returns `0` such that the direction always stays the same. This is also the case without this pull request, is this intended?

obneq:
really nice to see this long standing issue addressed so well, hope this can make it into main soon!

src/server.cpp
sfan5: For data in the middle of a packet that was removed, you need to keep a dummy value in place of that.
For new data at the end, the deserialization need to be wrapped into a `try {} catch(SerializationError &e) {};`

sfan5:
Still missing backwards compatiblity for the network protocol, should be easy to add.

DTA7:
@sfan5 `readF32` returns `0` if the `istringstream` is empty instead of throwing an exception, so I used `is.rdbuf()->in_avail() >= 4`.

src/network/clientpackethandler.cpp
```diff
@@ -942,12 +942,15 @@ void Client::handleCommand_SpawnParticle(NetworkPacket* pkt)
...
 		collision_removal = readU8(is);
 		animation.deSerialize(is, m_proto_ver);
 		glow = readU8(is);
 		object_collision = readU8(is);
-		bounce_fraction = readF32(is);
-		bounce_threshold = readF32(is);
+		if (is.rdbuf()->in_avail() >= 4)
+			bounce_fraction = readF32(is);
+		if (is.rdbuf()->in_avail() >= 4)
+			bounce_threshold = readF32(is);
```
sfan5: this is already wrapped in a `catch (...) {}`, so the `if`s are not needed

src/server.cpp
```diff
@@ -1676,7 +1678,8 @@ void Server::SendAddParticleSpawner(session_t peer_id, u16 protocol_version,
...
+	pkt << true; // Dummy value for vertical
```
sfan5: since the default is `vertical = false`, it should probably be used here instead (since old clients will still interpret this value)

sfan5:
One thing I wondered about removal of `vertical`:
does Irrlicht's particle system not support these? or is there another reason?

sfan5 (approved):
Code looks good
Tested to work (incl. backwards compatiblity)

src/server.cpp
DTA7: You're right, thanks

DTA7:
@sfan5 They are always facing the player as with `vertical=false`.
https://github.com/zaki/irrlicht/blob/1b0e050d90f280507ff248fc85897efe6737a13c/source/Irrlicht/CParticleSystemSceneNode.cpp#L310

sfan5:
So Irrlicht does not support `vertical = true`, is that correct?

DTA7:
Yes, but I'm not sure whether this option was a good idea. This is the old behavior with `vertical=true` when looking from above the particles:
screenshot_20190428_191632

doc/lua_api.txt
```diff
@@ -6831,8 +6831,14 @@ Used by `minetest.add_particle`.
         -- `physical = true,` and `collide_with_objects = true,`.
         -- Requires collisiondetection = true to have any effect.
 
-        vertical = false,
-        -- If true faces player using y axis only
```
DS-Minetest: It should probably be documented that this feature existed in the past and has been removed. Else some people might wonder when they see it still used in old mods.

src/client/particles.h
```diff
@@ -26,204 +26,48 @@ with this program; if not, write to the Free Software Foundation, Inc.,
...
-	 *
-	 * We don't need to check the particle spawner list because client ID will n
-	 * ever overlap (u64)
-	 * @return new id
+	* This function is only used by client particle spawners
+	*
+	* We don't need to check the particle spawner list because client ID will n
+	* ever overlap (u64)
+	* @return new id
 	 */
```
DS-Minetest:
```suggestion
	*/
```

src/network/clientpackethandler.cpp
```diff
@@ -991,15 +997,17 @@ void Client::handleCommand_AddParticleSpawner(NetworkPacket* pkt)
...
 
-	bool vertical          = false;
 	bool collision_removal = false;
 	u16 attached_id        = 0;
 	TileAnimationParams animation;
 	animation.type         = TAT_NONE;
 	u8 glow                = 0;
 	bool object_collision  = false;
+	f32 bounce_fraction = 1.f;
+	f32 bounce_threshold = 0.f;
```
DS-Minetest: Missing alignment.

src/client/particles.cpp
```diff
@@ -31,635 +29,795 @@ with this program; if not, write to the Free Software Foundation, Inc.,
...
-}
+	video::SColor getParticleLightColor(const v3f &position)
+	{
+		u8 light = 0;
+		bool pos_ok = true;
+
+		v3s16 p = v3s16(
+			floor(position.X+0.5),
+			floor(position.Y+0.5),
+			floor(position.Z+0.5));
```
DS-Minetest: 3*
```suggestion
			floor(position.Z + 0.5f));
```

rubenwardy:
From a quick github search, here is one instance where `vertical=true` is used: https://github.com/MarkuBu/farlands050/blob/9d77bcc889bc251502a1586162b282943003c007/mods/decoblocks/lantern.lua#L101

DTA7:
@rubenwardy Ok I see what you mean, this is from the mod of the second link.
screenshot_20190429_203719
But to change this the Irrlicht code would have to be changed, but it's an external dependency...

doc/lua_api.txt
```diff
@@ -6831,6 +6831,9 @@ Used by `minetest.add_particle`.
         -- `physical = true,` and `collide_with_objects = true,`.
         -- Requires collisiondetection = true to have any effect.
 
+	-- deprecated: `vertical = false`
+	-- Particles now always face player.
```
sfan5: indentation

DTA7: Oh right, sorry

doc/lua_api.txt
DS-Minetest: Please press grafik.
:arrow_down:

paramat:
'Vertical' is an important feature, many mods rely on this, for example most weather mods (like snowdrift and HW weather) use it for rain.
If at all possible the feature should be preserved.

rubenwardy:
Irrlicht sucks. The preferred way of extending a built-in class is to copy in the entire definition of the class into your own project to re-implement that interface

I had to do this in #8383

paramat:
Hmm i wonder if we can live without 'vertical' ... rain would have to become round drops unfortunately.
If i had to choose then faster particles has priority over 'vertical'.

Would it be practical to keep the old particle implementation as an option, at least for the medium term?

rubenwardy:
I'd be against keeping the old implementation as an option, and would prefer an Irrlicht extension

DTA7:
@rubenwardy @paramat In that case it would be easiest not to use the Irrlicht implementation, but to improve the existing implementation and of course keeping the collision changes.
I think the performance improvement without collision comes mainly from the rendering of particle spawners, which are rendered more efficiently in the Irrlicht case compared to the old implementation, where each particle is rendered separately (with a separate "material" for each particle).
The relevant parts in the Irrlicht implementation are not long:
https://github.com/zaki/irrlicht/blob/1b0e050d90f280507ff248fc85897efe6737a13c/source/Irrlicht/CParticleSystemSceneNode.cpp#L310
https://github.com/zaki/irrlicht/blob/1b0e050d90f280507ff248fc85897efe6737a13c/source/Irrlicht/CParticleSystemSceneNode.cpp#L411
https://github.com/zaki/irrlicht/blob/1b0e050d90f280507ff248fc85897efe6737a13c/source/Irrlicht/CParticleSystemSceneNode.cpp#L586

Should the particles of a particle spawner show the animation at the same time or separately? This should be easily doable using the texture coordinates.

paramat:
> I'd be against keeping the old implementation as an option, and would prefer an Irrlicht extension

That would be fine with me, if an Irrlicht extension is practical and worth the effort just for 'vertical'.

DTA7 to be clear, what you have done here may be fine, we may have to live without 'vertical', i just think we need to consider what to do.
Keeping 'vertical' in some way may end up too much complexity to be worth it.

paramat:
Hmm, 'vertical' is really important though, seeing how important weather mods are and how rain depends on 'vertical' to look ok. I think we should preserve 'vertical', despite this being a lot of work.

ExeVirus:
I would like to jump in late on this issue, but I have a suggestion for the near term that should dramatically decrease the frame-rate drop associated with particles.

Right now, all particles are rendered if a player is within range and there is no culling (non-rendering) of particles off-screen. 

To fix this in a way that requires minimal processing, I suggest a check of the x,y,z coordinate of the particle against the view of a given player. This view could be simplified to a bounding box, to speed up calculations, and would only include values starting at the player x,y,z and extending out in the direction of current view.

See below for my thinking:
Bounding Box Example
Here I have only given the x and z min/max values, but the point is this bounding box would take much less overhead than rendering every particle (a given particle could fail on the first min.x comparision and return) in range and thereby increase number of supported particles at a time by a minimum of 2x. (more likely 3x with some better math by people smarter than I)

paramat:
Possibly a good idea, although due to field of view the bounding box cannot have parallel sides, so i would simply render all particles that are on the 'player view' side of the perpendicular line through the player, this would halve the rendered particles.

ExeVirus:
I came up with a better solution that uses a bounding sphere instead to allow for optimized calculations: 

Step 1: Get player position, view vector and view distance. 

Step 2: find the point that is exactly 1/2 the view distance away from player position along the view vector. 

Step 3: Check if particle distance to this point is less than 1/2 the view distance away (plus some fudge value like 10 up to 30). 


I tested this method using world edit and the hollow sphere command. the only particles that would get cut off are those on your far left and right vision when playing on a extra wide-screen computer (i.e. 700x1300).

Depending on the size of that fudge value you cut the potential render area for entities and particles down by a factor of 4-5, with only a distance calculation and comparison (for each entity and particle).

I'll submit a PR soon with the idea and see if you all like it or not.

numberZero:
@ExeVirus Your suggestion may slow things down if used per-particle. All these calculations are done by the GPU anyway. Sure GPU gets 4 vertex per particle, but it is much faster than the CPU (on such parallelizable tasks), so should cope well. What may help is particle grouping and culling on per-group basis, so that the CPU has only to check the bounding box to prevent sending a bunch of particles to the GPU.

ExeVirus:
@numberZero while I agree and understand that the GPU is much faster than the CPU once the particles are sent there, we are doing a lot of C++ processing for every particle on CPU before sending. especially for "vertical" particles. Read through the particle:step() function and the update lighting and vertices functions to see what I mean: https://github.com/minetest/minetest/blob/47492386ece5b016a7a0ed06d6a44cc7d60adb55/src/client/particles.cpp#L133

I am not disagreeing that this will be inferior to using the Irrlicht engine particles, those will be much much faster, but in this comments thread I see a lot of concern about losing the "vertical" option and I was thinking this would be a way to still get some speedup without sacrificing wanted features. 

I still fully want to move to pure GPU processing for particles, that would make life so much better, but I see where others are concerned.

numberZero:
> I still fully want to move to pure GPU processing for particles

Is that possible at all without shaders? We still has that mode.

> Read through the particle:step() function and the update lighting and vertices functions to see what I mean

`step` remains there regardless of rendering.

https://github.com/minetest/minetest/blob/47492386ece5b016a7a0ed06d6a44cc7d60adb55/src/client/particles.cpp#L231-L238

For vertical particles, trigonometry can be replaced with square root. Or the angle might be replaced with `m_player->getYaw()`, not sure will that look OK. And rotating by such fixed angle may be replaced with matrix multiplication.

With good shaders, `updateVertices` is totally unnecessary. Animation would be easier with texture arrays, but current technique is totally usable from a shader (unlike that used for node animation, where the texture strip is pre-cut).

kilbith:
Your PR:
screenshot_20190731_142820

Minetest HEAD:
screenshot_20190731_143217

There is *much* less particles with some rain mod.

pyrollo:
In `particle.cpp`, class `ContinuousEmitter`, `emitt` only emits one particle per call. If particle spawning frequency is high, it misses spawning many particles. It should spawn as many particles as expected during the given time.

kilbith:
See: https://github.com/EvidenceBKidscode/kidscode/commit/43d781e5c3de3

paramat:
To clarify, i am unfortunately :-1: for the current state of this due to the breaking of the 'vertical' particle feature, which is essential for rain.

paramat:
It seems the way forward is to, as suggested by rubenwardy, use an Irrlicht extension to preserve the 'vertical' feature. In which case i assume this PR is still usable, but needs work on the Irrlicht extension.

paramat:
Seems to need adoption as it seems big changes are needed to preserve the 'vertical' feature.

rubenwardy:
It would be good to gather uses of vertical - is it actually needed by weather systems, or can it be worked around?

paramat:
I cannot see how it can be worked around for elongated raindrops, as used in 'snowdrift' mod and Hidden Worlds weather. Very few people find round raindrops acceptable.
The vertical feature has been present for a long time and is useful for other uses of particles, i do not think removing this feature can even be considered.

ExeVirus:
Why remove the feature, rather than separate it out?

I.e. Make a new way to access particles that would be used going forward
and would implement whatever changes you are thinking that wouldn't work
for vertical particles?

SmallJoker:
There was a major cleanup in the particle code recently. Rebasing the PR might not make sense any more?

rubenwardy:
We have forked Irrlicht, can this vertical issue be solved there?

sfan5:
Another thing I've been wondering about is if it wouldn't be possible to keep the old code exclusively for vertical particles and use Irrlicht's particle system for everything else.

Issue #8476, opened 2019-04-17
Changing password on mobile may put new password in chat


archfan7411:
##### Issue type

 - Bug report

 
##### Minetest version

```
5.0.1
```

##### OS / Hardware

Operating system: Android
CPU: Unknown

GPU model: N/A
OpenGL version: Unknown

##### Summary

On the server Frost Factions, mobile players have reported that changing their password inputs the new password in the chat. The players confirmed that they were running the official Android Minetest.

I cannot personally test this due to the lack of a mobile device with Minetest.

##### Steps to reproduce

Open the password changing dialogue on Android.
Change your password.

Wayward1:
I can confirm, running latest Minetest from the Play Store.
![Screenshot_2019-04-17-10-51-48](https://user-images.githubusercontent.com/8821916/56298152-a86aa880-60ff-11e9-9a53-2488103b11b0.png)

```
2019-04-17 10:50:41: ACTION[Server]: me joins game. List of players: me
2019-04-17 10:50:41: WARNING[Main]: Client side scripting is disabled by client.
2019-04-17 10:50:58: ACTION[Server]: CHAT: <me> test
2019-04-17 10:51:24: ACTION[Server]: CHAT: <me> hahaha
2019-04-17 10:51:31: ACTION[Server]: CHAT: <me> hahahalol
2019-04-17 10:51:39: ACTION[Server]: CHAT: <me> hahahalol

archfan7411:
Was this perhaps a feature for debugging that accidentally made it into release?

ClobberXD:
Regression probably introduced by 323c860. @stujones11

stujones11:
Perhaps a little unfair to call it a regression when this form never worked on android before that patch, however, I have found the actual cause and have submitted a PR

ClobberXD:
> Perhaps a little unfair to call it a regression when this form never worked on android before that patch...

Apologies for mis-wording this issue (didn't pay much attention to what I was typing). A bug it is... :)

PR #8478, opened 2019-04-18
Android: Clear chat open flag on cancel or completion

stujones11:
Fixes:  #8476

This will ensure that only legitimate chat input gets handled as such. Previously this flag was set but then never unset.

### DIFF
--- a/src/client/game.cpp
+++ b/src/client/game.cpp
@@ -1880,6 +1880,9 @@ void Game::processKeyInput()
 	} else if (wasKeyDown(KeyType::INVENTORY)) {
 		openInventory();
 	} else if (input->cancelPressed()) {
+#ifdef __ANDROID__
+		m_android_chat_open = false;
+#endif
 		if (!gui_chat_console->isOpenInhibited()) {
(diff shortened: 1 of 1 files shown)


paramat:
Thanks and well done finding the bug.

stujones11:
No problem, though I would appreciate some more dev input as I am really only guessing this was the intended purpose of that flag, though it would seem otherwise redundant. I have uploaded my debug apk if anyone cares to test it (with usual side-loading precautions!)

https://www.dropbox.com/s/0f2dpl4el396cgr/Minetest-debug.apk?dl=0

Wayward1:
Looks good here :)

Screenshot_20190418-164244

I haven't looked for any effect on performance yet, but I'll do some more testing to see if anything significant appears

stujones11:
@Wayward1 thanks for testing, I wouldn't expect any significant performance improvement, though JNI calls do carry some amount of overhead, I am guessing it may help older devices mostly.

nerzhul:
Okay for me

PR #8488, opened 2019-04-26
Improve readability of debug menu by using '|'

ClobberXD:
screenshot_20190426_221014

### DIFF
--- a/src/client/gameui.cpp
+++ b/src/client/gameui.cpp
@@ -97,17 +97,17 @@ void GameUI::update(const RunStats &stats, Client *client, MapDrawControl *draw_
 		std::ostringstream os(std::ios_base::binary);
 		os << std::fixed
 			<< PROJECT_NAME_C " " << g_version_hash
-			<< ", FPS: " << fps
+			<< " | FPS: " << fps
 			<< std::setprecision(0)
-			<< ", drawtime: " << drawtime_avg << "ms"
+			<< " | drawtime: " << drawtime_avg << "ms"
 			<< std::setprecision(1)
-			<< ", dtime jitter: "
+			<< " | dtime jitter: "
 			<< (stats.dtime_jitter.max_fraction * 100.0) << "%"
 			<< std::setprecision(1)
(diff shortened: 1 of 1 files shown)


DS-Minetest:
I have doubts that this is really more readable.

paramat:
debug

Master for comparison. I'm unsure it's an improvement.

paramat:
The current use of commas makes everything somewhat run into the next. I find the PR an improvement.

ClobberXD:
@DS-Minetest The `|` is always a great candidate for separators, as it spans the whole height of the line, and is also (generally) surrounded by spaces on both sides.

I myself should've also added a screenshot of the debug menu in master for comparison. Thanks, paramat.

ClobberXD:
I noticed that yaw and cardinal direction don't have a separator in between. Are they considered to be related, or should I add a separator there?

src/client/gameui.cpp
```diff
@@ -126,10 +126,10 @@ void GameUI::update(const RunStats &stats, Client *client, MapDrawControl *draw_
 			<< "pos: (" << (player_position.X / BS)
 			<< ", " << (player_position.Y / BS)
 			<< ", " << (player_position.Z / BS)
-			<< "), yaw: " << (wrapDegrees_0_360(cam.camera_yaw)) << "° "
+			<< ") | yaw: " << (wrapDegrees_0_360(cam.camera_yaw)) << "°"
```
Thomas--S: If you are not going to add a separator here, I suggest to keep at least the space as it is in master.

```suggestion
			<< ") | yaw: " << (wrapDegrees_0_360(cam.camera_yaw)) << "° "
```

ClobberXD: Will do, if reviewers don't see the need to separate yaw and cardinal direction.

ClobberXD: Restored the space. Thanks.

paramat:
Yaw, NSEW and +_XZ are all closely related so are in one section, no need to change that.

PR #8500, opened 2019-04-29
util/travis/: Allow common.sh and lint.sh to be executable

ClobberXD:
The lack of exec. permission on lint.sh caused quite a bit of confusion on my end, when I was trying to run LINT. Hopefully there's no risk in making these files executable.


nerzhul:
common.sh doesn't intend to be executed directly it's a "lib"

ClobberXD:
I see. Is it not possible to execute the scripts in the terminal, and then call the functions they define?

nerzhul:
sourcing doesn't require exec privs

ClobberXD:
Neato

PR #8501, opened 2019-04-29
Added ability to store and restore map backup to sqlite3 backend

EvidenceBKidscode:
**Purpose**
This PR adds ability, when using sqlite3 back end, to store, manage and restore map backups.
Creating, deleting and restore backups is instantaneous, regardless of the map size or the amount of changes done.

This is a functionality that we needed on our project but it could be useful to other projects like _Capture The Flag_. Likely to be totally useless on traditional Minetest servers as it may destroy player's builds.

**Drawbacks**
  * Works only with sqlite3. Should be very easy to port to other SQL databases, like PostgreSQL. I don't know how it could be difficult to port to non SQL databases.
  * Needs a map conversion that could be *very* long on big maps (but most likely to be useful on small maps like CTF) because ...:
  * It adds a field in the "blocks" table and changes its primary key (Sqlite does not support changing primary key, so this operation involves the copy of the whole "blocks" table into a new one). 

**Axes of improvements**
We did that change with a SQL oriented mind. It may not be the best way for Minetest. Some queries may look a bit complex and could be simplified by "de-SQLize" the thing (more code, less SQL).

The change consist in adding a "versions" table and a "version_id" field in the "blocks" table. The same position could have several versions. Switching from a backup to another consists only in changing versions status (and informing clients of the change). That's why it is instant.

The addition of the "version_id" field is not totally satisfying because :
  * It breaks map compatibility somehow 
  * It makes necessary to use database able to manipulate tables with fields (not like key-value databases)

We thought of adding "version_id" in the 64 bits of the "pos" field. Only 3x16 bits are actually used, so 16 bits could be available for version. This makes it possible to have 64000+ versions.

To keep it instant, we need to create a new version at each *restore*. So this means 64000 possible restore in the server life (if 1 restore each 5 minutes, 24/24 7/7, this makes only 220 days).

A reuse mechanisme could be implemented once unused blocks/versions are purged to get rid of this limitation.

**Known bugs**
  * Wont work / not tested on Android (sqlite3 is slightly different on this platform) but should work with some work done.
  * Loaded entities (alive entities) are not taken in account in backups. Only stored entities are stored/restored.

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -4589,6 +4589,22 @@ Storage API
     * returns reference to mod private `StorageRef`
     * must be called during mod load time
 
+Database Backups
+----------------
+
+* `minetest.map_list_backups()`:
+    * returns a list of existing backups names
+
+* `minetest.map_create_backup(backup_name)`:
+    * Creates a backup of the map
+
+* `minetest.map_restore_backup(backup_name)`:
+    * Restores an existing backup of the map. Warning: this may destroy a lot of
+      players's work
+
+* `minetest.map_delete_backup(backup_name)`:
+	* Deletes an existing backup of the map
+
 Misc.
 -----
 
--- a/src/database/database-sqlite3.cpp
+++ b/src/database/database-sqlite3.cpp
@@ -193,6 +193,23 @@ Database_SQLite3::~Database_SQLite3()
 	SQLOK_ERRSTREAM(sqlite3_close(m_database), "Failed to close database");
 }
 
+bool Database_SQLite3::tableExists(const std::string &table_name)
+{
+	sqlite3_stmt * stmt;
+
+	assert(m_database); // Pre-condition
+
+	SQLOK(sqlite3_prepare_v2(m_database,
+			"SELECT * FROM `sqlite_master` where type='table' and name=?;",
+			-1, &stmt, NULL),
+		"tableExists: Failed to verify existence of table (prepare)");
+
+	str_to_sqlite(stmt, 1, table_name);
+	bool res = (sqlite3_step(stmt) == SQLITE_ROW);
+	sqlite3_finalize(stmt);
+	return res;
+}
+
 /*
  * Map database
  */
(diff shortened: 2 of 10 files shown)


EvidenceBKidscode:
Here is a sample mod code for testing:

```lua
local function backup_exists(backup)
	local list = minetest.map_list_backups()
	for _, name in ipairs(list) do
		if name == backup then
			return true
		end
	end
	return false
end

minetest.register_chatcommand("list_backups", {
	params = "",
	description = "List available backups",
	func = function(name, param)
		local list = minetest.map_list_backups()
		if #list > 0 then
			minetest.chat_send_player(name, "Available backups:")

			for _, backup in ipairs(list) do
				minetest.chat_send_player(name, "  "..backup)
			end
		else
			minetest.chat_send_player(name, "No backups yet.")
		end
	end,
})

minetest.register_chatcommand("restore", {
	params = "<backup_name>",
	description = "Restore a map backup",
	func = function(name, param)
		if not backup_exists(param) then
			return false, string.format("Backup %s does not exist.", param)
		end
		minetest.map_restore_backup(param)
	end,
})

minetest.register_chatcommand("create_backup", {
	params = "<backup_name>",
	description = "Create a new map backuo for later restoration",
	func = function(name, param)
		if backup_exists(param) then
			return false, string.format("Backup %s already exists.", param)
		end
		minetest.map_create_backup(param)
	end,
})

minetest.register_chatcommand("delete_backup", {
	params = "<backup_name>",
	description = "Delete a map backup",
	func = function(name, param)
		if not backup_exists(param) then
			return false, string.format("Backup %s does not exist.", param)
		end
		minetest.map_delete_backup(param)
	end,
})
```

lhofhansl:
Any way to integrate this with #7544?

SmallJoker:
Do you intend to replace the rollback feature with these backups; or how well would these work when used in combination?

ClobberXD:
This does not seem to be performant enough to be used for backing up every tiny change. This seems to be better suited for manual, periodical backups.

pyrollo:
I don't think it could replace the rollback feature as it does not separate at all actions done per player. 
The idea is to be able to "mark" a specific point in time and be able to come back to it later.

Here is our use case:
We have an training on a pre-made map. Teacher makes some preparation and creates a backup. Student comes and play around (building and altering the map). The result of the training can be backed up also. 
Then the teacher has ability to bring map back in the past to make students do a second different attempt to build/solve the training. 
Different versions of the map could be compared by restoring them successively.

We thing that could be interesting for minigames servers, CTF servers or servers for demonstration/educational purpose.

This PR is more about opening a discussion about this "WIP" or "POC" stuff and make it evolve in the right way.

In terms of performances, the way things are organized, it is always instant. The "long" operations are done in a separate thread that cleans up unused data.

pyrollo:
> Any way to integrate this with #7544?

I guess both PR could lead to a clever map versioning system.

sorcerykid:
I'm curious if this feature would be optional for sqlite3 backends? If not, then does it add any additional overhead (in terms of memory, disk usage, or even performance)?

pyrollo:
Bug preventing loaded entities form  being taken in account when creating backup is now fixed

pyrollo:
Hi Sorcerykid, some answers:
- Memory : no impact, apart the storage, everything works as previously
- Disk usage : Blocks backed up and modified are stored twice. But if you don't use backups facility, the only extra disk usage is the new "version_id" column.
- Performances : the block read and write queries are a bit slowed down (theoretically) but not measured.

Summary: If you don't create backups, changes should not be measurable (if you do, there is only an impact on disk space).

SmallJoker:
Honestly I have zero clues about this rollback stuff. Is there any server owner who could share their opinions about this PR and (would be great if) they could give it a try.

pyrollo:
This PR gets a bit old. On our side we have made many changes on the sqlite backend to fit our needs, not only this backup/restore stuff (added a modification timestamp + x,y,z coordinates on each mapblock). There have been some bug fixes and evolution on it.

I'm wondering if it could be a specific database backend apart from usual sqlite backend. I'm interested also to have opinion from server owners, in particular those having to often "reset map".

VanessaE:
@ShadowNinja this does not appear to be intended to replace the rollback function.

But since I'm here, I'll offer my 3 cents (adjusted for inflation).

In my opinion, this proposal solves the question of backups in the wrong way.

"Instantaneous" implies it'll start a full copy of the map running in the background while players go about their business, or it'll start backing-up only the changes players are making in realtime, i.e. an incremental backup or some kind of copy-on-write scheme.  The wording here implies that the backups will be stored within the map database.

Backups of anything should be stored and managed separately from the program that created the live data, including Minetest.  Otherwise a corrupt Minetest folder or map database blows away your backup(s) too.  Backups of critical data should not be incremental, as damage to the "original" backup file will damage all of its instances, up to the first intentional modification.

My servers, altogether, take about 167 GB of storage at the moment; My backup strategy relies on a script I wrote, run by a `cron` job every morning:

* Notifications are sent to all logged-in players that a backup cycle has started.
* The Minetest directory is rsynced to the backup directory while the servers continue running.
* As many other tasks as possible are done here, while the servers are up.
* The servers are then shut down by SIGINT.
* The worlds are rsynced to the same target directory again, i.e. a second pass.
* Some misc cleanups are done, such as trimming rollback databases and clearing-out abandoned player accounts.
* The servers are brought back up.
* Background processes are started to generate the overview maps from that backup data.
* At the same time, the backup is pushed to an off-site file server, which keeps a weeks' worth.  At this point, users' home directories are backed-up as well.
 
Since I use LevelDB for the maps, the first backup pass takes 5-10 minutes, and the second pass is super fast.  The purpose of that second pass is to replace the invalid data that inevitably happens in the first pass, since the servers were running then.  On a good day, when there's not much to copy in the second pass, and not much work for the "misc cleanups" step to do, the servers are only down (as seen by IRC) for about 2 minutes.  30 seconds of that is just the script waiting for a graceful shutdown after SIGINT.

sorcerykid:
As a server owner, I don't really have the need for builtin rollback feature. I developed my own incremental backup system for the map.sqlite databases. Currently my backup strategy is three fold.

* I have a cronjob setup to do a full backup of the entire world directory every day to a temp folder at 1800 CST. This generally only takes about a 90 seconds to complete, so downtime is very brief.

* At 1810 CST, the incremental backup manager is executed. This generates a new incremental backup within my daily archive. Each daily archive contains one full backup of the database from Monday (~16 GB), and 6 incremental backups for every day thereafter (~10-15 megabytes).

* On Mondays at 18:20 CST, the map database, auth database, and the player database in the temp folder are gzipped and rsynced to a secure, offsite cloud backup service. These weekly backups serve as a second-line of defense in case of data-loss on the production server itself.

If I ever need to do a rollback, the incremental backup manager includes a restore command. So for example `lua restore.lua /pub/worlds/jt2 20200601 5` would restore the 5th version from the `20200601` archive (the full backup is always version `0`). There's also a third optional parameter that allows for specifying .txt file with a list of mapblock IDs to restore.

So I'm not really a candidate to test this feature myself.

pyrollo:
This PR is outdated. On Kidscode side, sqlite backend became something very different from MT one. Maybe a second sqlite backend with advanced features would be a better way to go.

Anyway, more explanations and more discussions needed before a new PR is proposed.

PR #8506, opened 2019-05-01
Server::handleChat: Add missing & to prevent unnecessary string copy

ClobberXD:


src/server.cpp
```diff
@@ -2859,7 +2859,7 @@ void Server::handleChatInterfaceEvent(ChatEvent *evt)
 }
 
 std::wstring Server::handleChat(const std::string &name, const std::wstring &wname,
-	std::wstring wmessage, bool check_shout_priv, RemotePlayer *player)
+	std::wstring &wmessage, bool check_shout_priv, RemotePlayer *player)
```
nerzhul: no, you will modify original string as it's not constant

ClobberXD: Yes, which is why I haven't made it a `const`. That being said, does our code-style even allow non-const references? :thinking:

nerzhul: it's not a problem of code style, you will modify the calling passed argument, which is not what we want here.

ClobberXD: I understand, thanks

PR #8515, opened 2019-05-08
Add matrix helpers to builtin

### LINKED ISSUE #7478: Have math matrix facilities built in Minetest itself
##### Issue type

 - Feature request

##### Summary

On a slight variation of https://github.com/minetest/minetest/issues/1157 my feature request is to actually include the math matrix facilites directly to the main Minetest source, or even better, implement it so that actual calculations are offloaded to C++ (just like suggested in the above issue), in order to provide such powerful facilities to modders out of the box.

According to the rules specified here...
https://github.com/minetest/contentdb/blob/master/app/flatpages/policy_and_guidance.md#31-right-to-a-name

...I have "locked" the "matrix" mod name and the use of the "matrix" variable name in the global space with this post on the official forums...

https://forum.minetest.net/viewtopic.php?f=9&t=20329

...and I will be more than happy to hand over such a global name to Minetest itself in case decision is taken to include / implement them.

Desour:
# matrices

#### summary
This PR adds helpers for using 3x3 matrices to builtin in lua.

#### what the helpers can do
These include not only operations like multiplying, adding or transposing matrices, but also making facedir or wallmounted values to rotation matrices and vice versa, getting rotation matrices for rotations around coordinate axes or around any given vector and conversion to and from euler angle vectors used for `obj:set_rotation(…)`.
The matrices can be used in modding for any vector transformation like rotating, scaling, ….

#### usecases
Usecases in detail:
- Mods like mesecons (or pipeworks or etc.) can rotate their "rules" easily.
- Objects can be rotated easily. Example: Rotate falling nodes in facedir orientation.
- Selection- and collisionboxes can be rotated easily.
- `fine_pointed` positions can be rotated.
- Other screwdrivers can be made very easily.
- Much more…

#### where I got the information for calculating and co. from
I got much information for making this PR from wikipedia, ~~are there problems with licencing?~~ (Edit: [no](http://irc.minetest.net/minetest-dev/2019-05-12#i_5543043).)
~~Also I used [this](http://www.gregslabaugh.net/publications/euler.pdf) as source (I got the calculations the same way as described there, I could show you a pdf, probably).~~
[This paper](https://www.semanticscholar.org/paper/Extracting-Euler-Angles-from-a-Rotation-Matrix-Day/668137fa4b875d890f446e689eea1e334bcf6bf6) was used. And [here](https://github.com/minetest/minetest/files/3170051/to_yaw_roll_pitch_show.pdf) is a pdf made by myself for you.

#### testing
Here is a mod that might help you to test everything: https://github.com/DS-Minetest/matrix_test
Play around with its code yourself.

#### some notes
Note that minetest has a left-handed coordinate system. Wikipedia says that these rotations will make rotations clockwise. You can get the rotation direction around a vector or an axis with the left hand rule.
The euler angles however are counter clockwise.
~~I could make it so that the rotation can be gotten with the right hand rule, if you want so. You have to decide.~~
~~I think, there might have been something else that I want you to decide upon, but I don't remember.~~
~~Edit: Should degree instead of radians be used?~~ Edit: [Q&A in irc](http://irc.minetest.net/minetest-dev/2019-05-12#i_5543025), it's ok as it is now.

(Btw. I found an interesting bug(?) in minetest's set_rotation while testing: there is always more than one way to rotate an object with euler angles, if you swap from one to another, you can see the object doing something weird at a moment. I guess this is because the anlges are not sent at once but each on its own? Try secondary using with the test item of the test mod I linked and wait and look.)

#### relations to other issues and PRs
Closes #7478.

### DIFF
--- a/builtin/common/matrix.lua
+++ b/builtin/common/matrix.lua
@@ -11,6 +11,15 @@ function matrix.new(a, ...)
 	end
 end
 
+function matrix.equals(m1, m2)
+	for i = 1, 9 do
+		if m1[i] ~= m2[i] then
+			return false
+		end
+	end
+	return true
+end
+
 function matrix.index(m, l, c)
 	return m[(l - 1) * 3 + c]
 end
--- a/builtin/common/matrix.lua
+++ b/builtin/common/matrix.lua
@@ -144,7 +144,7 @@ function matrix.rotation_around_z(angle)
 	        0,  0, 1}
 end
 
-function matrix.rotation_around_vector(v, angle) --todo: test this and correct it if necessary
+function matrix.rotation_around_vector(v, angle)
 	local length_v = vector.length(v)
 	v = vector.divide(v, length_v)
 	angle = angle or length_v
--- a/builtin/common/matrix.lua
+++ b/builtin/common/matrix.lua
@@ -43,12 +43,6 @@ local function multiply_scalar(m, a)
 end
 
 local function vector_matrix_multiply(m1, v)
-	--~ local m2 = matrix.new()
-	--~ m2[1] = v.x
-	--~ m2[4] = v.y
-	--~ m2[7] = v.z
-	--~ local m3 = matrix.multiply(m1, m2)
-	--~ return vector.new(m3[1], m3[4], m3[7])
 	return {
 		x = m1[1] * v.x + m1[2] * v.y + m1[3] * v.z,
 		y = m1[4] * v.x + m1[5] * v.y + m1[6] * v.z,
@@ -126,7 +120,7 @@ function matrix.rotation_around_z(angle)
 	        0,  0, 1}
 end
 
-function matrix.rotation_around_vector(v, angle)
+function matrix.rotation_around_vector(v, angle) --todo: test this and correct it if necessary
 	local length_v = vector.length(v)
 	v = vector.divide(v, length_v)
 	angle = angle or length_v
@@ -149,5 +143,5 @@ end
 
 function matrix.from_yaw_pitch_roll(v)
 	return matrix.multiply(matrix.rotation_around_y(v.y),
-			matrix.rotation_around_x(v.x),matrix.rotation_around_z(v.z)) -- the order might be wrong
+			matrix.rotation_around_x(v.x), matrix.rotation_around_z(v.z)) -- the order might be wrong; todo:test
 end
--- a/builtin/common/matrix.lua
+++ b/builtin/common/matrix.lua
@@ -119,15 +119,6 @@ function matrix3.invert(a)
 	return b
 end
 
--- note that minetest has a left-handed coordinate system.
--- ergo the rotation will be clockwise
--- (see wikipedia for details)
-
---~ local rounding_precision = 1000000
---~ local function round(a)
-	--~ return math.floor(a * rounding_precision + 0.5) / rounding_precision
---~ end
-
 local function sin(x)
 	if x % math.pi == 0 then
 		return 0
@@ -183,45 +174,6 @@ function matrix3.rotation_around_vector(v, angle)
 	}
 end
 
---~ function matrix3.to_pitch_yaw_roll(m)
-	--~ local r = vector.new()
-	--~ if round(math.abs(m[6])) ~= 1 then
-		--~ r.x = -math.asin(m[6])
-		--~ local cosx = math.cos(r.x)
-		--~ r.y = math.atan2(m[3] / cosx, m[9] / cosx)
-		--~ r.z = math.atan2(m[4] / cosx, m[5] / cosx)
-	--~ elseif m[6] < 0 then -- gimbal lock
-		--~ r.x = math.pi / 2
-		--~ r.z = 0
-		--~ r.y = math.atan2(m[2], m[1])
-	--~ else
-		--~ r.x = -math.pi / 2
-		--~ r.z = 0
-		--~ r.y = math.atan2(-m[2], m[1])
-	--~ end
-	--~ r.x = -r.x
-	--~ r.y = -r.y
-	--~ r.z = -r.z
-	--~ return r
---~ end
---~ function matrix3.to_pitch_yaw_roll(m)
-	--~ local r = vector.new()
-	--~ if round(math.abs(m[6])) ~= 1 then
-		--~ r.x = math.asin(m[6])
-		--~ local cosx = math.cos(r.x)
-		--~ r.y = math.atan2(-m[3] / cosx, m[9] / cosx)
-		--~ r.z = math.atan2(-m[4] / cosx, m[5] / cosx)
-	--~ elseif m[6] > 0 then -- gimbal lock
-		--~ r.x = math.pi / 2
-		--~ r.z = 0
-		--~ r.y = math.atan2(m[2], m[1])
-	--~ else
-		--~ r.x = -math.pi / 2
-		--~ r.z = 0
-		--~ r.y = math.atan2(m[7], m[8])
-	--~ end
-	--~ return r
---~ end
 function matrix3.to_pitch_yaw_roll(m)
 	local r = vector.new()
 	r.y = math.atan2(-m[3], m[9])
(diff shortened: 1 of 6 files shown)


builtin/common/matrix.lua
```diff
@@ -0,0 +1,205 @@
...
+	return {m[1], m[4], m[7],
+			m[2], m[5], m[8],
+			m[3], m[6], m[9]}
+end
+
+-- note that minetest has a left-handed coordinate system.
+-- ergo the rotation will be clockwise
+-- (see wikipedia for details)
+
+local round_exactness = 1000000
```
ClobberXD: Rename to something like `rounding_precision` or simply `precision`? Sounds more... "correct" that way. :)

builtin/game/item.lua
```diff
@@ -112,6 +112,61 @@ function core.facedir_to_dir(facedir)
...
+		--~ elseif m[i] == 1 then
+			--~ r = r + 2 * 3^(i-1)
+		--~ elseif m[i] ~= -1 then
+			--~ return -1
+		--~ end
+	end
+	return r
+end
+
+for i = 0, 23 do
```
ClobberXD: Don't table indexes start from 1 in Lua?

builtin/common/matrix.lua
```diff
@@ -0,0 +1,205 @@
...
+
+function matrix.transpose(m)
+	return {m[1], m[4], m[7],
+			m[2], m[5], m[8],
+			m[3], m[6], m[9]}
+end
+
+-- note that minetest has a left-handed coordinate system.
+-- ergo the rotation will be clockwise
+-- (see wikipedia for details)
```
ClobberXD: Minetest using a left-handed coordinate system has already been documented in multiple places, and this comment block could be removed, IMHO.

builtin/game/item.lua
DS-Minetest: Yes, but formspec param2 values start at 0.

builtin/common/matrix.lua
DS-Minetest: Yes, of course, I'll remove such unneeded comments soon. They are basically just notes to make reviewing easier.

DS-Minetest:
> Also, I notice the following functions do what's essentially a table copy, where the operations can be performed in-place instead:
> 
>     * `matrix.floor`
> 
>     * `matrix.round`
> 
>     * `matrix.apply`
> 
>     * `multiply_scalar`

The reason for this is the consistency to the corresponding `vector.*` functions. I didn't want these functions to make confusion.

builtin/common/matrix.lua
DS-Minetest: Done.

builtin/game/item.lua
ClobberXD: I see. Thanks for clarifying.

DS-Minetest: Perhaps it would be better if `facedir_matrices` was indexed from 1 to 24.
`matrix_facedirs`, however, has 0 to 23 as values, not keys.

ClobberXD:
> The reason for this is the consistency to the corresponding vector.* functions. I didn't want these functions to make confusion.

I think you misunderstood. I meant that the aforementioned functions unnecessarily create a new matrix for operations that can be performed in-place. e.g. the `matrix.apply` function should be defined like so:

```lua
function matrix.apply(m, func)
	for i = 1, 9 do
		m[i] = func(m[i])
	end
	return m
end
```

DS-Minetest:
I didn't misunderstand you, but my answer was imprecise, sorry for that.
The `vector.*` functions return new vectors, too:
https://github.com/minetest/minetest/blob/72feab081c336d2d71d300131d30e71694b485f3/builtin/common/vector.lua#L50-L56
Because of consistency I did the same for the `matrix.*` functions. Otherwise, modders will get confused.

paramat:
p_gimeno on IRC http://irc.minetest.net/minetest-dev/2019-05-11#i_5542547

17:45 p_gimeno   hm, DS-Minetest isn't on IRC, right? The Matrix-Euler conversion functions in #8515 should probably be using the core, or at least the same method as the core.
17:45 ShadowBot  https://github.com/minetest/minetest/issues/8515 -- Add matrix helpers to builtin by DS-Minetest
17:48 p_gimeno   The method is here: https://github.com/minetest/minetest/pull/8019#issuecomment-454976672
17:52 p_gimeno   in the core, they are functions setPitchYawRollRad and getPitchYawRollRad defined in numeric.cpp
18:03 p_gimeno   I also wonder why to deliberately drop precision when returning rotation matrices
18:19 p_gimeno   if the intention is to handle sin(pi/2) and cos(pi), special casing them would be better, rather than reducing the precision under that of 32-bit float

DS-Minetest:
Thanks for transferring these chat messages @paramat.

>18:03 p_gimeno I also wonder why to deliberately drop precision when returning rotation matrices
18:19 p_gimeno if the intention is to handle sin(pi/2) and cos(pi), special casing them would be better, rather than reducing the precision under that of 32-bit float

This is indeed the intention. I will do the special casing.

The function in numeric.cpp and the one at https://github.com/minetest/minetest/pull/8019#issuecomment-454976672 (it is probably the same) are both for matrices from the irrlicht coordinate system (where y and z is swapped), for matrices like this:
rotmat_from_wikipedia
Minetest has this:
other_rotmat_from_wikipedia
(not exactly this but the left handed version of this)
But the method they used seems to be better, according to what https://www.semanticscholar.org/paper/Extracting-Euler-Angles-from-a-Rotation-Matrix-Day/668137fa4b875d890f446e689eea1e334bcf6bf6 says.

paramat:
Replies here http://irc.minetest.net/minetest-dev/2019-05-11#i_5542603
Perhaps you can keep an eye on IRC minetest-dev channel logs. You probably know p_gimeno can't post on Github.

DS-Minetest:
>Perhaps you can keep an eye on IRC minetest-dev channel logs.

Yes, sure.

>02:56 	p_gimeno 	@DS-Minetest, I think I see what happens. We're using opposite matrix conventions. I'm using row vectors (because Irrlicht does too, see http://irrlicht.sourceforge.net/docu/classirr_1_1core_1_1_c_matrix4.html#details) and you're using column vectors. We're transposed with respect to each other :)
02:58 		Irrlicht weirdness should probably not be exposed to users where it can be avoided, so I like your convention better.

I agree that weirdness is bad.
I'll just try to use the same method as in numeric.cpp and implement it in lua. Transferring between conventions is much more confusing.

Sadly I don't really know anything about quaternions yet.

Btw. the X, Y and Z in the rotation matrices I showed up there are matrices, see here: https://en.wikipedia.org/wiki/Euler_angles#Rotation_matrix

Btw is it bad that in this PR there are only 3x3 matrices and not 4x4 matrices like in irrlicht?

DS-Minetest:
Ok, it's done.
[Here's a pdf for you that you can see that the calculations are correct.](https://github.com/minetest/minetest/files/3170051/to_yaw_roll_pitch_show.pdf)

DS-Minetest:
Irc comments are addressed (the ones starting here: http://irc.minetest.net/minetest-dev/2019-05-12#i_5542790).

DS-Minetest:
The new commits describe themselves well, for detailed information you might want to read the irc logs.

DS-Minetest:
New comments from p_gimeno (http://irc.minetest.net/minetest-dev/2019-05-14#i_5543844) were addressed.

DS-Minetest:
Ok, the code is cleaned up.
I've also updated the first comment.

Input about documentation, code style and code comments might be useful.
Fell free to test and review.

(Actually I don't know a reliable way to test `minetest.wallmounted_to_matrix3` and `minetest.matrix3_to_wallmounted`.)

DS-Minetest:
Rebased.

Andrey2470T:
Would it be nice to add *n* x *m* matrices with variable number of rows/columns or would it be redundant and useless?

Desour:
> Would it be nice to add _n_ x _m_ matrices with variable number of rows/columns or would it be redundant and useless?

Arbitrary sized matrices are in general of course useful. However, it doesn't make sense to use the arbitrary-sized matrix stuff implementation for fixed size 3x3 matrices. This PR is mainly for rotation matrices.

Closing.
- Minetest never gives you matrices. Mods can easily add such helpers themselves. Having such helpers in minetest's builtin just makes everything less flexible.
- There would be several things to improve in this PR:
  - Use metatables for matrices.
  - Unrolling some of the loops might be better for performance.
  - Split `matrix.new` into `new`, `zero` and `copy`.
- (Feel free to use any of this PR's changes that come from me under WTFPL, after all they are all trivial.)

PR #8524, opened 2019-05-12
Add compatible, consistent coordinate system to FormSpecs.

v-rob:
This PR gives all FormSpec elements consistent coordinating via a new tag `real_coordinates[<bool>]`.  When set to true, all following elements will use the new system.

**Example Screenshots:**

Shows all elements affected, with the exception of `background` for clarity.
new_all_elements

An example showing the simplicity of using the new coordinating system to make a complex FormSpec.
fake_brower

Both of these example FormSpecs can be found in the example game found at https://github.com/v-rob/minetest_formspec_game (may be outdated at times).

**Technical Stuff:**

One FormSpec coordinate in the new system is the same as the size of an image button in the old one.  There is no padding between elements or at the edge of the form.  `label`, `vertlabel`, and `checkbox` are positioned by the center of the text, and `label` newlines are half of a coordinate.  This was chosen because it would allow more flexibility and has a more normal line spacing, and if a full coordinate is necessary, two newlines will work perfectly.  Inventory lists are spaced at 1/4 a coordinate apart since this is very close to the original while still being unified.

The tag `real_coordinates` can be placed anywhere in the form to turn on or off the new coordinate system.  If it is placed directly after `size`, `position`, `anchor`, and `no_prepend`, then `size` will use the new coordinate system.

Formspec prepends are unaffected by the coordinate system in the main form.  If the coordinate system is changed in either one, it will not affect the other.

Two elements have changed syntax when `real_coordinates` is enabled.  `dropdown` has an extra parameter in the size section for height, and `tabheader` has an extra parameter after the size for the height and width of the entire tabheader.

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -1875,9 +1875,14 @@ is used when the server receives user input. You must not use the name
 Spaces and newlines can be inserted between the blocks, as is used in the
 examples.
 
-Position and size units are inventory slots, `X` and `Y` position the formspec
-element relative to the top left of the menu or container. `W` and `H` are its
-width and height values.
+Position and size units are inventory slots unless the new coordinating system
+is enabled. `X` and `Y` position the formspec element relative to the top left
+of the menu or container. `W` and `H` are its width and height values.
+
+If the new system is enabled, all elements have unified cooridinates for all
+elements with no padding or spacing in between. This is highly recommended
+for new forms. See `real_coordinates[<bool>]`.
+
 Inventories with a `player:<name>` inventory location are only sent to the
 player named `<name>`.
 
@@ -1951,6 +1956,14 @@ Elements
 * Must be used after the `size`, `position`, and `anchor` elements (if present).
 * Disables player:set_formspec_prepend() from applying to this formspec.
 
+### `real_coordinates[<bool>]`
+
+* Must be used either before or anywhere after `size`, `position`, `anchor`, and
+  `no_prepend` (if present) elements.
+* When set to true, all following formspec elements will use the new coordinate system.
+* **Note**: Formspec prepends are not affected by the coordinates in the main form.
+  They must enable it explicitly.
+
 ### `container[<X>,<Y>]`
 
 * Start of a container block, moves all physical elements in the container by
@@ -1968,11 +1981,15 @@ Elements
 
 * Show an inventory list if it has been sent to the client. Nothing will
   be shown if the inventory list is of size 0.
+* **Note**: The spacing between inventory places is not affected by the
+  new coordinate system.
 
 ### `list[<inventory location>;<list name>;<X>,<Y>;<W>,<H>;<starting item index>]`
 
 * Show an inventory list if it has been sent to the client. Nothing will
   be shown if the inventory list is of size 0.
+* **Note**: The spacing between inventory places is not affected by the
+  new coordinate system.
 
 ### `listring[<inventory location>;<list name>]`
 
@@ -2064,7 +2081,8 @@ Elements
 * Textual password style field; will be sent to server when a button is clicked
 * When enter is pressed in field, fields.key_enter_field will be sent with the
   name of this field.
-* Fields are a set height, but will be vertically centred on `H`
+* Fields are a set height, but will be vertically centred on `H` with the old
+  system.  With the new coordinate system, `H` will modify the height.
 * `name` is the name of the field as returned in fields to `on_receive_fields`
 * `label`, if not blank, will be text printed on the top left above the field
 * See `field_close_on_enter` to stop enter closing the formspec
@@ -2074,7 +2092,8 @@ Elements
 * Textual field; will be sent to server when a button is clicked
 * When enter is pressed in field, `fields.key_enter_field` will be sent with
   the name of this field.
-* Fields are a set height, but will be vertically centred on `H`
+* Fields are a set height, but will be vertically centred on `H` with the old
+  system.  With the new coordinate system, `H` will modify the height.
 * `name` is the name of the field as returned in fields to `on_receive_fields`
 * `label`, if not blank, will be text printed on the top left above the field
 * `default` is the default value of the field
@@ -2112,7 +2131,8 @@ Elements
 * The label formspec element displays the text set in `label`
   at the specified position.
 * The text is displayed directly without automatic line breaking,
-  so label should not be used for big text chunks.
+  so label should not be used for big text chunks.  Newlines can be
+  used to make labels multiline, but `textarea` is preferred.
 
 ### `vertlabel[<X>,<Y>;<label>]`
 
@@ -2122,12 +2142,14 @@ Elements
 ### `button[<X>,<Y>;<W>,<H>;<name>;<label>]`
 
 * Clickable button. When clicked, fields will be sent.
-* Fixed button height. It will be vertically centred on `H`
+* Fixed button height. It will be vertically centred on `H` with the old system.
+  With the new coordinate system, `H` will modify the height.
 * `label` is the text on the button
 
 ### `image_button[<X>,<Y>;<W>,<H>;<texture name>;<name>;<label>]`
 
 * `texture name` is the filename of an image
+* **Note**: Height can be modified on image buttons.
 
 ### `image_button[<X>,<Y>;<W>,<H>;<texture name>;<name>;<label>;<noclip>;<drawborder>;<pressed texture name>]`
 
@@ -2146,10 +2168,12 @@ Elements
 ### `button_exit[<X>,<Y>;<W>,<H>;<name>;<label>]`
 
 * When clicked, fields will be sent and the form will quit.
+* Same as `button` in all other respects.
 
 ### `image_button_exit[<X>,<Y>;<W>,<H>;<texture name>;<name>;<label>]`
 
 * When clicked, fields will be sent and the form will quit.
+* Same as `button` in all other respects.
 
 ### `textlist[<X>,<Y>;<W>,<H>;<name>;<listelem 1>,<listelem 2>,...,<listelem n>]`
 
@@ -2175,6 +2199,22 @@ Elements
 ### `tabheader[<X>,<Y>;<name>;<caption 1>,<caption 2>,...,<caption n>;<current_tab>;<transparent>;<draw_border>]`
 
 * Show a tab**header** at specific position (ignores formsize)
+* `x` and `y`: position of the tabheader
+* `name` fieldname data is transferred to Lua
+* `caption 1`...: name shown on top of tab
+* `current_tab`: index of selected tab 1...
+* `transparent` (optional): show transparent
+* `draw_border` (optional): draw border
+
+### `tabheader[<X>,<Y>;<H>;<name>;<caption 1>,<caption 2>,...,<caption n>;<current_tab>;<transparent>;<draw_border>]`
+
+* Show a tab**header** at specific position (ignores formsize)
+* **Important note**: This syntax for tabheaders can only be used if the new
+  coordinate system is enabled.
+* `x` and `y`: position of the tabheader
+* `h`: height of the tabheader
+* **Important Note**: The tabheader height can only be modified if the new
+  coordinate system is enabled.
 * `name` fieldname data is transferred to Lua
 * `caption 1`...: name shown on top of tab
 * `current_tab`: index of selected tab 1...
@@ -2193,8 +2233,22 @@ Elements
 * **Important note**: There are two different operation modes:
     1. handle directly on change (only changed dropdown is submitted)
     2. read the value on pressing a button (all dropdown values are available)
-* `x` and `y` position of dropdown
-* Width of dropdown
+* `x` and `y`: position of the dropdown
+* `w`: width of the dropdown
+* Fieldname data is transferred to Lua
+* Items to be shown in dropdown
+* Index of currently selected dropdown item
+
+### `dropdown[<X>,<Y>;<W>,<H>;<name>;<item 1>,<item 2>, ...,<item n>;<selected idx>]`
+
+* Show a dropdown field
+* **Important note**: This syntax for dropdowns can only be used if the new
+  coordinate system is enabled.
+* **Important note**: There are two different operation modes:
+    1. handle directly on change (only changed dropdown is submitted)
+    2. read the value on pressing a button (all dropdown values are available)
+* `x` and `y`: position of the dropdown
+* `w` and `h`: width and height of the dropdown
 * Fieldname data is transferred to Lua
 * Items to be shown in dropdown
 * Index of currently selected dropdown item
@@ -3734,7 +3788,6 @@ Call these functions only at load time!
     * Called after a new player has been created
 * `minetest.register_on_punchplayer(function(player, hitter, time_from_last_punch, tool_capabilities, dir, damage))`
     * Called when a player is punched
-    * Note: This callback is invoked even if the punched player is dead.
     * `player`: ObjectRef - Player that was punched
     * `hitter`: ObjectRef - Player that hit
     * `time_from_last_punch`: Meant for disallowing spamming of clicks
@@ -5399,11 +5452,6 @@ This is basically a reference to a C++ `ServerActiveObject`
     * in first person view
     * in third person view (max. values `{x=-10/10,y=-10,15,z=-5/5}`)
 * `get_eye_offset()`: returns `offset_first` and `offset_third`
-* `send_mapblock(blockpos)`:
-    * Sends a server-side loaded mapblock to the player.
-    * Returns `false` if failed.
-    * Resource intensive - use sparsely
-    * To get blockpos, integer divide pos by 16
 
 `PcgRandom`
 -----------
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2081,8 +2081,8 @@ Elements
 * Textual password style field; will be sent to server when a button is clicked
 * When enter is pressed in field, fields.key_enter_field will be sent with the
   name of this field.
-* Fields are a set height, but will be vertically centred on `H` with the old
-  system.  With the new coordinate system, `H` will modify the height.
+* With the old coordinate system, fields are a set height, but will be vertically
+  centred on `H`. With the new coordinate system, `H` will modify the height.
 * `name` is the name of the field as returned in fields to `on_receive_fields`
 * `label`, if not blank, will be text printed on the top left above the field
 * See `field_close_on_enter` to stop enter closing the formspec
@@ -2199,6 +2199,8 @@ Elements
 ### `tabheader[<X>,<Y>;<name>;<caption 1>,<caption 2>,...,<caption n>;<current_tab>;<transparent>;<draw_border>]`
 
 * Show a tab**header** at specific position (ignores formsize)
+* **Important note**: This syntax for tabheaders can only be used with the
+  old coordinate system.
 * `X` and `Y`: position of the tabheader
 * `name` fieldname data is transferred to Lua
 * `caption 1`...: name shown on top of tab
@@ -2209,8 +2211,8 @@ Elements
 ### `tabheader[<X>,<Y>;<H>;<name>;<caption 1>,<caption 2>,...,<caption n>;<current_tab>;<transparent>;<draw_border>]`
 
 * Show a tab**header** at specific position (ignores formsize)
-* **Important Note**: The tabheader height can only be modified if the new
-  coordinate system is enabled.
+* **Important note**: This syntax for tabheaders can only be used with the
+  new coordinate system.
 * `X` and `Y`: position of the tabheader
 * `H`: height of the tabheader
 * `name` fieldname data is transferred to Lua
@@ -2228,6 +2230,8 @@ Elements
 ### `dropdown[<X>,<Y>;<W>;<name>;<item 1>,<item 2>, ...,<item n>;<selected idx>]`
 
 * Show a dropdown field
+* **Important note**: This syntax for dropdowns can only be used with the
+  old coordinate system.
 * **Important note**: There are two different operation modes:
     1. handle directly on change (only changed dropdown is submitted)
     2. read the value on pressing a button (all dropdown values are available)
@@ -2240,8 +2244,8 @@ Elements
 ### `dropdown[<X>,<Y>;<W>,<H>;<name>;<item 1>,<item 2>, ...,<item n>;<selected idx>]`
 
 * Show a dropdown field
-* **Important note**: This syntax for dropdowns can only be used if the new
-  coordinate system is enabled.
+* **Important note**: This syntax for dropdowns can only be used with the
+  new coordinate system.
 * **Important note**: There are two different operation modes:
     1. handle directly on change (only changed dropdown is submitted)
     2. read the value on pressing a button (all dropdown values are available)
@@ -3786,7 +3790,7 @@ Call these functions only at load time!
     * Called after a new player has been created
 * `minetest.register_on_punchplayer(function(player, hitter, time_from_last_punch, tool_capabilities, dir, damage))`
     * Called when a player is punched
-	* Note: This callback is invoked even if the punched player is dead.
+    * Note: This callback is invoked even if the punched player is dead.
     * `player`: ObjectRef - Player that was punched
     * `hitter`: ObjectRef - Player that hit
     * `time_from_last_punch`: Meant for disallowing spamming of clicks
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2201,6 +2201,7 @@ Elements
 
 * Show a tab**header** at specific position (ignores formsize)
 * `X` and `Y`: position of the tabheader
+* *Note*: Width and height are automatically chosen with this syntax
 * `name` fieldname data is transferred to Lua
 * `caption 1`...: name shown on top of tab
 * `current_tab`: index of selected tab 1...
@@ -2213,7 +2214,20 @@ Elements
 * **Important note**: This syntax for tabheaders can only be used with the
   new coordinate system.
 * `X` and `Y`: position of the tabheader
-* `H`: height of the tabheader
+* `H`: height of the tabheader. Width is automatically determined with this syntax.
+* `name` fieldname data is transferred to Lua
+* `caption 1`...: name shown on top of tab
+* `current_tab`: index of selected tab 1...
+* `transparent` (optional): show transparent
+* `draw_border` (optional): draw border
+
+### `tabheader[<X>,<Y>;<W>,<H>;<name>;<caption 1>,<caption 2>,...,<caption n>;<current_tab>;<transparent>;<draw_border>]`
+
+* Show a tab**header** at specific position (ignores formsize)
+* **Important note**: This syntax for tabheaders can only be used with the
+  new coordinate system.
+* `X` and `Y`: position of the tabheader
+* `W` and `H`: width and height of the tabheader
 * `name` fieldname data is transferred to Lua
 * `caption 1`...: name shown on top of tab
 * `current_tab`: index of selected tab 1...
@@ -2233,7 +2247,7 @@ Elements
     1. handle directly on change (only changed dropdown is submitted)
     2. read the value on pressing a button (all dropdown values are available)
 * `X` and `Y`: position of the dropdown
-* `W`: width of the dropdown
+* `W`: width of the dropdown. Height is automatically chosen with this syntax.
 * Fieldname data is transferred to Lua
 * Items to be shown in dropdown
 * Index of currently selected dropdown item
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2130,14 +2130,21 @@ Elements
 
 * The label formspec element displays the text set in `label`
   at the specified position.
+* **Note**: If the new coordinate system is enabled, labels are
+  positioned from the center of the text, not the top.
 * The text is displayed directly without automatic line breaking,
   so label should not be used for big text chunks.  Newlines can be
-  used to make labels multiline, but `textarea` is preferred.
+  used to make labels multiline.
+* **Note**: With the new coordinate system, newlines are spaced with
+  half a coordinate.  With the old system, newlines are spaced 2/5 of
+  an inventory slot.
 
 ### `vertlabel[<X>,<Y>;<label>]`
 
 * Textual label drawn vertically
 * `label` is the text on the label
+* **Note**: If the new coordinate system is enabled, vertlabels are
+  positioned from the center of the text, not the left.
 
 ### `button[<X>,<Y>;<W>,<H>;<name>;<label>]`
 
@@ -2272,6 +2279,8 @@ Elements
 * `name` fieldname data is transferred to Lua
 * `label` to be shown left of checkbox
 * `selected` (optional): `true`/`false`
+* **Note**: If the new coordinate system is enabled, checkboxes are
+  positioned from the center of the checkbox, not the top.
 
 ### `scrollbar[<X>,<Y>;<W>,<H>;<orientation>;<name>;<value>]`
 
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2392,7 +2392,7 @@ offsets when migrating:
 |---------|------------|---------|-------
 | box     | +0.3, +0.1 | 0, -0.4 |
 | button  |            |         | Buttons now support height, so you will need to set h=1 and reposition if h!=1 before
-| list    |            |         | Note: spacing is now 0.25 for both directions, meaning lists will be taller in height
+| list    |            |         | Spacing is now 0.25 for both directions, meaning lists will be taller in height
 | label   |            |  +0.3   | The first line of text is now positioned centered exactly at the position specified
 
 
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -1879,7 +1879,7 @@ Position and size units are inventory slots unless the new coordinating system
 is enabled. `X` and `Y` position the formspec element relative to the top left
 of the menu or container. `W` and `H` are its width and height values.
 
-If the new system is enabled, all elements have unified cooridinates for all
+If the new system is enabled, all elements have unified coordinates for all
 elements with no padding or spacing in between. This is highly recommended
 for new forms. See `real_coordinates[<bool>]`.
 
@@ -1981,15 +1981,15 @@ Elements
 
 * Show an inventory list if it has been sent to the client. Nothing will
   be shown if the inventory list is of size 0.
-* **Note**: The spacing between inventory slots is not affected by the
-  new coordinate system.
+* **Note**: With the new coordinate system, the spacing between inventory
+  slots is one-fourth the size of an inventory slot.
 
 ### `list[<inventory location>;<list name>;<X>,<Y>;<W>,<H>;<starting item index>]`
 
 * Show an inventory list if it has been sent to the client. Nothing will
   be shown if the inventory list is of size 0.
-* **Note**: The spacing between inventory slots is not affected by the
-  new coordinate system.
+* **Note**: With the new coordinate system, the spacing between inventory
+  slots is one-fourth the size of an inventory slot.
 
 ### `listring[<inventory location>;<list name>]`
 
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -1981,14 +1981,14 @@ Elements
 
 * Show an inventory list if it has been sent to the client. Nothing will
   be shown if the inventory list is of size 0.
-* **Note**: The spacing between inventory places is not affected by the
+* **Note**: The spacing between inventory slots is not affected by the
   new coordinate system.
 
 ### `list[<inventory location>;<list name>;<X>,<Y>;<W>,<H>;<starting item index>]`
 
 * Show an inventory list if it has been sent to the client. Nothing will
   be shown if the inventory list is of size 0.
-* **Note**: The spacing between inventory places is not affected by the
+* **Note**: The spacing between inventory slots is not affected by the
   new coordinate system.
 
 ### `listring[<inventory location>;<list name>]`
@@ -2092,8 +2092,8 @@ Elements
 * Textual field; will be sent to server when a button is clicked
 * When enter is pressed in field, `fields.key_enter_field` will be sent with
   the name of this field.
-* Fields are a set height, but will be vertically centred on `H` with the old
-  system.  With the new coordinate system, `H` will modify the height.
+* With the old coordinate system, fields are a set height, but will be vertically
+  centred on `H`. With the new coordinate system, `H` will modify the height.
 * `name` is the name of the field as returned in fields to `on_receive_fields`
 * `label`, if not blank, will be text printed on the top left above the field
 * `default` is the default value of the field
@@ -2142,8 +2142,8 @@ Elements
 ### `button[<X>,<Y>;<W>,<H>;<name>;<label>]`
 
 * Clickable button. When clicked, fields will be sent.
-* Fixed button height. It will be vertically centred on `H` with the old system.
-  With the new coordinate system, `H` will modify the height.
+* With the old coordinate system, buttons are a set height, but will be vertically
+  centred on `H`. With the new coordinate system, `H` will modify the height.
 * `label` is the text on the button
 
 ### `image_button[<X>,<Y>;<W>,<H>;<texture name>;<name>;<label>]`
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2149,7 +2149,8 @@ Elements
 ### `image_button[<X>,<Y>;<W>,<H>;<texture name>;<name>;<label>]`
 
 * `texture name` is the filename of an image
-* **Note**: Height can be modified on image buttons.
+* **Note**: Height is supported on both the old and new coordinate systems
+  for image_buttons.
 
 ### `image_button[<X>,<Y>;<W>,<H>;<texture name>;<name>;<label>;<noclip>;<drawborder>;<pressed texture name>]`
 
@@ -2199,8 +2200,6 @@ Elements
 ### `tabheader[<X>,<Y>;<name>;<caption 1>,<caption 2>,...,<caption n>;<current_tab>;<transparent>;<draw_border>]`
 
 * Show a tab**header** at specific position (ignores formsize)
-* **Important note**: This syntax for tabheaders can only be used with the
-  old coordinate system.
 * `X` and `Y`: position of the tabheader
 * `name` fieldname data is transferred to Lua
 * `caption 1`...: name shown on top of tab
@@ -2230,8 +2229,6 @@ Elements
 ### `dropdown[<X>,<Y>;<W>;<name>;<item 1>,<item 2>, ...,<item n>;<selected idx>]`
 
 * Show a dropdown field
-* **Important note**: This syntax for dropdowns can only be used with the
-  old coordinate system.
 * **Important note**: There are two different operation modes:
     1. handle directly on change (only changed dropdown is submitted)
     2. read the value on pressing a button (all dropdown values are available)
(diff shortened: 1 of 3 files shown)


src/gui/guiFormSpecMenu.cpp
```diff
@@ -2065,6 +2263,19 @@ void GUIFormSpecMenu::regenerateGui(v2u32 screensize)
 		}
 	}
 
+	/* Copy of the "real_coordinates" element for before the form size. */
+	mydata.real_coordinates = false;
+	for (; i< elements.size(); i++) {
```
SmallJoker: This is an optional formspec element. Use a copy of `i` in case this element is not provided.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -1776,6 +1964,11 @@ void GUIFormSpecMenu::parseAnchor(parserData *data, const std::string &element)
 			<< "'" << std::endl;
 }
 
+void GUIFormSpecMenu::parseRealCoordinates(parserData* data, const std::string
+	&element) {
+	data->real_coordinates = is_yes(element);
```
SmallJoker: Is the parameter needed? Even if `false` would be used in mods to ensure compatibility in the future; old mods would not make use of this element anyway.

src/gui/guiFormSpecMenu.cpp
v-rob: `parseElement` only parses elements after `size`, `position`, and `anchor`.  So, if it is only parsed there, then the formsize will use the old coordinates since this tag only applies to things after it.

As for a copy of `i`, this isn't necessary because it checks if the element is `size` or `invsize` after checking for `real_coordinates`, and it isn't incremented if it is.  This check can't be removed even with a copy of `i` because it might parse a `real_coordinates` after the size, giving an incorrect result.

src/gui/guiFormSpecMenu.cpp
v-rob: There might be possible use cases for the old system, and it doesn't hurt even if it isn't ever used.  I'll wait for more input before changing this.

Inlining sounds good, though.  The function was from before I knew about `is_yes`.

src/gui/guiFormSpecMenu.cpp
SmallJoker: Sorry, my bad. The loop breaks before `i` is incremented - exactly as it should be. Since the order of these initial elements is fixed, it might be safer to `break;` in any case. In fact, `!elements[i].empty()` allows incorrectly written formspecs, such as `formspec_version[1]]]]]real_coordinates[true]`.

v-rob: So, are you saying I should make it like https://github.com/v-rob/minetest/blob/formspec-real-coordinates/src/gui/guiFormSpecMenu.cpp#L2298-#L2300 instead of what I'm using now (`!elements[i].empty()`)?

DS-Minetest:
I guess, this also fixes #6047.

src/gui/guiFormSpecMenu.cpp
SmallJoker: Yes, exactly. That could look like this:
```C++
mydata.real_coordinates = false;
for (; i < elements.size(); i++) {
	std::vector<std::string> parts = split(elements[i], '[');
	std::string name = trim(parts[0]);
	if (name != "real_coordinates"parts.size() != 2)
		break; // Invalid format

	mydata.real_coordinates = is_yes(trim(parts[1]));
}
```

v-rob:
> I guess, this also fixes #6047.

Pretty much.  But to be absolutely technical, it doesn't because the old system still has the problem, but the likeliness of people using the old system is highly unlikely and no one's going to modify it.

> You could change the list element that it has an extra optional parameter (if real coordinates is enabled) which specifies the spacing between the slots.

Not necessary.  There's already a PR for that which looks pretty good: https://github.com/minetest/minetest/pull/7971.

////////////////////

I want to say that this pull request should ideally be merged with all other important formspec changes to group them all together in one Minetest release so that all formspec breaking changes can be grouped together.  I don't think that this PR actually breaks anything, but formspecs designed for this system would look weird in the old system, so it should be grouped with the rest.

Some related PRs are https://github.com/minetest/minetest/pull/8383 and https://github.com/minetest/minetest/pull/8507.

And related to styling that Rubenwardy has expressed he wishes to make PRs for with `style` tags are https://github.com/minetest/minetest/pull/5385 and https://github.com/minetest/minetest/pull/5361

Lastly, we might want to do something about the things expressed by Wuzzy in https://github.com/minetest/minetest/issues/8339.

And, of course, any other things I might have missed and other issues/styling people think up.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -1382,23 +1533,35 @@ void GUIFormSpecMenu::parseTabHeader(parserData* data, const std::string &elemen
 {
 	std::vector<std::string> parts = split(element,';');
 
-	if (((parts.size() == 4) || (parts.size() == 6)) ||
-		((parts.size() > 6) && (m_formspec_version > FORMSPEC_API_VERSION)))
+	if (((parts.size() == 4) || (parts.size() == 6)) || (parts.size() == 7 &&
+		data->real_coordinates) || ((parts.size() > 6) &&
```
SmallJoker: Adding a new parameter without increasing the formspec version will throw the error `Invalid TabHeader element ...` and become unusable on older clients. It sucks.

doc/lua_api.txt
```diff
@@ -2162,6 +2186,22 @@ Elements
...
+
+### `tabheader[<X>,<Y>;<H>;<name>;<caption 1>,<caption 2>,...,<caption n>;<current_tab>;<transparent>;<draw_border>]`
+
+* Show a tab**header** at specific position (ignores formsize)
+* **Important note**: This syntax for tabheaders can only be used if the new
+  coordinate system is enabled.
+* `X` and `Y`: position of the tabheader
+* `H`: height of the tabheader
+* **Important Note**: The tabheader height can only be modified if the new
+  coordinate system is enabled.
```
DS-Minetest: The important note is given twice.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: With 7 parameters it's a new element. It is normal that new elements don't work with older clients. The old element is persisted, ergo this is ok imo.

v-rob: The dropdown does this as well in the geometry section by allowing height.  But if you notice in the code, both of them only allow the extra parameter if `real_coordinates` is enabled.

Breaking these elements doesn't even matter anyway because `real_coordinates` itself breaks forms anyway when put before the form size on older clients.  The clients think that size doesn't exist and spits out a funky FormSpec, as can be seen here:

image

I really wish we could just forget about backwards compatibility and just tell people to update.  But unfortunately, that's not how it works unless you're Microsoft.

SmallJoker: Right, it's guarded by the `real_coordinates` mode, more than that can't be done here.
It's fine to add new elements, as they don't make the entire formspec unusable. In this case it's a little bit different. The players might wonder why the formspec does not show up (correctly, if even). I have no idea on how to make this more user-friendly because node meta formspecs cannot be checked against the presence of `real_coordinates` to send the outdated client a message.
Leave it as-is. Players got to update.

Thomas--S:
I tested this on my Android device, it seems to work well.

doc/lua_api.txt
```diff
@@ -1968,11 +1981,15 @@ Elements
 
 * Show an inventory list if it has been sent to the client. Nothing will
   be shown if the inventory list is of size 0.
+* **Note**: The spacing between inventory places is not affected by the
```
DS-Minetest: `inventory places` sounds a bit weird to me. Maybe `inventory slots` or something like `item rectangles` would be better. But note that I'm not a native speaker, so don't trust me.
(The same is some lines below in the other `list` element.)

doc/lua_api.txt
```diff
@@ -2051,7 +2068,8 @@ Elements
 * Textual password style field; will be sent to server when a button is clicked
 * When enter is pressed in field, fields.key_enter_field will be sent with the
   name of this field.
-* Fields are a set height, but will be vertically centred on `H`
+* Fields are a set height, but will be vertically centred on `H` with the old
+  system.  With the new coordinate system, `H` will modify the height.
```
DS-Minetest: Suggestion:
With the old coordinate system, fields are…. With the new system, ….
This might be clearer.

v-rob:
> I tested this on my Android device, it seems to work well.

Thanks for that, I have no way of testing it on Android so I didn't know if it would work well.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -1936,6 +2124,11 @@ void GUIFormSpecMenu::parseElement(parserData* data, const std::string &element)
 		return;
 	}
 
+	if (type == "real_coordinates") {
+		data->real_coordinates = is_yes(description);
```
DS-Minetest: I think, it would be better having here a `parseRealCoordinates` function as every other element type does the same.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -1382,23 +1533,35 @@ void GUIFormSpecMenu::parseTabHeader(parserData* data, const std::string &elemen
...
+		std::string str_index = parts[i+3];
 		bool show_background = true;
 		bool show_border = true;
 		int tab_index = stoi(str_index) -1;
 
 		MY_CHECKPOS("tabheader",0);
 
-		if (parts.size() == 6) {
-			if (parts[4] == "true")
+		if (parts.size() >= 6) {
```
DS-Minetest: You could do `== 6 + i`.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -1382,23 +1533,35 @@ void GUIFormSpecMenu::parseTabHeader(parserData* data, const std::string &elemen
...
-		std::string str_index = parts[3];
+
+		// If we're using real coordinates, add an extra field for height.
+		// Width is not here because tabs are the width of the text, and
+		// there's no reason to change that.
+		int i = 0;
+		std::string h_geom;
+		if (data->real_coordinates) {
+			i++;
+			h_geom = parts[1];
```
DS-Minetest: What is if the 6 parameter element is used with `real_coordinates`?

src/gui/guiFormSpecMenu.cpp
```diff
@@ -1412,15 +1575,22 @@ void GUIFormSpecMenu::parseTabHeader(parserData* data, const std::string &elemen
...
+			std::vector<std::string> v_geom = {"1", h_geom}; // Dummy value
+			geom = getRealCoordinateGeometry(&v_geom);
+			pos.Y -= geom.Y; // TabHeader base pos is the bottom, not the top.
+		} else {
 			v2f32 pos_f = pos_offset * spacing;
 			pos_f.X += stof(v_pos[0]) * spacing.X;
 			pos_f.Y += stof(v_pos[1]) * spacing.Y - m_btn_height * 2;
 			pos = v2s32(pos_f.X, pos_f.Y);
+
+			geom.Y = m_btn_height*2;
```
```suggestion
			geom.Y = m_btn_height * 2;
```
In this code block, spaces are used.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -1256,15 +1381,33 @@ void GUIFormSpecMenu::parseVertLabel(parserData* data, const std::string &elemen
...
+		if (data->real_coordinates) {
+			pos = getRealCoordinateBasePos(false, &v_pos);
+
+			// Add offset for smaller rect.
+			pos.X += (imgsize.X / 4);
+
+			rect = core::rect<s32>(pos.X, pos.Y,
+				pos.X + imgsize.Y / 2,
+				pos.Y + font_line_height(m_font) *
+				(text.length() + 1));
```
DS-Minetest: > ```
> 			// Actually text.length() would be correct but adding +1 avoids
> 			// breaking mods.
> ```

It might make sense adding this comment here too or not adding `1`.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -1220,12 +1322,34 @@ void GUIFormSpecMenu::parseLabel(parserData* data, const std::string &element)
...
-				pos.X + m_font->getDimension(wlabel.c_str()).Width,
-				posy + m_btn_height);
+
+			core::rect<s32> rect;
+
+			if (data->real_coordinates) {
+				v2s32 pos = getRealCoordinateBasePos(false, &v_pos);
+
+				// Add offset for smaller rect.
+				pos.Y += (imgsize.Y / 4) + ( ((float)imgsize.Y) * i / 2);
```
```suggestion
				pos.Y += (imgsize.Y / 4) + (((float)imgsize.Y) * i / 2);
```

Is it good to change the line distance? Before, it was `spacing.Y * 2.0 / 5.0 = imgsize.Y * 15/13 *  2/5 = imgsize.Y * 6 / 13` which is a bit less than `imgsize.Y / 2`.
Probably it is good. But you should change the code comments up there.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -1220,12 +1322,34 @@ void GUIFormSpecMenu::parseLabel(parserData* data, const std::string &element)
...
+					pos.X, pos.Y,
+					pos.X + m_font->getDimension(wlabel.c_str()).Width,
+					pos.Y + imgsize.Y / 2);
+
+			} else {
+				v2s32 pos = getElementBasePos(false, nullptr);
+				pos.X += stof(v_pos[0]) * spacing.X;
+				pos.Y += (stof(v_pos[1]) + 7.0f / 30.0f) * spacing.Y;
+
+				s32 posy = pos.Y + ((float)i) * spacing.Y * 2.0 / 5.0;
```
DS-Minetest: There's no reason for an extra variable anymore.
```suggestion
				pos.Y += ((float) i) * spacing.Y * 2.0 / 5.0;
```
(Also change `posy` to `pos.Y` below.)

src/gui/guiFormSpecMenu.cpp
```diff
@@ -270,6 +270,25 @@ v2s32 GUIFormSpecMenu::getElementBasePos(bool absolute,
 	return v2s32(pos_f.X, pos_f.Y);
 }
 
+v2s32 GUIFormSpecMenu::getRealCoordinateBasePos(bool absolute,
+		const std::vector<std::string> *v_pos)
```
nerzhul: ref not pointer

src/gui/guiFormSpecMenu.cpp
```diff
@@ -868,12 +947,26 @@ void GUIFormSpecMenu::parseDropDown(parserData* data, const std::string &element
...
+		if (data->real_coordinates) {
 
-		core::rect<s32> rect = core::rect<s32>(pos.X, pos.Y,
-				pos.X + width, pos.Y + (m_btn_height * 2));
+			std::vector<std::string> v_geom = split(parts[1],',');
+			MY_CHECKGEOM("dropdown",1);
+
+			pos = getRealCoordinateBasePos(false, &v_pos);
+			geom = getRealCoordinateGeometry(&v_geom);
+			rect = core::rect<s32>(pos.X, pos.Y, pos.X+geom.X, pos.Y+geom.Y);
```
DS-Minetest: Using dropdowns with only width set is impossible with `real_coordinates` change or document this behaviour.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -868,12 +947,26 @@ void GUIFormSpecMenu::parseDropDown(parserData* data, const std::string &element
 
 		MY_CHECKPOS("dropdown",0);
 
-		v2s32 pos = getElementBasePos(false, &v_pos);
+		v2s32 pos;
+		v2s32 geom;
+		core::rect<s32> rect;
 
-		s32 width = stof(parts[1]) * spacing.Y;
+		if (data->real_coordinates) {
```
DS-Minetest: Why is here an empty line?

doc/lua_api.txt
```diff
@@ -3718,7 +3770,7 @@ Call these functions only at load time!
     * Called after a new player has been created
 * `minetest.register_on_punchplayer(function(player, hitter, time_from_last_punch, tool_capabilities, dir, damage))`
     * Called when a player is punched
-    * Note: This callback is invoked even if the punched player is dead.
+	* Note: This callback is invoked even if the punched player is dead.
```
```suggestion
    * Note: This callback is invoked even if the punched player is dead.
```

src/gui/guiFormSpecMenu.cpp
DS-Minetest: `getElementBasePos` uses a pointer, too. I guess both should be changed.

src/gui/guiFormSpecMenu.cpp
v-rob: It was a function before, but SmallJoker suggested that it be inlined since it's only a single line.  I could make it a macro, but I don't see the point, really.

src/gui/guiFormSpecMenu.cpp
v-rob: Yes, I pretty much copied and pasted and changed the logic.  If I was to change them, how should I do it?  I'm still not very experienced in C++.

src/gui/guiFormSpecMenu.cpp
v-rob: Actually, `text.length()` without adding 1 makes the `vertlabel` too short and cuts off the text, so that comment is actually incorrect.  I should probably remove that completely.

src/gui/guiFormSpecMenu.cpp
v-rob: You shouldn't be allowed to not set the height.  All elements have explicitly set heights with `real_coordinates` because otherwise dropdowns will have a weird height that doesn't match up with everything else.

I'll document this.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: (Or you could define the function as `inline`.)
But you are right, this is probably ok like it is now.
https://github.com/minetest/minetest/pull/8524#discussion_r283117155

src/gui/guiFormSpecMenu.cpp
DS-Minetest: You could also comment that the `+ 1` is for not cutting off the label.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Make the asterisk(*) here before `v_pos` to an ampersand(&) and treat the parameter as if it was a "normal" parameter, ie. remove the dereferencing asterisks in the function body and the referencing ampersands in the function calls.

SmallJoker:
```diff
-	pos = getRealCoordinateBasePos(false, &v_pos);
+	pos = getRealCoordinateBasePos(false, v_pos);

-	const std::vector<std::string> *v_pos)
+	const std::vector<std::string> &v_pos)

-	pos_f.X += stof((*v_pos)[0]) + pos_offset.X;
-	pos_f.Y += stof((*v_pos)[1]) + pos_offset.Y;
+	pos_f.X += stof(v_pos[0]) + pos_offset.X;
+	pos_f.Y += stof(v_pos[1]) + pos_offset.Y;
```

`getElementBasePos()` takes a pointer so that it can be used without a position value (`parseSimpleField()` and `parseLabel()`).

src/gui/guiFormSpecMenu.cpp
v-rob: `imgsize.Y / 2` will line up with the new elements properly so that newlines can be used instead of multiple label elements with ease.  Just `imgsize` would have lines that are too far apart, and if the person decided to use elements spaced half the normal height (which is probably a likely occurrence, since spacings of one is too big and anything else is tricky to work with), then a whole `imgsize` wouldn't line up at all.

The old spacing, though, doesn't make sense at all with the new system, which is why I am changing it.

src/gui/guiFormSpecMenu.cpp
v-rob: There's a lot of things like this, and I don't want to fix them all because that would bloat the PR and make it harder to review.  I fix some of them, though.

src/gui/guiFormSpecMenu.cpp
v-rob: I don't see the problem.  All other parameters are offset by one, and you have to use the sixth (well, actually seventh because of the height) parameter.  In fact, using less than six parameters isn't even documented in `lua_api.txt`, so we don't actually have to worry about it.

DS-Minetest: With "6 parameter element" I was referring to the tabheader element that only needs 6 parameters and ergo has no height.
But as you have added the following lines to the documentation, this is now fixed.
```
* **Important note**: This syntax for tabheaders can only be used with the
  old coordinate system.
```

doc/lua_api.txt
```diff
@@ -1968,11 +1981,15 @@ Elements
 
 * Show an inventory list if it has been sent to the client. Nothing will
   be shown if the inventory list is of size 0.
+* **Note**: The spacing between inventory slots is not affected by the
+  new coordinate system.
```
DS-Minetest: This is no more true.

doc/lua_api.txt
```diff
@@ -1968,11 +1981,15 @@ Elements
...
 
 ### `list[<inventory location>;<list name>;<X>,<Y>;<W>,<H>;<starting item index>]`
 
 * Show an inventory list if it has been sent to the client. Nothing will
   be shown if the inventory list is of size 0.
+* **Note**: The spacing between inventory slots is not affected by the
+  new coordinate system.
```
DS-Minetest: Same as above.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -66,8 +66,8 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 
 #define MY_CHECKGEOM(a,b)													\
 	if (v_geom.size() != 2) {												\
-		errorstream<< "Invalid pos for element " << a << "specified: \""	\
-			<< parts[b] << "\"" << std::endl;								\
+		errorstream<< "Invalid geometry for element " << a <<				\
+			"specified: \"" << parts[b] << "\"" << std::endl;				\
```
DS-Minetest: Because of consistency with `MY_CHECKPOS`, `"specified: \""` should be moved back up.

DS-Minetest:
:+1: 
It's not easy to find any more mistakes by reviewing, even with the most pedantic sight.

The travis check failed because of some outdating stuff, it isn't the fault of this PR.

This change is very important.

doc/lua_api.txt
v-rob: I'm still tuning this.  I got it totally wrong last time (they were overlapping instead of spaced), so I haven't changed the docs yet.

I decided to change it because even though #7971 can edit the spacing, it will edit it relative to the current values.  I'm currently testing a spacing of half an inventory slot both horizontally and vertically.  This is probably best since it will follow the coordinate system well enough without being too far apart, but I haven't tested it (just started compiling).

src/gui/guiFormSpecMenu.cpp
v-rob: Moving `"specified: \""` back up makes it too long and it hits the backslash (using 4 space long tabs which the dev wiki code style guidelines suggest).  I could increase all the indents, but I don't know if I should.

v-rob:
I've changed the inventory slots to have a one-fourth space in between so that they're more unified without changing the look a huge amount (it's fairly hard to tell the difference).  I've updated the docs accordingly as well as the PR screenshot with the inventory list.  To illustrate the difference, here are before and after pics of the inventory slots.

Before:
all_elements

After:
new_all_elements

And for fun, this formspec with `real_coordinates` disabled:
old_elements


I think that this PR is pretty much almost ready to merge.  Thanks for all the reviewing you've been doing!

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Oh, ok, then this seems ok as it currently is.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -1412,15 +1582,23 @@ void GUIFormSpecMenu::parseTabHeader(parserData* data, const std::string &elemen
...
+		} else {
 			v2f32 pos_f = pos_offset * spacing;
 			pos_f.X += stof(v_pos[0]) * spacing.X;
 			pos_f.Y += stof(v_pos[1]) * spacing.Y - m_btn_height * 2;
 			pos = v2s32(pos_f.X, pos_f.Y);
+
+			geom.Y = m_btn_height * 2;
 		}
-		v2s32 geom;
 		geom.X = DesiredRect.getWidth();
```
rubenwardy: you can add support for W with this element here

src/gui/guiFormSpecMenu.cpp
```diff
@@ -348,13 +367,19 @@ void GUIFormSpecMenu::parseList(parserData* data, const std::string &element)
...
+		if (location == "context" || location == "current_name")
 			loc = m_current_inventory_location;
 		else
 			loc.deSerialize(location);
 
-		v2s32 pos = getElementBasePos(true, &v_pos);
+		v2s32 pos;
 		v2s32 geom;
+
+		if (data->real_coordinates)
```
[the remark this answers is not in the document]
v-rob: I would prefer not to do this because there are so many irregularities and exceptions.  First off, there is the `getRealCoordinateGeometry`, which has no equivalent function for the older system, so the `if` would be required.  Also, there are some elements (like fields) which modify the position again after calling `getElementBasePos`.  The end result would likely be that the `if` would still be used nearly as much.

doc/lua_api.txt
```diff
@@ -2162,6 +2186,22 @@ Elements
...
+* `transparent` (optional): show transparent
+* `draw_border` (optional): draw border
+
+### `tabheader[<X>,<Y>;<H>;<name>;<caption 1>,<caption 2>,...,<caption n>;<current_tab>;<transparent>;<draw_border>]`
+
+* Show a tab**header** at specific position (ignores formsize)
+* **Important note**: This syntax for tabheaders can only be used with the
+  new coordinate system.
+* `X` and `Y`: position of the tabheader
+* `H`: height of the tabheader
```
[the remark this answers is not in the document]
v-rob: I agree in a way, but width for tabs would be strange since it would apply to all tabs, so text might be cut off on some and have a lot of padding on another.  Another potential solution is to make the width add padding to the sides of the text.

I'd like more input on how I should do this.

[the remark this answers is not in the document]
rubenwardy: surely it would be the width of the tab header? So it would stop and add the `<>` at a smaller width, as shown here:

image

src/gui/guiFormSpecMenu.cpp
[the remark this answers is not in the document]
rubenwardy: Ok, this is fine then

doc/lua_api.txt
```diff
@@ -2177,11 +2217,27 @@ Elements
 ### `dropdown[<X>,<Y>;<W>;<name>;<item 1>,<item 2>, ...,<item n>;<selected idx>]`
 
 * Show a dropdown field
+* **Important note**: This syntax for dropdowns can only be used with the
```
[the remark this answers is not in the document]
v-rob: A good idea; of course, I'll use a full coordinate as the default, not the not-quite-full-but-more-than-half thing dropdowns use now.

doc/lua_api.txt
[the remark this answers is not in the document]
v-rob: Ahh, that makes sense.  I'll implement this then.  I think that it should be allowed to be left as default as well, though, since that's good enough for most people.

DS-Minetest:
You forgot the tooltip in rect_mode:
https://github.com/minetest/minetest/blob/d31f6074e0a8a2f7a27d26ab3be70d4344284b7b/src/gui/guiFormSpecMenu.cpp#L1851-L1866

SmallJoker:
I successfully tested the PR with your game. Works nicely, just one question:

> parseElement only parses elements after size, position, and anchor. So, if it is only parsed there, then the formsize will use the old coordinates since this tag only applies to things after it.

`::parseSize` does not make use of the new coordinate system, so wouldn't it be possible to show the formspecs to older clients by moving `real_coordinates` below `size`?

v-rob:
> `::parseSize` does not make use of the new coordinate system, so wouldn't it be possible to show the formspecs to older clients by moving `real_coordinates` below `size`?

In other words, you're saying that if `real_coordinates` is _directly_ after `size`, `position`, `anchor`, and `no_prepend`, it will affect the size?  That could easily work, although it would be strange.  But hey, the formspec syntax is already crazy :-)

@DS-Minetest Thank you very much for noticing the tooltip.  I had totally forgotten!

//////////////////////////////

I was just wondering, should new elements for formspecs (e.g. scrollable containers) try to emulate inventory slot behavior as well as `real_coordinates` or only use `real_coordainates`?

v-rob:
Ok, tooltips are fixed and `real_coordinates` has been moved after `size`, `position`, `anchor`, and `no_prepend` elements.  Formspec now mostly show on older clients.  Of course, the old coordinates are used, but unfortunately, tabheaders will not show because of the added argument.  Unfortunately, there's no way to fix this as far as I can see.

DS-Minetest:
> I was just wondering, should new elements for formspecs (e.g. scrollable containers) try to emulate inventory slot behavior as well as `real_coordinates` or only use `real_coordainates`?

Afaik, the old system was not consistent and ergo not really a system. New elements should simply use a system, the new, "real" system. If the real_coordinates are not used, it's simply another inconsistent element. Also, new formspecs that can have new elements will for sure use `real_coordainates`.

Can we have maybe a flag in `minetest.features` or somewhere else, that tells the mods whether the `real_coordinates` element is possible? (Maybe it would make more sense to make an entire `formspec` table in `minetest.features`, that has for example for each element whether it can be used or not. Maybe also a flag that tells whether the real coordinates are used by default.) Hm, better forget this for now, this is something for another PR. Do not make this PR too complex, imo we need it into master, changes slow down the reviewing and merging process.

Maybe you should add a note in the docs that if the size should use the old system, something like `size[...]real_coordinates[false]real_coordinates[true]some_other_element[...]` can be done.
(Perhaps say it more general, "If **enabled** immediately after `size`, ... elements, the form size will use the new coordinate system.".)

rubenwardy:
> Can we have maybe a flag in minetest.features or somewhere else, that tells the mods whether the real_coordinates element is possible? 

No, because the feature is per-client

rubenwardy:
It would be handy to have a tag like `size_rc[...]` or `size_real[...]` to do the equivilant of this:

```
real_coordinates[true]
size[...]
```

rubenwardy:
Whilst testing this, I'm creating a migration document. This should go lua_api when complete:

| Element      | Position | Size | Notes                                                                                 |
|--------------|----------|------|---------------------------------------------------------------------------------------|
| box          | +0.3     |      |                                                                                       |
| button       |          |      | Buttons now support height, so you will need to set h=1 and reposition if h!=1 before |
| size         |          |      |                                                                                       |
| container    | -        | -    |                                                                                       |
| list         |          |      |                                                                                       |
| tooltip      |          |      |                                                                                       |
| image        |          |      |                                                                                       |
| item_image   |          |      |                                                                                       |
| image_button |          |      |                                                                                       |
| background   |          |      |                                                                                       |
| field        |          |      |                                                                                       |
| textarea     |          |      |                                                                                       |
| label        | +0.1     |      |                                                                                       |
| vertlabel    |          |      |                                                                                       |
| textlist     |          |      |                                                                                       |
| tabheader    |          |      |                                                                                       |
| dropdown     |          |      |                                                                                       |
| checkbox     |          |      |                                                                                       |
| scrollbar    |          |      |                                                                                       |
| table        |          |      |                                                                                       |

v-rob:
> It would be handy to have a tag like `size_rc[...]` or `size_real[...]` to do the equivalent of this:
> 
> ```
> real_coordinates[true]
> size[...]
> ```

This would create the same problem as why I removed it.  Since it's an unknown tag (and the most important one at that, since other elements won't work without it), the form won't have a size.  It stinks.

So where should I put the list when done?  Just after all the formspec stuff like at [Line 2271](https://github.com/minetest/minetest/blob/master/doc/lua_api.txt#L2271)?  Or somewhere else?

rubenwardy:
> So where should I put the list when done? Just after all the formspec stuff like at Line 2271? Or somewhere else?

Yeah, create a section called "Real Coordinates"

v-rob:
It probably should be put on all the elements that have padding either in a new column or as part of the size column.  Some of them (`field`, `textarea`, and a few others) don't have the padding, so that must be documented.

rubenwardy:
## Migrating to Real Coordinates

In the old system, positions included paddding. For example, two 1x1 elements at 0,0 and 1,1 would have padding between them. To recreate an old layout with padding, you'll need put the positions and sizes through the following formula:

```
new = (old + 1)*spacing + padding*2
where
    padding = 0.75, 0.75
    spacing = 5/4, 15/13
```

A few elements had random offsets in the old system. Here is a table which shows
what offsets when migrating:

| Element      | Position   | Size   | Notes                                                                                 |
|--------------|------------|--------|---------------------------------------------------------------------------------------|
| box          | +0.3, +0.1 | 0,-0.4 |                                                                                       |
| button       |            |        | Buttons now support height, so you will need to set h=1 and reposition if h!=1 before |
| size         | -          | -      |                                                                                       |
| container    | -          | -      |                                                                                       |
| list         |            |        |                                                                                       |
| tooltip      |            |        |                                                                                       |
| image        |            |        |                                                                                       |
| item_image   |            |        |                                                                                       |
| image_button |            |        |                                                                                       |
| background   |            |        |                                                                                       |
| field        |            |        |                                                                                       |
| textarea     |            |        |                                                                                       |
| label        | +0.1       |        |                                                                                       |
| vertlabel    |            |        |                                                                                       |
| textlist     |            |        |                                                                                       |
| tabheader    |            |        |                                                                                       |
| dropdown     |            |        |                                                                                       |
| checkbox     |            |        |                                                                                       |
| scrollbar    |            |        |                                                                                       |
| table        |            |        |                                                                                       |

rubenwardy:
Labels are positioned incorrectly. This is 0,0:

image

SmallJoker:
@v-rob 
> In other words, you're saying that if real_coordinates is directly after size, position, anchor, and no_prepend, it will affect the size?

Yes sure there was a catch, I just didn't see it when I wrote this request. If `position` and `anchor` are already scaled correctly, I'd say that's fine. Otherwise the tag would miss its purpose: to fix the positioning and scaling of all elements.

rubenwardy:
> If position and anchor are already scaled correctly, I'd say that's fine. Otherwise the tag would miss its purpose: to fix the positioning and scaling of all elements.

Position and anchor relate to where the formspec window is on the screen, not to the positioning of elements

v-rob:
Yes, `anchor` and `position` are not affected by `real_coordinates` and don't need to be.

@rubenwardy Labels are fine; this is intended behavior.  The text is meant to be centered in a single coordinate.  If you place an element behind the label, it's easier to see what I mean (both the box and the label are at 0,0):

label

rubenwardy:
This is inconsistent behaviour. As a modder, I expect the label to be positioned at the position I specify. Labels don't have sizes, and sizes used are internal implementation details

Maybe later, support for labels with sizes and specified alignment can be added

v-rob:
I think they should be left as-is.  Labels are most commonly used next to or above elements.  As it's impossible to know how tall text is since it's a client option, it is impossible to center text next to an element or to be certain that the text doesn't conflict with things below it.

Here's an example:

align

The first one is easy; just place them at the same coordinate height.  For the half-height one, since labels don't have changeable height, just place it `0.25` higher than the button next to it is.  If it's properly documented, it won't feel inconsistent.

Now take the last one.  Since text height can be changed, you'll have to guess where to place the text, and then it will only look exactly right with the font, font size, formspec scaling, and resolution on your computer.  This principle is also affected when figuring out how far above a different element you have to place your text.

rubenwardy:
> * The label formspec element displays the text set in `label`
  at the specified position.
> * The text is displayed directly without automatic line breaking,
so label should not be used for big text chunks.

The documentation doesn't mention any sizes, and heavily implies that the text will be rendered at the position that you set.

Also, the text isn't center aligned - it's just offset:

image

rubenwardy:
Another thing: the offset assumes a particular element size, which is incorrect. Better to let modders position them

v-rob:
> The documentation doesn't mention any sizes, and heavily implies that the text will be rendered at the position that you set.

Yeah, it hadn't been updated yet.

> You could just move the origin of the text to the left-middle of the text, then to align with an element you'd use a .25 position like:
> 
> ```
> button[2,0;3,0.5;a;Hello]
> label[0,0.25;Label]
> ```

That's a much better idea than what I currently have.  I wasn't perfectly happy with my current system, but I certainly didn't want it aligned to the top.  I'll do that immediately (as well as for `vertlabel` which does the same thing).

rubenwardy:
> You could just move the origin of the text to the left-middle of the text, then to > align with an element you'd use a .25 position like:
>
> ```
> button[2,0;3,0.5;a;Hello]
> label[0,0.25;Label]
> ```

The left is the behaviour of the PR before, and the right us what I meant in the quote:

image

v-rob:
> The left is the behaviour of the PR before, and the right is what I meant in the quote. The idea is that you get easy placement and alignment whilst maintaining font size compatibility
> 
> image

Did you notice the two commits I did yesterday?  That's exactly what I did ;-)

rubenwardy:
Oh, I must have misread. I thought you center aligned everything. Will check it out

v-rob:
I center aligned `label` as well as `vertlabel` (for the X coordinate instead of Y, of course) and `checkbox` because they all had the sizeless problem.  It's also documented in `lua_api.txt`.

rubenwardy:
Ok cool, that works as I expected. At this point in time, I have no further issues with this PR. However, I'd like to keep converting my game over to real coordinates as a way of testing

rubenwardy:
:+1: when appropriate documentation is added. See https://github.com/minetest/minetest/pull/8524#issuecomment-501783669

v-rob:
I had a thought recently.  This PR adds a unified system, but there are other potential systems people might want to have.  For instance, the HUD uses coordinates from `0` to `1`, which represent placement on the screen.  So, what if, instead of `real_coordinates[<bool>]`, this PR did `coordinates[<type>]`?  These would look something like these:
```
coordinates[old]      -- Current system
coordinates[real]     -- This PR
coordinates[relative] -- HUD-ish
```
This could make the coordinate system extensible.  It would be trivial to modify this PR to do this, and new systems could be added without major hacky changes like this does.  Obviously, I wouldn't add the HUD-ish system in this PR, but it could easily be done later.

DS-Minetest:
For such other types, `imgsize` could be changed, I guess. (This might however maybe make problems with lists and co. as long as you can't set its sizes.)
(Another coordinates type that would make sense might be to make it relative to font size.)

v-rob:
On second thought, that idea wouldn't be that useful.  Although it might be nice to have e.g. a font height system, you would probably want to combine these, like perhaps `font-height` for the height, `relative` for the X and Y coordinates and `real` for the width, which would be quite impossible in any case with the current formspecs.

So, I'll leave this as-is.  This is something for the new formspec system to solve.

v-rob:
I've rebased (of course the 9-slice backgrounds had to be merged first and mess everything up!) and added the `Migrating to Real Coordinates`.  You can remove the `Documentation needed` label; I was just getting to it! :-)

Anyhow, I don't see anything left to do besides reviews before this is merged.  Thanks for all the work and help!

doc/lua_api.txt
```diff
@@ -2284,6 +2363,44 @@ reserved to pass key press events to formspec!
 
 
 
+Migrating to Real Coordinates
+=============================
```
pauloue: Make this be a 2nd level heading?

rubenwardy: oops. Yeah, this should be a heading below the formspec one

doc/lua_api.txt
```diff
@@ -2281,6 +2360,41 @@ Elements
...
+```
+
+A few elements had random offsets in the old system. Here is a table which shows these
+offsets when migrating:
+
+| Element |  Position  |  Size   | Notes
+|---------|------------|---------|-------
+| box     | +0.3, +0.1 | 0, -0.4 |
+| button  |            |         | Buttons now support height, so you will need to set h=1 and reposition if h!=1 before
+| list    |            |         | Note: spacing is now 0.25 for both directions, meaning lists will be taller in height
```
DS-Minetest: - `Note:` in `Notes` is redundant.
- The lines are a bit long (also in the other lines).

Maybe do something like this?
```suggestion
| list    |            |         | spacing is now 0.25 for both directions, meaning lists
|         |            |         | will be taller in height
```

v-rob: You can't place it on two lines like that because it will add another row to the table, like so:

```
| Element |  Position  |  Size   | Notes
|---------|------------|---------|-------
| list    |            |         | spacing is now 0.25 for both directions, meaning
|         |            |         | lists will be taller in height
```

| Element |  Position  |  Size   | Notes
|---------|------------|---------|-------
| list    |            |         | spacing is now 0.25 for both directions, meaning
|         |            |         | lists will be taller in height

Even if you add a plain newline without pipes, it still doesn't work:

```
| Element |  Position  |  Size   | Notes
|---------|------------|---------|-------
| list    |            |         | spacing is now 0.25 for both directions, meaning
lists will be taller in height
| bla     |            |         |
```
| Element |  Position  |  Size   | Notes
|---------|------------|---------|-------
| list    |            |         | spacing is now 0.25 for both directions, meaning
lists will be taller in height
| bla     |            |         |

Markdown really didn't plan for this.  I don't think we should change this to have improper markdown.

Although yes, `Note:` is redundant.  I just copied and pasted what Rubenwardy wrote (with a single correction).  I'll fix that.

DS-Minetest: Oh, I forgot that `lua_api.txt` is written in a way that it could be a markdown file.
You are right.

doc/lua_api.txt
```diff
@@ -2281,6 +2360,41 @@ Elements
...
+
+A few elements had random offsets in the old system. Here is a table which shows these
+offsets when migrating:
+
+| Element |  Position  |  Size   | Notes
+|---------|------------|---------|-------
+| box     | +0.3, +0.1 | 0, -0.4 |
+| button  |            |         | Buttons now support height, so you will need to set h=1 and reposition if h!=1 before
+| list    |            |         | Spacing is now 0.25 for both directions, meaning lists will be taller in height
+| label   |            |  +0.3   | The first line of text is now positioned centered exactly at the position specified
```
rubenwardy: `| 0, +0.3    |         |`

doc/lua_api.txt
```diff
@@ -2281,6 +2360,41 @@ Elements
...
+size = (oldsize-1)*spacing + padding*2 + 1
+```
+
+A few elements had random offsets in the old system. Here is a table which shows these
+offsets when migrating:
+
+| Element |  Position  |  Size   | Notes
+|---------|------------|---------|-------
+| box     | +0.3, +0.1 | 0, -0.4 |
+| button  |            |         | Buttons now support height, so you will need to set h=1 and reposition if h!=1 before
```
rubenwardy: Buttons now support height, so you'll need to set h=0.5, and also reposition if h!=1 before

SmallJoker: To keep the line length limit: How about putting a note `*²` there and `2) Buttons now support height ...` below the table?

DS-Minetest: Shouldn't it be like this?
> Buttons now support height, so you'll need to set `h = 2 * 15/13 * 0.35`, and also reposition if `h ~= 15/13 * 0.35` before

(Actually repositioning has to be done anyway.)

v-rob: I'm not sure what I'm supposed to put here:  DS-Minetest's or Rubenwardy's?  I think DS-Minetest's is the accurate one.

rubenwardy: DS-Minetest's is better.

However, I do think that 0.5 looks better than 0.8076923076923076 as a button height :D

paramat:
Is anything delaying merge?

v-rob:
Nothing.

PR #8527, opened 2019-05-12
add a placenode command

Niwla23:
A /placenode command for setting nodes by command. Code from spawnentity used

### DIFF
--- a/builtin/game/chatcommands.lua
+++ b/builtin/game/chatcommands.lua
@@ -704,7 +704,7 @@ core.register_chatcommand("placenode", {
 				return false, "Invalid parameters ('" .. param .. "')"
 			end
 		end
-		core.add_node(p, {name=nodename})
+		core.add_node(p, { name = nodename })
 		return true, ("%q placed."):format(nodename)
 	end,
 })
--- a/builtin/game/chatcommands.lua
+++ b/builtin/game/chatcommands.lua
@@ -677,6 +677,38 @@ core.register_chatcommand("spawnentity", {
 	end,
 })
 
+core.register_chatcommand("placenode", {
+	params = "<NodeName> [<X>,<Y>,<Z>]",
+	description = "Places node at given (or your) position",
+	privs = {give=true, interact=true},
+	func = function(name, param)
+		local nodename, p = string.match(param, "^([^ ]+) *(.*)$")
+		if not nodename then
+			return false, "NodeName required"
+		end
+		core.log("action", ("%s invokes /placenode, nodename=%q")
+				:format(name, nodename))
+		local player = core.get_player_by_name(name)
+		if player == nil then
+			core.log("error", "Unable to spawn node, player is nil")
+			return false, "Unable to spawn node, player is nil"
+		end
+		if not core.registered_nodes[nodename] then
+			return false, "Cannot place an unknown node"
+		end
+		if p == "" then
+			p = player:get_pos()
+		else
+			p = core.string_to_pos(p)
+			if p == nil then
+				return false, "Invalid parameters ('" .. param .. "')"
+			end
+		end
+		core.add_node(p, {name=nodename})
+		return true, ("%q placed."):format(nodename)
+	end,
+})
+
 core.register_chatcommand("pulverize", {
 	params = "",
 	description = "Destroy item in hand",
--- a/builtin/game/chatcommands.lua
+++ b/builtin/game/chatcommands.lua
@@ -706,7 +706,7 @@ core.register_chatcommand("placenode", {
 		end
 		core.add_node(p, { name = nodename })
 		return true, ("%q placed at %s"):format(nodename, core.pos_to_string(p))
-	end,
+	end
 })
 
 core.register_chatcommand("pulverize", {
--- a/builtin/game/chatcommands.lua
+++ b/builtin/game/chatcommands.lua
@@ -690,8 +690,8 @@ core.register_chatcommand("setnode", {
 				:format(name, nodename))
 		local player = core.get_player_by_name(name)
 		if not player then
-			core.log("error", "Unable to spawn node, player is nil")
-			return false, "Unable to spawn node, no player"
+			core.log("error", "Unable to place node, player is nil")
+			return false, "Unable to set node, no player"
 		end
 		if not core.registered_nodes[nodename] then
 			return false, "Cannot place an unknown node"
--- a/builtin/game/chatcommands.lua
+++ b/builtin/game/chatcommands.lua
@@ -691,7 +691,7 @@ core.register_chatcommand("placenode", {
 		local player = core.get_player_by_name(name)
 		if not player then
 			core.log("error", "Unable to spawn node, player is nil")
-			return false, "Unable to spawn node, player is nil"
+			return false, "Unable to spawn node, no player"
 		end
 		if not core.registered_nodes[nodename] then
 			return false, "Cannot place an unknown node"
--- a/builtin/game/chatcommands.lua
+++ b/builtin/game/chatcommands.lua
@@ -689,7 +689,7 @@ core.register_chatcommand("placenode", {
 		core.log("action", ("%s invokes /placenode, nodename=%q")
 				:format(name, nodename))
 		local player = core.get_player_by_name(name)
-		if player == nil then
+		if not player then
 			core.log("error", "Unable to spawn node, player is nil")
 			return false, "Unable to spawn node, player is nil"
 		end
--- a/builtin/game/chatcommands.lua
+++ b/builtin/game/chatcommands.lua
@@ -686,7 +686,7 @@ core.register_chatcommand("setnode", {
 		if not nodename then
 			return false, "Node name required"
 		end
-		core.log("action", ("%s invokes /placenode, nodename=%q")
+		core.log("action", ("%s invokes /setnode, nodename=%q")
 				:format(name, nodename))
 		local player = core.get_player_by_name(name)
 		if not player then
@@ -705,7 +705,7 @@ core.register_chatcommand("setnode", {
 			end
 		end
 		core.add_node(pos, { name = nodename })
-		return true, ("%q placed at %s"):format(nodename, core.pos_to_string(pos))
+		return true, ("%q setted at %s"):format(nodename, core.pos_to_string(pos))
 	end
 })
 
--- a/builtin/game/chatcommands.lua
+++ b/builtin/game/chatcommands.lua
@@ -680,7 +680,7 @@ core.register_chatcommand("spawnentity", {
 core.register_chatcommand("placenode", {
 	params = "<NodeName> [<X>,<Y>,<Z>]",
 	description = "Places node at given (or your) position",
-	privs = { give = true, interact = true, protection_bypass, server },
+	privs = {interact = true, server = true},
 	func = function(name, param)
 		local nodename, p = string.match(param, "^([^ ]+) *(.*)$")
 		if not nodename then
--- a/builtin/game/chatcommands.lua
+++ b/builtin/game/chatcommands.lua
@@ -700,7 +700,7 @@ core.register_chatcommand("placenode", {
 			p = player:get_pos()
 		else
 			p = core.string_to_pos(p)
-			if p == nil then
+			if not p then
 				return false, "Invalid parameters ('" .. param .. "')"
 			end
 		end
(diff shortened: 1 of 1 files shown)


builtin/game/chatcommands.lua
```diff
@@ -677,6 +677,38 @@ core.register_chatcommand("spawnentity", {
...
+		end
+		if p == "" then
+			p = player:get_pos()
+		else
+			p = core.string_to_pos(p)
+			if p == nil then
+				return false, "Invalid parameters ('" .. param .. "')"
+			end
+		end
+		core.add_node(p, {name=nodename})
```
DS-Minetest: Check for protection with `minetest.is_protected(pos, name)`.

builtin/game/chatcommands.lua
```diff
@@ -677,6 +677,38 @@ core.register_chatcommand("spawnentity", {
 	end,
 })
 
+core.register_chatcommand("placenode", {
+	params = "<NodeName> [<X>,<Y>,<Z>]",
+	description = "Places node at given (or your) position",
+	privs = {give=true, interact=true},
+	func = function(name, param)
+		local nodename, p = string.match(param, "^([^ ]+) *(.*)$")
```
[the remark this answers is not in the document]
ClobberXD: I suggest renaming `p` to `pos` for better readability.

builtin/game/chatcommands.lua
ClobberXD: ... only if player does not have `protection_bypass` priv. :)

builtin/game/chatcommands.lua
```diff
@@ -677,6 +677,38 @@ core.register_chatcommand("spawnentity", {
 	end,
 })
 
+core.register_chatcommand("placenode", {
+	params = "<NodeName> [<X>,<Y>,<Z>]",
+	description = "Places node at given (or your) position",
+	privs = {give=true, interact=true},
```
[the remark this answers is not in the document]
DS-Minetest: Note that all the other chatcommands in the file have the same style.

Also https://dev.minetest.net/Lua_code_style_guidelines#Spaces says:
>Spaces are not used around parenthesis, brackets, or curly braces. 

=>
```diff
-	privs = { give = true, interact = true },
+	privs = {give = true, interact = true},
```

paramat:
I suspect this neds extra checks to not replace nodes that would otherwise be unremovable, like npdes with no dig groups, unlocked chests containing items, maybe more ...

builtin/game/chatcommands.lua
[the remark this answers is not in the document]
paramat: 'add node' is deprecated, use 'set node' instead.

paramat:
Please can you explain why this is needed or why this will be very useful?

Some more thoughts:

The 'give' priv allows a player to give an item to a player, but that is very different from being able to place that item. Many more checks will be needed, and/or, the privs may need to be 'server'.

For a single placed node this will be roughly the same effort as typing 'giveme xxx' and placing the node. For multiple identical nodes this will be far more effort as every placed node needs a command, so i'm not sure this is worth adding.

This command doesn't allow choosing the rotation, many nodes are placed with a rotation set by player view direction. A parameter for setting param2 may be needed.

builtin/game/chatcommands.lua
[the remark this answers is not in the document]
paramat: In fact, this may need to be 'place node' to mimick a player placing.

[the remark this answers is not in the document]
ClobberXD: Agreed with both the points above.

builtin/game/chatcommands.lua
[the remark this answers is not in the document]
Niwla23: You can do that, but i just copied it from spawnentity so its exactly the same style but i also like pos more than p

builtin/game/chatcommands.lua
```diff
@@ -677,6 +677,38 @@ core.register_chatcommand("spawnentity", {
...
+	func = function(name, param)
+		local nodename, p = string.match(param, "^([^ ]+) *(.*)$")
+		if not nodename then
+			return false, "NodeName required"
```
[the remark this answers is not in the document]
Niwla23: Also, thats like in Spawn Entity its called EntityName there

builtin/game/chatcommands.lua
```diff
@@ -677,6 +677,38 @@ core.register_chatcommand("spawnentity", {
...
+		end
+		core.log("action", ("%s invokes /placenode, nodename=%q")
+				:format(name, nodename))
```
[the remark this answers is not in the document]
Niwla23: Because its not logged in spawnentity but great idea

Niwla23:
> 
> 
> Please can you explain why this is needed or why this will be very useful?
> 
For example when you are at 100 nodes and the server has no WE just type it in and you have a block! Or when you want to use a commandblock that builds something, it could be good.

Niwla23:
And why is should it be something worth? its worth is a mouseclick. Wheres the problem in adding a command?

Niwla23:
added server and protection_bypass as requiered privs

builtin/game/chatcommands.lua
```diff
@@ -677,6 +677,38 @@ core.register_chatcommand("spawnentity", {
 	end,
 })
 
+core.register_chatcommand("placenode", {
+	params = "<NodeName> [<X>,<Y>,<Z>]",
+	description = "Places node at given (or your) position",
+	privs = { give = true, interact = true, protection_bypass, server },
```
ClobberXD:
```suggestion
	privs = {interact = true, server = true},
```

- You missed `= true`.
- Checking if the player has `protection_bypass` should be done within `func` and it shouldn't be a required priv (not that it really matters - the caller would have the `server` priv anyway).

builtin/game/chatcommands.lua
```diff
@@ -677,6 +677,38 @@ core.register_chatcommand("spawnentity", {
...
+	func = function(name, param)
+		local nodename, p = string.match(param, "^([^ ]+) *(.*)$")
+		if not nodename then
+			return false, "Node name required"
+		end
+		core.log("action", ("%s invokes /placenode, nodename=%q")
+				:format(name, nodename))
+		local player = core.get_player_by_name(name)
+		if player == nil then
```
ClobberXD:
```suggestion
		if not player then
```

Wuzzy2:
Nice. But given that you call `core.add_node`, shouldn't the command be called `/addnode` or `/setnode` instead?

`/placenode` implies the node is placed with the same effects as if the player would have placed it with rightclick. Note that placing a node and setting a node have different implications in the Lua API as those trigger different callbacks.

paramat:
I'm neutral, i can't judge the usefulness, input from server admin would be useful.
Placing a node in mid-air is one use.

I agree with Wuzzy2, the command should be called /addnode or /setnode,
or /placenode (if minetest.place_node() is used).
I'm not actually sure if one of add_node and set_node is deprecated, neither is documented as deprecated.

Since this is 'server' priv, i guess 'add_node()' or 'set_node()' may be better than using 'place_node()'?

paramat:
```
int ModApiEnvMod::l_add_node(lua_State *L)
{
	return l_set_node(L);
}
```
I think `minetest.set_node()` should be used, as it is now the common practice in mods, and because 'add node' redirects to 'set node'.
If so the command should be called '/setnode'.

ClobberXD:
Being able to "place" nodes this way is much more useful than just "setting" the nodes there, IMHO, for a multitude of use-cases, incl. 
- Place nodes mid-air without having to build a tall tower to point at
- Place nodes a little farther away from the player, to mark a distance of n nodes or smth., for example.
- Mesecons command-blocks.

I think anything that can be done by a player should be doable via commands or at least the API, which would allow for fake players, bots, etc.

DS-Minetest:
Placing nodes in air might cause problems for some nodes because of `pointed_thing.type = "nothing"`.

Just setting the node is good imo.

A client-side chatcommand `placenode` might be worth consideration when/if this will be possible.

Niwla23:
What should be changed now?

Niwla23:
Ok, i changed that

benrob0329:
Shouldn't commands like this be defined in the game, not the engine? Not every game needs WE-type commands, and might implement different checks.

builtin/game/chatcommands.lua
```diff
@@ -677,6 +677,38 @@ core.register_chatcommand("spawnentity", {
 	end,
 })
 
+core.register_chatcommand("setnode", {
+	params = "<NodeName> [<X>,<Y>,<Z>]",
+	description = "Places node at given (or your) position",
+	privs = {interact = true, server = true},
```
paramat: If 'server' is a required priv, is 'interact' redundant and unnecessary?

builtin/game/chatcommands.lua
```diff
@@ -677,6 +677,38 @@ core.register_chatcommand("spawnentity", {
...
+		end
+		if pos == "" then
+			pos = player:get_pos()
+		else
+			pos = core.string_to_pos(pos)
+			if not pos then
+				return false, "Invalid parameters ('" .. param .. "')"
+			end
+		end
+		core.add_node(pos, { name = nodename })
```
paramat: set_node

builtin/game/chatcommands.lua
```diff
@@ -677,6 +677,38 @@ core.register_chatcommand("spawnentity", {
...
+core.register_chatcommand("setnode", {
+	params = "<NodeName> [<X>,<Y>,<Z>]",
+	description = "Places node at given (or your) position",
+	privs = {interact = true, server = true},
+	func = function(name, param)
+		local nodename, pos = string.match(param, "^([^ ]+) *(.*)$")
+		if not nodename then
+			return false, "Node name required"
+		end
+		core.log("action", ("%s invokes /placenode, nodename=%q")
```
paramat: 'placenode' doesn't match the name of the command.

builtin/game/chatcommands.lua
```diff
@@ -677,6 +677,38 @@ core.register_chatcommand("spawnentity", {
...
+				:format(name, nodename))
+		local player = core.get_player_by_name(name)
+		if not player then
+			core.log("error", "Unable to spawn node, player is nil")
+			return false, "Unable to spawn node, no player"
```
paramat: 'spawn' is the wrong word, 'spawn' implies an entity.
Should be 'place' to match the rest of the code.

builtin/game/chatcommands.lua
paramat: If 'interact' is enough to use this command, i think it is wrong that the average player can place nodes absolutely anywhere on the map. Only being able to place nodes where a player can reach is an important safeguard.
I think only those with 'server' priv should be able to use this command due to its power.

Niwla23:
fixed @paramat

builtin/game/chatcommands.lua
ClobberXD: Privs are always AND. Chat commands use `core.check_player_privs`, which as you might already know, returns a boolean and a list of missing privs, and the first value is true only if all the required privs are satisfied.

paramat: Thanks, 'interact' seems unnecessary then.

builtin/game/chatcommands.lua
paramat: ^

Niwla23:
wahts a rebase? @SmallJoker

DS-Minetest:
@Niwla23 https://forum.minetest.net/viewtopic.php?f=3&t=14262&p=211498#p211502

builtin/game/chatcommands.lua
```diff
@@ -677,6 +677,38 @@ core.register_chatcommand("spawnentity", {
...
+		local nodename, pos = string.match(param, "^([^ ]+) *(.*)$")
+		if not nodename then
+			return false, "Node name required"
+		end
+		core.log("action", ("%s invokes /setnode, nodename=%q")
+				:format(name, nodename))
+		local player = core.get_player_by_name(name)
+		if not player then
+			core.log("error", "Unable to place node, player is nil")
+			return false, "Unable to set node, no player"
```
paramat: `Unable to place node`
For consistency.

paramat:
Author cannot rebase so needs adoption, if this feature is desired.
I think this feature is low priority.

rubenwardy:
Is this supported by anyone? The feature seems to be of minimal value. The only one I can think of is having an initial node in a singlenode mapgen, but the UX of that is terrible compared to Sky Lands or [The Origin](https://content.minetest.net/packages/Wuzzy/origin/)

The implementation for this should be very simple, and isn't worth the amount of review that has gone into it

paramat:
I agree with closing.

PR #8530, opened 2019-05-14
Add scrollbaroptions FormSpec element

### LINKED ISSUE #8575: Formspec: add min, max and pagesize parameters to scrollbar element
##### Issue type
 - Feature request
 
##### Minetest version
```
Using Irrlicht 1.8.4
BUILD_TYPE=Release
RUN_IN_PLACE=1
USE_GETTEXT=1
USE_SOUND=1
USE_CURL=1
USE_FREETYPE=1
USE_LUAJIT=0
STATIC_SHAREDIR="."
```

##### Summary
Currently the `scrollbar` formspec element sill uses the `gui::IGUIScrollBar` class and the `min` and `max` values are set to `0` and `1000`:
https://github.com/minetest/minetest/blob/7379aa74cf98c7e4c7aa5325ef1531d412a0abac/src/gui/guiFormSpecMenu.cpp#L516-L520
https://github.com/minetest/minetest/blob/7379aa74cf98c7e4c7aa5325ef1531d412a0abac/doc/lua_api.txt#L2196-L2206

I request that the scrollbar element uses the new `guiScrollBar` class (from #8507) and that you can set the `min`, the `max` and the `pagesize`.

Note that there's #8530 which would already cover `min` and `max`, but not `pagesize`.

v-rob:
This PR adds a `scrollbaroptions` element for formspecs.  Similar to `tableoptions`, it must be placed before the scrollbars it modifies.  Old scrollbars are not affected.  Closes #8575.

Valid options are:

- `min` - Minimum value of the scrollbar (default 0)
- `max` - Maximum value of the scrollbar (default 1000)
- `smallstep` - Scroll amount when using arrows or scroll wheel (default 10)
- `largestep` - Scroll amount when clicking the bar itself (default 100)
- `thumbsize` - Size of the thumb determined by the number of unit the thumb spans out of the range of values (max - min).
- `arrows` - Whether to make the arrows hidden, shown, or automatic.

Some potential examples:

`scrollbaroptions[min=0;max=255;smallstep=8;largestep=16;thumbsize=8;arrows=hide]`
`scrollbaroptions[max=10000;largestep=1000]`

If the step value is negative, it will use Irrlicht's defaults.  If the max is less than or equal to the min, the scrollbar element will have no bar and do nothing, i.e. disabled.  If thumbsize is zero or less, it defaults to 1 (which usually will appear as a square).

This PR is Ready for Review.

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2368,6 +2368,9 @@ Elements
     * scrolled, defaults to `10`
 * `largestep=<int>`
     * Sets scrollbar step value when the bar itself is clicked, defaults to `100`
+* `pagesize=<int>`
+    * Sets size of a single screen for the scrollbar. Setting this to -1 lets the
+    * scrollbar calculate this automatically.
 
 ### `table[<X>,<Y>;<W>,<H>;<name>;<cell 1>,<cell 2>,...,<cell n>;<selected idx>]`
 
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2362,6 +2362,7 @@ Elements
     * Sets scrollbar minimum value, defaults to `0`
 * `max=<int>`
     * Sets scrollbar maximum value, defaults to `1000`
+	* If the max is equal to the min, the scrollbar will be disabled
 * `smallstep=<int>`
     * Sets scrollbar step value when the arrows are clicked, defaults to `10`
 * `largestep=<int>`
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2359,18 +2359,20 @@ Elements
 ### `scrollbaroptions[opt1;opt2;...]`
 * Sets options for all following `scrollbar[]` elements
 * `min=<int>`
-    * Sets scrollbar minimum value, defaults to `0`
+    * Sets scrollbar minimum value, defaults to `0`.
 * `max=<int>`
-    * Sets scrollbar maximum value, defaults to `1000`
-    * If the max is equal to the min, the scrollbar will be disabled
+    * Sets scrollbar maximum value, defaults to `1000`.
+      If the max is equal to the min, the scrollbar will be disabled.
 * `smallstep=<int>`
     * Sets scrollbar step value when the arrows are clicked or the mouse wheel is
-    * scrolled, defaults to `10`
+      scrolled, defaults to `10`.
 * `largestep=<int>`
-    * Sets scrollbar step value when the bar itself is clicked, defaults to `100`
-* `pagesize=<int>`
-    * Sets size of a single screen for the scrollbar. Setting this to -1 lets the
-    * scrollbar calculate this automatically.
+    * Sets scrollbar step value when the bar itself is clicked, defaults to `100`.
+* `thumbsize=<int>`
+    * Sets size of the thumb on the scrollbar. Size is calculated in the number of
+      units the scrollbar spans out of the range of the scrollbar values.
+      Example: If a scrollbar has a min of 1 and a max of 1000, a thumbsize of 10
+      would span a tenth of the scrollbar space.
 
 ### `table[<X>,<Y>;<W>,<H>;<name>;<cell 1>,<cell 2>,...,<cell n>;<selected idx>]`
 
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2365,14 +2365,17 @@ Elements
       If the max is equal to the min, the scrollbar will be disabled.
 * `smallstep=<int>`
     * Sets scrollbar step value when the arrows are clicked or the mouse wheel is
-      scrolled, defaults to `10`.
+      scrolled.
+    * If this is set to a negative number, the value will be reset to `10`.
 * `largestep=<int>`
-    * Sets scrollbar step value when the bar itself is clicked, defaults to `100`.
+    * Sets scrollbar step value when the bar itself is clicked.
+    * If this is set to a negative number, the value will be reset to `100`.
 * `thumbsize=<int>`
     * Sets size of the thumb on the scrollbar. Size is calculated in the number of
-      units the scrollbar spans out of the range of the scrollbar values.
-      Example: If a scrollbar has a min of 1 and a max of 1000, a thumbsize of 10
+      units the thumb spans out of the range of the scrollbar values.
+    * Example: If a scrollbar has a `min` of 1 and a `max` of 100, a thumbsize of 10
       would span a tenth of the scrollbar space.
+    * If this is set to zero or less, the value will be reset to `1`.
 
 ### `table[<X>,<Y>;<W>,<H>;<name>;<cell 1>,<cell 2>,...,<cell n>;<selected idx>]`
 
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2368,7 +2368,7 @@ Elements
       scrolled.
     * If this is set to a negative number, the value will be reset to `10`.
 * `largestep=<int>`
-    * Sets scrollbar step value when the bar itself is clicked.
+    * Sets scrollbar step value used by page up and page down.
     * If this is set to a negative number, the value will be reset to `100`.
 * `thumbsize=<int>`
     * Sets size of the thumb on the scrollbar. Size is calculated in the number of
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2346,15 +2346,26 @@ Elements
 
 ### `scrollbar[<X>,<Y>;<W>,<H>;<orientation>;<name>;<value>]`
 
-* Show a scrollbar
+* Show a scrollbar using options defined by the previous `scrollbaroptions[]`
 * There are two ways to use it:
     1. handle the changed event (only changed scrollbar is available)
     2. read the value on pressing a button (all scrollbars are available)
 * `orientation`:  `vertical`/`horizontal`
 * Fieldname data is transferred to Lua
-* Value this trackbar is set to (`0`-`1000`)
+* Value of this trackbar is set to (`0`-`1000`) by default
 * See also `minetest.explode_scrollbar_event`
   (main menu: `core.explode_scrollbar_event`).
+
+### `scrollbaroptions[opt1;opt2;...]`
+* Sets options for all following `scrollbar[]` elements
+* `min=<int>`
+    * Sets scrollbar minimum value, defaults to `0`
+* `max=<int>`
+    * Sets scrollbar maximum value, defaults to `1000`
+* `smallstep=<int>`
+    * Sets scrollbar step value when the arrows are clicked, defaults to `10`
+* `largestep=<int>`
+    * Sets scrollbar step value when the bar itself is clicked, defaults to `100`
 
 ### `table[<X>,<Y>;<W>,<H>;<name>;<cell 1>,<cell 2>,...,<cell n>;<selected idx>]`
 
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2362,7 +2362,7 @@ Elements
     * Sets scrollbar minimum value, defaults to `0`
 * `max=<int>`
     * Sets scrollbar maximum value, defaults to `1000`
-	* If the max is equal to the min, the scrollbar will be disabled
+    * If the max is equal to the min, the scrollbar will be disabled
 * `smallstep=<int>`
     * Sets scrollbar step value when the arrows are clicked, defaults to `10`
 * `largestep=<int>`
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2376,6 +2376,9 @@ Elements
     * Example: If a scrollbar has a `min` of 1 and a `max` of 100, a thumbsize of 10
       would span a tenth of the scrollbar space.
     * If this is set to zero or less, the value will be reset to `1`.
+* `arrows=<show/hide/default>`
+    * Whether to show the arrow buttons on the scrollbar. `default` hides the arrows
+      when the scrollbar gets too small, but shows them otherwise.
 
 ### `table[<X>,<Y>;<W>,<H>;<name>;<cell 1>,<cell 2>,...,<cell n>;<selected idx>]`
 
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2364,7 +2364,8 @@ Elements
     * Sets scrollbar maximum value, defaults to `1000`
     * If the max is equal to the min, the scrollbar will be disabled
 * `smallstep=<int>`
-    * Sets scrollbar step value when the arrows are clicked, defaults to `10`
+    * Sets scrollbar step value when the arrows are clicked or the mouse wheel is
+    * scrolled, defaults to `10`
 * `largestep=<int>`
     * Sets scrollbar step value when the bar itself is clicked, defaults to `100`
 
--- a/src/gui/guiFormSpecMenu.cpp
+++ b/src/gui/guiFormSpecMenu.cpp
@@ -2651,7 +2651,6 @@ void GUIFormSpecMenu::parseElement(parserData* data, const std::string &element)
 
 	if (type == "scrollbaroptions") {
 		parseScrollBarOptions(data, description);
-
 		return;
 	}
 
--- a/src/gui/guiFormSpecMenu.cpp
+++ b/src/gui/guiFormSpecMenu.cpp
@@ -623,6 +623,7 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
 
 		auto style = getStyleForElement("scrollbar", name);
 		e->setNotClipped(style.getBool(StyleSpec::NOCLIP, false));
+		e->setArrowsVisible(data->scrollBarOptions.showArrows);
 
 		s32 max = data->scrollBarOptions.max;
 		s32 min = data->scrollBarOptions.min;
@@ -647,7 +648,8 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
 		m_fields.push_back(spec);
 		return;
 	}
-	errorstream<< "Invalid scrollbar element(" << parts.size() << "): '" << element << "'"  << std::endl;
+	errorstream << "Invalid scrollbar element(" << parts.size() << "): '" << element
+		<< "'" << std::endl;
 }
 
 void GUIFormSpecMenu::parseScrollBarOptions(parserData* data, const std::string &element)
@@ -669,26 +671,37 @@ void GUIFormSpecMenu::parseScrollBarOptions(parserData* data, const std::string 
 			continue; // Go to next option
 		}
 
-		int value = stoi(options[1]);
-
 		if (options[0] == "max") {
-			data->scrollBarOptions.max = value;
+			data->scrollBarOptions.max = stoi(options[1]);
 			continue;
 		}
-		if (options[0] == "min") {
-			data->scrollBarOptions.min = value;
+		else if (options[0] == "min") {
+			data->scrollBarOptions.min = stoi(options[1]);
 			continue;
 		}
-		if (options[0] == "smallstep") {
+		else if (options[0] == "smallstep") {
+			int value = stoi(options[1]);
 			data->scrollBarOptions.smallStep = value < 0 ? 10 : value;
 			continue;
 		}
-		if (options[0] == "largestep") {
+		else if (options[0] == "largestep") {
+			int value = stoi(options[1]);
 			data->scrollBarOptions.largeStep = value < 0 ? 100 : value;
 			continue;
 		}
-		if (options[0] == "thumbsize") {
+		else if (options[0] == "thumbsize") {
+			int value = stoi(options[1]);
 			data->scrollBarOptions.thumbSize = value <= 0 ? 1 : value;
+			continue;
+		}
+		else if (options[0] == "arrows") {
+			std::string value = trim(options[1]);
+			if (value == "hide")
+				data->scrollBarOptions.showArrows = 0;
+			else if (value == "show")
+				data->scrollBarOptions.showArrows = 1;
+			else // Auto hide/show
+				data->scrollBarOptions.showArrows = 2;
 			continue;
 		}
 
--- a/src/gui/guiFormSpecMenu.cpp
+++ b/src/gui/guiFormSpecMenu.cpp
@@ -618,17 +618,62 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
 		auto style = getStyleForElement("scrollbar", name);
 		e->setNotClipped(style.getBool(StyleSpec::NOCLIP, false));
 
-		e->setMax(1000);
-		e->setMin(0);
+		e->setMax(data->scrollBarOptions.max);
+		e->setMin(data->scrollBarOptions.min);
+
 		e->setPos(stoi(parts[4]));
-		e->setSmallStep(10);
-		e->setLargeStep(100);
+		e->setSmallStep(data->scrollBarOptions.smallStep);
+		e->setLargeStep(data->scrollBarOptions.largeStep);
 
 		m_scrollbars.emplace_back(spec,e);
 		m_fields.push_back(spec);
 		return;
 	}
 	errorstream<< "Invalid scrollbar element(" << parts.size() << "): '" << element << "'"  << std::endl;
+}
+
+void GUIFormSpecMenu::parseScrollBarOptions(parserData* data, const std::string &element)
+{
+	std::vector<std::string> parts = split(element, ';');
+
+	if (parts.size() > 0) {
+		for (unsigned int i = 0; i < parts.size(); i++) {
+
+			std::vector<std::string> options = split(parts[i], '=');
+
+			if (options.size() != 2) {
+				errorstream << "Invalid scrollbaroptions option syntax: '" <<
+					element << "'" << std::endl;
+				continue; // Go to next option
+			}
+
+			int value = stoi(options[1]);
+
+			if (options[0] == "max") {
+				data->scrollBarOptions.max = value;
+				continue;
+			}
+			if (options[0] == "min") {
+				data->scrollBarOptions.min = value;
+				continue;
+			}
+			if (options[0] == "smallstep") {
+				data->scrollBarOptions.smallStep = value;
+				continue;
+			}
+			if (options[0] == "largestep") {
+				data->scrollBarOptions.largeStep = value;
+				continue;
+			}
+
+			errorstream << "Invalid scrollbaroptions option(" << options[0] <<
+				"): '" << element << "'" << std::endl;
+		}
+		return;
+	}
+
+	errorstream << "Invalid scrollbaroptions element(" << parts.size() << "): '" <<
+		element << "'"  << std::endl;
 }
 
 void GUIFormSpecMenu::parseImage(parserData* data, const std::string &element)
@@ -2581,6 +2626,12 @@ void GUIFormSpecMenu::parseElement(parserData* data, const std::string &element)
 
 	if (type == "style_type") {
 		parseStyle(data, description, true);
+		return;
+	}
+
+	if (type == "scrollbaroptions") {
+		parseScrollBarOptions(data, description);
+
 		return;
 	}
 
--- a/src/gui/guiFormSpecMenu.cpp
+++ b/src/gui/guiFormSpecMenu.cpp
@@ -35,6 +35,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include <IGUIStaticText.h>
 #include <IGUIFont.h>
 #include <IGUITabControl.h>
+#include "guiScrollBar.h"
 #include "client/renderingengine.h"
 #include "log.h"
 #include "client/tile.h" // ITextureSource
@@ -618,22 +619,24 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
 		auto style = getStyleForElement("scrollbar", name);
 		e->setNotClipped(style.getBool(StyleSpec::NOCLIP, false));
 
-		e->setMax(data->scrollBarOptions.max);
-		e->setMin(data->scrollBarOptions.min);
+		s32 max = data->scrollBarOptions.max;
+		s32 min = data->scrollBarOptions.min;
+
+		e->setMax(max);
+		e->setMin(min);
 
 		e->setPos(stoi(parts[4]));
+
 		e->setSmallStep(data->scrollBarOptions.smallStep);
 		e->setLargeStep(data->scrollBarOptions.largeStep);
 
-		if (data->scrollBarOptions.pageSize < 0) {
-			if (is_horizontal) {
-				e->setPageSize(dim.X);
-			} else {
-				e->setPageSize(dim.Y);
-			}
-		} else {
-			e->setPageSize(data->scrollBarOptions.pageSize);
-		}
+		s32 scrollbar_size;
+		if (is_horizontal)
+			scrollbar_size = dim.X;
+		else
+			scrollbar_size = dim.Y;
+
+		e->setPageSize(scrollbar_size * (max-min+1) / data->scrollBarOptions.thumbSize);
 
 		m_scrollbars.emplace_back(spec,e);
 		m_fields.push_back(spec);
@@ -652,7 +655,7 @@ void GUIFormSpecMenu::parseScrollBarOptions(parserData* data, const std::string 
 			std::vector<std::string> options = split(parts[i], '=');
 
 			if (options.size() != 2) {
-				errorstream << "Invalid scrollbaroptions option syntax: '" <<
+				warningstream << "Invalid scrollbaroptions option syntax: '" <<
 					element << "'" << std::endl;
 				continue; // Go to next option
 			}
@@ -668,33 +671,34 @@ void GUIFormSpecMenu::parseScrollBarOptions(parserData* data, const std::string 
 				continue;
 			}
 			if (options[0] == "smallstep") {
-				if (value < 0) {
+				if (value < 0)
 					data->scrollBarOptions.smallStep = 10;
-				} else {
+				else
 					data->scrollBarOptions.smallStep = value;
-				}
 				continue;
 			}
 			if (options[0] == "largestep") {
-				if (value < 0) {
+				if (value < 0)
 					data->scrollBarOptions.largeStep = 100;
-				} else {
+				else
 					data->scrollBarOptions.largeStep = value;
-				}
 				continue;
 			}
-			if (options[0] == "pagesize") {
-				data->scrollBarOptions.pageSize = value;
+			if (options[0] == "thumbsize") {
+				if (value < 0)
+					data->scrollBarOptions.thumbSize = 1;
+				else
+					data->scrollBarOptions.thumbSize = value;
 				continue;
 			}
 
-			errorstream << "Invalid scrollbaroptions option(" << options[0] <<
+			warningstream << "Invalid scrollbaroptions option(" << options[0] <<
 				"): '" << element << "'" << std::endl;
 		}
 		return;
 	}
 
-	errorstream << "Invalid scrollbaroptions element(" << parts.size() << "): '" <<
+	warningstream << "Invalid scrollbaroptions element(" << parts.size() << "): '" <<
 		element << "'"  << std::endl;
 }
 
--- a/src/gui/guiFormSpecMenu.cpp
+++ b/src/gui/guiFormSpecMenu.cpp
@@ -625,6 +625,16 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
 		e->setSmallStep(data->scrollBarOptions.smallStep);
 		e->setLargeStep(data->scrollBarOptions.largeStep);
 
+		if (scrollBarOptions.pageSize < 0) {
+			if (is_horizontal) {
+				e->setPageSize(dim.X);
+			} else {
+				e->setPageSize(dim.Y);
+			}
+		} else {
+			e->setPageSize(data->scrollBarOptions.pageSize);
+		}
+
 		m_scrollbars.emplace_back(spec,e);
 		m_fields.push_back(spec);
 		return;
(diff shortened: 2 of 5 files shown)


DS-Minetest:
Imo it would be good if #8507 is merged before this, so that here the option sliderlength can be added.

v-rob:
I'm trying to incorporate the new `guiScrollBar` class in this PR, and I can't figure out what's going incorrectly.  The [error code](https://travis-ci.org/minetest/minetest/jobs/542035206#L1154) is rather undecipherable, and I can't figure anything out from it.  I think the error lies in line [516](https://github.com/minetest/minetest/pull/8530/files#diff-074850a88796e13dad12dd7ae1a8d9f7R516), but I'm not certain.  I based this `guiScrollBar` on the one in `guiEditBoxWithScrollbar.cpp`

src/gui/guiFormSpecMenu.cpp
```diff
@@ -513,14 +513,23 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
...
 		e->setMin(data->scrollBarOptions.min);
 		e->setPos(stoi(parts[4]));
 		e->setSmallStep(data->scrollBarOptions.smallStep);
 		e->setLargeStep(data->scrollBarOptions.largeStep);
+		
+		if (scrollBarOptions.pageSize < 0) {
+			if (is_horizontal) {
+				e->setPageSize(dim.X);
+			} else {
+				e->setPageSize(dim.Y);
```
DS-Minetest: Shouldn't `dim.X` and `dim.Y` be swapped?

DS-Minetest:
Have you rebased already?

src/gui/guiFormSpecMenu.cpp
```diff
@@ -545,25 +554,38 @@ void GUIFormSpecMenu::parseScrollBarOptions(parserData* data, const std::string
...
 			}
-			if (options[0] == "largestep") {
-				data->scrollBarOptions.largeStep = value;
+			if (option == "pagesize") {
+				data->scrollBarOptions.pageSize = value;
 				continue;
 			}
 
-			errorstream << "Invalid scrollbaroptions option(" << options[0] <<
+			errorstream << "Invalid scrollbaroptions option(" << option <<
```
SmallJoker: For future releases: Please change this to `warningstream` or `infostream`.

src/gui/guiFormSpecMenu.cpp
v-rob: No, since if the scrollbar is horizontal, the page size should be the width, X.  Otherwise, it should be the height, Y.

v-rob:
I have rebased as evidenced by this file: https://github.com/v-rob/minetest/blob/scrollbaroptions/src/gui/guiScrollBar.cpp.  I believe the problem lies in the instantiation of `guiScrollBar`, but I don't have much to work off of.  Maybe @stujones11 can shed some light on this, since he made it.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Oh, soory, I misunderstood pagesize.

DS-Minetest:
@v-rob `"guiScrollBar.h"` is not yet included.

rubenwardy:
This is really cool!

v-rob:
Well, it works now (almost).  I just needed to include `guiScrollBar.h` (duh!) as well as change a few `IGUIScrollBar` things to the new `guiScrollBar`.  Now I just need to make the bar change size properly -- see the screenshot :)

large_scrollbar

v-rob:
I still have two bugs to iron out before this PR can be merged:  First, I _still_ don't think that I have `pagesize` quite right.  The other bug is that trying to get the value from the scrollbar always returns `INV` (which is failure).  I need to do some more testing.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -513,14 +514,30 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
 
 		spec.ftype = f_ScrollBar;
 		spec.send  = true;
-		gui::IGUIScrollBar* e =
-				Environment->addScrollBar(is_horizontal,rect,this,spec.fid);
+		guiScrollBar* e = new guiScrollBar(Environment, this, spec.fid, rect,
+				is_horizontal, true);
+
+		int max = data->scrollBarOptions.max;
```
DS-Minetest: As far as I have seen, minetest seems to use `s32`, not `int`.

DS-Minetest:
If you want the default quadratic slider rectangle, you have to set page_size as large as possible, this results in a thumb_size of (nearly) 0, which is smaller than thumb_min and will ergo be set to thumb_min.

If the length of the scrollbar (including borders) is the length of the visible part of the whole page that you scroll over, page_size is the length of the whole page.
(All this lengths are in pixels.)

The length of the scrollbar over page_size is kept equal to thumb_size over thumb_area.

Imo, this is very confusing and not really useful for modders.
There should be something else that you can set.

Maybe just give the thumb size in scrollbar units.

Example 1: The whole page is 1000 scrollbar units long (max-min). (In pixels it's `1000*some_factor`.) The visible part is `10*same_factor_as_before` pixels long. => The thumb has to be 10 long (in scrollbar units).

Example 2: You can choose one number out of 10 with a scrollbar (from 0 to 9). Ergo min is 0 and max is 9. And thumb_size in scrollbar units is 1 because it's always one number at a time.

Now, the question is: How to convert thumb_size in scrollbar units into page_size in pixels?

src/gui/guiFormSpecMenu.cpp
```diff
@@ -513,14 +514,30 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
 
 		spec.ftype = f_ScrollBar;
 		spec.send  = true;
-		gui::IGUIScrollBar* e =
-				Environment->addScrollBar(is_horizontal,rect,this,spec.fid);
+		guiScrollBar* e = new guiScrollBar(Environment, this, spec.fid, rect,
```
stujones11: These references must be 'dropped' in the formspec destructor otherwise it will lead to memory leaks!

DS-Minetest: [Destructer](https://github.com/minetest/minetest/blob/5106ea9e39a4ac34e645baae419a5ad1f2a5cb91/src/gui/guiFormSpecMenu.cpp#L119) and [regenerateGui](https://github.com/minetest/minetest/blob/5106ea9e39a4ac34e645baae419a5ad1f2a5cb91/src/gui/guiFormSpecMenu.cpp#L2057) call [`removeChildren`](https://github.com/minetest/minetest/blob/5106ea9e39a4ac34e645baae419a5ad1f2a5cb91/src/gui/guiFormSpecMenu.cpp#L159) [which drops all children](http://irrlicht.sourceforge.net/docu/_i_g_u_i_element_8h_source.html#l00304) of `this` which include `e`.

stujones11: No, the children are removed, not dropped. Elements that are 'added' with `Evironment-add..` should not be dropped.

DS-Minetest: Oh, [`addChildToEnd`](http://irrlicht.sourceforge.net/docu/_i_g_u_i_element_8h_source.html#l00818) grabs the element. Well, then I think, it would be ok to drop `e` directly after constructing it.

stujones11: I suggest you look at how tables are handled `m_tables` there are also created with the `new` operator, hence why they are being 'dropped' here: https://github.com/minetest/minetest/blob/5106ea9e39a4ac34e645baae419a5ad1f2a5cb91/src/gui/guiFormSpecMenu.cpp#L122

Essentially, the author needs to do the same for `m_scrollbars`. It may be possible to drop them before hand, however, it is more usual to do this in the destructor.

DS-Minetest: Maybe just drop the element directly after creating it if it is never used again directly by this class.

stujones11: Are you sure, or is this a guess? I would prefer to err on the side of caution and consistency and do it like they did it for tables :)

DS-Minetest: I'm not really a trustworthy source here. Other humans are more experienced.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -2932,9 +3007,9 @@ void GUIFormSpecMenu::acceptInput(FormspecQuitMode quitmode=quit_mode_no)
 					// no dynamic cast possible due to some distributions shipped
 					// without rtti support in irrlicht
 					IGUIElement * element = getElementFromId(s.fid);
-					gui::IGUIScrollBar *e = NULL;
+					guiScrollBar *e = NULL;
 					if ((element) && (element->getType() == gui::EGUIET_SCROLL_BAR)) {
```
DS-Minetest: `guiScrollBar` is not of type `gui::EGUIET_SCROLL_BAR`.
It is `EGUIET_ELEMENT`:
https://github.com/minetest/minetest/blob/429a98964859b83016f2eb47a47a08ab8dc3c57e/src/gui/guiScrollBar.cpp#L19

src/gui/guiFormSpecMenu.cpp
```diff
@@ -2932,9 +3007,9 @@ void GUIFormSpecMenu::acceptInput(FormspecQuitMode quitmode=quit_mode_no)
...
-						e = static_cast<gui::IGUIScrollBar*>(element);
+						e = static_cast<guiScrollBar*>(element);
```
```suggestion
						e = static_cast<guiScrollBar *>(element);
```

src/gui/guiFormSpecMenu.h
```diff
@@ -405,7 +405,7 @@ class GUIFormSpecMenu : public GUIModalMenu
 	std::vector<std::pair<FieldSpec,gui::IGUICheckBox*> > m_checkboxes;
 	std::map<std::string, TooltipSpec> m_tooltips;
 	std::vector<std::pair<irr::core::rect<s32>, TooltipSpec>> m_tooltip_rects;
-	std::vector<std::pair<FieldSpec,gui::IGUIScrollBar*> > m_scrollbars;
+	std::vector<std::pair<FieldSpec,guiScrollBar*> > m_scrollbars;
```
DS-Minetest:
```suggestion
	std::vector<std::pair<FieldSpec, guiScrollBar *> > m_scrollbars;
```

src/gui/guiFormSpecMenu.h
```diff
@@ -455,6 +455,15 @@ class GUIFormSpecMenu : public GUIModalMenu
 		std::string focused_fieldname;
 		GUITable::TableOptions table_options;
 		GUITable::TableColumns table_columns;
+
+		struct {
+			int max = 1000;
+			int min = 0;
+			int smallStep = 10;
+			int largeStep = 100;
+			int thumbSize = 1;
```
DS-Minetest: Maybe use `s32` instead of `int`.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -529,6 +542,63 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
 	errorstream<< "Invalid scrollbar element(" << parts.size() << "): '" << element << "'"  << std::endl;
 }
 
+void GUIFormSpecMenu::parseScrollBarOptions(parserData* data, const std::string &element)
+{
+	std::vector<std::string> parts = split(element, ';');
+
+	if (parts.size() > 0) {
+		for (unsigned int i = 0; i < parts.size(); i++) {
```
DS-Minetest: Maybe `u32` instead of `unsigned int`.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -529,6 +542,63 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
 	errorstream<< "Invalid scrollbar element(" << parts.size() << "): '" << element << "'"  << std::endl;
 }
 
+void GUIFormSpecMenu::parseScrollBarOptions(parserData* data, const std::string &element)
+{
+	std::vector<std::string> parts = split(element, ';');
+
+	if (parts.size() > 0) {
```
```suggestion
	if (parts.size() <= 0) {
		warningstream << ...;
		return;
	}
	...
```

src/gui/guiFormSpecMenu.cpp
```diff
@@ -529,6 +542,63 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
...
+			}
+			if (options[0] == "min") {
+				data->scrollBarOptions.min = value;
+				continue;
+			}
+			if (options[0] == "smallstep") {
+				if (value < 0)
+					data->scrollBarOptions.smallStep = 10;
+				else
+					data->scrollBarOptions.smallStep = value;
```
DS-Minetest: This would be shorter:
```cpp
data->scrollBarOptions.smallStep = MYMAX(value, 10);
```
(Also everywhere else.)

src/gui/guiFormSpecMenu.cpp
```diff
@@ -513,14 +514,26 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
 
 		spec.ftype = f_ScrollBar;
 		spec.send  = true;
-		gui::IGUIScrollBar* e =
-				Environment->addScrollBar(is_horizontal,rect,this,spec.fid);
+		guiScrollBar* e = new guiScrollBar(Environment, this, spec.fid, rect,
```
```suggestion
		guiScrollBar *e = new guiScrollBar(Environment, this, spec.fid, rect,
```

src/gui/guiFormSpecMenu.h
```diff
@@ -508,6 +517,7 @@ class GUIFormSpecMenu : public GUIModalMenu
 	bool parseVersionDirect(const std::string &data);
 	bool parseSizeDirect(parserData* data, const std::string &element);
 	void parseScrollBar(parserData* data, const std::string &element);
+	void parseScrollBarOptions(parserData* data, const std::string &element);
```
DS-Minetest:
```suggestion
	void parseScrollBarOptions(parserData *data, const std::string &element);
```

src/gui/guiFormSpecMenu.h
DS-Minetest: Does this even work without including `"guiScrollBar.h"` in `guiFormSpecMenu.h`?

src/gui/guiFormSpecMenu.cpp
```diff
@@ -529,6 +542,63 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
...
+			}
+			if (options[0] == "largestep") {
+				if (value < 0)
+					data->scrollBarOptions.largeStep = 100;
+				else
+					data->scrollBarOptions.largeStep = value;
+				continue;
+			}
+			if (options[0] == "thumbsize") {
+				if (value < 0)
```
```suggestion
				if (value <= 0)
```
Else you would divide by 0.

src/gui/guiFormSpecMenu.h
stujones11: I agree, either the include should be moved to the header or use a forward declaration, not sure which is preferred in MT but since `guiTable.h` is included in the header, I would go with the former.

src/gui/guiFormSpecMenu.cpp
v-rob: I don't think it's necessary to use this for a counter.

src/gui/guiFormSpecMenu.cpp
v-rob: I think I have this done.  I just copied the table and changed it a bit, so I want someone to check it over.

DS-Minetest: It looks good now.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Actually here you could just do something like `for ( const std::string &i : parts)`.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -529,6 +545,54 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
...
+	std::vector<std::string> parts = split(element, ';');
+
+	if (parts.size() == 0) {
+		warningstream << "Invalid scrollbaroptions element(" << parts.size() << "): '" <<
+			element << "'"  << std::endl;
+		return;
+	}
+
+	for (unsigned int i = 0; i < parts.size(); i++) {
+
```
DS-Minetest: It might be that empty lines after things like `for` are not wanted.
```suggestion
```

src/gui/guiFormSpecMenu.cpp
v-rob: I suppose you could.  C++ has many strange syntactical things you can do.

DS-Minetest: A ranged for loop isn't that strange…

src/gui/guiFormSpecMenu.h
DS-Minetest: This is now `parserData * data`.

DS-Minetest:
Code looks nice now.

Can this PR be added to `Proposed order of merging formspec PRs` at https://dev.minetest.net/Meetings#2019-06-X (between style and real_coordinates)?

v-rob:
Yes, I think it's ready to be merged now.  It can be after or before `style`, but I think it should be before `scroll_container`, since this directly modifies scrollbars.

DS-Minetest:
Tested. Almost everything works nicely.

But if you set `min`, it's drawn incorrectly.
Maybe this can be fixed by using `scroll_pos-min` instead of `scroll_pos` here:
https://github.com/minetest/minetest/blob/bb14b7e9564b982ddcecb972218bfdae00cb4de4/src/gui/guiScrollBar.cpp#L292

v-rob:
Rebased and fixed improper scrollbar drawing.  Travis CI complains about this line, although it adheres to the Minetest code conventions: https://travis-ci.org/minetest/minetest/jobs/555143252#L522

DS-Minetest:
Tested, it works pretty well.
However, if I set `min` and `max` to nearly equal values (eg. `0` and `3`), I can't get the scroller to be quadratic anymore.

(I would just add the extra tab to please travis.)

Your rebase is a little weird. The reviewing diff shows all your PR's changes and also the changes of recent master commits. Reviewing won't be easily possible like this.

SmallJoker:
Something went wrong in the rebase. Not all of your commits were applied to the top of your branch.

v-rob:
I don't know why the rebase did what it did, and I have no idea how to fix it.  Git is mostly black magic to me still.  I would appreciate any help I could get on this.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -2180,6 +2243,11 @@ void GUIFormSpecMenu::parseElement(parserData* data, const std::string &element)
 		return;
 	}
 
+	if (type == "scrollbaroptions") {
```
SmallJoker: Move below "real_coodinates".

SmallJoker:
Rebased & squashed a few similar commits. This PR would fit very well into `style[]`, though.

doc/lua_api.txt
```diff
@@ -2277,16 +2277,37 @@ Elements
...
+    * Sets scrollbar minimum value, defaults to `0`.
+* `max=<int>`
+    * Sets scrollbar maximum value, defaults to `1000`.
+      If the max is equal to the min, the scrollbar will be disabled.
+* `smallstep=<int>`
+    * Sets scrollbar step value when the arrows are clicked or the mouse wheel is
+      scrolled.
+    * If this is set to a negative number, the value will be reset to `10`.
+* `largestep=<int>`
+    * Sets scrollbar step value when the bar itself is clicked.
```
DS-Minetest: `largestep` is no longer used if you click on the tray.
```suggestion
    * Sets scrollbar step value used by page up and page down.
```

v-rob:
Rebased (and didn't mess up this time).  Sorry it took so long.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -556,17 +560,30 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
 
 		spec.ftype = f_ScrollBar;
 		spec.send  = true;
-		gui::IGUIScrollBar* e =
-				Environment->addScrollBar(is_horizontal,rect,this,spec.fid);
+		guiScrollBar *e = new guiScrollBar(Environment, this, spec.fid, rect,
```
DS-Minetest: I see you didn't compile yet.
```suggestion
		GUIScrollBar *e = new GUIScrollBar(Environment, this, spec.fid, rect,
```

src/gui/guiFormSpecMenu.cpp
```diff
@@ -3384,9 +3453,9 @@ void GUIFormSpecMenu::acceptInput(FormspecQuitMode quitmode=quit_mode_no)
 					// no dynamic cast possible due to some distributions shipped
 					// without rtti support in irrlicht
 					IGUIElement * element = getElementFromId(s.fid);
-					gui::IGUIScrollBar *e = NULL;
-					if ((element) && (element->getType() == gui::EGUIET_SCROLL_BAR)) {
-						e = static_cast<gui::IGUIScrollBar*>(element);
+					guiScrollBar *e = NULL;
+					if ((element) && (element->getType() == EGUIET_ELEMENT)) {
+						e = static_cast<guiScrollBar *>(element);
```
DS-Minetest: Here GUI, too.

src/gui/guiFormSpecMenu.h
```diff
@@ -27,6 +27,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include "inventorymanager.h"
 #include "modalMenu.h"
 #include "guiTable.h"
+#include "guiScrollBar.h"
```
DS-Minetest: Use forward declaration:
- Move this line to `guiFormSpecMenu.cpp`.
- Add below there (line 36 and further) a line with `class GUIScrollBar;`.

src/gui/guiFormSpecMenu.h
```diff
@@ -438,7 +439,7 @@ class GUIFormSpecMenu : public GUIModalMenu
 	std::vector<std::pair<FieldSpec,gui::IGUICheckBox*> > m_checkboxes;
 	std::map<std::string, TooltipSpec> m_tooltips;
 	std::vector<std::pair<irr::core::rect<s32>, TooltipSpec>> m_tooltip_rects;
-	std::vector<std::pair<FieldSpec,gui::IGUIScrollBar*> > m_scrollbars;
+	std::vector<std::pair<FieldSpec, guiScrollBar *>> m_scrollbars;
```
DS-Minetest:
```suggestion
	std::vector<std::pair<FieldSpec, GUIScrollBar *>> m_scrollbars;
```

src/gui/guiFormSpecMenu.h
```diff
@@ -542,6 +552,7 @@ class GUIFormSpecMenu : public GUIModalMenu
 	bool parseVersionDirect(const std::string &data);
 	bool parseSizeDirect(parserData* data, const std::string &element);
 	void parseScrollBar(parserData* data, const std::string &element);
+	void parseScrollBarOptions(parserData * data, const std::string &element);
```
DS-Minetest:
```suggestion
	void parseScrollBarOptions(parserData *data, const std::string &element);
```

src/gui/guiFormSpecMenu.cpp
```diff
@@ -26,6 +26,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include "guiButton.h"
 #include "guiFormSpecMenu.h"
 #include "guiTable.h"
+#include "guiScrollBar.h"
```
DS-Minetest: \*cough\* \*cough\* _…pqr**st**uvwxyz_

v-rob: They're supposed to be in alphabetical order?  I didn't know that.

DS-Minetest:
I'd like to point out that this PR is the highest element in a priority list.

v-rob:
All this PR needs is reviews, unless people want this integrated into `style`, which would likely be trivial.  But it really should be merged soon, since it's holding everything else up in the queue, and much of those are much more important.

DS-Minetest:
https://github.com/minetest/minetest/pull/8740#issuecomment-519513766 also happens in this PR. Here's the image from that comment:

I request that the visibility of the arrow buttons should be settable with scrollbar options.

DS-Minetest:
Another issue is that a scrollbar with `"scrollbaroptions[min=0;max=2;thumbsize=1]"` feels weird when dragging, as if there was some offset. This is probably caused by something in the `GUIScrollBar` class.

v-rob:
I'm working on the button visibility, although it's a little trickier than I thought.

> Another issue is that a scrollbar with `"scrollbaroptions[min=0;max=2;thumbsize=1]"` feels weird when dragging, as if there was some offset. This is probably caused by something in the `GUIScrollBar` class.

This completely depends on where you grab the thumb.  If you grab it near the edge, it will move when the mouse approaches the edge of the next position.  Similarly, if you grab it at the center, it will move when the mouse approaches the center of the next position.  So, this is beyond the scope of this PR.

doc/lua_api.txt
```diff
@@ -2284,16 +2284,40 @@ Elements
...
+    * If this is set to a negative number, the value will be reset to `100`.
+* `thumbsize=<int>`
+    * Sets size of the thumb on the scrollbar. Size is calculated in the number of
+      units the thumb spans out of the range of the scrollbar values.
+    * Example: If a scrollbar has a `min` of 1 and a `max` of 100, a thumbsize of 10
+      would span a tenth of the scrollbar space.
+    * If this is set to zero or less, the value will be reset to `1`.
+* `arrows=<show/hide/default>`
+    * Whether to show the arrow buttons on the scrollbar. `default` hides the arrows
+      when the scrollbar gets too small, but shows them otherwise.
```
DS-Minetest: The arrow buttons are not hidden if the scrollbar is too small, but if it's too broad. (If it's 4 times wider than long, the buttons are omitted.)

I don't think that the default option is needed. The default can be `show`.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -556,23 +561,98 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
...
+			data->scrollBarOptions.largeStep = value < 0 ? 100 : value;
+			continue;
+		}
+		else if (options[0] == "thumbsize") {
+			int value = stoi(options[1]);
+			data->scrollBarOptions.thumbSize = value <= 0 ? 1 : value;
+			continue;
+		}
+		else if (options[0] == "arrows") {
+			std::string value = trim(options[1]);
```
```suggestion
			const std::string &value = trim(options[1]);
```

src/gui/guiScrollBar.cpp
```diff
@@ -315,6 +316,12 @@ void GUIScrollBar::setPageSize(const s32 &size)
 	setPos(scroll_pos);
 }
 
+void GUIScrollBar::setArrowsVisible(bool visible)
+{
+	up_button->setVisible(visible);
+	down_button->setVisible(visible);
```
DS-Minetest: I think, it would be better to add a button visibility attribute (bool or enum).
At least set boarder to width, height or 0.

doc/lua_api.txt
v-rob: The `show` options will _always_ show the arrows, whereas `default` lets the scrollbar choose.  The `show` option can be used, for instance, to only show arrows and no bar, e.g. this code will make a spinbox:
```
scrollbaroptions[arrows=show]
field[0,0;3,1;field;SpinBox;5]
scrollbar[0,3;0.5,1;vertical;sbar;0]
```
If we didn't have `default`, though, then some scrollbars might end up this way accidentally.

src/gui/guiFormSpecMenu.cpp
SmallJoker: `trim` returns a temporary string, hence const ref is either useless or will even cause issues.

DS-Minetest: Oh, sorry.

src/gui/guiScrollBar.cpp
```diff
@@ -419,7 +425,12 @@ void GUIScrollBar::refreshControls()
 		down_button->setAlignment(EGUIA_UPPERLEFT, EGUIA_LOWERRIGHT,
 				EGUIA_LOWERRIGHT, EGUIA_LOWERRIGHT);
 	}
-	bool visible = (border_size != 0);
+	bool visible;
+	if (is_arrows_visible == 2)
+		visible = (border_size != 0);
+	else
+		visible = (bool) is_arrows_visible;
```
DS-Minetest: Maybe you also have to set `border_size`.

src/gui/guiScrollBar.h
```diff
@@ -47,6 +48,7 @@ class GUIScrollBar : public IGUIElement
 
 	IGUIButton *up_button;
 	IGUIButton *down_button;
+	u8 is_arrows_visible;
```
DS-Minetest: An enum might be better.

rubenwardy:
I don't think the properties you state in the OP should be part of style because they're not directly styling or theme related. Most scrollbars should define their own min and max, for example, and having global settings for that doesn't make sense

v-rob:
Well, this is a global setting like `tableoptions`, but it can be changed as often as you like.  Global settings make for less code and faster runtime.  Also, if two scrollbars are identical (I can see some possible situations, e.g. RGB color sliders from 0-255), one global setting will work for all of them as long as it's placed right before.

This is no longer WIP, unless anyone has anything else to add/change.

DS-Minetest:
There's a weird file called `get_data()`.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -122,6 +123,10 @@ GUIFormSpecMenu::~GUIFormSpecMenu()
 		table_it.second->drop();
 	}
 
+	for (auto &scrollbar_it : m_scrollbars) {
+		scrollbar_it.second->drop();
+	}
```
DS-Minetest: Please also drop in `regenerateGui` (see line 2466).
(And the curly braces aren't needed.)

src/gui/guiFormSpecMenu.cpp
```diff
@@ -556,23 +561,96 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
...
-		e->setLargeStep(100);
+
+		e->setSmallStep(data->scrollBarOptions.smallStep);
+		e->setLargeStep(data->scrollBarOptions.largeStep);
+
+		s32 scrollbar_size;
+		if (is_horizontal)
+			scrollbar_size = dim.X;
+		else
+			scrollbar_size = dim.Y;
```
DS-Minetest: `s32 scrollbar_size = is_horizontal ? dim.X : dim.Y;`

src/gui/guiFormSpecMenu.cpp
```diff
@@ -556,23 +561,96 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
...
+	std::vector<std::string> parts = split(element, ';');
+
+	if (parts.size() == 0) {
+		warningstream << "Invalid scrollbaroptions element(" << parts.size() << "): '" <<
+			element << "'"  << std::endl;
+		return;
+	}
+
+	for (const std::string &i : parts) {
+		std::vector<std::string> options = split(i, '=');
```
DS-Minetest: Would the following be possible?
```suggestion
		const std::vector<std::string> &options = split(i, '=');
```

src/gui/guiFormSpecMenu.cpp
```diff
@@ -556,23 +561,96 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
...
 		return;
 	}
-	errorstream<< "Invalid scrollbar element(" << parts.size() << "): '" << element << "'"  << std::endl;
+	errorstream << "Invalid scrollbar element(" << parts.size() << "): '" << element
+		<< "'" << std::endl;
+}
+
+void GUIFormSpecMenu::parseScrollBarOptions(parserData* data, const std::string &element)
+{
+	std::vector<std::string> parts = split(element, ';');
```
DS-Minetest: Would the following be possible?
```suggestion
	const std::vector<std::string> &parts = split(element, ';');
```

src/gui/guiFormSpecMenu.cpp
```diff
@@ -556,23 +561,96 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
...
+			warningstream << "Invalid scrollbaroptions option syntax: '" <<
+				element << "'" << std::endl;
+			continue; // Go to next option
+		}
+
+		if (options[0] == "max") {
+			data->scrollBarOptions.max = stoi(options[1]);
+			continue;
+		}
+		else if (options[0] == "min") {
```
DS-Minetest: `}` and `else` in one line:
`} else if (..) {`
(See https://www.kernel.org/doc/html/latest/process/coding-style.html#placing-braces-and-spaces.)

src/gui/guiFormSpecMenu.cpp
```diff
@@ -3384,9 +3467,9 @@ void GUIFormSpecMenu::acceptInput(FormspecQuitMode quitmode=quit_mode_no)
 					// no dynamic cast possible due to some distributions shipped
 					// without rtti support in irrlicht
 					IGUIElement * element = getElementFromId(s.fid);
-					gui::IGUIScrollBar *e = NULL;
-					if ((element) && (element->getType() == gui::EGUIET_SCROLL_BAR)) {
-						e = static_cast<gui::IGUIScrollBar*>(element);
+					GUIScrollBar *e = NULL;
```
```suggestion
					GUIScrollBar *e = nullptr;
```

DS-Minetest:
> Another issue is that a scrollbar with `"scrollbaroptions[min=0;max=2;thumbsize=1]"` feels weird when dragging, as if there was some offset. This is probably caused by something in the `GUIScrollBar` class.
> Edit: This is indeed out of scope, sorry.

Btw. I've found the reason and a fix for this:
`GUIScrollBar::getPosFromMousePos` makes rounding errors. To fix this, change the following line:
https://github.com/minetest/minetest/blob/e892bcdcd7c4144acfe562e16e7add38797fa184/src/gui/guiScrollBar.cpp#L250
to:
`	return core::isnotzero(range()) ? s32(f32(p) / f32(w) * range() + 0.5f) + min_pos : 0;`
(Note the added `0.5f`.)

DS-Minetest:
:+1: 
Tested again, works.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -556,23 +559,96 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
 
 		spec.ftype = f_ScrollBar;
 		spec.send  = true;
-		gui::IGUIScrollBar* e =
-				Environment->addScrollBar(is_horizontal,rect,this,spec.fid);
+		GUIScrollBar *e = new GUIScrollBar(Environment, this, spec.fid, rect,
```
SmallJoker: Needs `e->drop()`.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -556,23 +559,96 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
...
+			warningstream << "Invalid scrollbaroptions option syntax: '" <<
+				element << "'" << std::endl;
+			continue; // Go to next option
+		}
+
+		if (options[0] == "max") {
+			data->scrollBarOptions.max = stoi(options[1]);
+			continue;
+		}
+		else if (options[0] == "min") {
```
SmallJoker: Please put `else if` onto the same line as `}` above.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: It is already dropped, see https://github.com/minetest/minetest/pull/8530/files#diff-074850a88796e13dad12dd7ae1a8d9f7R126 and https://github.com/minetest/minetest/pull/8530/files#diff-074850a88796e13dad12dd7ae1a8d9f7R2465.

src/gui/guiFormSpecMenu.h
```diff
@@ -489,6 +490,16 @@ class GUIFormSpecMenu : public GUIModalMenu
...
 		GUITable::TableColumns table_columns;
+
+		struct {
+			s32 max = 1000;
+			s32 min = 0;
+			s32 smallStep = 10;
+			s32 largeStep = 100;
+			s32 thumbSize = 1;
+			u8 showArrows = 2;
+		} scrollBarOptions;
```
SmallJoker: This is a variable name. Please use `snake_case`.

src/gui/guiScrollBar.cpp
```diff
@@ -419,7 +426,22 @@ void GUIScrollBar::refreshControls()
...
-	bool visible = (border_size != 0);
+
+	bool visible;
+	if (is_arrows_visible == 2)
+		visible = (border_size != 0);
+	else if (is_arrows_visible == 0) {
+		visible = false;
+		border_size = 0;
+	}
+	else {
```
SmallJoker: `else` on same line as `}`. Please see coding style guidelines.

src/gui/guiScrollBar.h
```diff
@@ -47,6 +48,7 @@ class GUIScrollBar : public IGUIElement
 
 	IGUIButton *up_button;
 	IGUIButton *down_button;
+	u8 is_arrows_visible = 2;
```
SmallJoker: Why not a boolean?

src/gui/guiFormSpecMenu.cpp
```diff
@@ -556,23 +559,96 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
...
+		e->setSmallStep(data->scrollBarOptions.smallStep);
+		e->setLargeStep(data->scrollBarOptions.largeStep);
+
+		s32 scrollbar_size;
+		if (is_horizontal)
+			scrollbar_size = dim.X;
+		else
+			scrollbar_size = dim.Y;
+
+		e->setPageSize(scrollbar_size * (max-min+1) / data->scrollBarOptions.thumbSize);
```
SmallJoker: Spaces between binary operators (`a - b`).

src/gui/guiFormSpecMenu.cpp
```diff
@@ -556,23 +559,96 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
...
-		e->setMax(1000);
-		e->setMin(0);
 		e->setPos(stoi(parts[4]));
-		e->setSmallStep(10);
-		e->setLargeStep(100);
+
+		e->setSmallStep(data->scrollBarOptions.smallStep);
+		e->setLargeStep(data->scrollBarOptions.largeStep);
+
+		s32 scrollbar_size;
```
SmallJoker: Could be combined to `s32 scrollbar_size = is_horizontal ? dim.X : dim.Y;`

src/gui/guiFormSpecMenu.cpp
SmallJoker: Thanks. Although it's currently inconsistent this should work as intended.

src/gui/guiScrollBar.h
```diff
@@ -17,6 +17,12 @@ the arrow buttons where there is insufficient space.
 using namespace irr;
 using namespace gui;
 
+enum ScrollBarArrowVisibility {
+	HIDE,
+	SHOW,
+	DEFAULT
+};
```
DS-Minetest: You can put that enum into the `GUIScrollBar` class:
```cpp
class GUIScrollBar : public IGUIElement
{
public:
	enum ArrowVisibility {
		HIDE,
		SHOW,
		DEFAULT
	};
//...
};
```
Outside of this class you can access it with `GUIScrollBar::ArrowVisibility` and its values eg. with `GUIScrollBar::HIDE`.
The forward declaration (`guiFormSpecMenu.h`, line 40)  might be possible with `enum GUIScrollBar::ArrowVisibility;` or with
```cpp
enum GUIScrollBar::ArrowVisibility {
	HIDE,
	SHOW,
	DEFAULT
};
```
(or just include the header in the header).

SmallJoker:
@v-rob Due to @DS-Minetest's formspec changes you might need to move some code.

I lost track of this PR, but it apparently only needs re-reviewing after rebase and some quick tests (which I actually wanted to do now).

v-rob:
Sorry it took so long to rebase.  I forget how to rebase properly every time I try it, so I wrote it down this time.

src/gui/guiFormSpecMenu.cpp
SmallJoker: No. The function returns a temporary variable.

PR #8535, opened 2019-05-15
Save forceloaded blocks file periodically

thomasrudin:
saves the forceloaded blocks periodically.
checks every 10 seconds if the forceloaded blocks got changed in-game
and persists them on-disk if thats the case

### DIFF
--- a/builtin/game/forceloading.lua
+++ b/builtin/game/forceloading.lua
@@ -8,6 +8,9 @@ local blocks_forceloaded
 local blocks_temploaded = {}
 local total_forceloaded = 0
 
+-- true, if the forceloaded blocks got changed (flag for persistence on-disk)
+local forceload_blocks_changed = false
+
 local BLOCKSIZE = core.MAP_BLOCKSIZE
 local function get_blockpos(pos)
 	return {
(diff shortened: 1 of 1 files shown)


builtin/game/forceloading.lua
```diff
@@ -31,6 +34,9 @@ local function get_relevant_tables(transient)
 end
 
 function core.forceload_block(pos, transient)
+	-- set changed flag
+	forceload_blocks_changed = true
```
SmallJoker: It is only changed after `forceload_block(blockpos)`

DS-Minetest:
Wouldn't it make more sense to save always when forceload or free block is called with `not transient`?

Oh, and the PR title has a typo.

thomasrudin:
> Wouldn't it make more sense to save always when forceload or free block is called with `not transient`?

No, because the function may (and will be) called multiple
times in a row...

> Oh, and the PR title has a typo.

Thats intentional, of course ;)

thomasrudin:
Just fyi: works "in production" (pandorabox) stable the last 2 weeks :smile:

PR #8537, opened 2019-05-15
Improve buildbot script

EvidenceBKidscode:
Here are some small buildbot improvements:
 
- Variables to set git repositories urls (allows to build from another repo);
- Variables to set git branch to build from (allows to build from non-master branch);
- Variables to change built name;
- Make -j uses the proper number of processors (autodetect);
- Changed toolchain from win32 to adpat it to non-debian distros (not quite satisfied of this);


SmallJoker:
How will debian platforms know whether they should use `toolchain_mingw_debian.cmake`?

pyrollo:
I don't know what should be the best practice to achieve that. Create a third buildbotXX.sh ?

SmallJoker:
@pyrollo The best solution would of course to automatically load the right toolchain file, depending on the platform. A third buildbot script is just yet another code duplicate, which is helpful in terms of maintenance.

sfan5:
The `toolchain_mingw.cmake` files are just intended to be examples and not necessarily ready-to-use on any platform.

pyrollo:
New version detecting which mingw32 compiler to use to build 32bits executable.

util/buildbot/buildwin32.sh
```diff
@@ -12,7 +19,16 @@ builddir="$( cd "$builddir" && pwd )"
 packagedir=$builddir/packages
 libdir=$builddir/libs
 
-toolchain_file=$dir/toolchain_mingw.cmake
+# Test which win32 compiler is present
+which i686-w64-mingw32-windres &&	toolchain_file=$dir/toolchain_i646-w64-mingw32.cmake
+which i586-mingw32msvc-windres &&	toolchain_file=$dir/toolchain_i586-mingw32msvc.cmake
+
+if [ ! $toolchain_file ]
```
sfan5: `if [ -z "$toolchain_file" ]; then`

pyrollo:
Seems better for Travis now :)

sfan5:
@pyrollo do you use MinGW with `i586-mingw32msvc-gcc` yourself?
If not I'd like to drop that toolchain file and add detection for the `i686-w64-mingw32-c++-posix` variant Debian/Ubuntu use. (which MT also requires by default, possibly causing confusing errors)

PR #8538, opened 2019-05-15
Add content_rating tag to appdata

cpba:
This tag allows software stores (KDE discover, etc) to filter and know whether applications are suitable for young users.

### DIFF
--- a/misc/net.minetest.minetest.appdata.xml
+++ b/misc/net.minetest.minetest.appdata.xml
@@ -3,6 +3,10 @@
 	<id>net.minetest.minetest.desktop</id>
 	<metadata_license>CC0-1.0</metadata_license>
 	<project_license>LGPL-2.1+ and CC-BY-SA-3.0 and MIT and Apache-2.0</project_license>
+	<content_rating type="oars-1.0">
+		<content_attribute id="violence-cartoon">mild</content_attribute>
+		<content_attribute id="social-chat">intense</content_attribute>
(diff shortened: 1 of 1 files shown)


Calinou:
Flathub now requires AppData files to contain `<content_rating>` tags, so this is a good idea.

sfan5:
By using the [OARS generator](https://hughsie.github.io/oars/generate.html) myself, I got to this result:
```xml
  <content_rating type="oars-1.0">
    <content_attribute id="violence-cartoon">mild</content_attribute>
    <content_attribute id="violence-fantasy">mild</content_attribute>
    <content_attribute id="social-chat">intense</content_attribute>
    <content_attribute id="social-info">mild</content_attribute>
  </content_rating>
```

Thoughts?

cpba:
It's a bit subjective, minetest characters consist of some cubes stacked
together so I considered them cartoon characters. However, they are meant
to represent human characters, so we could add mild fantasy violence if you
believe that's more accurate.

Social info is also server dependent, but we could add it to err on the
side of caution if you believe that's for the best.

sfan5:
Regarding `social-info`, the dropdown says "Mild: Using any online API, e.g. a user-counter"
There is one API Minetest uses by default, which is fetching a list of servers from servers.minetest.net.

cpba:
@sfan5 That's a good point, I have updated the branch with your suggestions.

PR #5852, opened 2017-05-30
Improve the path select GUI

red-001:
- Allow lua to chose whatever directories or files can be selected
- Fix selecting directories
- Rename dialog to `guiPathSelectMenu` from `guiFileSelectMenu`
- Rename lua function for opening the menu from `show_file_open_dialog` to `show_path_select_dialog`
- Remove duplicate code and fix code style.

This fixes #5833

PR #8572, opened 2019-06-02
Remove modstore leftovers. Fix core.show_path_select_dialog documentation

DS-Minetest:
In #5852, it was forgotten to update the documentation and to replace `show_file_open_dialog` in  `builtin/mainmenu/tab_content.lua`. This PR fixes this.

Notes:
- I did not test this.
- I wonder why there was no crash yet. Or is there an issue?
- It might be that in `tab_content.lua` it's not a file selector, please check.

### DIFF
--- a/builtin/mainmenu/tab_content.lua
+++ b/builtin/mainmenu/tab_content.lua
@@ -153,11 +153,6 @@ local function handle_buttons(tabview, fields, tabname, tabdata)
 		return true
 	end
 
-	if fields["btn_mod_mgr_install_local"] ~= nil then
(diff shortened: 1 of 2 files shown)


SmallJoker:
This dead code. I cannot find any definition of `btn_mod_mgr_install_local` because it was removed from `tab_content.lua`/`get_formspec` somewhen.

pauloue:
It was removed in https://github.com/minetest/minetest/commit/dc3ca09e0e0e25f40cccae12c73830b0334ffb60.
So the checks for `btn_mod_mgr_install_local` and `mod_mgt_open_dlg_accepted` in `handle_buttons` can be removed.

DS-Minetest:
Ok, the checks are removed.

PR #8573, opened 2019-06-02
HTTP API: Allow binary downloads and headers

SmallJoker:
This patch allows to download binary files.
Needed for https://github.com/minetest-mods/skinsdb/pull/27

How to test:
1) Request the HTTP functions
2) Download s binary file
3) Write it to the disk

Previously the contents were only copied to the first NULL character (C-strings).

### DIFF
--- a/src/script/common/c_converter.cpp
+++ b/src/script/common/c_converter.cpp
@@ -540,9 +540,9 @@ v3s16 getv3s16field_default(lua_State *L, int table,
 }
 
 void setstringfield(lua_State *L, int table,
-		const char *fieldname, const char *value)
+		const char *fieldname, const std::string &value)
 {
-	lua_pushstring(L, value);
+	lua_pushlstring(L, value.c_str(), value.length());
 	if(table < 0)
 		table -= 1;
 	lua_setfield(L, table, fieldname);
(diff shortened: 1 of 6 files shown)


sfan5:
This would be a good candidate to add to `minetest.features` so mods can check whether binary requests are supported.

thomasrudin:
Does this also support binary uploads?

SmallJoker:
@thomasrudin Yes, after this PR you can pass them in POST fields. I don't think it would be a good idea to allow that in the URL (i.e. GET fields) as well. I did not test this functionality but `HTTPFetchOngoing::HTTPFetchOngoing` clearly passes cURL the length of the C-string which then is used instead of the null termination.

@sfan5 Thanks for the suggestion. Added.

src/script/common/c_converter.cpp
```diff
@@ -540,9 +540,9 @@ v3s16 getv3s16field_default(lua_State *L, int table,
 }
 
 void setstringfield(lua_State *L, int table,
-		const char *fieldname, const char *value)
+		const char *fieldname, const std::string &value)
```
sfan5: does this not introduce some/many(?) unneeded conversions to std::string and back?
how often is setstringfield used with `char*`?

SmallJoker: It's used exactly 1 times: `l_http.cpp`

Issue #4844, opened 2016-12-04
It's not possible anymore to write text with label[ ... ] over item images or inventory slots in formspecs


Miner59:
This would be nice if this would work again.
I found issue  #3672 which fixes the text on the item image button itself. But I think other text still can't overlap images on image buttons or inventory slots. It seems inventory slots now always hide text written over it. I think it should depend on the order the elements are written in the formspec string, which element overlaps the other.

For comparison:

On actual minetest version:
![hiddennumbers](https://cloud.githubusercontent.com/assets/10222419/20868857/63be0f0c-ba66-11e6-9212-3067395764c4.png)

On older version:
http://i58.tinypic.com/w172bt.png

What I talk about are the numbers for counting the items which you can see were working in older versions. I point on the position the number should be with the cursor in upper image.
The code used to do this is the same in both images, it uses label[ ... ] to display numbers over the pictures of the items.

Miner59:
Hello,
I want to bump this issue because I got no response on it.
I made a better screenshot and wrote a little test mod to give better impression of the current behavior.
In older versions, before changes from RBA on generating item images, text written with label in formspecs was in front of images and inventory slots. Now it is always hidden behind.
Truely buggy is the interaction between image buttons and text written with label.
The text is in front of the button, but behind the image depicted on the button. (???)

![example](https://cloud.githubusercontent.com/assets/10222419/21290874/eb9959bc-c4cb-11e6-83be-16b0d31fc26d.png)

test mod:
load the mod, then type /label_test in chat and watch your inventory screen
[label_test.zip](https://github.com/minetest/minetest/files/659194/label_test.zip)

dsohler:
[Reference link](https://github.com/cornernote/minetest-craft_guide/issues/18)

PR #8591, opened 2019-06-10
Add scroll_container formspec element

### LINKED ISSUE #8336: Fullscreen bgcolor ignores color
##### Issue type
 - Bug report
 
##### Minetest version
```
5.0.0
```

##### Summary
The color of `bgcolor[]` does not work when fullscreen. It's always a translucent black, no matter what color you choose.

##### Steps to reproduce
Use `luacmd` and enter this command:

`/lua minetest.show_formspec("singleplayer", "hey", "size[4,4]bgcolor[red;true]")`

Expected: Red fullscreen. Actual: Translucent black fullscreen.

Compare to:

`/lua minetest.show_formspec("singleplayer", "hey", "size[4,4]bgcolor[red]")`

This works, a red window is opened.

### LINKED ISSUE #8576: Formspec: scrollable container
##### Issue type
 - Feature request
 
##### Minetest version
```
Using Irrlicht 1.8.4
BUILD_TYPE=Release
RUN_IN_PLACE=1
USE_GETTEXT=1
USE_SOUND=1
USE_CURL=1
USE_FREETYPE=1
USE_LUAJIT=0
STATIC_SHAREDIR="."

```

##### Summary
I request a new formspec element that should work like this:

Two formspec elements are added:
```
scrollbar_container[<X>,<Y>;<W>,<H>;<name of a scrollbar>;<multiplier>]
scrollbar_container_end[]
```
All formspec elements that are defined after a `scrollbar_container` and before a `scrollbar_container_end` should be
- offset by `<X>` and `<Y>`,
- additionally offset by the current value of the scrollbar with the name `<name of a scrollbar>` ~~(or the value of the scrollbar of the element with the name `<name of a scrollbar>` (eg. `textarea`))~~ times `<multiplier>` (the direction depends on whether the scrollbar is vertical or horizontal) and
- not drawn outside of the window defined by `<X>,<Y>;<W>`.

If your mouse is in the scrollbar container, the corresponding scrollbar should be scrolled with the mouse wheel.

This would allow small and large windows with relatively much information.

##### Example
The following formspec:
```
size[6,4]
scrollbar[5,0;0.2,4;vertical;scrbar;0]
scrollbar_container[0,0;5,4;scrbar;0.025]
button[0,0;2,1;btn;Button]
field[4,4;3,1;fld;field;empty]
scrollbar_container_end[]
```
Should look approximately like this:
example
example2

##### Usecases
I can see very many usecases for this. Here are just very few examples:
- A key change menu where you do not have to compress a limited number of buttons to small space.
- A creative inventory where you just scroll through the list.
- A formspec listing rules where you can't mark the text and where the button to accept appears when you at the bottom of the text.

DS-Minetest:
Does what is described in #8576 and closes #8576.
Also fixes #8336. (Closes #8716, closes #8740.)
I recommend reviewing and merging #8716 and #8740 first.

- There are a three new classes (`GUIBox`, `GUIItemImage` and `GUIScrollContainer`).
- All elements are realized with or bound to `IGUIElement`s. This is to make use of parents.
- There's no need to get an absolute element position as every element is a child.
- Lists are still drawn after everything else, ergo #4844 is not entirely fixed. (I've tried to add a `GUIList` class which draws a list, but it was too hard and would have gone too far.)
- The scroll_container consists of two elements. One clips every child, the other moves them. The clipper is the parent of the mover.
- The scroll container mover (of class `GUIScrollContainer`) is moved on event.
- I guess some things can be optimized, I'm very inexperienced.

<details><summary>Here's a formspec that I've used for testing.</summary>

```lua
local my_formspec =
	"size[10,8]"..
	--~ "position[0.2,0.7]"..
	--~ "anchor[0,1]"..
	"button[8,1;1,1;bla;Bla]"..
	"scrollbar[7.5,0;0.3,4;vertical;scrbar;0]"..
	"scroll_container[1,1;8,6;scrbar;down;0.06]"..
		"button[0,1;1,1;baum;Baum]"..
		--~ "container[0,0]"..
		"button[0,10;1,1;planze;Pflanze]"..
		"pwdfield[2,2;1,1;baum2;Baum]"..
		--~ "container_end[]"..
		"list[current_player;main;4,4;1,5;]"..
		"box[2,5;3,2;#ffff00]"..
		"image[1,10;3,2;bubble.png]"..
		"image[3,1;bubble.png]"..
		"item_image[2,6;3,2;default:mese]"..
		"label[2,15;bla baum\nfoo bar]"..
		--~ "textarea[2,10;5,6;txtarea;Baum;bla]"..
		"item_image_button[2,3;1,1;default:dirt_with_grass;itemimagebutton;ItemImageButton]"..
		--~ "tooltip[planze;Blatt;#f00;#000]"..
		"tooltip[0,11;3,2;Blatt;#f00;#000]"..
		"box[0,11;3,2;#00ff00]"..
		--~ "checkbox[2,25;chb;ChB;false]",
		"container[0,18]"..
			"scroll_container[1,2;3,2;scrbar2;right;0.1]"..
				"label[10,0;nest]"..
				"button[0,0;6,1;butnest;Nest]"..
				"bgcolor[#aa0a]"..
			"scroll_container_end[]"..
			"scrollbar[1,0;3.5,0.3;horizontal;scrbar2;0]"..
		"container_end[]"..
		--~ "tablecolumns[text,bla]"..
		--~ "table[1,23;3,1;tbl;a,b,c,d;2]"..
		--~ "textlist[0,5;2,0.5;txtlst;a,b,c]"..
		"dropdown[0,6;2;hmdrpdwn;apfel,birne;1]"..
		"image_button[0,4;2,2;bubble.png;bubblebutton;bbbbtt;false;true;heart.png]"..
		--~ "tabheader[-1,9;mytabheader;a,b,c;2;false;false]"..
		--~ "background[2,4;2,2;default_wood.png;true]"..
		"bgcolor[#00aa]"..
	"scroll_container_end[]"..
	--~ "textarea[5,0.5;1,1;scrtxtarea;Baum;blablaajk adfkjgn \n akfjb]"..
	--~ "bgcolor[;true]"..
	--~ "bgcolor[#0000]"..
	""
```
</details>

Here's a demo video: [scroll_container_show.m4v.zip](https://github.com/minetest/minetest/files/3273388/scroll_container_show.m4v.zip)

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -1882,6 +1882,11 @@ When displaying text which can contain formspec code, e.g. text set by a player,
 use `minetest.formspec_escape`.
 For coloured text you can use `minetest.colorize`.
 
+Elements drawn in the order they are defined. All background elements are drawn
+before all other elements.
+`list` elements are an exception here. They are drawn last. This, however, might
+be changed at any time.
+
 WARNING: Minetest allows you to add elements to every single formspec instance
 using `player:set_formspec_prepend()`, which may be the reason backgrounds are
 appearing when you don't expect them to. See [`no_prepend[]`].
@@ -1971,6 +1976,23 @@ Elements
 * End of a container, following elements are no longer relative to this
   container.
 
+### `scroll_container[<X>,<Y>;<W>,<H>;<scrollbar name>;<direction>;<multiplier>]`
+
+* Start of a scroll_container block, all elements in a scroll_container block will
+  * take the block's coordinate as position origin,
+  * be additionally moved by the current value of the scrollbar with the name
+    `scrollbar name` times `multiplier` in the direction `direction` and
+  * be clipped to the rectangle defined by `X`, `Y`, `W` and `H`.
+* Possible values for `direction` are `down` and `right`.
+* `multiplier` is optional and defaults to `0.1`.
+* Nesting is possible.
+* Some elements might work a little different if they are in a scroll_container.
+
+### `scroll_container_end[]`
+
+* End of a scroll_container, following elements are no longer bound to this
+  container.
+
 ### `list[<inventory location>;<list name>;<X>,<Y>;<W>,<H>;]`
 
 * Show an inventory list if it has been sent to the client. Nothing will
@@ -2037,11 +2059,15 @@ Elements
 
 * Show an inventory image of registered item/node
 
-### `bgcolor[<color>;<fullscreen>]`
+### `bgcolor[<color>]`
 
 * Sets background color of formspec as `ColorString`
-* If `true`, the background color is drawn fullscreen (does not affect the size
-  of the formspec).
+
+### `bgcolor[<color>;<fullscreen>]`
+
+* If `color` is a valid `ColorString`, the fullscreen background color
+  is set to `color`.
+* If `fullscreen` is a true value, the fullscreen background color is drawn.
 
 ### `background[<X>,<Y>;<W>,<H>;<texture name>]`
 
--- a/src/gui/guiBox.cpp
+++ b/src/gui/guiBox.cpp
@@ -1,0 +1,38 @@
+/*
+Minetest
+Copyright (C) 2013 celeron55, Perttu Ahola <celeron55@gmail.com>
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation; either version 2.1 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License along
+with this program; if not, write to the Free Software Foundation, Inc.,
+51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+#include "guiBox.h"
+
+GUIBox::GUIBox(gui::IGUIEnvironment *env, gui::IGUIElement *parent, s32 id,
+		core::rect<s32> &rectangle, video::SColor &color) :
+		gui::IGUIElement(gui::EGUIET_ELEMENT, env, parent, id, rectangle),
+		m_color(color)
+{
+}
+
+void GUIBox::draw()
+{
+	if (!IsVisible)
+		return;
+
+	Environment->getVideoDriver()->draw2DRectangle(
+			m_color, AbsoluteRect, &AbsoluteClippingRect);
+
+	IGUIElement::draw();
+}
--- a/src/gui/guiBox.h
+++ b/src/gui/guiBox.h
@@ -1,0 +1,34 @@
+/*
+Minetest
+Copyright (C) 2013 celeron55, Perttu Ahola <celeron55@gmail.com>
+
+This program is free software; you can redistribute it and/or modify
+it under the terms of the GNU Lesser General Public License as published by
+the Free Software Foundation; either version 2.1 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public License along
+with this program; if not, write to the Free Software Foundation, Inc.,
+51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+#pragma once
+
+#include "irrlichttypes_extrabloated.h"
+
+class GUIBox : public gui::IGUIElement
+{
+public:
+	GUIBox(gui::IGUIEnvironment *env, gui::IGUIElement *parent, s32 id,
+			core::rect<s32> &rectangle, video::SColor &color);
+
+	virtual void draw();
+
+private:
+	video::SColor m_color;
+};
--- a/src/gui/guiFormSpecMenu.cpp
+++ b/src/gui/guiFormSpecMenu.cpp
@@ -54,6 +54,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include "util/string.h" // for parseColorString()
 #include "irrlicht_changes/static_text.h"
 #include "client/guiscalingfilter.h"
+#include "guiBackgroundImage.h"
 #include "guiEditBoxWithScrollbar.h"
 #include "intlGUIEditBox.h"
 
@@ -255,14 +256,9 @@ std::vector<std::string>* GUIFormSpecMenu::getDropDownValues(const std::string &
 	return NULL;
 }
 
-v2s32 GUIFormSpecMenu::getElementBasePos(bool absolute,
-		const std::vector<std::string> *v_pos)
+v2s32 GUIFormSpecMenu::getElementBasePos(const std::vector<std::string> *v_pos)
 {
-	v2s32 pos = padding;
-	if (absolute)
-		pos += AbsoluteRect.UpperLeftCorner;
-
-	v2f32 pos_f = v2f32(pos.X, pos.Y) + pos_offset * spacing;
+	v2f32 pos_f = v2f32(padding.X, padding.Y) + pos_offset * spacing;
 	if (v_pos) {
 		pos_f.X += stof((*v_pos)[0]) * spacing.X;
 		pos_f.Y += stof((*v_pos)[1]) * spacing.Y;
@@ -270,18 +266,10 @@ v2s32 GUIFormSpecMenu::getElementBasePos(bool absolute,
 	return v2s32(pos_f.X, pos_f.Y);
 }
 
-v2s32 GUIFormSpecMenu::getRealCoordinateBasePos(bool absolute,
-		const std::vector<std::string> &v_pos)
+v2s32 GUIFormSpecMenu::getRealCoordinateBasePos(const std::vector<std::string> &v_pos)
 {
-	v2f32 pos_f = v2f32(0.0f, 0.0f);
-
-	pos_f.X += stof(v_pos[0]) + pos_offset.X;
-	pos_f.Y += stof(v_pos[1]) + pos_offset.Y;
-
-	if (absolute)
-		return v2s32(pos_f.X * imgsize.X + AbsoluteRect.UpperLeftCorner.X,
-				pos_f.Y * imgsize.Y + AbsoluteRect.UpperLeftCorner.Y);
-	return v2s32(pos_f.X * imgsize.X, pos_f.Y * imgsize.Y);
+	return v2s32((stof(v_pos[0]) + pos_offset.X) * imgsize.X,
+		(stof(v_pos[1]) + pos_offset.Y) * imgsize.Y);
 }
 
 v2s32 GUIFormSpecMenu::getRealCoordinateGeometry(const std::vector<std::string> &v_geom)
@@ -342,6 +330,100 @@ void GUIFormSpecMenu::parseContainerEnd(parserData* data)
 	}
 }
 
+void GUIFormSpecMenu::parseScrollContainer(parserData *data, const std::string &element)
+{
+	std::vector<std::string> parts = split(element, ';');
+
+	if (parts.size() < 3) {
+		errorstream << "Invalid scroll_container start element (" << parts.size()
+				<< "): '" << element << "'" << std::endl;
+		return;
+	}
+
+	std::vector<std::string> v_pos  = split(parts[0], ',');
+	std::vector<std::string> v_geom = split(parts[1], ',');
+	std::string scrollbar_name = parts[2];
+	std::string direction = parts[3];
+	f32 multiplier = 0.1f;
+	if (parts.size() >= 5 && !parts[4].empty())
+		multiplier = stof(parts[4]);
+
+	MY_CHECKPOS("scroll_container", 0);
+	MY_CHECKGEOM("scroll_container", 1);
+
+	v2s32 pos = getRealCoordinateBasePos(v_pos);
+	v2s32 geom = getRealCoordinateGeometry(v_geom);
+
+	// old parent (this)
+	// ↓ is parent of
+	// clipper
+	// ↓ is parent of
+	// mover
+	// ↓ is parent of
+	// other elements
+
+	// make clipper
+	FieldSpec specc(
+		"",
+		L"",
+		L"",
+		258+m_fields.size()
+	);
+
+	if (direction == "down")
+		multiplier *= -imgsize.Y;
+	else if (direction == "right")
+		multiplier *= -imgsize.X;
+
+	core::rect<s32> rectc = core::rect<s32>(pos, pos + geom);
+
+	current_parent = new gui::IGUIElement(EGUIET_ELEMENT, Environment,
+			current_parent, specc.fid, rectc);
+	current_parent->drop();
+
+	m_fields.push_back(specc);
+
+	// make mover
+	FieldSpec specm(
+		"",
+		L"",
+		L"",
+		258+m_fields.size()
+	);
+
+	core::rect<s32> rectm = core::rect<s32>(0, 0, geom.X, geom.Y);
+
+	GUIScrollContainer *mover = new GUIScrollContainer(Environment,
+			current_parent, specm.fid, rectm, direction, multiplier);
+	mover->drop();
+	current_parent = mover;
+
+	m_scroll_containers.emplace_back(scrollbar_name, mover);
+
+	m_fields.push_back(specm);
+
+	// remove interferring offset of normal containers
+	container_stack.push(pos_offset);
+	pos_offset.X = 0.0f;
+	pos_offset.Y = 0.0f;
+}
+
+void GUIFormSpecMenu::parseScrollContainerEnd(parserData *data)
+{
+	if (current_parent == this || current_parent->getParent() == this ||
+			container_stack.empty()) {
+		errorstream << "Invalid scroll_container end element, "
+				<< "no matching scroll_container start element" << std::endl;
+	} else if (pos_offset.getLengthSQ() != 0.0f) {
+		errorstream << "Invalid scroll_container end element, "
+				<< "try to close normal container start element" << std::endl;
+	} else {
+		current_parent = current_parent->getParent()->getParent();
+		pos_offset = container_stack.top();
+		container_stack.pop();
+	}
+}
+
 void GUIFormSpecMenu::parseList(parserData* data, const std::string &element)
 {
 	if (m_client == 0) {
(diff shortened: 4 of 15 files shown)


stujones11:
I see you are creating a number of references using the `new` operator but I do not see where these references are being 'dropped'. This is necessary to free up the memory when the form closes.

DS-Minetest:
Yes, sorry, I'm inexperienced in such stuff. I'll make a function that recursively drops all children, their children, ….

stujones11:
> Please tell me if I made a mistake here

I just did :P

Maybe this helps: http://www.irrlicht3d.org/pivot/entry.php?id=436

DS-Minetest:
[`addChildToEnd`](http://irrlicht.sourceforge.net/docu/_i_g_u_i_element_8h_source.html#l00818) which is used in `IGUIElement`'s constructor grabs the child. Ergo it was needed to drop all elements made with `new`. This is done now.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -323,6 +318,98 @@ void GUIFormSpecMenu::parseContainerEnd(parserData* data)
 	}
 }
 
+void GUIFormSpecMenu::parseScrollContainer(parserData* data, const std::string &element)
+{
+	std::vector<std::string> parts = split(element, ';');
+
+	if (parts.size() >= 3) {
```
SmallJoker:
```C++
if (parts.size() < 3 && m_formspec_version <= FORMSPEC_API_VERSION) {
	errorstream <<  "Invalid scroll_container star..."
	return;
}
```

src/gui/guiFormSpecMenu.cpp
```diff
@@ -323,6 +318,98 @@ void GUIFormSpecMenu::parseContainerEnd(parserData* data)
...
+		f32 multiplier = 0.1f;
+		if (parts.size() >= 5 && !parts[4].empty())
+			multiplier = stof(parts[4]);
+
+		MY_CHECKPOS("scroll_container", 0);
+		MY_CHECKGEOM("scroll_container", 1);
+
+		v2s32 pos = v2s32((stof(v_pos[0]) + pos_offset.X) * imgsize.X,
+			(stof(v_pos[1]) + pos_offset.Y) * imgsize.Y);
+		v2s32 geom = v2s32(stof(v_geom[0]) * imgsize.X, stof(v_geom[1]) * imgsize.Y);
```
SmallJoker: According to the "bla" button's position (example formspec), this scaling does not match the formspec coordinates. Or did I miss anything?

src/gui/guiFormSpecMenu.cpp
```diff
@@ -323,6 +318,98 @@ void GUIFormSpecMenu::parseContainerEnd(parserData* data)
...
+		// remove interferring offset of normal containers
+		container_stack.push(pos_offset);
+		pos_offset.X = 0.0f;
+		pos_offset.Y = 0.0f;
+		return;
+	}
+	errorstream<< "Invalid scroll_container start element (" << parts.size() << "): '" << element << "'"  << std::endl;
+}
+
+void GUIFormSpecMenu::parseScrollContainerEnd(parserData* data)
```
SmallJoker: `parserData *data`

src/gui/guiFormSpecMenu.cpp
```diff
@@ -323,6 +318,98 @@ void GUIFormSpecMenu::parseContainerEnd(parserData* data)
...
+{
+	if (current_parent == this || current_parent->getParent() == this || container_stack.empty()) {
+		errorstream<< "Invalid scroll_container end element, no matching scroll_container start element"  << std::endl;
+	} else if (pos_offset.getLengthSQ() != 0.0f) {
+		errorstream<< "Invalid scroll_container end element, try to close normal container start element"  << std::endl;
```
SmallJoker: Spaces around binary operators (`a << b`)

src/gui/guiFormSpecMenu.cpp
```diff
@@ -681,8 +835,47 @@ void GUIFormSpecMenu::parseBackground(parserData* data, const std::string &eleme
...
+				// clipped to scroll_container => clip to clipper rect
+				const core::rect<s32> &window_rect =
+						current_parent->getParent()->getRelativePosition();
+				core::rect<s32> rect(window_rect.UpperLeftCorner - pos,
+						window_rect.LowerRightCorner + pos);
+				e = Environment->addImage(rect,
+						current_parent->getParent()->getParent(), spec.fid, 0, true);
+			}
+
+			e->setImage(texture);
```
SmallJoker: For safety, please add a `nullptr` check here (`FATAL_ERROR_IF` or similar)

doc/lua_api.txt
```diff
@@ -1964,6 +1964,23 @@ Elements
 * End of a container, following elements are no longer relative to this
   container.
 
+### `scroll_container[<X>,<Y>;<W>,<H>;<scrollbar name>;<direction>;<multiplier>]`
+
+* Start of a scroll_container block, all elements in a scroll_container block are
+  * moved by `(X, Y),
+  * additionally moved by the current value of the scrollbar with the name
+    `scrollbar name` times `multiplier` in the direction `direction` and
+  * clipped to the rectangle defined by `X`, `Y`, `W` and `H`.
```
SmallJoker: Not sure if this is even needed. Textareas clip too, since it's what you'd expect from scrollbar-containers.

doc/lua_api.txt
```diff
@@ -1964,6 +1964,23 @@ Elements
...
+* Possible values for `direction` are `down` and `right`.
+* `multiplier` is optional and defaults to `0.1`.
+* Nesting is possible.
+* Some elements might work a little different if they are in a scroll_container.
```
SmallJoker: What would you like to say here? Different scaling?

src/client/hud.cpp
```diff
@@ -630,6 +630,15 @@ void drawItemStack(video::IVideoDriver *driver,
 		return;
 	}
 
+	// do not draw anything if most of the rect should be clipped
+	// TODO: find a way to clip properly
+	if (clip != nullptr) {
```
SmallJoker: You could use `if (ptr) {` everywhere. Not necessary to check against `nullptr` or `NULL`. Both are interpreted as `false`.

src/gui/guiBox.cpp
```diff
@@ -0,0 +1,20 @@
+
```
SmallJoker: Needs copyright header

src/gui/guiBox.h
```diff
@@ -0,0 +1,16 @@
+
```
SmallJoker: Copyright header here as well.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -353,7 +440,7 @@ void GUIFormSpecMenu::parseList(parserData* data, const std::string &element)
 		else
 			loc.deSerialize(location);
 
-		v2s32 pos = getElementBasePos(true, &v_pos);
```
SmallJoker: Is `pos += AbsoluteRect.UpperLeftCorner;` not necessary here?

src/gui/guiFormSpecMenu.cpp
```diff
@@ -458,7 +562,7 @@ void GUIFormSpecMenu::parseCheckbox(parserData* data, const std::string &element
 
 		spec.ftype = f_CheckBox;
 
-		gui::IGUICheckBox* e = Environment->addCheckBox(fselected, rect, this,
+		gui::IGUICheckBox* e = Environment->addCheckBox(fselected, rect, current_parent,
```
SmallJoker: Touched line. Needs pointer correction.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -636,7 +790,7 @@ void GUIFormSpecMenu::parseButton(parserData* data, const std::string &element,
 		spec.ftype = f_Button;
 		if(type == "button_exit")
 			spec.is_exit = true;
-		gui::IGUIButton* e = Environment->addButton(rect, this, spec.fid,
+		gui::IGUIButton* e = Environment->addButton(rect, current_parent, spec.fid,
```
SmallJoker: Pointer correction needed.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -1564,11 +1782,34 @@ void GUIFormSpecMenu::parseBackgroundColor(parserData* data, const std::string &
...
 			((parts.size() > 2) && (m_formspec_version > FORMSPEC_API_VERSION))) {
-		parseColorString(parts[0], m_bgcolor, false);
-
-		if (parts.size() == 2) {
-			std::string fullscreen = parts[1];
-			m_bgfullscreen = is_yes(fullscreen);
+		if (parts.size() == 1) {
+			if (current_parent == this)
+				parseColorString(parts[0], m_bgcolor, false);
+			else {
```
SmallJoker: Brackets for upper condition required (by code style).

src/gui/guiFormSpecMenu.cpp
```diff
@@ -2506,110 +2795,17 @@ void GUIFormSpecMenu::drawMenu()
 	/*
 		Draw backgrounds
 	*/
-	for (const GUIFormSpecMenu::ImageDrawSpec &spec : m_backgrounds) {
```
SmallJoker: Wouldn't it be possible to use `m_backgrounds`, but with a custom "container rect" as offset?

src/gui/guiFormSpecMenu.cpp
```diff
@@ -3665,6 +3842,12 @@ bool GUIFormSpecMenu::OnEvent(const SEvent& event)
 			}
 		}
 
+		if (event.GUIEvent.EventType == gui::EGET_SCROLL_BAR_CHANGED) {//hier
```
SmallJoker: // redundant comment

src/gui/guiItemImage.cpp
```diff
@@ -0,0 +1,43 @@
...
+		gui::IGUIFont *font, Client *client) :
+		gui::IGUIElement(gui::EGUIET_ELEMENT, env, parent, id, rectangle),
+		m_item_name(item_name), m_font(font), m_client(client),
+		m_label(core::stringw())
+{
+}
+
+void GUIItemImage::draw()
+{
+	if (m_client == 0) {
```
SmallJoker: `if (!m_client)` comparing a pointer with `0` works but is unusual.

src/gui/guiItemImage.cpp
```diff
@@ -0,0 +1,43 @@
+
+#include "guiItemImage.h"
```
SmallJoker: Header notice missing

src/gui/guiItemImage.h
```diff
@@ -0,0 +1,28 @@
+
+#pragma once
```
SmallJoker: Header notice missing

src/gui/guiScrollContainer.cpp
```diff
@@ -0,0 +1,38 @@
+
+#include "guiScrollContainer.h"
```
SmallJoker: Header notice missing

src/gui/guiScrollContainer.h
```diff
@@ -0,0 +1,25 @@
+
+#pragma once
```
SmallJoker: Header notice missing

SmallJoker:
Background or container side-effect:

grafik

rubenwardy:
The background there is actually a box element

src/gui/guiFormSpecMenu.cpp
DS-Minetest: There's already a `if (texture != nullptr) {` before, so I guess, this is about `e`.

doc/lua_api.txt
DS-Minetest: Normal containers do not clip. So I think, it makes sense pointing this out.

doc/lua_api.txt
DS-Minetest: For example `bgcolor[color]` doesn't set the background of the whole formspec window but only of the scroll container. Also some elements that normally are never clipped are clipped in scroll_containers. Everything is intuitive, however, some people might get confused if they don't look whether the element is inside a scroll_container and expect the same to happen.

src/gui/guiBox.cpp
DS-Minetest: Should I just copy it from another file?

src/gui/guiFormSpecMenu.cpp
DS-Minetest: No, absolute positions aren't necessary anywhere anymore as every element is now a child and as such has the `AbsoluteRect` of its parent as offset (if parent is `this`, the offset will still be `AbsoluteRect.UpperLeftCorner`).

src/gui/guiFormSpecMenu.cpp
DS-Minetest: What would be the benefit of this?

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Sorry, that was a marker to jump around in code.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Done.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: This is the same as in #8524. 
That PR will very probably be merged before this one. Supporting the old system with new elements doesn't make sense imo. Modders that use a new element have `real_coordinates` available and will use it.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Done.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Done.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Done.

doc/lua_api.txt
```diff
@@ -1964,6 +1964,23 @@ Elements
 * End of a container, following elements are no longer relative to this
   container.
 
+### `scroll_container[<X>,<Y>;<W>,<H>;<scrollbar name>;<direction>;<multiplier>]`
+
+* Start of a scroll_container block, all elements in a scroll_container block are
+  * moved by `(X, Y),
```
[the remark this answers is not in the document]

src/client/hud.cpp
DS-Minetest: Done.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Done.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Done.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Done.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Done.

src/gui/guiItemImage.cpp
DS-Minetest: Done.
(This was copied from the old code.)

DS-Minetest:
Thanks for the review!

src/gui/guiFormSpecMenu.cpp
SmallJoker: Yes, I meant `e`. Just in case the code is either touched or Irrlicht cannot add the image for some reason.

src/gui/guiBox.cpp
SmallJoker: Yes, or directly from the source: https://www.gnu.org/licenses/old-licenses/lgpl-2.1.txt

src/gui/guiFormSpecMenu.cpp
SmallJoker: It would mean that at least these controls are drawn in the same function, separated from the parsing code. However, not many other elements do it this way so I guess it's OK to keep this change.

DS-Minetest:
The box is now fixed. The main menu looks normal again.

src/gui/guiBox.cpp
DS-Minetest: If so, done.

DS-Minetest:
I've just copied the copyright header from `guiFormSpecMenu.h` now. (I'm not sure if this is correct.)

Note that in the `fix code style and improve documentation` I've also added to the documentation a hint in what order the elements are drawn. This can be given now that there aren't anymore so many elements that are drawn manually.

src/gui/guiScrollContainer.h
```diff
@@ -0,0 +1,43 @@
...
+			f32 scrollfactor);
+
+	virtual bool OnEvent(const SEvent &event);
+
+	void OnScrollEvent(gui::IGUIElement *caller);
+
+	inline void setScrollBar(guiScrollBar *scrollbar) { m_scrollbar = scrollbar; }
+
+private:
+	guiScrollBar *m_scrollbar;
```
stujones11: Pointer should be initialized with `nullptr`, preferably in the ctor

DS-Minetest: Done.

DS-Minetest:
I've rebased.

src/gui/guiBox.cpp
DS-Minetest: Is it correct how it is currently?

DS-Minetest:
`regster_on_receive_fields` is tested now, too. (I did not test every single element.) Dropdowns and scrollbars didn't work properly (also checkboes, fields, ...) because it was not searched for their ids in the children. This is fixed now.
And you were able to see rect tooltips by hovering on the border of the scroll container. This is fixed now, too.

stujones11:
I am still a little uncertain about the way you handling the reference counting. It is not really goo practice to be reusing these refs once you have dropped them, which you are in a number of cases.

When you `drop` reference in Irrlicht, you the then set its reference count to zero (assuming it has not been 'grabbed' elsewhere). This effective says it is safe for removal but this is not done right away since it is still attached to the scene graph via the the gui environment.

I have a sneaking suspicion that Android drop the current scene graph when entering the lock screen but have not had a chance to test this there, yet.

I would prefer to see the references dropped in the formspec destructor, as is done for tables. However if the MT devs are happy with your methods and it does not cause a segfault then you may just ignore concerns :)

DS-Minetest:
I've squashed and rebased again.

pyrollo:
Hi DS-Minetest, that's a very good idea and good work!

For some other devs, I got the same problem as you got with item_image_buttons.

You may be interested in that PR : https://github.com/minetest/minetest/pull/8652

DS-Minetest:
> Hi DS-Minetest, that's a very good idea and good work!

Thanks! ^_^


Nice PR that you've got there! Sadly I don't really understand much about this.

pyrollo:
>  Sadly I don't really understand much about this.
;)
Only important thing to catch is that it provides a good answer to this TODO:
https://github.com/minetest/minetest/commit/a3182b899bdaf6f85867191afa8081809134aea1#diff-0d07b27d5a432ddb5b488a3c3f2a0dd0R634

DS-Minetest:
Rebased:
2019-07-26-230444_1920x1080_scrot
Thanks, @pyrollo!

DS-Minetest:
I've copied the change of bgcolor to #8716 to give the possibility to make this PR less bloated.

SmallJoker:
@DS-Minetest The other PR was merged. Needs rebase.

DS-Minetest:
I've made a new PR, rebasing would have been too difficult, probably. See #9101.

PR #8603, opened 2019-06-15
Mods: Improve mod loading priorities

### LINKED ISSUE #8601: Optional dependencies being overridden/turned into hard dependencies
This issue is very similar to #3136
(very nearly a duplicate, in fact, but that one's old, and long since closed)

Using 5.0.1-release, singleplayer

Consider this example dependency tree:

Mod A hard depends on mod C.  
B hard depends on D, optional-depends on F
C depends on nothing
D optional-depends on E
E hard depends on F

Now disable mod F.  Let's say you're doing this to test a feature in B that changes behavior depending on the presence of F.

In this case, E of course doesn't load, but D won't load either, because now the engine treats it as if it *hard-depends* on E, contrary to its `depends.txt`/`mod.conf`.  This causes B not to load as well, since it needs D.

So while one expects A, B, C, and D to load in this example, only A and C get loaded, thus preventing the modder from doing the test they wanted (to check B's behavior without F).

OK, now try it in a real setup:

* Download Dreambuilder
* Create a new world,
* Enable the whole modpack in that world's config
* Now disable just `mesecons` (the one base mod from that project), leaving everything else enabled
* Start the world.
* Many mods that should have loaded won't, for example `homedecor_lighting` (it has a number of dependencies, but mesecons is strictly optional)

Now, go back into the world config and disable all of the `mesecons*` mods (there are about 30, all grouped together), then save and start the world.  You'll find that many of the items that should have loaded the first time work now, even though you were turning things *off* in the config.

It's as though hard dependencies propagate upwards through the dependency tree once a mod can't be satisfied, overriding cases where mods are explicitly set to be optional.

SmallJoker:
Fixes #8601

## How it works
_Documentation in the code™_
There is now a new list to determinate how many mods depend on the mod:
   * deps_chain: Contains all mod names which either need or may need the mod

1) Dependencies are read from the mod's config
2) Mods with incomplete hard-dependencies are disabled (same as before)
3) Each mod adds itself to the list of each mod it needs
   * ...and then adds itself to the list of that mod's dependencies and so on
4) Sort the final list: Sort the `ModSpec`s according to the amount of mods depending on this mod
5) Load the mods

FYI: The code was last modified 7 years ago, in 6af8a34. Apparently it worked quite reliable so far.

## To do

- [x] Some local tests
- [ ] Submit YOUR mod dependency issues, or test the PR to see whether it works as expected

## How to test

1) Download [`mod_sorting_pack.zip`](https://github.com/minetest/minetest/files/3293490/mod_sorting_pack.zip)
2) Enable all mods inside the "Mod sorting modpack"
3) Join the world (maybe check `verbose` output)
4) Test with your own complicated dependency issues

### DIFF
--- a/src/content/mods.cpp
+++ b/src/content/mods.cpp
@@ -333,51 +333,119 @@ void ModConfiguration::checkConflictsAndDeps()
 	resolveDependencies();
 }
 
+struct PrioritySortedMod
+{
+	PrioritySortedMod(ModSpec *mod) : spec(mod) {}
+
+	static bool sorter(const PrioritySortedMod *a, const PrioritySortedMod *b)
+	{
+		// Return true means: move 'a' towards begin() in std::sort
+		return a->deps_chain.size() > b->deps_chain.size();
+	}
+
+	void addDependingMod(const std::string &modname,
+			std::unordered_map<std::string, PrioritySortedMod> &mods_by_name);
+
+	std::unordered_set<std::string> deps_chain;
+	ModSpec *spec;
+};
+
+void PrioritySortedMod::addDependingMod(const std::string &modname,
+		std::unordered_map<std::string, PrioritySortedMod> &mods_by_name)
+{
+	if (deps_chain.find(modname) != deps_chain.end()) {
+		// clang-format off
+		if (modname == spec->name)
+			warningstream << "PrioritySortedMod::addDependingMod(): Detected" <<
+					" circular dependencies in mod '" << modname << "'" << std::endl;
+		// clang-format on
+		return;
+	}
+
+	deps_chain.emplace(modname);
+
+	for (const std::string &dependency : spec->depends) {
+		auto &mod = mods_by_name.find(dependency)->second;
+		mod.addDependingMod(modname, mods_by_name);
+	}
+
+	for (const std::string &dependency : spec->optdepends) {
+		auto &mod = mods_by_name.find(dependency)->second;
+		mod.addDependingMod(modname, mods_by_name);
+	}
+}
+
 void ModConfiguration::resolveDependencies()
 {
-	// Step 1: Compile a list of the mod names we're working with
-	std::set<std::string> modnames;
-	for (const ModSpec &mod : m_unsatisfied_mods) {
-		modnames.insert(mod.name);
+	std::unordered_map<std::string, PrioritySortedMod> mods_by_name;
+	for (ModSpec &mod : m_unsatisfied_mods) {
+		// Reduce memory copy time by passing 'mod' as pointer
+		// 'm_unsatisfied_mods' owns the value!
+		mods_by_name.emplace(std::pair<std::string, PrioritySortedMod>(
+				mod.name, &mod));
 	}
 
-	// Step 2: get dependencies (including optional dependencies)
-	// of each mod, split mods into satisfied and unsatisfied
-	std::list<ModSpec> satisfied;
-	std::list<ModSpec> unsatisfied;
-	for (ModSpec mod : m_unsatisfied_mods) {
-		mod.unsatisfied_depends = mod.depends;
-		// check which optional dependencies actually exist
-		for (const std::string &optdep : mod.optdepends) {
-			if (modnames.count(optdep) != 0)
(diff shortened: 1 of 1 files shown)


src/content/mods.cpp
```diff
@@ -333,51 +333,119 @@ void ModConfiguration::checkConflictsAndDeps()
...
 {
-	// Step 1: Compile a list of the mod names we're working with
-	std::set<std::string> modnames;
-	for (const ModSpec &mod : m_unsatisfied_mods) {
-		modnames.insert(mod.name);
+	PrioritySortedMod(ModSpec *mod) : spec(mod) {}
+
+	static bool sorter(const PrioritySortedMod *a, const PrioritySortedMod *b)
+	{
+		// Return true means: move towards begin() in std::sort
```
DS-Minetest:
```suggestion
		// Return true means: move a towards begin() in std::sort
```

DS-Minetest:
```suggestion
		// Return true means: move a towards begin() in std::sort
```

SmallJoker:
@DS-Minetest Removed the duplicated change suggestion. Were you able to get any impression on the PR's functionality?

DS-Minetest:
> Removed the duplicated change suggestion.

Sorry, github didn't show it as commented, so I pressed again.

> Were you able to get any impression on the PR's functionality?

Yes, indeed.
First, all mods are checked whether they exist, if not (or if dependencies are missing), they are _removed_. This is repeated until there is no dependency that could not be satisfied.
Then, for each mod all mods, that depend on the mod, are recursively saved.
Therefor, if any mod A depends on another mod B, B has at least one more mods that depend on it than A (all mods that depend on A + 1 (B) + all mods that depend on B but not on A).
Hence, the mods can be sorted for the number of mods that depend on them. This is done and should work.

However, I'm not sure if there aren't other ways to do this that are simpler and more efficient.

SmallJoker:
@DS-Minetest Yes, that's exactly how the PR works. Unless there are circular dependencies I cannot find any side-effects or misbehaviours with this approach. There are surely other ways, but proper sorting will always involve getting the priority of a mod which happens to be the amount of mods which depend on it. 
Maybe there's somewhere a way to optimize this (caching chain parts?), but I believe the straight-forward method here should do its job just fine. Efficiency is not a major concern since this sorting only happens once when loading a new world.

src/content/mods.cpp
```diff
@@ -333,51 +333,119 @@ void ModConfiguration::checkConflictsAndDeps()
...
+};
+
+void PrioritySortedMod::addDependingMod(const std::string &modname,
+		std::unordered_map<std::string, PrioritySortedMod> &mods_by_name)
+{
+	if (deps_chain.find(modname) != deps_chain.end()) {
+		// clang-format off
+		if (modname == spec->name)
+			warningstream << "PrioritySortedMod::addDependingMod(): Detected" <<
+					" circular dependencies in mod '" << modname << "'" << std::endl;
```
DS-Minetest: It might be good to stop the entire mod loading process if there are circular dependencies. Else it might be that a mod is loaded but not the mod it depends on.
Example:
A depends on B, B depends on A. Both mods do exist, ergo they will both be loaded as they are not removed in step 1. In step 2 `addDependingMod` is used. A and B each have only one mod in their `deps_chain` set (themselves). Hence one of them is loaded before the other. This mustn't happen.

SmallJoker:
@DS-Minetest Thanks for your feedback. The new commit now integrates the circular dependency handling into the dependency solving loop. The problem is now solved, but the code a _little bit_  more complicated.

SmallJoker:
Found pipeworks to be a special case. It assumes that `screwdriver` is always present (it was), hence does not include it in the dependency list.

DS-Minetest:
Code still looks good.

Pipeworks should probably just add `screwdriver` as optional dependency.

DS-Minetest:
Btw. #8059 seems related.

SmallJoker:
@DS-Minetest Thanks for pointing that out. It will be entirely compatible with this PR, no conflicts or whatsoever.

PR #8623, opened 2019-06-24
CAO: Add Colored Text Sprite

EvidenceBKidscode:
This PR offered by [Kidscode](https://kidscode.com) allows to create entity sprites with **native** colored text.

Demo: https://www.youtube.com/watch?v=to2BMSdGwlo

## Documentation
<details>
<summary>Click to expand</summary>

#### Sample
```Lua
initial_properties = {
	visual = "text_sprite",
	visual_size = {x = 1.0, y = 1.0},
	mesh = "{orange}{1.5}K{1}idscode rocks\n{#0FF/#00F}Érîthréà\n{#ABCDEF/#456789}Eûphôriâ\n\nÉqt{0.5}Éqt{1.2}Éqt\n{orange/red}\\\\\\{\\^\\}^0.1;0.2;0.0;0.0;0.0;1.0;0.22",
	colors = { "#F080F0FF", "#F01020FF", "#80808080", "#40404080" },
},

...
	
obj:set_sprite_text(""\\\\\\{\\^\\}\n\n{orange}Kidscode\n{#0FF/#00F}Érîthréà\n{#ABCDEF/#456789}Eûphôriâ");
```

#### Parameters

- `visual_size`: X scale, Y scale
- `mesh`: text, border size, X padding, Y padding, X offset, Y offset, spacing, base line offset
- `colors`: text top color, text bottom color, background color, border color

The text separator is the circumflex.

The value separator is the semicolon.

#### Values

- `border size`: tells the size of the border (in character height unit)
- `X/Y padding`: tells how much padding is added between the text and the border (in character size unit)
- `X/Y offset`: tells how much offset is added to the text position inside the frame (in character size unit)
- `spacing`: tells how much space is added between contiguous characters (`1.0` by default)
- `base line offset`: tells where the base line is located (from `0.0` to `1.0`)

The actual text frame width is computed from the frame height and longest text line width.

The text can contains named or hexadecimal colors between accolades.
Hexadecimal colors start by a hash mark, and can contain 3 or 6 hexadecimal digits.
Both the top and bottom colors can be provided, by separating them with a slash character.

The text can also contain scales between accolades.
The default scale is `1.0`.

Special characters can be escaped by preceding them with a backslash character.
</details>


SmallJoker:
So the current trend is to modify & migrate all Irrlicht sources to our code base?
Don't get me wrong - entities with text that do not require a bitmap font (and a mod to "render") is a great idea and looks amazing. How much of the Irrlicht code was changed to achieve this?

DS-Minetest:
Maybe related: #1367

src/irrlicht_changes/CTextSpriteSceneNode.cpp
```diff
@@ -0,0 +1,925 @@
...
+	bottomColor = BottomColor;
+}
+
+//! Parses an hexadecimal color
+void CTextSpriteSceneNode::parseColor(const std::string& color_string,
+	video::SColor& color, f32& scale)
+{
+	static std::map<std::string, u32> colors;
+	
+	if (colors.size() == 0)
```
SmallJoker: Can be imported from `string.cpp` -> make `named_colors` a global variable

src/client/content_cao.cpp
```diff
@@ -506,6 +511,10 @@ void GenericCAO::removeFromScene(bool permanent)
 		m_spritenode->remove();
 		m_spritenode->drop();
 		m_spritenode = nullptr;
+	} else if (m_textspritenode) {
+		m_textspritenode->remove();
+		m_textspritenode->drop();
+		m_textspritenode = NULL;
```
ClobberXD: If I'm not mistaken, shouldn't this be `delete m_textspritenode` instead?

SmallJoker: No, the pointer is owned by Irrlicht.

src/util/string.cpp
```diff
@@ -467,6 +467,27 @@ u64 read_seed(const char *str)
 	return num;
 }
 
+void parseTextString(const std::string &value, std::string &text, std::string &params,
```
SmallJoker: Leave the declaration here and keep the definition in the source file. Headers should be lightweight as possible to for faster compiling.
```C++
void parseTextString(const std::string &value, std::string &text, std::string &params,
		const char sep, const char esc);
```

src/util/string.h
```diff
@@ -85,9 +85,166 @@ std::string writeFlagString(u32 flags, const FlagDesc *flagdesc, u32 flagmask);
 size_t mystrlcpy(char *dst, const char *src, size_t size);
 char *mystrtok_r(char *s, const char *sep, char **lasts);
 u64 read_seed(const char *str);
+void parseTextString(const std::string &value, std::string &text, std::string &params, 
+		const char sep, const char esc='\\');
 bool parseColorString(const std::string &value, video::SColor &color, bool quiet,
 		unsigned char default_alpha = 0xff);
 
+struct ColorContainer {
```
SmallJoker: Same here. Struct declaration is OK in the header, but the definition should really be in `string.cpp`.

src/client/content_cao.cpp
ClobberXD: Ah ok, thanks

SmallJoker:
What kind of color/size encoding is this? It doesn't look like anything known in Minetest.
There are text colorizing functions for both - foreground and background (latter kinda broken). Also HTML tags are now a thing. Why did you decide on this brackets-value notation?

celeron55:
http://irc.minetest.net/minetest-dev/2020-03-06#i_5648818

PR #8630, opened 2019-06-26
Add Full Screen Sized FormSpecs

v-rob:
This PR give FormSpecs the ability to be the full size of the screen instead of a box in the center.  This solves problems like https://github.com/minetest/minetest/issues/6733#issuecomment-506000467, which says that FormSpecs (the main menu in this case) are crammed into a small box.  It requires `real_coordinates` to be enabled (https://github.com/minetest/minetest/pull/8524 is required).

The syntax uses the third parameter of the `size` element, like so: `size[5,5,stretch]`.  The original behavior of `true` and `false` for the third parameter for fixed size is untouched for compatibility.  On older clients, this will appear to be a normal scaled FormSpec.

When stretching is enabled, coordinates will be stretched to fit the FormSpec size.  For instance, if the FormSpec uses `size[10,5,stretch]`, the X coordinates will be a tenth of the screen, and elements one wide will be a tenth of the screen.  Likewise, the Y coordinates will be a fifth of the screen.

### Screenshots

Replace X and Y with the numbers above the screenshots
```
size[<X>,<Y>,stretch]
real_coordinates[true]
box[0,0;0.5,0.5;#000000]
box[0.5,0.5;0.5,0.5;#FFFFFF]
```

X, Y = 1

X, Y = 10

Notice how in the second screenshot, there is a slight border at the top and bottom.  This is because the screen might not divide evenly, especially with larger numbers, so a size of `1,1` is recommended or the background should be removed.  If I did make the form fill the whole screen, there would be small inaccuracies with the coordinates.

This PR is Ready for Review.

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -1875,9 +1875,14 @@ is used when the server receives user input. You must not use the name
 Spaces and newlines can be inserted between the blocks, as is used in the
 examples.
 
-Position and size units are inventory slots, `X` and `Y` position the formspec
-element relative to the top left of the menu or container. `W` and `H` are its
-width and height values.
+Position and size units are inventory slots unless the new coordinating system
+is enabled. `X` and `Y` position the formspec element relative to the top left
+of the menu or container. `W` and `H` are its width and height values.
+
+If the new system is enabled, all elements have unified cooridinates for all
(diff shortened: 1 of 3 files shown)


DS-Minetest:
This looks great!

rubenwardy:
image

PR #8640, opened 2019-06-29
 Trigger on_place in many situations even if prediction failed 

### LINKED ISSUE #8517: `on_place` is not called when trying to place a node into another node
##### Issue type
 - Bug report
 
##### Minetest version
```
Minetest 5.1.0-dev-e32a630 (Linux)
Using Irrlicht 1.8.4
BUILD_TYPE=Release
RUN_IN_PLACE=1
USE_GETTEXT=1
USE_SOUND=1
USE_CURL=1
USE_FREETYPE=1
USE_LUAJIT=0
STATIC_SHAREDIR="."
```

##### Summary
If you set `on_place` in a node's node definition, the callback is not called if there's already a node where the new node would be.

##### Steps to reproduce
- Register a **node** with `on_place = function() minetest.chat_send_all("on_place") end,`.
- In minetest place a full node and next to it a slab or something similar.
- Move you point to the full node on the side at which the other node is.
It should look like this:
```
/-----\
|     |←you look at this face
|     |/-----\
\-----/\-----/
```
- Rightclick.
- Look into chat and see that there's no message.
- Try the same with an item that is not a node and get a message.

DS-Minetest:
- ~~Note that there are two commits, the first is just for removal of an indentation level.~~
- Fixes #8517.
- Some other things that were probably not reported yet are fixed, too. Eg. `on_place` is triggered now if you want to place with a torch to a rail.

## Detailed description

The function `Game::nodePlacementPrediction` is renamed to / replaced with `Game::nodePlacement`. `nodePlacement` reports the interact (`INTERACT_PLACE`) to the server (triggering `on_rightclick` and/or `on_place`), chooses the right sound and does the placement prediction (the prediction only happens in the same situations as before). If the prediction was made (a node was placed client-side), the function returns `true` and the csm callback is called.

## To do

(Use split mode for reviewing.)

## How to test

<details><summary>I've used a node with many callbacks. (Much of this is not needed.)</summary>

```lua
minetest.register_node("test_callbacks:node", {
	description = "callback node",
	tiles = {"default_wood.png^heart.png"},
	groups = {choppy = 3, oddly_breakable_by_hand = 3},

	on_construct = function(pos)
		minetest.chat_send_all("on_construct")
	end,
	after_place_node = function(pos, placer, itemstack, pointed_thing)
		minetest.chat_send_all("after_place_node")
	end,
		minetest.chat_send_all("on_rightclick")
	end,
	on_place = function(itemstack, placer, pointed_thing)
		minetest.chat_send_all("on_place")
		if placer and placer:is_player() and placer:get_player_control().aux1 then
			return minetest.item_place(itemstack, placer, pointed_thing, 0)
		end
	end,
	on_secondary_use = function(itemstack, user, pointed_thing)
		minetest.chat_send_all("on_secondary_use")
	end,
})
```

</details>

### DIFF
--- a/src/client/game.cpp
+++ b/src/client/game.cpp
@@ -801,8 +801,8 @@ private:
 
 	void updateChat(f32 dtime, const v2u32 &screensize);
 
-	bool nodePlacementPrediction(const ItemDefinition &selected_def,
-		const ItemStack &selected_item, const v3s16 &nodepos, const v3s16 &neighbourpos);
+	bool nodePlacement(const ItemDefinition &selected_def, const ItemStack &selected_item,
+		const v3s16 &nodepos, const v3s16 &neighbourpos, const PointedThing &pointed);
 	static const ClientEventHandler clientEventHandler[CLIENTEVENT_MAX];
 
 	InputHandler *input = nullptr;
@@ -1083,7 +1083,7 @@ void Game::run()
 		//    RenderingEngine::run() from this iteration
 		//  + Sleep time until the wanted FPS are reached
 		limitFps(&draw_times, &dtime);
-		
+
 		// Prepare render data for next iteration
 
 		updateStats(&stats, draw_times, dtime);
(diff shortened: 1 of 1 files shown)


DS-Minetest:
Rebased.

Ugh, it is awful how this actually pretty small change looks so big.

I should probably describe the changes:
The function `nodePlacementPrediction` (It is a function that is only called at one place.) has a new parameter `bool *send_interact`. Every time before returning, the `bool` to where this parameter points is set to `true` or `false`. It determines whether the placement should also be sent as interact. There are some situations where this shouldn't happen. Eg. when the player wants to place the node into himself or if the placement position is invalid. In these cases, (if there isn't a prediction either, which is mostly the case,) the place_failed sound is played. Before this PR, there were only two special situations where interact is sent even though no placement prediction was made, an empty placement prediction or a rightclickable node.

It might be easier to use the split mode for reviewing.

src/client/game.cpp
```diff
@@ -3281,36 +3282,36 @@ void Game::handlePointingAtNode(const PointedThing &pointed,
 
 			// If the wielded item has node placement prediction,
 			// make that happen
+			bool send_interact;
```
SmallJoker: Needs default value for safety.

DS-Minetest:
Thanks for reviewing!
I've done your second suggestion (sounds are set in the function, too) and renamed the function.

DS-Minetest:
Rebased.

DS-Minetest:
Rebased.

src/client/game.cpp
```diff
@@ -3266,144 +3251,162 @@ bool Game::nodePlacementPrediction(const ItemDefinition &selected_def,
...
-		if (is_valid_position)
-		{
-			if (nodedef->get(n_under).buildable_to)
-				p = nodepos;
-			else {
-				node = map.getNode(p, &is_valid_position);
-				if (is_valid_position &&!nodedef->get(node).buildable_to)
-					return false;
+		// Report to server
+		client->interact(INTERACT_PLACE, pointed);
```
sfan5: since it needs to do this anyway, why is this not a the top-level of the function?

DS-Minetest: It isn't done anyway, see https://github.com/minetest/minetest/pull/8640/files#diff-7f71a92de20504e8a5636b4317ea67b2L3270, https://github.com/minetest/minetest/pull/8640/files#diff-7f71a92de20504e8a5636b4317ea67b2R3400 and https://github.com/minetest/minetest/pull/8640/files#diff-7f71a92de20504e8a5636b4317ea67b2R3408.

sfan5: Hmm ok, I'm not 100% happy with the code but it should be fine.

DS-Minetest:
Thanks!
But maybe should I have moved the [meta formspec code](https://github.com/DS-Minetest/minetest/blob/873ac56e6c49a067c3e51bb0362da4d9a72def46/src/client/game.cpp#L3202-L3222) into the new function. :thinking: (Maybe I'll open a new PR for that.)

PR #8641, opened 2019-06-29
Clean up and fix freetype=false crashes

SmallJoker:
A IGUIFont of type bitmap/vector cannot be converted to CGUITTFont
Fixes various segfaults in gameplay
Shorter font cache code, cleaned up (?)

This was broken since colored text became a thing. Tracking down this bug took a horrible long time, but now it finally works.

**Test 1: USE_FREETYPE=1, `freetype = true`**
Compiled-in TTF support. Same functionality as in master.
1) Modify `font_path` and/or `mono_font_path` to invalid paths
2) Minetest loads the fallback path, or the setting default if everything else fails

**Test 2 Bitmap: USE_FREETYPE=1, `freetype = false`**
TTF support, but using images. Works in PR, does not in master.
1) Get the font images from https://github.com/minetest/minetest/tree/25945dc5/fonts (removed in dceb9f7)
2) Adjust `font_path` and `mono_font_path`
3) Minetest starts, no text colors but it runs

**Test 3 Vectors: USE_FREETYPE=1, `freetype = false`**
TTF support, but using xml files. Works in PR, does not in master.
1) `font_path = fonts/mono_dejavu_sans.xml`
2) `mono_font_path = fonts/mono_dejavu_sans.xml`
3)  Minetest starts, no text colors but it runs

**USE_FREETYPE=0**
Requires many guiFormSpecMenu changes, thus it should be fixed in a later PR. There's yet no client option without freetype built-in.

### DIFF
--- a/src/client/fontengine.cpp
+++ b/src/client/fontengine.cpp
@@ -55,36 +55,7 @@ FontEngine::FontEngine(Settings* main_settings, gui::IGUIEnvironment* env) :
 	assert(m_env != NULL); // pre-condition
 	assert(m_env->getSkin() != NULL); // pre-condition
 
-	m_currentMode = FM_Simple;
-
-#if USE_FREETYPE
-	if (g_settings->getBool("freetype")) {
-		m_default_size[FM_Standard] = m_settings->getU16("font_size");
-		m_default_size[FM_Fallback] = m_settings->getU16("fallback_font_size");
-		m_default_size[FM_Mono]     = m_settings->getU16("mono_font_size");
-
-		if (is_yes(gettext("needs_fallback_font"))) {
-			m_currentMode = FM_Fallback;
-		}
-		else {
-			m_currentMode = FM_Standard;
-		}
-	}
-
-	// having freetype but not using it is quite a strange case so we need to do
-	// special handling for it
-	if (m_currentMode == FM_Simple) {
-		std::stringstream fontsize;
-		fontsize << DEFAULT_FONT_SIZE;
-		m_settings->setDefault("font_size", fontsize.str());
-		m_settings->setDefault("mono_font_size", fontsize.str());
-	}
-#endif
-
-	m_default_size[FM_Simple]       = m_settings->getU16("font_size");
-	m_default_size[FM_SimpleMono]   = m_settings->getU16("mono_font_size");
-
-	updateSkin();
+	readSettings();
 
 	if (m_currentMode == FM_Standard) {
 		m_settings->registerChangedCallback("font_size", font_setting_changed, NULL);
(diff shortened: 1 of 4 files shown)


src/client/fontengine.cpp
```diff
@@ -260,81 +222,55 @@ void FontEngine::updateFontCache()
...
-			font_config_prefix = "";
-
+			break;
 	}
 
-	if (m_font_cache[mode].find(basesize) != m_font_cache[mode].end())
-		return;
-
-	if ((mode == FM_Simple) || (mode == FM_SimpleMono)) {
-		initSimpleFont(basesize, mode);
```
SmallJoker: Moved to `getFont` for a saner call structure.

src/client/fontengine.cpp
```diff
@@ -260,81 +222,55 @@ void FontEngine::updateFontCache()
...
-			font_shadow_alpha =
-					g_settings->getU16(font_config_prefix + "font_shadow_alpha");
-		} catch (SettingNotFoundException&) {}
+	u16 font_shadow       = 0;
+	u16 font_shadow_alpha = 0;
+	g_settings->getU16NoEx(setting_prefix + "font_shadow", font_shadow);
+	g_settings->getU16NoEx(setting_prefix + "font_shadow_alpha", font_shadow_alpha);
 
-		std::string font_path = g_settings->get(font_config_prefix + "font_path");
+	std::string fallback_settings[] = {
```
SmallJoker: Same functionality, just shorter code.

src/client/fontengine.cpp
```diff
@@ -439,59 +324,35 @@ void FontEngine::initSimpleFont(unsigned int basesize, FontMode mode)
...
-				break;
-			}
-		}
+	irr::gui::IGUIFont *font = nullptr;
+	std::string font_extensions[] = { ".png", ".xml" };
 
-		std::stringstream fontsize_plus_xml;
-		fontsize_plus_xml << basename << "_" << (size + offset) << ".xml";
+	// Find nearest matching font scale
+	// Does a "zig-zag motion" (positibe/negative), from 0 to MAX_FONT_SIZE_OFFSET
```
SmallJoker: Also same functionality here. Positive offset it checked first, then goes from 0 to -MAX_FONT_SIZE_OFFSET and +MAX_FONT_SIZE_OFFSET

src/client/fontengine.h
```diff
@@ -112,15 +112,6 @@ class FontEngine
 	/** current font engine mode */
 	FontMode m_currentMode = FM_Standard;
 
-	/** font mode of last request */
```
SmallJoker: Replaced by replacing several `m_font_cache[mode]` lookups in `getFont()`. Roughly the same performance, I guess.

MoNTE48:
Forgive my incompetence. Any usage scenario? MT is built with freetype support, but the user turns off freetype. What for? This should not work in principle.

SmallJoker:
@MoNTE48 Using bitmaps to render text gives you a few FPS more on old (real old) hardware. This PR is just one of the performance tweaks I wrote to make the game playable even on my worst rig.

SmallJoker:
Comparison: Freetype enabled on the right side.
GPU: Quadro FX 4600 (nouveau)
CPU: Core 2 Quad Q6700
Full render size: 1280x978 (image cut)
comparison

MoNTE48:
Wow, I didn’t know that Freetype reduced performance so much. Try this font, please. Just wondering
https://github.com/MultiCraftProject/MultiCraft/blob/DEV/fonts/Retron2000.ttf

SmallJoker:
50% scaled for uploading. Same rendering resolution as before.
grafik

MoNTE48:
Relative gains from using bitmaps is minimal. By the way, why does MT use DroidSansFallbackFull.ttf? There is DroidSansFallback.ttf, which is less than 1 MB, but has no differences, as far as I could find in Google.
[offtop]Also MT could use the system font. Many fonts in Windows, Linux, Android are required. for example, MultiCraft uses the Android system font, which reduced the APK by 5 MB compared to Minetest. Never heard of any problems.[/offtop]

SmallJoker:
@MoNTE48 I updated the older post with some comparisons I made. What matters is not the file size, but the amount of characters drawn on screen.

MoNTE48:
@SmallJoker I understand, thank you.
The second part of my previous post was not about performance, but about file size. Why increase the size of the game if can not do this? Even if it does not affect performance. If I am not mistaken, MultiCraft for Android, along with analytics, advertising and other additional libs, took up less space than Minetest without all this.

src/client/fontengine.cpp
```diff
@@ -439,59 +325,35 @@ void FontEngine::initSimpleFont(unsigned int basesize, FontMode mode)
...
+		s32 sign = (zoffset & 1) ? -1 : 1;
+		s32 offset = zoffset >> 1;
 
-			if (font) {
-				verbosestream << "FontEngine: found font: " << fontsize_plus_xml.str() << std::endl;
-				break;
-			}
-		}
+		for (const std::string &ext : font_extensions) {
+			path.str(std::string()); // Clear
```
sfan5: `path.str("")` would be shorter

src/client/fontengine.cpp
```diff
@@ -211,20 +176,17 @@ unsigned int FontEngine::getDefaultFontSize()
 /******************************************************************************/
 void FontEngine::readSettings()
 {
-#if USE_FREETYPE
```
AlexeySachkov: @SmallJoker, by removing this `#ifdef` you broke possibility to build minetest without freetype and gettext:

Right now I'm trying to build the very minimal configuration and I've set `ENABLE_FREETYPE=OFF` and `ENABLE_GETTEXT=OFF` and I see the following error:

```
minetest\src\client\fontengine.cpp(184): error C3861: 'gettext': identifier not found
```

SmallJoker: Please explain further. This preprocessor checks against `USE_FREETYPE`, not `USE_GETTEXT`.
There's a dummy function available for `gettext()` if gettext is not available:
https://github.com/minetest/minetest/blob/008b80fe1ca9d46610b8971e2ac92fb56da8e69f/src/gettext.h#L38

SmallJoker: fbc8724

AlexeySachkov: Thanks a lot, this fixes the build!

PR #8652, opened 2019-07-02
Fix missing item images clipping in formspecs

pyrollo:
When using item_image_button in formspec with a node, item image is not clipped at all.

This PR makes drawItemStack clip drawn image against clip rectangle even with 3D rendered items.

## How to test
You can easily test result of this PR with this Lua code :
```lua
minetest.register_chatcommand("t", {
	params = "",
	description = "Test",
	func = function(name, param)
		minetest.show_formspec(name, "test", [[size[5,5]
		item_image_button[0,-0.75;1,1;default:stone;test;]
		item_image_button[1,-0.5;1,1;default:stone;test;]
		item_image_button[2,-0.25;1,1;default:stone;test;]
		item_image_button[3,0;1,1;default:stone;test;]
		item_image_button[1,5;1,1;default:stone;test;]
		item_image_button[-0.75,1;1,1;default:stone;test;]
		item_image_button[-0.5,2;1,1;default:stone;test;]
		item_image_button[-0.25,3;1,1;default:stone;test;]
		item_image_button[4.5,1;1,1;default:stone;test;]
		item_image_button[4.75,2;1,1;default:stone;test;]
		item_image_button[5,3;1,1;default:stone;test;]
		]])
	end,
})
```
Result before the modification:
screenshot_20190702_234303

Result after it:
screenshot_20190702_234055

### DIFF
--- a/src/client/hud.cpp
+++ b/src/client/hud.cpp
@@ -652,23 +652,28 @@ void drawItemStack(video::IVideoDriver *driver,
 		core::rect<s32> viewrect = rect;
 		if (clip)
 			viewrect.clipAgainst(*clip);
+
 		core::matrix4 ProjMatrix;
+		ProjMatrix.buildProjectionMatrixOrthoLH(2.0f, 2.0f, -1.0f, 100.0f);
+
 		core::matrix4 ViewMatrix;
 		ViewMatrix.buildProjectionMatrixOrthoLH(
 			2.0f * viewrect.getWidth() / rect.getWidth(),
 			2.0f * viewrect.getHeight() / rect.getHeight(),
-			-1, 100);
+			-1.0f,
+			100.0f);
 		ViewMatrix.setTranslation(core::vector3df(
 			1.0f * (rect.LowerRightCorner.X + rect.UpperLeftCorner.X -
 					viewrect.LowerRightCorner.X - viewrect.UpperLeftCorner.X) /
 					viewrect.getWidth(),
 			1.0f * (viewrect.LowerRightCorner.Y + viewrect.UpperLeftCorner.Y -
 					rect.LowerRightCorner.Y - rect.UpperLeftCorner.Y) /
-					viewrect.getHeight(), 0));
+					viewrect.getHeight(),
+			0.0f));
 
-		ProjMatrix.buildProjectionMatrixOrthoLH(2, 2, -1, 100);
 		driver->setTransform(video::ETS_PROJECTION, ProjMatrix);
 		driver->setTransform(video::ETS_VIEW, ViewMatrix);
+
 		core::matrix4 matrix;
 		matrix.makeIdentity();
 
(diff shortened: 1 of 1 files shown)


src/client/hud.cpp
```diff
@@ -649,10 +649,25 @@ void drawItemStack(video::IVideoDriver *driver,
 		core::rect<s32> oldViewPort = driver->getViewPort();
 		core::matrix4 oldProjMat = driver->getTransform(video::ETS_PROJECTION);
 		core::matrix4 oldViewMat = driver->getTransform(video::ETS_VIEW);
+		core::rect<s32> viewrect = rect;
+		if (clip) viewrect.clipAgainst(*clip);
```
DS-Minetest:
```suggestion
		if (clip)
			viewrect.clipAgainst(*clip);
```

src/client/hud.cpp
```diff
@@ -649,10 +649,25 @@ void drawItemStack(video::IVideoDriver *driver,
...
 		core::matrix4 ProjMatrix;
+		core::matrix4 ViewMatrix;
+		ViewMatrix.buildProjectionMatrixOrthoLH(
+			2.0 * viewrect.getWidth() / rect.getWidth(),
```
DS-Minetest:
```suggestion
			2.0f * viewrect.getWidth() / rect.getWidth(),
```
(Not only here.)

pyrollo: Thanks ! Updates done.

src/client/hud.cpp
```diff
@@ -649,10 +649,26 @@ void drawItemStack(video::IVideoDriver *driver,
...
 		core::matrix4 oldViewMat = driver->getTransform(video::ETS_VIEW);
+		core::rect<s32> viewrect = rect;
+		if (clip)
+			viewrect.clipAgainst(*clip);
 		core::matrix4 ProjMatrix;
+		core::matrix4 ViewMatrix;
+		ViewMatrix.buildProjectionMatrixOrthoLH(
+			2.0f * viewrect.getWidth() / rect.getWidth(),
+			2.0f * viewrect.getHeight() / rect.getHeight(),
+			-1, 100);
```
DS-Minetest:
```suggestion
			-1.0f, 100.0f);
```
See http://irrlicht.sourceforge.net/docu/classirr_1_1core_1_1_c_matrix4.html#ae4a0618e2da724a26a5d8a201a63d8a5, these should be floats, too.

src/client/hud.cpp
```diff
@@ -649,10 +649,26 @@ void drawItemStack(video::IVideoDriver *driver,
...
+		ViewMatrix.setTranslation(core::vector3df(
+			1.0f * (rect.LowerRightCorner.X + rect.UpperLeftCorner.X -
+					viewrect.LowerRightCorner.X - viewrect.UpperLeftCorner.X) /
+					viewrect.getWidth(),
+			1.0f * (viewrect.LowerRightCorner.Y + viewrect.UpperLeftCorner.Y -
+					rect.LowerRightCorner.Y - rect.UpperLeftCorner.Y) /
+					viewrect.getHeight(), 0));
```
DS-Minetest:
```suggestion
					viewrect.getHeight(), 0.0f));
```
It's a vector3d**f**.

pyrollo: Ahah done :)

src/client/hud.cpp
pyrollo: Yes, that's a good thing to make a new pass on this code, it solves old style issues ;)

PR #8673, opened 2019-07-13
New jitter, RTT maths for object position prediction

SmallJoker:
* Real average RTT and jitter calculations
  * Does #8183 occur again? I hope not. **<< Needs testing**
* Remove min/max RTT and jitter values
  * Useless. Missing cooldown they were only used in `l_get_player_information`
* Accept client-side object position drift within the network jitter
  * Smoother object movement with less "jump-back"s

## To do
Ready to test and review.

- [ ] Test whether this weird network regression occurs again

## How to test

Spawn lua entities on a server and compare the movement with before
 * If smoother: Nice
 * If it's somewhat inaccurate: Trash this PR.

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -3663,11 +3663,7 @@ Utilities
       {
           address = "127.0.0.1",     -- IP address of client
           ip_version = 4,            -- IPv4 / IPv6
-          min_rtt = 0.01,            -- minimum round trip time
-          max_rtt = 0.2,             -- maximum round trip time
           avg_rtt = 0.02,            -- average round trip time
-          min_jitter = 0.01,         -- minimum packet time jitter
-          max_jitter = 0.5,          -- maximum packet time jitter
           avg_jitter = 0.03,         -- average packet time jitter
           connection_uptime = 200,   -- seconds since client connected
           protocol_version = 32,     -- protocol version used by client
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -3665,6 +3665,10 @@ Utilities
           ip_version = 4,            -- IPv4 / IPv6
           avg_rtt = 0.02,            -- average round trip time
           avg_jitter = 0.03,         -- average packet time jitter
+          min_rtt = 0.01,            -- DEPRECATED: equals to (avg_rtt - avg_jitter)
+          max_rtt = 0.2,             -- DEPRECATED: equals to (avg_rtt + avg_jitter)
+          min_jitter = 0.01,         -- DEPRECATED: constant (0.01)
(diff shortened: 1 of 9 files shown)


thomasrudin:
> Test whether this weird network regression occurs again

I had some network related issues a few days ago:
https://github.com/pandorabox-io/pandorabox.io/issues/223#issuecomment-512892310

It was a combination of the `mesecons` globalstep taking quite long (>100ms) and less calls to `Server::Receive()` (which only receives one packet per call as far as i see)

It had almost no impact on the server lag though: The server hummed along happily without the possibility to interact with it :smile:

doc/lua_api.txt
```diff
@@ -3663,11 +3663,7 @@ Utilities
       {
           address = "127.0.0.1",     -- IP address of client
           ip_version = 4,            -- IPv4 / IPv6
-          min_rtt = 0.01,            -- minimum round trip time
```
thomasrudin: Uhm, won't that break backwards compatibility?
I'm using that, for example:
https://github.com/thomasrudin-mt/monitoring/blob/master/builtin/player_rtt.lua#L30

SmallJoker: Yes, you're right, it might break server mods.
Should I just let it return `avg_rtt - avg_jitter`? `min_rtt`/`max_rtt` will always have the value of the most extreme RTTs measured since joining, which is neither precise nor helpful.

SmallJoker:
@thomasrudin Thanks a lot for testing this server-side. The most changes were done on client-side, though. If you happen to be (or know) a Minetest PvP addict, it would be the best situation to test the new accuracy of this PR.
I tested it on the local network plus on a few servers, but can't see much of a difference - even though the code is run for ~1/3 of the position updates.

thomasrudin:
> @thomasrudin Thanks a lot for testing this server-side. The most changes were done on client-side, though. If you happen to be (or know) a Minetest PvP addict, it would be the best situation to test the new accuracy of this PR.
> I tested it on the local network plus on a few servers, but can't see much of a difference - even though the code is run for ~1/3 of the position updates.

No, i did not test anything on that PR, sorry if i worded that wrong.
I was just commenting about that spurious network issue you mentioned.


> Yes, you're right, it might break server mods.

Thats alright, i don't have problems with breaking changes, it keeps the code alive :smile:

ClobberXD:
Tested in CTF - it seems to me that player movements are a lot more jerkier. I really hope this PR actually works, and it was I who made a mistake in testing. Oh, but I was using a debug build, while my regular client is a release build. So I'll also test this PR in a release build when I get the time.

SmallJoker:
Thanks for testing. I actually marked this PR as WIP because of that issue. I'll have to remove the RTT prediction and just use the jitter to ensure it's not too jerky.

PR #8674, opened 2019-07-13
Optimize getting active objects a bit.

lhofhansl:
This is a (slight) optimization when retrieving active objects.

Objects are retrieved (and sometimes sorted) by their distance. We can save some cycles by comparing (and sorting by) the square of the distance.

I noticed that with many active objects the various getActiveObjects and getObjectsInsideRadius methods show up prominently in profiler logs.

Note the change in DistanceSortedActiveObject. Since this now stores the square of the distance (order is still the same) I made d private so that nobody can access it and mistake it for an actual distance.

**Update:** I removed the change in serverenvironment.cpp. This one actually did not show in the profiler, and it would have led to overflows with radius' over 128, which can happen with zooming. The other methods use floats and hence are safe.

### DIFF
--- a/src/client/clientobject.h
+++ b/src/client/clientobject.h
@@ -90,10 +90,10 @@ private:
 	static std::unordered_map<u16, Factory> m_types;
 };
 
-struct DistanceSortedActiveObject
+class DistanceSortedActiveObject
 {
+public:
 	ClientActiveObject *obj;
-	f32 d;
 
 	DistanceSortedActiveObject(ClientActiveObject *a_obj, f32 a_d)
 	{
@@ -105,4 +105,7 @@ struct DistanceSortedActiveObject
 	{
 		return d < other.d;
 	}
+
+private:
+	f32 d;
 };
--- a/src/serverenvironment.cpp
+++ b/src/serverenvironment.cpp
@@ -290,17 +290,16 @@ void LBMManager::applyLBMs(ServerEnvironment *env, MapBlock *block, u32 stamp)
 void fillRadiusBlock(v3s16 p0, s16 r, std::set<v3s16> &list)
 {
 	v3s16 p;
-	s32 r2 = r * r;
 	for(p.X=p0.X-r; p.X<=p0.X+r; p.X++)
-	for(p.Y=p0.Y-r; p.Y<=p0.Y+r; p.Y++)
-	for(p.Z=p0.Z-r; p.Z<=p0.Z+r; p.Z++)
-	{
-		// limit to a sphere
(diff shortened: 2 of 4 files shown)


src/serverenvironment.cpp
```diff
@@ -290,16 +290,17 @@ void LBMManager::applyLBMs(ServerEnvironment *env, MapBlock *block, u32 stamp)
...
-			for(p.Z=p0.Z-r; p.Z<=p0.Z+r; p.Z++)
-			{
-				// limit to a sphere
-				if (p.getDistanceFrom(p0) <= r) {
-					// Set in list
-					list.insert(p);
-				}
-			}
+	for(p.Y=p0.Y-r; p.Y<=p0.Y+r; p.Y++)
+	for(p.Z=p0.Z-r; p.Z<=p0.Z+r; p.Z++)
```
SmallJoker: Needs code style fixes (lines were touched).

lhofhansl: Yep.

Actually... I'm going to take back this change anyway for two reasons: **(1)** This did not show in the profiler - I just fixed it ...  because and **(2)** I'm worried about overflow. getDistanceFromSQ is declared as `T getDistanceFromSQ (const vector3d< T > &other)` so for v3s16 I'd be worried about overflow.

src/client/clientobject.h
```diff
@@ -90,10 +90,10 @@ class ClientActiveObject : public ActiveObject
 	static std::unordered_map<u16, Factory> m_types;
 };
 
-struct DistanceSortedActiveObject
+class DistanceSortedActiveObject
```
sfan5: I've been wondering, does doing this introduce any overhead over `struct`s?

lhofhansl: I don't think so. AFAIK a struct is the same as a class, just that all access defaults to public.

lhofhansl:
I looked into this a bit further. These are called mostly for collision testing. In my case I have a few thousand active objects. Each checking all others for collisions quickly going into the millions.
It would be better to filter out objects further away by some cheap spatial means (rather than checking them all for distance).

src/client/clientobject.h
lhofhansl: I could have left it a struct and just mark d private. Would you prefer that? It does not make a difference, though.

lhofhansl:
What I have here is good to merge?

src/client/clientobject.h
sfan5: No, it's fine.

lhofhansl:
9fe32461

PR #8677, opened 2019-07-14
Unify wield item handling

SmallJoker:
This moves the wield item functions to Player and the tool utils for range calculation
Also 'local_inventory' was removed due to redundancy in Client
Removing this duplicated code reduces the possibility of different handling in server and client.

## To do
This PR is Ready for Review.

## How to test
Create a `hand` slot, put a tool or anything else in.
Click randomly in your world. If your hand is not capable of digging the node, it will try to fall back to `hand` (same as in master).

### DIFF
--- a/src/client/client.cpp
+++ b/src/client/client.cpp
@@ -935,7 +935,7 @@ void Client::interact(u8 action, const PointedThing& pointed)
 	NetworkPacket pkt(TOSERVER_INTERACT, 1 + 2 + 0);
 
 	pkt << action;
-	pkt << (u16)myplayer->getWieldIndex();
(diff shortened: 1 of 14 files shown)


src/client/client.cpp
```diff
@@ -935,7 +935,7 @@ void Client::interact(u8 action, const PointedThing& pointed)
 	NetworkPacket pkt(TOSERVER_INTERACT, 1 + 2 + 0);
 
 	pkt << action;
-	pkt << (u16)getPlayerItem();
+	pkt << (u16)myplayer->getWieldIndex();
```
sfan5: `static_cast<>`

PR #8680, opened 2019-07-15
Formspec: Add HTML-like element

kilbith:
This PR offered by [Kidscode](https://kidscode.com) allows to format a formspec content almost like HTML.

**Demo video:** https://www.youtube.com/watch?v=_AtifpDQ-5M

**Demo mod:** https://paste.ubuntu.com/p/Nf6XfXnSZX/

-------------------

screenshot_20190715_163150


v-rob:
This looks pretty nice from the screenshot and video, but I have a few suggestions and questions:

Questions:
- Does it collapse multiple whitespaces into one?  (I hope it doesn't, since that is so annoying IMO.)
- Are the `width` and `height` parameters in pixels or coordinates?  Since this is scalable, it should probably be in coordinates.
- In the `<style>` tag, when setting the font size, is the value the height in pixels or is it relative to the current size?

And a few suggestions:
- I noticed that margins and things like that are in pixels.  Since formspecs are scalable, I think that they should be coordinates
- Other things should be relative like text size (if it isn't already) so that the form can fit everybody's screen size and settings.
- Add support for `real_coordinates` in the `text` element, or just drop support for the old system and always use `real_coordinates` (which reduces code size, and the old system doesn't matter much any more since there's no practical use for it anymore.)
- Use `guiScrollBar` instead of `IGUIScrollBar` since the former has a modifiable thumb height.
- Perhaps rename it from `text` to something more descriptive, since this is much more than plain text.  Perhaps something like `markup`, since that is more descriptive and self-explanatory?

I personally would have chosen a syntax less like HTML, but that's just personal preference, so you don't have to listen to that.

Anyway, that's just a few quick thoughts I had on the spot.

pyrollo:
Hi v-rob,

>     * Does it collapse multiple whitespaces into one?  (I hope it doesn't, since that is so annoying IMO.)
No it does not. It also keeps new lines. Of course in "justify" mode, it adds some extra space to them.

>     * Are the `width` and `height` parameters in pixels or coordinates?  Since this is scalable, it should probably be in coordinates.
Now they are in irrlicht coordinates (I suppose they are pixels). 

>     * In the `<style>` tag, when setting the font size, is the value the height in pixels or is it relative to the current size?
Same as above.

>     * I noticed that margins and things like that are in pixels.  Since formspecs are scalable, I think that they should be coordinates
Yes that's a good idea. I don't know if it's possible (element aware of scaling?). Nothing scales but the text is ajusted to fit. Like in other text displaying softwares. If you change window size, text is layed out again to fit but font size is not changed.

>     * Other things should be relative like text size (if it isn't already) so that the form can fit everybody's screen size and settings.
Same as above, char size is given as is to font APIs.
 
>     * Add support for `real_coordinates` in the `text` element, or just drop support for the old system and always use `real_coordinates` (which reduces code size, and the old system doesn't matter much any more since there's no practical use for it anymore.)
Yes, that's something we planed to do. We were waiting for you PR to be accepted (I guess it is done now).
I hope the "old" system will be abandoned. I took textarea as template because it's close to this new element.

>     * Use `guiScrollBar` instead of `IGUIScrollBar` since the former has a modifiable thumb height.
Ok!

>     * Perhaps rename it from `text` to something more descriptive, since this is much more than plain text.  Perhaps something like `markup`, since that is more descriptive and self-explanatory?
The name is always a problem, it has to be short and precise. We got through many names* and back to `text` even it's not satisfying. `markup` is short but it does not satisfy me neither.

> I personally would have chosen a syntax less like HTML, but that's just personal preference, so you don't have to listen to that.
I would also. First I though of something like BBCode but it would have been a pain to escape all brackets not to conflict with formspecs (specially in short texts, long text are likely to use `formspec_escape`). BBCode would have been fine as it is aimed to be basic and simple.
Then I went to html/xml/svg/... syntax, not wanting to use an exotic syntax, even if it's somehow misleading. HTML is complex and huge, here we are in a much simpler and basic stufff.
But yeah, if we could have a syntax fitting more to our needs, I take it ! (actually if its made of opening/closing tags, it's easy to change)

Thanks for your feed back !

v-rob:
Thanks for getting back!  Answering a few questions and adding a few more comments:

* For the size of images, when I mean coordinates, I mean `imgsize` which is calculated for each formspec's size.  So, for instance, [line 415](https://github.com/minetest/minetest/pull/8680/files#diff-48e7da6df9c6e0ee3bfb75cafdbae418R415) could instead be 
    ```c++
    int width = strtol(attrs["width"].c_str(), NULL, 10) * imgsize.X;
    ```
    You'd also have to pass `imgsize` to the constructor of GUIText, of course.

* Yes, the old coordinate system is pretty much abandoned now.  New elements really don't have to work with it anymore since that would add extra code for no real benefit.

* For `guiScrollBar`, use `setPageSize` to set the height of the thumb.

* I was also thinking BBCode for the syntax since it's easier to understand.  I was trying to make something like this PR a while back, but it didn't work very well at all.  During that, I came up with a syntax that was short and simple, very similar to BBCode.  Here it is (although modified for this PR):

<details>
<summary><b>Tags</b></summary>

Note that, although I used `[]`, `<>` could be used just as well.  I prefer `[]` because it fits in with the formspec syntax better.  I also left out the closing tags because it's just `[/tagname]`.

`[align=<left/right/center/justify>]`: Aligns text

`[action=<name>]`: A link.  I called it action because it doesn't necessarily have to work like a link; it could do a different action

`[img=<name>;float=<left/right>,width=<#>,height=<#>]`: Image.  After semicolon are the optional arguments, separated by commas.

`[item=<name>;float=<right/left>,width=<#>,height=<#>,rotate=<true/false>]`: Item image

`[color=<color>]`: Set text color

`[font=<mono/normal>]`: Set font to monospace or back to normal

`[size=<big/bigger/normal/#>`: Set font size.  Works as `<big>`, `<bigger>`, `<normal>`, and `<style size=...>` all in one.

`[global;op1=<value>,op2=<value>]`: Just like `global`, but with comma separated values.

I'm not very satisfied with `img`, `item`, or `global` since they all take multiple values, so I tried to make them as close to normal formspec syntax as possible.  But I think the other tags are alright.
</details>

rubenwardy:
I prefer XML tags to bbcode.

Also, I think the element should be called `hypertext[]`, as that's what it is

pyrollo:
This PR could be labelled as WIP, there are still many improvements that can be done.

Regarding the tags, I think we will keep xml style syntax. 
We adopted the "action" tag name which is much better and clearer.
I think thinks should be kept simple and our PR is already too rich in term of styling.
I'm wondering if having just 3 sizes of chars, 2 styles (like it is yet) and a few automatic styles (actions, "bold" or "strong", and another one) would be sufficient. May be giving the possibility of choosing any font size would lead to a risk of having very heterogeneous formspecs. Anyway the full possibilities will be kept in C++ code for later use.

so tags could be : 
* Align tags : `align` or `center` / `justify` / `right` / `left`
* Size tags : `normal` / `big` / `bigger` ?
* `action`, `img`, `item`
* `font` or `mono` / `proportionnal` (two possibilities)
* `b`, `i` with color/style set in `global`
* `global` : very convenient way to configure global stuff : action, action hover, b, i styles.

And that's it, keeping it simple for non programmers.

SmallJoker:
You implemented an entire HTML-like parser and render (using Irrlicht ofc) in a thousand lines. Wow, that's impressive! It won't only improve the capabilities of wiki mods, but also make formspecs visually more attractive.
I see that this PR yet needs a few touches to complete, like real coordinates and rendering speed tests using large content (clipping content).

Sokomine:
Please resolve the merge conflict with src/client/hud.cpp. I tried it in my local branch but am not sure if I did that right. Fixing that might attract a lot more testers.

The image of the furnace did not show up in my local client. Other than that, the test command worked fine and showed the formspec as displayed above. Hovering over a link did highlight it, and clicking on "using" displayed the help page as it ought to. Scrolling worked fine as well.

Even adding a classic formspec button to the new formspec is supported.

The font size seems to be independent of the font size set in minetest.conf. Perhaps this is a good idea. It may make formspec design easier as long as the fonts do not get too small.

I did have one big problem though: Opening the inventory formspec turned out to be a bad idea as I found no way to close it again. ESC did not work. I had to kill the client.

Inventory items, items in chests and images in unified_inventory where not shown. Perhaps that is related to the furnace not showing up and caused by my inproper merge.

doc/lua_api.txt
```diff
@@ -2385,6 +2395,86 @@ offsets when migrating:
...
+Make that text a clickable link.
+
+* `name`: Name of the link (mandatory).
+
+When clicked, the formspec is send to the server. The value of the text field
+sent to `on_player_receive_fields` will be "link:" concatenated to the link name.
+
+`<img name=... float=... width=... height=...>`
+
+Draws an image. Image must be in game media.
```
ClobberXD:
```suggestion
Draws an image. Image must be in the `textures/` sub-directory of a mod.
```

src/gui/guiFormSpecMenu.h
```diff
@@ -591,4 +592,4 @@ class FormspecFormSource: public IFormSource
 	}
 
 	std::string m_formspec;
-};
+};
```
ClobberXD: Newline seems to be inadvertently removed from EOF

src/gui/guiText.h
```diff
@@ -0,0 +1,178 @@
...
+		bool OnEvent(const SEvent& event);
+
+	protected:
+
+		// GUI members
+		Client *m_client;
+		gui::IGUIScrollBar *m_vscrollbar;
+		TextDrawer m_drawer;
+
+		// Positionning
```
ClobberXD:
```suggestion
		// Positioning
```

ClobberXD (changes requested):
Great work on this PR! I've suggested some minor changes as follows.

src/gui/guiText.cpp
```diff
@@ -0,0 +1,1000 @@
...
+	return text;
+}
+
+bool check_color(const std::string &str) {
+	video::SColor color;
+	return parseColorString(str, color, false);
+}
+
+bool check_integer(const std::string &str) {
+	if (str == "")
```
nerzhul: if (str.empty())

src/gui/guiText.cpp
```diff
@@ -0,0 +1,1000 @@
...
+	return *endptr == '\0';
+}
+
+// -----------------------------------------------------------------------------
+// ParsedText - A text parser
+
+void ParsedText::Element::setStyle(StyleList &style)
+{
+	if (style["valign"] == "middle")
+		this->valign = VALIGN_MIDDLE;
```
nerzhul: unneeded this in this function

src/gui/guiText.cpp
```diff
@@ -0,0 +1,1000 @@
...
+		m_paragraph->setStyle(m_style);
+	}
+}
+
+void ParsedText::enterElement(ElementType type)
+{
+	enterParagraph();
+
+	if (!m_element || m_element->type != type)
+	{
```
nerzhul: code style

src/gui/guiText.cpp
```diff
@@ -0,0 +1,1000 @@
...
+	// New word if needed
+	if (c == L' ' || c == L'\t')
+		enterElement(ELEMENT_SEPARATOR);
+	else
+		enterElement(ELEMENT_TEXT);
+
+	m_element->text += c;
+}
+
+ParsedText::Tag* ParsedText::newTag(std::string name, AttrsList attrs)
```
nerzhul: const std::string &name

src/gui/guiText.cpp
```diff
@@ -0,0 +1,1000 @@
...
+{
+	endElement();
+	Tag* newtag = new Tag();
+	newtag->name = name;
+	newtag->attrs = attrs;
+	m_tags.push_back(newtag);
+	return newtag;
+}
+
+ParsedText::Tag* ParsedText::openTag(std::string name, AttrsList attrs)
```
nerzhul: const std::string &name

nerzhul: const AttrsList &

src/gui/guiText.cpp
nerzhul: const AttrsList &

src/gui/guiText.cpp
```diff
@@ -0,0 +1,1000 @@
...
+{
+	Tag* newtag = newTag(name, attrs);
+	m_active_tags.push_front(newtag);
+	return newtag;
+}
+
+bool ParsedText::closeTag(std::string name)
```
nerzhul: const std::string &

src/gui/guiText.cpp
```diff
@@ -0,0 +1,1000 @@
...
+			found = true;
+			break;
+		}
+	return found;
+}
+
+void ParsedText::globalTag(AttrsList &attrs)
+{
+	for (auto const& attr : attrs)
+	{
```
nerzhul: code style too

src/gui/guiText.cpp
```diff
@@ -0,0 +1,1000 @@
...
+	for (auto id = m_active_tags.begin(); id != m_active_tags.end(); ++id)
+		if ((*id)->name == name) {
+			m_active_tags.erase(id);
+			found = true;
+			break;
+		}
+	return found;
+}
+
+void ParsedText::globalTag(AttrsList &attrs)
```
nerzhul: missing const

src/gui/guiText.cpp
```diff
@@ -0,0 +1,1000 @@
...
+	m_text(text),
+	m_client(client),
+	m_environment(environment)
+{
+	// Size all elements
+	for (auto & p : m_text.m_paragraphs) {
+		for (auto & e : p.elements) {
+			if (e.type == ParsedText::ELEMENT_SEPARATOR ||
+				e.type == ParsedText::ELEMENT_TEXT)
+			{
```
nerzhul: multiple code style issues

src/gui/guiText.cpp
```diff
@@ -0,0 +1,1000 @@
...
+						e.dim.Width = dim.Width * e.dim.Height / dim.Height;
+				}
+			}
+		}
+	}
+}
+
+// Get element at given coordinates. Coordinates are inner coordinates (starting
+// at 0,0).
+ParsedText::Element* TextDrawer::getElementAt(core::position2d<s32> pos)
```
nerzhul: const &

src/gui/guiText.cpp
```diff
@@ -0,0 +1,1000 @@
...
+			return (height - m_height) / 2;
+		case ParsedText::VALIGN_TOP:
+		default:
+			return 0;
+	}
+}
+
+// Draw text in a rectangle with a given offset. Items are actually placed in
+// relative (to upper left corner) coordinates.
+void TextDrawer::draw(
```
nerzhul: i think we miss const ref here too

src/gui/guiText.cpp
```diff
@@ -0,0 +1,1000 @@
...
+GUIText::GUIText(
+	const wchar_t* text,
+	IGUIEnvironment* environment,
+	IGUIElement* parent, s32 id,
+	const core::rect<s32>& rectangle,
+	Client *client,
+	ISimpleTextureSource *tsrc) :
+	IGUIElement(EGUIET_ELEMENT, environment, parent, id, rectangle),
+	m_client(client),
+	m_vscrollbar(nullptr),
```
nerzhul: should be inited in the def not the constructor

src/gui/guiText.cpp
```diff
@@ -0,0 +1,1000 @@
...
+bool GUIText::OnEvent(const SEvent& event) {
+	if (event.EventType == EET_GUI_EVENT) {
+		if (event.GUIEvent.EventType == EGET_SCROLL_BAR_CHANGED)
+			if (event.GUIEvent.Caller == m_vscrollbar)
+				m_text_scrollpos.Y = -m_vscrollbar->getPos();
+	}
+
+	if (event.EventType == EET_MOUSE_INPUT_EVENT)
+	{
+		if (event.MouseInput.Event == EMIE_MOUSE_WHEEL)
```
nerzhul: switch case is better here

nerzhul (dismissed):
A very good feature. A little bit codestyle to fix and many performance issues to fix, this can be memory intensive (especially for android) on many parts

src/gui/guiText.cpp
ClobberXD: i.e.

```cpp
if (!m_element || m_element->type != type) {
```

src/gui/guiText.cpp
ClobberXD: i.e. double-indents for continuation lines and opening brace on the same line

```cpp
if (e.type == ParsedText::ELEMENT_SEPARATOR ||
		e.type == ParsedText::ELEMENT_TEXT) {
```

Sokomine:
Later on, some limited support of tables would be nice.  Sometimes - even in the example in the introduction - some texts require a tabular setup. This may be very difficult to implement and is a wish for further development once this got merged (which I hope will be possible soon!)

paramat:
See the output of the clang format build check for many valid codestyle errors.
Of course, many of clang format's suggestions are invalid or bad suggestions and shouldn't be implemented, it's fairly obvious which.

kilbith:
- `text` element renamed to `hypertext`
- `link` becomes `action`
- Fixed code style, bug with line height, bug with link hovering and item rotation now works
- Changed tag management and added new tags
- Demo mod updated accordingly

doc/lua_api.txt
```diff
@@ -2386,6 +2396,104 @@ offsets when migrating:
...
+
+* `name`: Name of the action (mandatory).
+
+When clicked, the formspec is send to the server. The value of the text field
+sent to `on_player_receive_fields` will be "action:" concatenated to the action
+name.
+
+`<img name=... float=... width=... height=...>`
+
+Draws an image. Image must be in the `textures/` sub-directory of a mod.
```
SmallJoker: Not directly. It has to be in the client's media cache.
```suggestion
Draws an image which is present in the client media cache.
```
Or similar. Maybe just copy the short `image[]` description. Even if it's not present server-side it can be taken from the client's texture pack.

doc/lua_api.txt
```diff
@@ -2386,6 +2396,104 @@ offsets when migrating:
...
+
+`<left>...</left>`: Left-aligned text
+
+`<right>...</right>`: Right-aligned text
+
+`<justify>...</justify>`: Justified text
+
+`<mono>...</mono>`: Monospaced font
+
+`<b>...</b>`, `<i>...</i>`: Supposed to be bold and italic font but actually a text with different color.
```
SmallJoker: "Supposed to be"? Because it's not yet supported? What does this have to do with colors?

doc/lua_api.txt
```diff
@@ -2386,6 +2396,104 @@ offsets when migrating:
...
+
+`<item name=... float=... width=... height=... rotate=...>`
+
+Draws an item image.
+
+* `name`: Item string of the item to draw (mandatory).
+* `float`: If present, makes the image floating (`left` or `right`).
+* `width`: Item image width.
+* `height`: Item image height.
+* `rotate`: Rotate item image if set to `yes`.
```
SmallJoker: Is this meant to switch between a simple 2D front texture and the rendered cube?

doc/lua_api.txt
```diff
@@ -2416,6 +2524,17 @@ Player Inventory lists
 Colors
 ======
 
+Formatted Text Format
+---------------------
+
+Formatted text format is similar to, but much simpler than, HTML.
+
+Available tags are:
+`<style color=... size=... font=... >`
```
SmallJoker: List notation would fit here.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -1326,6 +1327,50 @@ void GUIFormSpecMenu::parseField(parserData* data, const std::string &element,
 	errorstream<< "Invalid field element(" << parts.size() << "): '" << element << "'"  << std::endl;
 }
 
+void GUIFormSpecMenu::parseHyperText(parserData* data, const std::string &element)
```
SmallJoker:
```suggestion
void GUIFormSpecMenu::parseHyperText(parserData *data, const std::string &element)
```

src/gui/guiFormSpecMenu.cpp
```diff
@@ -1326,6 +1327,50 @@ void GUIFormSpecMenu::parseField(parserData* data, const std::string &element,
...
+{
+	std::vector<std::string> parts = split(element,';');
+	if (parts.size() == 4)
+	{
```
SmallJoker: Misplaced bracket. Error & return would also reduce the indent level.
Consider adding a forwards-compatibility mode like the other formspec elements. In theory it is possible to let mods specify the `formspec_version[n]` element, so for future changes that would be handy to not require adding a new element.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -1326,6 +1327,50 @@ void GUIFormSpecMenu::parseField(parserData* data, const std::string &element,
...
+		std::vector<std::string> v_pos = split(parts[0],',');
+		std::vector<std::string> v_geom = split(parts[1],',');
+		std::string name = parts[2];
+		std::string text = parts[3];
+
+		MY_CHECKPOS("text",0);
+		MY_CHECKGEOM("text",1);
+
+		// [ Taken from textarea
```
SmallJoker: Needs real_coordinates integration. Just copy the new textarea code again?

src/gui/guiFormSpecMenu.cpp
```diff
@@ -1326,6 +1327,50 @@ void GUIFormSpecMenu::parseField(parserData* data, const std::string &element,
...
+
+		FieldSpec spec(
+			name,
+			utf8_to_wide(unescape_string(text)),
+			L"",
+			258+m_fields.size()
+		);
+
+		spec.ftype = f_Unknown;
+		new GUIHyperText(spec.flabel.c_str(), Environment, this, spec.fid, rect, m_client, m_tsrc);
```
SmallJoker: Where is it destructed? Usually you have to reduce the reference counter by 1 so that Irrlicht can destroy the element as soon it's not displayed any more.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -3315,17 +3365,18 @@ bool GUIFormSpecMenu::preprocessEvent(const SEvent& event)
 			}
 		}
 	}
-	// Mouse wheel events: send to hovered element instead of focused
-	if(event.EventType==EET_MOUSE_INPUT_EVENT
-			&& event.MouseInput.Event == EMIE_MOUSE_WHEEL) {
+	// Mouse wheel and move events: send to hovered element instead of focused
+	if (event.EventType == EET_MOUSE_INPUT_EVENT &&
+		(event.MouseInput.Event == EMIE_MOUSE_WHEEL ||
```
SmallJoker: +1 indent.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -3961,6 +4012,11 @@ bool GUIFormSpecMenu::OnEvent(const SEvent& event)
 					s.fdefault = L"Changed";
 					acceptInput(quit_mode_no);
 					s.fdefault = L"";
+				} else if ((s.ftype == f_Unknown) &&
+					(s.fid == event.GUIEvent.Caller->getID())) {
+						s.send = true;
```
SmallJoker: Please swap the indent levels of this code block with the one from the continued `if` condition.

src/gui/guiFormSpecMenu.h
```diff
@@ -520,6 +520,7 @@ class GUIFormSpecMenu : public GUIModalMenu
 	void parseSimpleField(parserData* data,std::vector<std::string> &parts);
 	void parseTextArea(parserData* data,std::vector<std::string>& parts,
 			const std::string &type);
+	void parseHyperText(parserData* data, const std::string &element);
```
SmallJoker:
```suggestion
	void parseHyperText(parserData *data, const std::string &element);
```

src/gui/guiHyperText.cpp
```diff
@@ -0,0 +1,1069 @@
...
+#include "client/fontengine.h"
+#include <SColor.h>
+#include "client/tile.h"
+#include "IVideoDriver.h"
+#include "client/client.h"
+#include "client/renderingengine.h"
+#include "hud.h"
+#include "guiHyperText.h"
+
+std::string strwtostr(const core::stringw &str) {
```
SmallJoker: Bracket placement does not follow the Minetest code style guidelines. Please put them into a new line for functions.

src/gui/guiHyperText.cpp
```diff
@@ -0,0 +1,1069 @@
...
+			if (attr.second == "bottom")
+				valign = ParsedText::VALIGN_BOTTOM;
+			if (attr.second == "middle")
+				valign = ParsedText::VALIGN_MIDDLE;
+		}
+
+		if (attr.first == "background") {
+			irr::video::SColor color;
+			if (attr.second == "none")
+				background_type = BACKGROUND_NONE;
```
SmallJoker: First code block also needs braces. 
https://www.kernel.org/doc/html/v5.0/process/coding-style.html#placing-braces-and-spaces

src/gui/guiHyperText.cpp
```diff
@@ -0,0 +1,1069 @@
...
+void ParsedText::parseStyles(AttrsList &attrs, StyleList &style)
+{
+	if (attrs.count("color") && check_color(attrs["color"]))
+		style["color"] = attrs["color"];
+
+	if (attrs.count("hovercolor") && check_color(attrs["hovercolor"]))
+		style["hovercolor"] = attrs["hovercolor"];
+
+	if (attrs.count("font") &&
+		(attrs["font"] == "mono" || attrs["font"] == "normal"))
```
SmallJoker: +1 indent please.

src/gui/guiHyperText.cpp
```diff
@@ -0,0 +1,1069 @@
...
+	StyleList style;
+
+	if (name == "global") {
+		if (end)
+			return 0;
+		globalTag(attrs);
+
+	} else if (name == "style") {
+		if (end)
+			closeTag(name);
```
SmallJoker: Please add braces for this block here too.

src/gui/guiHyperText.cpp
```diff
@@ -0,0 +1,1069 @@
...
+
+	return cursor;
+}
+
+// -----------------------------------------------------------------------------
+// Text Drawer
+
+TextDrawer::TextDrawer(
+	const wchar_t* text,
+	Client* client,
```
SmallJoker: Misplaced pointers.

src/gui/guiHyperText.cpp
```diff
@@ -0,0 +1,1069 @@
...
+	m_text(text), m_client(client), m_environment(environment)
+{
+	// Size all elements
+	for (auto &p : m_text.m_paragraphs) {
+		for (auto &e : p.elements) {
+			if (e.type == ParsedText::ELEMENT_SEPARATOR ||
+					e.type == ParsedText::ELEMENT_TEXT) {
+				if (e.font) {
+					e.dim.Width = e.font->getDimension(e.text.c_str()).Width;
+					e.dim.Height = e.font->getDimension(L"Yy").Height;
```
SmallJoker: There are fontengine helper functions to determine the text dimensions.

src/gui/guiHyperText.cpp
```diff
@@ -0,0 +1,1069 @@
...
+						e.dim.Width = dim.Width * e.dim.Height / dim.Height;
+				}
+			}
+		}
+	}
+}
+
+// Get element at given coordinates. Coordinates are inner coordinates (starting
+// at 0,0).
+ParsedText::Element* TextDrawer::getElementAt(const core::position2d<s32> &pos)
```
SmallJoker: Misplaced pointer.

src/gui/guiHyperText.cpp
```diff
@@ -0,0 +1,1069 @@
...
+		for (auto &el : p.elements) {
+			core::rect<s32> rect(el.pos, el.dim);
+			if (rect.isPointInside(pos))
+				return &el;
+		}
+	}
+	return 0;
+}
+
+void TextDrawer::place(s32 width)
```
SmallJoker: Would you please be so nice to include a quick summary about what this function in general does?

src/gui/guiHyperText.cpp
```diff
@@ -0,0 +1,1069 @@
...
+						}
+						else if (f.rect.LowerRightCorner.X + f.margin >= right &&
+								f.rect.UpperLeftCorner.X - f.margin > left) { // float on right
+							if (f.rect.UpperLeftCorner.X -
+									std::max(f.margin, p.margin) < right)
+								right = f.rect.UpperLeftCorner.X -
+										std::max(f.margin, p.margin);
+
+						}
+						else if (f.rect.UpperLeftCorner.X - f.margin <= left &&
```
SmallJoker: Combine with line above. Same for L687.

src/gui/guiHyperText.cpp
```diff
@@ -0,0 +1,1069 @@
...
+			s32 nexty = y;
+			do {
+				y = nexty;
+				nexty = 0;
+
+				// Inner left & right
+				left = m_text.margin;
+				right = width - m_text.margin;
+
+				for (auto f : m_floating) {
```
SmallJoker: Could be a const ref.

src/gui/guiHyperText.cpp
```diff
@@ -0,0 +1,1069 @@
...
+
+			// Third pass, actually place everything
+			for (auto e = linestart; e != lineend; ++e) {
+				if (e->floating != ParsedText::FLOAT_NONE)
+					continue;
+
+				e->pos.X = x;
+				e->pos.Y = y;
+
+				switch(e->type) {
```
SmallJoker: Missing space after some `switch` and `while` keywords.

src/gui/guiHyperText.cpp
```diff
@@ -0,0 +1,1069 @@
...
+	core::position2d<s32> offset = dest_rect.UpperLeftCorner + dest_offset;
+
+	if (m_text.background_type == ParsedText::BACKGROUND_COLOR)
+		m_environment->getVideoDriver()->draw2DRectangle(
+				m_text.background_color, dest_rect);
+
+	for (auto &p : m_text.m_paragraphs) {
+		for (auto &el : p.elements) {
+			core::rect<s32> rect(el.pos + offset, el.dim);
+			if (rect.isRectCollided(dest_rect)) {
```
SmallJoker: Please invert & continue to reduce the indent levels.

src/gui/guiHyperText.h
```diff
@@ -0,0 +1,182 @@
...
+protected:
+
+	// Parser functions
+	void enterElement(ElementType type);
+	void endElement();
+	void enterParagraph();
+	void endParagraph();
+	void pushChar(wchar_t c);
+	ParsedText::Tag* newTag(const std::string &name, const AttrsList &attrs);
+	ParsedText::Tag* openTag(const std::string &name, const AttrsList &attrs);
```
SmallJoker: Bunch of misplaced pointers in this file.

SmallJoker (commented):
Long review. The PR works, so I think the leftover is to make the code style perfect?

doc/lua_api.txt
kilbith: No.

SmallJoker: Would you please be so nice and extend the description a bit? If this is just ordinary rotation, would it be possible to change `yes` into a float, specifying the angle?

doc/lua_api.txt
kilbith: Example please?

SmallJoker:
```
* `<style color=... size=... font=... >`
* ..
* ..
```

doc/lua_api.txt
kilbith: No, it's not yet supported.

doc/lua_api.txt
benrob0329: `client's media cache` is more confusing, even if slightly more accurate. It'd also be a very bad practice to use a texture assumed to be cached on the client.

src/gui/guiHyperText.cpp
ClobberXD: Does this function convert a wide string to a narrow string? Note that MT already has `wide_to_narrow` and `narrow_to_wide` declared in `util/string.h`.

src/gui/guiHyperText.cpp
kilbith: For some reasons `g_fontengine->getTextWidth(e.text.c_str());` and `g_fontengine->getTextHeight();` change the content positioning.

src/gui/guiFormSpecMenu.cpp
kilbith: Destructing it in the dtor causes a double-free. Looks like it's dropped somewhere else (possibly in `removeChildren`?)

kilbith:
- Support for `real_coordinates`
- Use of `guiScrollBar` and `setPageSize`
- Code style cleaning

kilbith:
Change to the `<item image>` tag: its attribute `rotate` now accepts a vector to set a custom angle. Examples:

- `<item image=default:furnace rotate=30,100,-50>`: Rotate item with 30° on X axis, 100% speed on Y axis (*clockwise*), -50° on Z axis.
- `<item image=default:furnace rotate=0,-10,0>`: Rotate item with 0° on X axis, 10% speed on Y axis (*anticlockwise*), 0° on Z axis.

src/gui/guiHyperText.cpp
pyrollo: The thing is we need to know the text dimension using its given font. I looked most appropriate to me.
If Truetype is in use, it looks like getDimention always gives the same height (not depending on string) so it could be simplified, but I'm not sure of the behavior if truetype not in use.

src/gui/guiHyperText.cpp
pyrollo: This function places all elements according to given width. Elements have been previously sized by constructor and will be later drawed by draw. It may be called each time width changes and resulting height is stored in m_height. See GUIHyperText constructor, it uses it once to test if text fits in window and eventually another time if width is reduced because of scrollbar added.

(this comment will be added to code)

kilbith:
**Update**
Bold, italic, bold-italic and underline are now supported. This works for both regular and mono fonts.

screenshot_20190816_152851

v-rob:
I must say, I am very happy with this PR.  I didn't expect bold, italic, or underlined, so I was quite happily surprised.  Guess I should try my hand a some reviewing because this is so important.  I did notice that there is both the `<mono>` tag and the `<style font=mono>` which is rather redundant.

I like the idea of having tables like Sokomine said, but I do agree that tables should be a project for further development.

Another idea I have it to place arbitrary elements like buttons or checkboxes with everything else.  No idea how the syntax would work, but it would be cool to have other things than just links, but like above, probably a project for further development afterwords.

pyrollo:
> I did notice that there is both the `<mono>` tag and the `<style font=mono>` which is rather redundant.

Yes it's redundant. The idea beneath that is to have a style tag capable of everything about styling chars and have simpler b / i / mono / u tags for convenience purpose.

And you are right about buttons, tables and placement, these should all be later projects (they may imply lot of new code).

Sokomine:
Any progress here? I'm still hoping for better formspecs and a way to chat with mobs. This PR can do that. Please continue working on it and eventually merging!

pyrollo:
Hi, some work has been done on our side but we have to prepare it for the PR.

Also, some recent changes of fontengine have to be taken in account.

v-rob:
I was thinking something:  Is there any way to escape the tags?  Compatibility might be broken with labels for user-inputted data since tags won't be escaped.  Adding this functionality to `minetest.formspec_escape` wouldn't be a good idea because it would invalidly escape things for all elements, including those that don't use tags.  I think a new function, `minetest.hypertext_escape` which runs `minetest.formspec_escape` and then escapes the tags would be a good idea.

kilbith:
> I was thinking something: Is there any way to escape the tags?

Yes, see the demo mod (2nd formspec).

pyrollo:
Hi all,
PR has been rebased and should be now ready for merging (WIP label can be removed). 
No extra feature will be added before it is merged, no code change planned (except needed in order to merge).

src/gui/guiHyperText.cpp
```diff
@@ -0,0 +1,1165 @@
...
+#include <SColor.h>
+#include "client/tile.h"
+#include "IVideoDriver.h"
+#include "client/client.h"
+#include "client/renderingengine.h"
+#include "hud.h"
+#include "guiHyperText.h"
+#include "util/string.h"
+
+std::string strwtostr(const core::stringw &str)
```
rubenwardy: this should be in util/string

src/gui/guiHyperText.cpp
```diff
@@ -0,0 +1,1165 @@
...
+	if (str.empty())
+		return false;
+
+	char *endptr = nullptr;
+	strtol(str.c_str(), &endptr, 10);
+
+	return *endptr == '\0';
+}
+
+bool check_bool(const std::string &str)
```
rubenwardy: is_yes already exists

src/gui/guiHyperText.cpp
```diff
@@ -0,0 +1,1165 @@
...
+	wchar_t *text = new wchar_t[size + sizeof(wchar_t)];
+	const char *data = &str[0];
+
+	mbsrtowcs(text, &data, size, NULL);
+
+	text[size] = L'\0';
+	return text;
+}
+
+bool check_color(const std::string &str)
```
rubenwardy: perhaps an anonymous namespace would be good here?

src/gui/guiHyperText.cpp
```diff
@@ -0,0 +1,1165 @@
...
+{
+	return str == "true" || str == "false";
+}
+
+// -----------------------------------------------------------------------------
+// ParsedText - A text parser
+
+void ParsedText::Element::setStyle(StyleList &style)
+{
+	this->underline = (style["underline"] == "true");
```
rubenwardy: is_yes

rubenwardy (commented):
Look good

pyrollo:
Hi Rubenwardy,

Changes have been made accordingly.

src/gui/guiHyperText.cpp
pyrollo: Actually it was not exactly the same but check_bool removed (that makes hypertext syntax more permissive but still ok).

Sokomine:
Thanks for keeping working on this! I'm glad it has been added to 5.2.0. Hopefully it'll come then. I'm eagerly awaiting this vast improvement of the game.

SmallJoker:
The testing code works, although this part is a bit misleading:
> <action name="default:cobblestone">cobblestone</action>

It implies that spaces are possible within the `name` field, but it only accepts a single word.
Apart from that - the demo code works perfectly, and the callbacks as well. There's some space below the formspec without real coordinates but I don't think anyone cares about scaling there. It's an entire mess anyway.

paramat:
Please add the files to the clang format whitelist https://github.com/minetest/minetest/blob/master/util/travis/clang-format-whitelist.txt to silence the build fail. Most of the suggestions made by clang format are (as usual) useless and make the code far less readable. Some of what it reports may be valid so i suggest looking through it first for any valid suggestions.

SmallJoker:
Started Minetest with freetype disabled:
```
Thread 1 "minetest" received signal SIGSEGV, Segmentation fault.
0x00005555556f1fcc in FontEngine::getFont(unsigned int, FontMode, bool, bool) ()
(gdb) bt
#0  0x00005555556f1fcc in FontEngine::getFont(unsigned int, FontMode, bool, bool) ()
#1  0x00005555556f30af in FontEngine::readSettings() ()
#2  0x00005555556f3d21 in FontEngine::FontEngine(Settings*, irr::gui::IGUIEnvironment*) ()
#3  0x00005555556a7e45 in ClientLauncher::run(GameParams&, Settings const&) ()
```

pyrollo:
Thanks for the fixing! Introducing FontSpec is very clever :)

I cant find a way to mark "change requested" as done (they all seem to be outdated).

SmallJoker (approved):

rubenwardy (approved):

SmallJoker:
Rebased.

paramat:
I dismissed nerzhul's change request as is apparently resolved now, according to pyrollo on IRC.

paramat:
What is the delay? No conflicts, 2 approvals. Will merge in a few hours if no-one else has.

paramat:
I do not officially oppose this, but i must admit i am uncomfortable with this feature. It is impressive, but seems rather overkill and unnecessary. Normal formspecs can provide information pages, i do not consider hyperlinks enough to justify this amount of code and added maintenance.
It also seems odd to add so much to formspecs when we are unsure about the future and are considering replacing them.

Problem is, this was submitted without the author discussing with us first whether we wanted it, and now i feel that core devs are feeling slightly under pressure to accept this due to the amount of effort put into it. This is why it is important for large features to be discussed first, as the contributing guidelines recommend.

So, what do other core devs actually think about this, ignoring the issues of wasting a PR or offending the author? Is it actually justified and suitable?

kilbith, in future please consider discussing large contributions with us first by opening an issue. Also, i am sorry if my opinion offends. I still appreciate so much effort to contribute.

nerzhul (approved):

pyrollo:
Hi Paramat,

Yes formspecs can already display information pages but with a very poor text layout. That was the main purpose of this PR. The hyperlinks were then easy to add.

Sorry we did not discuss about it before propose it but we developed it based on a need we had, and then we thought it could really interest community (and according to modders it does).

In addition, I wrote the parser and renderer in a way that it could be reused elsewhere if needed.

PR #8682, opened 2019-07-16
Make players immortal using immortal=1

ClobberXD:
Adoption of #8266, originally by @Wuzzy.

****

Players too, are now immune to damage if `immortal` group is set to 1.

#### Implementation details

- Added `UnitSAO::isImmortal` returns true if special `immortal` group is set.

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -1532,7 +1532,9 @@ Another example: Make red wool from white wool and red dye:
 Special groups
 --------------
 
-* `immortal`: Disables the group damage system for an entity
+* `immortal`: Skips all damage and breath handling for an object. This group
+  will also hide the integrated HUD status bars for players, and is
+  automatically set to all players when damage is disabled on the server.
 * `punch_operable`: For entities; disables the regular damage mechanism for
   players punching it by hand or a non-tool item, so that it can do something
   else than take damage.
--- a/src/content_sao.cpp
+++ b/src/content_sao.cpp
@@ -134,7 +134,7 @@ void UnitSAO::setArmorGroups(const ItemGroupList &armor_groups)
 	m_armor_groups_sent = false;
 }
 
-const ItemGroupList &UnitSAO::getArmorGroups()
+const ItemGroupList &UnitSAO::getArmorGroups() const
 {
 	return m_armor_groups;
 }
@@ -1013,14 +1013,14 @@ void PlayerSAO::step(float dtime, bool send_recommended)
 		}
 	}
 
-	if (m_breathing_interval.step(dtime, 0.5f)) {
+	if (m_breathing_interval.step(dtime, 0.5f) && !isImmortal()) {
 		// Get nose/mouth position, approximate with eye position
 		v3s16 p = floatToInt(getEyePosition(), BS);
 		MapNode n = m_env->getMap().getNodeNoEx(p);
 		const ContentFeatures &c = m_env->getGameDef()->ndef()->get(n);
-		// If player is alive & no drowning & not in ignore, breathe
-		if (m_breath < m_prop.breath_max &&
-				c.drowning == 0 && n.getContent() != CONTENT_IGNORE && m_hp > 0)
+		// If player is alive & not drowning & not in ignore & not immortal, breathe
+		if (m_breath < m_prop.breath_max && c.drowning == 0 &&
+				n.getContent() != CONTENT_IGNORE && m_hp > 0)
 			setBreath(m_breath + 1);
 	}
 
@@ -1067,6 +1067,7 @@ void PlayerSAO::step(float dtime, bool send_recommended)
 		// create message and add to list
 		ActiveObjectMessage aom(getId(), true, str);
 		m_messages_out.push(aom);
+		m_env->getScriptIface()->player_event(this, "properties_changed");
 	}
 
 	// If attached, check that our parent is still there. If it isn't, detach.
@@ -1285,8 +1286,8 @@ int PlayerSAO::punch(v3f dir,
 
 	FATAL_ERROR_IF(!puncher, "Punch action called without SAO");
 
-	// No effect if PvP disabled
-	if (!g_settings->getBool("enable_pvp")) {
+	// No effect if PvP disabled or if immortal
+	if (isImmortal() || !g_settings->getBool("enable_pvp")) {
 		if (puncher->getType() == ACTIVEOBJECT_TYPE_PLAYER) {
 			std::string str = gob_cmd_punched(getHP());
 			// create message and add to list
--- a/src/content_sao.h
+++ b/src/content_sao.h
@@ -45,11 +45,12 @@ public:
 
 	inline bool isAttached() const
 	{ return getParent(); }
+
 	inline bool isImmortal() const
-	{ return itemgroup_get(m_armor_groups, "immortal"); }
+	{ return itemgroup_get(getArmorGroups(), "immortal"); }
 
 	void setArmorGroups(const ItemGroupList &armor_groups);
-	const ItemGroupList &getArmorGroups();
+	const ItemGroupList &getArmorGroups() const;
 	void setAnimation(v2f frame_range, float frame_speed, float frame_blend, bool frame_loop);
 	void getAnimation(v2f *frame_range, float *frame_speed, float *frame_blend, bool *frame_loop);
 	void setAnimationSpeed(float frame_speed);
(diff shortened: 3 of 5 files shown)


ClobberXD:
Fixed the merge conflicts, but I'm yet to work on [SmallJoker's suggestions](https://github.com/minetest/minetest/pull/8266#issuecomment-471178242)

src/script/lua_api/l_object.cpp
```diff
@@ -235,8 +235,14 @@ int ObjectRef::l_set_hp(lua_State *L)
 		return 0;
 
 	// Get HP
+	int old_hp = co->getHP();
 	int hp = lua_tonumber(L, 2);
 
+	// Protect from damage if immortal
+	if ((itemgroup_get(co->getArmorGroups(), "immortal") == 1) && (hp < old_hp)) {
```
sfan5: can't you use co->isImmortal() here?

ClobberXD: Yes, that's what I intend to do. This PR _should_ be ready for review by the evening.

ClobberXD:
PR's ready for review. Also implemented @SmallJoker's suggestions. Testing required.

src/script/lua_api/l_object.cpp
ClobberXD: Done

src/content_sao.h
```diff
@@ -251,6 +258,13 @@ class PlayerSAO : public UnitSAO
 	void setWantedRange(const s16 range);
 	s16 getWantedRange() const { return m_wanted_range; }
 
+	bool isImmortal()
+	{
+		return itemgroup_get(getArmorGroups(), "immortal") ||
+				!g_settings->getBool("enable_damage") ||
+				!g_settings->getBool("enable_pvp");
```
sfan5: players will no longer die from falling damage when PvP is disabled.
yeah, no

src/content_sao.h
```diff
@@ -106,31 +106,38 @@ class LuaEntitySAO : public UnitSAO
...
 	void moveTo(v3f pos, bool continuous);
 	float getMinimumSavedMovement();
 	std::string getDescription();
 	void setHP(s32 hp, const PlayerHPChangeReason &reason);
 	u16 getHP() const;
+
+	bool isImmortal()
+	{
+		return itemgroup_get(getArmorGroups(), "immortal") ||
+				!g_settings->getBool("enable_damage");
```
sfan5: I don't know if this works well either.
There may be value in allowing entities to be damaged even when players can't be.

src/script/lua_api/l_object.cpp
sfan5: still doesn't use `co->isImmortal()`

sfan5: Also, shouldn't this be in RemotePlayer/PlayerSAO instead of the script api?

src/content_sao.h
ClobberXD: Hmm, true. So PvP would have to be checked separately, I suppose. Good catch!

src/content_sao.h
ClobberXD: Agreed.

src/script/lua_api/l_object.cpp
ClobberXD: Oops :)

Yeah, I'll move this to PlayerSAO. I don't think the `hp < old_hp` check is required either - if an player is immortal, their HP shouldn't increase either. Or... should it?

ClobberXD: Nvm, the exact code already exists in `PlayerSAO::setHP` since a687efa6

src/content_sao.cpp
```diff
@@ -1020,7 +1020,7 @@ void PlayerSAO::step(float dtime, bool send_recommended)
 		const ContentFeatures &c = m_env->getGameDef()->ndef()->get(n);
 		// If player is alive & no drowning & not in ignore, breathe
 		if (m_breath < m_prop.breath_max &&
-				c.drowning == 0 && n.getContent() != CONTENT_IGNORE && m_hp > 0)
+				(isImmortal() || c.drowning == 0) && n.getContent() != CONTENT_IGNORE && m_hp > 0)
```
SmallJoker: This is a bit weird. I think breath values should freeze like the HP as soon the player becomes immortal. Mods can still re-fill the breath if they want to restore the player's mortality.

doc/lua_api.txt
```diff
@@ -1532,7 +1532,8 @@ Another example: Make red wool from white wool and red dye:
 Special groups
 --------------
 
-* `immortal`: Disables the group damage system for an entity
+* `immortal`: Disables the group damage system for an entity.
+  Disables all damage for a player.
```
SmallJoker: `Skips all damage and breath handling for a player. This group will also hide the integrated HUD status bars.`

Wuzzy2:
Note that the refilling breath was intentionally allowed in the original PR, only the decrease of breath was stopped.

Otherwise, the breath meter would oddly freeze at like 5 bubbles and neither decrease nor increase if you dive in water and drink the Potion of Immortability.

SmallJoker:
@Wuzzy2 It's not an issue if the breath stops re-filling because the HUD bars will hide automatically thanks to the new trigger called upon player object properties changes.

src/content_sao.cpp
ClobberXD: I agree. I'll remove this.

ClobberXD: Here's the modified condition:

```cpp
if (m_breath < m_prop.breath_max && c.drowning == 0 &&
		n.getContent() != CONTENT_IGNORE && m_hp > 0 && !isImmortal())
	setBreath(m_breath + 1);
```

Player "breathes" only if they're _not_ immortal.

ClobberXD:
@sfan5 Please update/dismiss your review.

src/content_sao.h
```diff
@@ -45,8 +45,9 @@ class UnitSAO: public ServerActiveObject
 
 	inline bool isAttached() const
 	{ return getParent(); }
-	inline bool isImmortal() const
-	{ return itemgroup_get(m_armor_groups, "immortal"); }
+
+	const bool isImmortal()
```
sfan5: the const here is useless
did you mean `bool isImmortal() const`?

also why not inline anymore?

src/server.cpp
```diff
@@ -1429,7 +1429,7 @@ void Server::SendPlayerHPOrDie(PlayerSAO *playersao, const PlayerHPChangeReason
 	if (playersao->isImmortal())
 		return;
 
-	session_t peer_id   = playersao->getPeerID();
+	session_t peer_id = playersao->getPeerID();
```
sfan5: should be reverted (no other changes in this file)

SmallJoker: It's partially related to the PR so why not improve this tiny bit too? :man_shrugging:

src/content_sao.cpp
```diff
@@ -1018,9 +1018,9 @@ void PlayerSAO::step(float dtime, bool send_recommended)
 		v3s16 p = floatToInt(getEyePosition(), BS);
 		MapNode n = m_env->getMap().getNodeNoEx(p);
 		const ContentFeatures &c = m_env->getGameDef()->ndef()->get(n);
-		// If player is alive & no drowning & not in ignore, breathe
-		if (m_breath < m_prop.breath_max &&
-				c.drowning == 0 && n.getContent() != CONTENT_IGNORE && m_hp > 0)
+		// If player is alive & not drowning & not in ignore & not immortal, breathe
+		if (m_breath < m_prop.breath_max && c.drowning == 0 &&
+				n.getContent() != CONTENT_IGNORE && m_hp > 0 && !isImmortal())
```
SmallJoker: `isImmortal()` check could be moved to the condition above in L1016.

src/server.cpp
ClobberXD: I agree with SmallJoker. I did initially touch some code in this file, which I later reverted. But I didn't revert this change, because god knows when this area of code would be touched next. ;)

src/content_sao.cpp
ClobberXD: Good idea!

src/content_sao.h
ClobberXD: I accidentally removed `inline`. Re-added, sorry.

I used `const bool` because the return value doesn't need to be modified. Doesn't it work that way?

And `isImmortal() const` throws the following error:

```
src/content_sao.h:50:40: error: passing ‘const UnitSAO’ as ‘this’ argument discards qualifiers [-fpermissive]
  { return itemgroup_get(getArmorGroups(), "immortal"); }
                                        ^
```

... which is why I removed it. It seems to make sense too; I just don't understand why it worked fine before.

ClobberXD: Fixed by making `getArmorGroups` a `const` method.

sfan5: > I used `const bool` because the return value doesn't need to be modified. Doesn't it work that way?

No. Even if it did, forbidding the caller from changing the returned value at some point afterwards would be stupid.

ClobberXD: I see... I'll make the return value non-const then.

PR #8688, opened 2019-07-17
[WIP] Sort server list by latency estimation

sfan5:
Sort the server list in the mainmenu by (estimated) latency from client to server.
This makes more sense since the ping reported by the serverlist is only useful for its own location (Netherlands, EU).
see also minetest/master-server#29

## To do

This PR is a Work in Progress

- [ ] figure out and implement determining own continent
- [ ] decide how exactly sorting should work, currently populated servers will be pushed down because they are further away (see below)

screenshot

## How to test

Edit `builtin/mainmenu/common.lua` line 66 with the continent code you reside on.
Then launch Minetest, switch to the multiplayer tab and see how the list is sorted.

<b>Please comment if you have suggestions regarding the open ToDo points.</b>

### DIFF
--- a/builtin/mainmenu/common.lua
+++ b/builtin/mainmenu/common.lua
@@ -63,20 +63,61 @@ function image_column(tooltip, flagname)
 end
 
 --------------------------------------------------------------------------------
-function order_favorite_list(list)
-	local res = {}
-	--orders the favorite list after support
-	for i = 1, #list do
-		local fav = list[i]
-		if is_server_protocol_compat(fav.proto_min, fav.proto_max) then
-			res[#res + 1] = fav
-		end
+local OWN_CONTINENT = "EU" -- TODO
+local latency_matrix = {
+	["AF"] = { ["AS"]=420, ["EU"]=160, ["NA"]=280, ["OC"]=450, ["SA"]=360, },
+	["AS"] = { ["EU"]=190, ["NA"]=170, ["OC"]=120, ["SA"]=300, },
+	["EU"] = { ["NA"]=120, ["OC"]=380, ["SA"]=200, },
+	["NA"] = { ["OC"]=200, ["SA"]=130, },
+	["OC"] = { ["SA"]=330, },
+}
+
+function estimate_favorite_latency(fav)
+	if OWN_CONTINENT == nil then return nil end
+	local here, there = OWN_CONTINENT, fav.geo_continent
+	if here == there then
+		return 0
 	end
+	return here > there and latency_matrix[there][here] or latency_matrix[here][there]
+end
+
+--------------------------------------------------------------------------------
+local function favorite_list_sort_key(fav)
+	local k = ""
+	-- protocol support
+	if is_server_protocol_compat(fav.proto_min, fav.proto_max) then
+		k = k .. "0"
+	else
+		k = k .. "1"
+	end
+	-- estimated latency (geographic)
+	if fav.geo_continent ~= nil then
+		local ms = estimate_favorite_latency(fav) or 999
+		k = k .. string.format("%03d", ms)
+	else
+		k = k .. "999"
+	end
+	return k
+end
+
+function order_favorite_list(list)
+	-- sort list of indexes first
+	local idx = {}
 	for i = 1, #list do
-		local fav = list[i]
-		if not is_server_protocol_compat(fav.proto_min, fav.proto_max) then
-			res[#res + 1] = fav
+		idx[#idx + 1] = i
+	end
+	table.sort(idx, function(a, b)
+		local key1 = favorite_list_sort_key(list[a])
+		local key2 = favorite_list_sort_key(list[b])
+		if key1 == key2 then
+			return a < b -- preserve original order
 		end
+		return key1 < key2
+	end)
+	-- assemble result table
+	local res = {}
+	for _, i in ipairs(idx) do
+		res[#res + 1] = list[i]
 	end
 	return res
 end
@@ -104,6 +145,11 @@ function render_serverlist_row(spec, is_favorite)
 
 	if spec.ping then
 		local ping = spec.ping * 1000
+		if spec.geo_continent ~= nil and ping < 400 then
+			-- with a ping over 400ms, the server obviously has latency problems
+			-- no matter which latency we estimate
+			ping = estimate_favorite_latency(spec) or ping
+		end
 		if ping <= 50 then
 			details = details .. "2,"
 		elseif ping <= 100 then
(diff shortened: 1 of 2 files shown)


ClobberXD:
IMO, latency shouldn't be the only factor that decides the order of servers displayed. AFAIK, there's some currently smart code that ranks servers based on uptime, number of players, etc. Servers should be sorted like how they are now, and then servers with (near-)equal scores could be sorted according to latency.

builtin/mainmenu/common.lua
```diff
@@ -63,20 +63,61 @@ function image_column(tooltip, flagname)
...
+local latency_matrix = {
+	["AF"] = { ["AS"]=420, ["EU"]=160, ["NA"]=280, ["OC"]=450, ["SA"]=360, },
+	["AS"] = { ["EU"]=190, ["NA"]=170, ["OC"]=120, ["SA"]=300, },
+	["EU"] = { ["NA"]=120, ["OC"]=380, ["SA"]=200, },
+	["NA"] = { ["OC"]=200, ["SA"]=130, },
+	["OC"] = { ["SA"]=330, },
+}
+
+function estimate_favorite_latency(fav)
+	if OWN_CONTINENT == nil then return nil end
```
ClobberXD:
```suggestion
	if not OWN_CONTINENT then
		return nil
	end
```

Also are the `== nil` checks necessary? Could `not <var>` be used instead?

sfan5: I prefer being explicit about checking for `nil`
also that code is subject to change anyway

ClobberXD: Ok, that's fine with me then.

HybridDog:
> figure out and implement determining own continent

I have seen country detection in sauerbraten.
https://github.com/pisto/spaghettimod/blob/26e3578e74ccd8a4d9f6e2fa60c609760ce4443c/script/load.d/100-geoip.lua#L24
https://github.com/pisto/spaghettimod/blob/26e3578e74ccd8a4d9f6e2fa60c609760ce4443c/updategeoip

sfan5:
> AFAIK, there's some currently smart code that ranks servers based on uptime, number of players, etc.

Yes, that code is in the server list. However the client does not have access to the calculated points, only the sorted result.

> I have seen country detection in sauerbraten.

<i>how</i> isn't the problem, what this needs is an additional API with the master server and fetching (and caching) this in the client

sofar:
> +	["AF"] = { ["AS"]=420, ["EU"]=160, ["NA"]=280, ["OC"]=450, ["SA"]=360, },
> +	["AS"] = { ["EU"]=190, ["NA"]=170, ["OC"]=120, ["SA"]=300, },
> +	["EU"] = { ["NA"]=120, ["OC"]=380, ["SA"]=200, },
> +	["NA"] = { ["OC"]=200, ["SA"]=130, },
> +	["OC"] = { ["SA"]=330, },

This should use real world values and not rounded guessed values.

We previously discussed in IRC using https://wondernetwork.com/pings.
This would give us (1) *actual* data and (2) a way to correct or update
the values in a consistent way in the future.

We did this before and it clearly shows that the above table has really
inaccurate values. For instance, Lagos-Brasilia = 180ms, not 360.
Amsterdam-Auckland = 273ms, not 380.

Cities I used (mostly "central" to the region/continent):
Amsterdam
Auckland
Brasilia
Denver
Lagos
Singapore

Result:

from/to:   Amsterd.  Auckland  Brasilia  Denver    Lagos    Singap.
Amsterdam  —         273.04    229.211   122.401   107.028  166.288
Auckland   273.134   —         325.579   164.543   378.538  114.919
Brasilia   223.223   326.015   —         219.826   186.98   376.485
Denver     122.334   164.522   214.528   —         225.117  212.997
Lagos      107.24    378.493   182.005   226.16    —        356.011
Singapore  165.895   114.894   374.544   212.662   355.977  —

This is quality data that's been filtered and should be reliably
available in the future should new cables be pulled :)

sofar:
BTW, I'd really like the location to be one factor in sorting - IOW
highly scoring servers should, no matter the location, still show higher
on the list.

The simple reason is that aside NA/EU, I wouldn't want the server list
to appear with just empty servers entirely for players in e.g. OC and
AF.

sfan5:
> This should use real world values and not rounded guessed values.

These are neither "rounded", "guessed" nor not "real world" values.
But I agree that it's worth using accurate, high quality values.

> I'd really like the location to be one factor in sorting - IOW
highly scoring servers should, no matter the location, still show higher
on the list.

This wasn't my intention, just the quickest thing I could make work as a PoC.
The client (currently) has no access to the calculated score by the server list, so "weighing in" the estimated latency is not easily possible.

MoNTE48:
A possible option (what I thought to use in MC) is to transfer the score to the client and sort the servers by the client. Then the client will deduct points for very distant servers and add 1+ points for local servers.
Then the player from France will have more chances to get to the French server, the player from Russia - to Russian.

sofar:
https://github.com/minetest/master-server/pull/34 implements a client lookup mechanism. The interface returns `NA` or `EU` etc. and a HTTP status code.

sofar:
> The client (currently) has no access to the calculated score by the server list, so "weighing in" the estimated latency is not easily possible.

I have a solution: weighted normalized ranking. It's relatively easy to implement, but requires a bit of iterating. math is trivial.

Step 1: calculate the normalized ranking based on order from the master server. The first one returned is given `1.0`, the last one is given `0.0`. If there would be 3 servers in the list, the middle one would receive `0.5`. If there are exactly 100 servers in the list, they would be assigned `1.0`, `0.99`, `0.98` etc. You get the idea, hopefully - this is a "normalized" ranking. each server will have it's own score - there will not be any servers with the same value here.

Step 2: calculate the from-to delay for each server.

Step 3: rank each server according to the from-to delay, lowest first. Then, normalize this ranking as well just like step 1. Note: here lots of servers will now have the SAME ranking - e.g. 1.0 for all servers that are in the closest continent, then 0.8 for the next contentent servers, etc..

Step 4: weigh the 2 rankings to create a new, final ranking. This can be 1:1 weighed, which means that both rankings contribute as much "weight" to the end ranking. Or it could be that you weigh it 2:1 or 1:2, or any other number. Then sort the list according to the final ranking.

However you choose, there will be an obvious preference in the list for servers that score high on both rankings.

This exact method is used on ITB to create https://minetest.foo-projects.org/rankings.html. I do not foresee issues doing this in lua for ~500 servers, since all the data is known and ready to iterate.

Some hypothetical examples based on 1:1 (adding the 2 rank values)
server A ranks 1.0 on the server list, but 0.2 on the geoip ranking. the end rank is 1.2.
server B ranks 0.90 on the server list, but 1.0 on the geoip ranking, the end rank is 1.9
server C ranks 0.1 on the server list, but 1.0 on the geoip ranking, the end rank is 1.1
server D ranks 0.3 on the server list, but 0.4 on the geoip ranking, the end rank is 0.7
End result sort order to the client: B, A, C, D

PR #8689, opened 2019-07-17
Don't unnecessarily clamp HP and breath to their maximums in set_properties().

shacknetisp:
This PR's change prevents `set_properties()` calls that have nothing to do with `hp_max` or `breath_max` overriding the saved hp before another mod has the chance to set a player's intended `hp_max` (such as in `on_joinplayer`).

In my case, `player_api` in MTG calls `set_properties()` on the player when they join, overwriting player HP before I had a chance to set the `hp_max` property.

### DIFF
--- a/src/content_sao.cpp
+++ b/src/content_sao.cpp
@@ -331,7 +331,7 @@ void LuaEntitySAO::addedToEnvironment(u32 dtime_s)
 	if(m_registered){
 		// Get properties
 		m_env->getScriptIface()->
-			luaentity_GetProperties(m_id, &m_prop);
+			luaentity_GetProperties(m_id, this, &m_prop);
 		// Initialize HP from properties
 		m_hp = m_prop.hp_max;
 		// Activate entity, supplying serialized state
--- a/src/script/common/c_content.cpp
+++ b/src/script/common/c_content.cpp
@@ -21,6 +21,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include "common/c_types.h"
 #include "nodedef.h"
 #include "object_properties.h"
+#include "content_sao.h"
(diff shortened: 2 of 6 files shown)


SmallJoker:
I see. This was really missed in a previous PR.
For a clean solution I recommend to do one of these:

* Pass `ServerActiveObject` to `read_object_properties` so that important properties can be applied instantly. Sending happens on update
* Move this HP and breath handling partially to `PlayerSAO::getPropertyPacket()`. That's when the player receives all other property information. (preferred solution)

shacknetisp:
@SmallJoker like this?

shacknetisp:
hmm, objects other than player's won't be covered here...

SmallJoker:
@shacknetisp I think you forgot to commit your changes before pushing. You can handle other object types in `LuaEntitySAO::getPropertyPacket()` if you decided for the 2nd approach. The first one will need casting to PlayerSAO (like it's already done in `l_set_properties`) to "send HP or die".

shacknetisp:
2nd approach does break immediate application of properties, so I'll implement the 1st.
(PR is currently reset back to the original solution.)

SmallJoker:
FYI: For the client the properties aren't applied until after they receive the packet from `getPropertyPacket`.

shacknetisp:
Yeah, I'm thinking about server-side like
```
obj:set_properties{hp_max = 5}
obj:set_hp(5) -- HP is 5 at this point.
obj:set_properties{hp_max = 2} -- Assuming properties are not immediately applied.
obj:get_hp() -- Still 5, HP is greater than hp_max!
```

shacknetisp:
Ok, here's one that immediately applies in read_object_properties(). Hopefully I'm doing everything right.

PR #8691, opened 2019-07-19
Give the online lua mainmenu also the client_list and mods

Lejo1:
The lua Mainmenu interface now gets also the clients and the mods!
This allows new Features just like listing players(maybe tooltip) or searching.
Finding servers with the wanted mods and more...

### DIFF
--- a/src/script/lua_api/l_mainmenu.cpp
+++ b/src/script/lua_api/l_mainmenu.cpp
@@ -265,8 +265,8 @@ int ModApiMainMenu::l_get_favorites(lua_State *L)
 {
 	std::string listtype = "local";
 
-	if (!lua_isnone(L,1)) {
-		listtype = luaL_checkstring(L,1);
+	if (!lua_isnone(L, 1)) {
+		listtype = luaL_checkstring(L, 1);
 	}
 
 	std::vector<ServerListSpec> servers;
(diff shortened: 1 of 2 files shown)


src/script/lua_api/l_mainmenu.cpp
```diff
@@ -391,6 +391,37 @@ int ModApiMainMenu::l_get_favorites(lua_State *L)
 			lua_settable(L, top_lvl2);
 		}
 
+		if (server["clients_list"].isArray()) {
+			unsigned int index_lvl2 = 1;
+			lua_pushstring(L,"clients_list");
```
SmallJoker: Space after comma. The old code wasn't checked for code style.

SmallJoker:
Sure, why not? Please update the docs accordingly:
https://github.com/minetest/minetest/blob/a6f784835244111c90c05b6aae47092c2c5763a7/doc/menu_lua_api.txt#L153-L155

Lejo1:
Done.

Enrikoo:
I have no Idea how this works I have tried to search for clients and It doesnt find. So how does it work?

Lejo1:
> I have no Idea how this works I have tried to search for clients and It doesnt find. So how does it work?

It's just a way this change can be used for.

Enrikoo:
So I tried to find for players with the search button but it didnt list players, only servers. How can I find for something different?

Lejo1:
> So I tried to find for players with the search button but it didnt list players, only servers. How can I find for something different?

You understand me wrong. You can't find a player list. But maybe a playerlist will be added soon.

Enrikoo:
Yes I cant understand this, so only the code has been changed or does it recognize game versions better, but it does not have to do with mainmenu.

Lejo1:
It **maybe** if added will be available in the next release.

Enrikoo:
Okay I have to wait, I can't do it yet.

PR #8709, opened 2019-07-24
Unify OpenGL ES to be a platform-independent cmake switch

sfan5:
This allows you to use OpenGL ES on desktop machines, for whatever reason you'd want that.
Would also help to support platforms such as the Raspberry Pi.

## To do
This PR is Ready for Review.

## How to test
I hate this section.
* grab irrlicht ogles from `svn://svn.code.sf.net/p/irrlicht/code/branches/ogl-es`
* edit `IrrCompileConfig.h` to disable opengl and ogles1 drivers
* compile and fix the egl.h include path to be `<EGL/egl.h>` if an error occurrs
* compile minetest with `-DENABLE_GLES=1` and the paths to your built irrlicht
* copy shaders from `<irrlicht>/media/Shaders` to `client/shaders/Irrlicht`
* start minetest and observe that it works

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -102,10 +102,18 @@ if(BUILD_CLIENT AND ENABLE_SOUND)
 endif()
 
 
-option(ENABLE_GLES "Enable OpenGL ES support" FALSE)
+option(ENABLE_GLES "Use OpenGL ES instead of OpenGL" FALSE)
 mark_as_advanced(ENABLE_GLES)
 if(ENABLE_GLES)
-	find_package(OpenGLES2)
+	find_package(OpenGLES2 REQUIRED)
+elseif()
+	if(NOT WIN32) # Unix probably
+		set(OPENGL_GL_PREFERENCE "LEGACY" CACHE STRING
+			"See CMake Policy CMP0072 for reference. GLVND is broken on some nvidia setups")
+		set(OpenGL_GL_PREFERENCE ${OPENGL_GL_PREFERENCE})
+
+		find_package(OpenGL REQUIRED)
+	endif()
 endif()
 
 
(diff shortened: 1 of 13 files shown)


src/CMakeLists.txt
```diff
@@ -543,6 +544,18 @@ if(BUILD_CLIENT)
...
+	if(ENABLE_GLES)
+		target_link_libraries(
+			${PROJECT_NAME}
+			${OPENGLES2_LIBRARIES}
+			${EGL_LIBRARIES}
+		)
+	else()
+		target_link_libraries(
+			${PROJECT_NAME}
+			${OPENGL_LIBRARIES}
```
SmallJoker: OpenGL could theoretically still be used on OGLES2 builds. How about including both, if they're provided?

SmallJoker:
My experimental patch isn't that far off from this PR. OGLES2 also works (but is quite slow) on desktop/laptop computers.

sfan5:
> OGLES2 also works (but is quite slow) on desktop/laptop computers

Not for me. `ogles2` yields identical performance to OpenGL on my system with the NVIDIA proprietary driver.

src/CMakeLists.txt
sfan5: The problem is that `ENABLE_GLES` changes some parts of the code, whether you use GLES at runtime or not.
It'd probably still work, but I don't really want to enable this until Minetest (one day) properly has coexisting GL/GLES support.

DanielTheGreatinc:
I will try this on my Raspberry Pi.

DanielTheGreatinc:
I have encountered an error: `minetest-stable-5/src/client/sky.cpp:53:21: error: cannot convert ‘bool’ to ‘irr::video::E_ZWRITE’ in assignment
  mat.ZWriteEnable = false;`

sfan5:
Sounds like Irrlicht is too new, try downgrading to revision `5811`.
Also please open an issue about Irrrlicht 1.9 support so I don't forget.

Issue #8402, opened 2019-03-19
Recipes removed by clear_craft are still returned by get_all_craft_recipes


FaceDeer:
##### Issue type
 - Bug report
 
##### Minetest version
```
minetest-5.0.0-win64
```

##### OS / Hardware
Operating system: Windows 10 Home
CPU: Intel i7-3770K @ 3.5GHz, 8.00 GB ram

##### Summary

I've found that using minetest.clear_craft to remove a recipe doesn't seem to actually remove it as far as minetest.get_all_craft_recipes is concerned, subsequent calls to minetest.get_all_craft_recipes will still return the "cleared" recipe.

##### Steps to reproduce

Attached is a simple mod that clears all already-registered recipes by calling get_all_craft_recipes for each registered item and clearing the returned recipe, and then immediately repeats the task. The second time it runs through the process it produces a spew of "WARNING[Main]: No craft recipe matches input" log outputs.

[clear_craft_test.zip](https://github.com/minetest/minetest/files/2981484/clear_craft_test.zip)

This is a snippet taken from my [simplecrafting_lib](https://github.com/minetest-mods/simplecrafting_lib/tree/toward_1.0) if anyone's curious about the context. I'm "importing" recipes into a custom crafting system and the import process can in some circumstances be run multiple times.

PR #8712, opened 2019-07-25
Fix some issues with minetest.clear_craft

pauloue:
Fixes https://github.com/minetest/minetest/issues/8402 and https://github.com/minetest/minetest/issues/8427

- Fix memory leak (checked with Valgrind).
- Fix crafts with an output count not being cleared when clearing by input (#8402).
- Fix recipe list being reversed when clearing by input (#8427).

## To do

This PR is Ready for Review.

## How to test

- Run the minimal game tests.


src/craftdef.cpp
```diff
@@ -963,84 +961,48 @@ class CCraftDefManager: public IWritableCraftDefManager
...
-			/* We will preallocate necessary memory addresses, so we don't need to reallocate them later.
-				This would save us some performance. */
-			new_vec_by_input.reserve(unhashed_inputs_vec.size());
-			for (auto &i2 : unhashed_inputs_vec) {
-				if (def != i2) {
-					new_vec_by_input.push_back(i2);
-				}
-			}
-			m_craft_defs[(int) CRAFT_HASH_TYPE_UNHASHED][0].swap(new_vec_by_input);
+			std::vector<CraftDefinition *> &defs = m_craft_defs[CRAFT_HASH_TYPE_UNHASHED][0];
```
SmallJoker: I believe `(int)` was required here to not break MSVC.

src/craftdef.cpp
```diff
@@ -963,84 +961,48 @@ class CCraftDefManager: public IWritableCraftDefManager
...
+				new_defs.push_back(def);
 				continue;
 			}
-			CraftOutput output = def->getOutput(input, gamedef);
 			got_hit = true;
-			auto vec_iter = m_output_craft_definitions.find(output.item);
-			if (vec_iter == m_output_craft_definitions.end())
+			std::string output = def->getOutput(input, gamedef).item;
+			delete def;
+			auto it = m_output_craft_definitions.find(craftGetItemName(output, gamedef));
```
SmallJoker: Wouldn't converting `output` to an itemstack to get its name suffice?

pauloue: Isn't that just what `craftGetItemName` does?
It would be better if `CraftOutput` and `CraftDefinition` stored itemstacks instead of strings.

pauloue:
Rebased (moved crafting tests to new file)

PR #8716, opened 2019-07-27
Fix the bgcolor formspec element

### LINKED ISSUE #8308: //usr/lib/i386-linux-gnu/libharfbuzz.so.0: Warnung: undefinierter Verweis auf »FT_Get_Var_Blend_Coordinates«
##### Issue type

 - Build issue
 
##### Minetest version

```
actual Git Version from today
```

##### OS / Hardware

Operating system: Kubuntu 18.04 LTS (32Bit)
CPU: Intel

GPU model:
OpenGL version:

##### Summary

//usr/lib/i386-linux-gnu/libharfbuzz.so.0: Warnung: undefinierter Verweis auf »FT_Get_Var_Blend_Coordinates«
collect2: error: ld returned 1 exit status
src/CMakeFiles/minetest.dir/build.make:7108: recipe for target 'bin/minetest' failed
make[2]: *** [bin/minetest] Error 1
CMakeFiles/Makefile2:211: recipe for target 'src/CMakeFiles/minetest.dir/all' failed
make[1]: *** [src/CMakeFiles/minetest.dir/all] Error 2
Makefile:151: recipe for target 'all' failed
make: *** [all] Error 2

##### Steps to reproduce

after updating my Kubuntu 16.04 to 18.04 I can´t compile anymore.

Desour:
- Goal of the PR: making `bgcolor` more useful and powerful
- The PR works by handling the bgcolor of the window separated from the fullscreen bgcolor. The old behaviour, where the window bgcolor was not shown if fullscreen bgcolor was activated (which is weird and unintuitive), can still be achieved by setting the window bgcolor to "#0000".
- ~~Fixes #8308.~~ Edit: Oops, wrong issue. I wanted to refer to #8336.
The change is taken form #8591.

## To do

As it is now possible to set window and fullscreen bgcolor at once, minetest_game formspec, for example, might look a little different, but this is only a little visual change and can be fixed easily. 

## How to test

```lua
local my_formspec =
	"size[10,8]"..
	"bgcolor[;true]".. -- modify this however you want
	"bgcolor[#0000]"..
	""

minetest.register_node("test:node", {
	description = "test node",
	tiles = {"default_wood.png^heart.png"},
	groups = {choppy = 3, oddly_breakable_by_hand = 3},
		minetest.show_formspec(clicker:get_player_name(), "my_formspec", my_formspec)
	end,
})
```

### DIFF
--- a/src/gui/guiFormSpecMenu.cpp
+++ b/src/gui/guiFormSpecMenu.cpp
@@ -1795,17 +1795,17 @@ void GUIFormSpecMenu::parseBox(parserData* data, const std::string &element)
 	errorstream<< "Invalid Box element(" << parts.size() << "): '" << element << "'"  << std::endl;
 }
 
-void GUIFormSpecMenu::parseBackgroundColor(parserData* data, const std::string &element)
+void GUIFormSpecMenu::parseBackgroundColor(parserData *data, const std::string &element)
 {
 	std::vector<std::string> parts = split(element,';');
 
 	if (((parts.size() == 1) || (parts.size() == 2)) ||
 			((parts.size() > 2) && (m_formspec_version > FORMSPEC_API_VERSION))) {
-		parseColorString(parts[0], m_bgcolor, false);
-
-		if (parts.size() == 2) {
-			std::string fullscreen = parts[1];
-			m_bgfullscreen = is_yes(fullscreen);
+		if (parts.size() == 1) {
+			parseColorString(parts[0], m_bgcolor, false);
+		} else if (parts.size() == 2) {
+			parseColorString(parts[0], m_fullscreen_bgcolor, false);
+			m_bgfullscreen = is_yes(parts[1]);
 		}
 
 		return;
@@ -2775,8 +2775,7 @@ void GUIFormSpecMenu::drawMenu()
 
 	if (m_bgfullscreen)
 		driver->draw2DRectangle(m_fullscreen_bgcolor, allbg, &allbg);
-	else
-		driver->draw2DRectangle(m_bgcolor, AbsoluteRect, &AbsoluteClippingRect);
+	driver->draw2DRectangle(m_bgcolor, AbsoluteRect, &AbsoluteClippingRect);
 
 	m_tooltip_element->setVisible(false);
 
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2037,11 +2037,15 @@ Elements
 
 * Show an inventory image of registered item/node
 
-### `bgcolor[<color>;<fullscreen>]`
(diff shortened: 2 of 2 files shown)


SmallJoker:
After `no_prepend[]` I can confirm this works just fine. A beauty.

grafik

SmallJoker (approved):

DS-Minetest:
If it's wanted, I can keep the old behaviour in formspec version < 2. But this would mean that the old code has to be maintained and that the api reference gets bloated. Furthermore, the old behaviour is very buggy and not really correctly documented. And the change only affects visuals.

rubenwardy:
I think that this doesn't need to be formspec_version only

rubenwardy (approved):

DS-Minetest:
I just realized that I wrote "Fixes #8308." instead of "Fixes #8336.", oops. Latter issue can be closed.

Wuzzy2:
Thank you. Although easiliy fixable by game/mod authors, this is technically a breaking change. This should be mentioned and highlighted in the changelog so that modders are not caught by surprise.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -1795,17 +1795,17 @@ void GUIFormSpecMenu::parseBox(parserData* data, const std::string &element)
...
 			((parts.size() > 2) && (m_formspec_version > FORMSPEC_API_VERSION))) {
-		parseColorString(parts[0], m_bgcolor, false);
-
-		if (parts.size() == 2) {
-			std::string fullscreen = parts[1];
-			m_bgfullscreen = is_yes(fullscreen);
+		if (parts.size() == 1) {
+			parseColorString(parts[0], m_bgcolor, false);
+		} else if (parts.size() == 2) {
+			parseColorString(parts[0], m_fullscreen_bgcolor, false);
```
SmallJoker: `m_bgcolor` should likely be used if `m_bgfullscreen` below turns out to be `false`.

DS-Minetest: See https://github.com/minetest/minetest/issues/8985#issuecomment-534739062.

The requested change could be done in a new PR.

kilbith:
So since this commit this has become impossible to hide this from every formspec: :-1: 

screenshot_20190926_155651

Code: https://github.com/minetest-mods/craftguide/blob/master/init.lua#L707

rubenwardy:
See this issue https://github.com/minetest/minetest/issues/8985

DS-Minetest:
>impossible to hide

`bgcolor[#0000]`

kilbith:
Thanks, but then some mods are broken anyway.

PR #8717, opened 2019-07-27
Rename guiScrollBar to GUIScrollBar

DS-Minetest:
According to https://dev.minetest.net/Code_style_guidelines#Classes, class names should be PascalCase. Ergo the class `guiScrollBar` has to be renamed to `GUIScrollBar`.

## To do

## How to test

Just compile.
(It compiles for me, so I guess there's no mistake.)


DS-Minetest:
Rebased.

PR #8725, opened 2019-07-31
Add support for mesh items

### LINKED ISSUE #3551: Add support for mesh items

ClobberXD:
(See #3551)

This PR allows items to have a mesh visual by optionally defining the `mesh` field and a `textures` table in the item definition (like in the `initial_properties` table of the entity definition).

e.g. 
```lua
minetest.register_item("test:mesh", {
	description = "This is a mesh item :D",
	mesh = "character.b3d",
	textures = { "character.png" }
})
```

This PR is strictly WIP, and I haven't compiled the code even once. I've just made a draft PR for discussion, and for obtaining inputs and guidance.

This PR includes a bunch of code-style fixes. I promise I'll move these into a separate commit once this PR becomes ready for review. ;)

### TODO

- Actually finish implementing mesh items.

Closes #3551.

### DIFF
--- a/builtin/game/features.lua
+++ b/builtin/game/features.lua
@@ -13,6 +13,7 @@ core.features = {
 	object_use_texture_alpha = true,
 	object_independent_selectionbox = true,
 	httpfetch_binary_data = true,
+	mesh_items = true,
 }
 
 function core.has_feature(arg)
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -3657,6 +3657,7 @@ Utilities
           -- Specifies whether binary data can be uploaded or downloaded using
           -- the HTTP API (5.1.0)
           httpfetch_binary_data = true,
+          mesh_items = true,
       }
 
 * `minetest.has_feature(arg)`: returns `boolean, missing_features`
@@ -6015,6 +6016,11 @@ Used by `minetest.register_node`, `minetest.register_craftitem`, and
         --      {soil = 2, outerspace = 1, crumbly = 1}
         --      {bendy = 2, snappy = 1},
         --      {hard = 1, metal = 1, spikes = 1}
+
+        mesh = "",
+        -- Requires `textures` to be defined.
+
+        textures = {},
 
         inventory_image = "default_tool_steelaxe.png",
 
--- a/src/client/camera.cpp
+++ b/src/client/camera.cpp
@@ -56,8 +56,8 @@ Camera::Camera(MapDrawControl &draw_control, Client *client):
 	// all other 3D scene nodes and before the GUI.
 	m_wieldmgr = smgr->createNewSceneManager();
 	m_wieldmgr->addCameraSceneNode();
-	m_wieldnode = new WieldMeshSceneNode(m_wieldmgr, -1, false);
-	m_wieldnode->setItem(ItemStack(), m_client);
+	m_wieldnode = new WieldMeshSceneNode(m_wieldmgr, m_client, -1, false);
+	m_wieldnode->setItem(ItemStack());
 	m_wieldnode->drop(); // m_wieldmgr grabbed it
 
 	/* TODO: Add a callback function so these can be updated when a setting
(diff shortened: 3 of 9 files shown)


ClobberXD:
I've moved some changes (incl. the majority of the code-style fixes atm) to #8726

DELTA-FORCE-CTF:
I would love to see this implemented. I assume you would also have to modify the wielded shader, or no?

ClobberXD:
The shaders don't need to be modified at all. Clients, however, do need to be up to date, and older clients could be broken if the server supports mesh items.

DS-Minetest:
May I request that it should possible to define the mesh used for wielding with the hand separate to the mesh used in inventories?

ClobberXD:
> May I request that it should possible to define the mesh used for wielding with the hand separate to the mesh used in inventories?

That's exactly what I initially set as my objective when starting out, but as I progressed, I noticed that a separate `wield_mesh` requires a separate set of textures (`wield_textures = { "wield_texture.png" }`). This seemed to be somewhat redundant to me, so I dropped it. I'll continue to finish implementing _one_ mesh and its textures first, and then add a second `wield_mesh` if it gains enough support.

paramat:
Random thought:
Should / could nodebox items be supported too? Nodeboxes are much easier to create (meshes are beyond the ability of most modders) so perhaps are higher in priority for supporting?

theviper121:
I am interested in this feature! Any chance of it being completed?

ClobberXD:
> I am interested in this feature! Any chance of it being completed?

This should be ready in a couple of weeks, if everything goes as planned IRL. :) (although it never does)

ClobberXD:
Small update: Documentation has been added (I'm open to requests for changes in documentation). I've also added `mesh_items = true` to `minetest.features`, because there's no other way to check that mesh items are supported. Do let me know if this is not the case.

Still working on the core of this PR, btw. :)

ClobberXD:
Very close to completion! In the meanwhile, is it a good idea to use the `wield_image` as the texture for the mesh instead of adding a new `textures` field? Note that `wield_image` won't be used anyway if `mesh` is defined.

I'm also assuming one texture would suffice, for the moment. I can add support for multiple textures at a later point in time, if the feature is requested.

ClobberXD:
Got the core functionality to work, but I'm unable to get the inventory mesh render properly. This is my first time working with Irrlicht code, and I don't understand much of it. Any help, however small, is much much appreciated. The following screenshots were taken on the latest force-pushed version of the PR:

screenshot_20200129_133142
screenshot_20200129_133147
screenshot_20200129_133225

All of the wielded items in the screenshots are actual meshes, present in MTG. The mod I've used to test this can be found at https://github.com/ClobberXD/wield_mesh

One other trivial issue I faced is that I wasn't able to arrive upon a correct default wield scale for mesh items, since they vary wildly unlike cube-nodes and extruded meshes. So I settled upon a default scale of 5.

ClobberXD:
Greetings. It'd be great if someone could help me with the following:

> ... but I'm unable to get the inventory mesh render properly. ...

(See previous comment)

LoneWolfHT:
Can't borrow from node rendering code?

ClobberXD:
> Can't borrow from node rendering code?

Do you mean the code that renders nodes in inventory slots? As far as I can tell, inventory mesh rendering (of all types, be it extruded mesh, mesh items, nodes, nodeboxes, etc.) uses the mesh returned by the standalone function `getItemMesh` in `wieldmesh.cpp`. I can make out the most obvious stuff, like the fact that `ItemMesh::mesh` corresponds to the mesh to be displayed in the inv. slot. What I don't understand is the stuff like `ItemMesh::buffer_colors` and the like.

Now, assuming the best case scenario, I suspect I've made small mistakes here and there in spite of getting the overall code right. Worst comes to worst, I've probably implemented this feature in the worst way possible, owing to my lack of understanding. Which is why I'd like someone a little more familiar with IrrLicht (or rendering in general) to take a look at the code and provide inputs and suggestions.

appgurueu:
> Can't borrow from node rendering code?

Take a look at `drawItemStack`, you might need to modify it.

PR #8726, opened 2019-07-31
ItemDefinition: Fix inventory_overlay and wield_overlay not being (de)serialized

ClobberXD:
While working on #8725, I noticed that `inventory_overlay` and `wield_overlay` aren't serialized while (de)serializing an `ItemDefinition` struct. Here's the commit that added `inventory_overlay` and `wield_overlay`: f6a33a1

This trivial PR:

- Adds the missing serialize and deserialize calls
- Fixes some code-style in `itemdef.h` and `itemdef.cpp`

Please let me know if these fields shouldn't be (de)serialized for some reason, or if I have to modify a couple other places that depend on the serialization of this struct.

****

I'd prefer to have this PR merged without squashing, if possible.

### DIFF
--- a/src/itemdef.cpp
+++ b/src/itemdef.cpp
@@ -56,7 +56,7 @@ ItemDefinition::ItemDefinition(const ItemDefinition &def)
 	*this = def;
 }
 
-ItemDefinition& ItemDefinition::operator=(const ItemDefinition &def)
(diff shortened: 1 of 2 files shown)


ClobberXD:
Aha, stupid me :D

`inventory_overlay` and `wield_overlay` are indeed (de)serialized, but a little further down the (de)serialize method, after checking for version.

PR #8733, opened 2019-08-01
Correct ServerActiveObject's virtual getArmorGroups() to be const.

shacknetisp:
Due to commit ec3142a , UnitSAO's getArmorGroups() did not match
ServerActiveObject's, notably resulting in the lua get_armor_groups() call
returning nothing.


SmallJoker:
@ClobberXD :angry: :anger:

ClobberXD:
Argh, I should've been more careful, sorry...

PR #8746, opened 2019-08-04
Use `InteractAction` enum to denote interaction actions

ClobberXD:
This PR replaces the magic numbers used as interaction modes both client-side and server-side, primarily for the sake of ease-of-readability.

### How does the PR work?

- New enum `InteractAction` has been added to `src/network/networkprotocol.h`
- `Client::interact` and `Server::handleCommand_Interact` have been modified to work with this enum instead of magical `u8` values.

### Why is this PR needed?

Readability. Without the enum, someone new to this area of code might not immediately understand the magic numbers passed to `Client::interact`. These magic numbers _are_ documented, but one would still have to search for the documentation before understanding what it's supposed to mean.

### How to test

- Fire up Minetest.
- Run around the world punching, digging, placing, and using stuff (while pointing at something and while pointing at nothing). No behaviour should be different from how it is in `master`.

****

This PR is Ready for Review. Tested; works.

### DIFF
--- a/src/client/client.cpp
+++ b/src/client/client.cpp
@@ -903,7 +903,7 @@ void writePlayerPos(LocalPlayer *myplayer, ClientMap *clientMap, NetworkPacket *
 	*pkt << fov << wanted_range;
 }
 
-void Client::interact(u8 action, const PointedThing& pointed)
+void Client::interact(InteractAction action, const PointedThing& pointed)
 {
 	if(m_state != LC_Ready) {
 		errorstream << "Client::interact() "
@@ -923,18 +923,11 @@ void Client::interact(u8 action, const PointedThing& pointed)
 		[5] u32 length of the next item (plen)
 		[9] serialized PointedThing
 		[9 + plen] player position information
-		actions:
-		0: start digging (from undersurface) or use
-		1: stop digging (all parameters ignored)
-		2: digging completed
-		3: place block or item (to abovesurface)
-		4: use item
-		5: perform secondary action of item
 	*/
 
 	NetworkPacket pkt(TOSERVER_INTERACT, 1 + 2 + 0);
 
-	pkt << action;
+	pkt << (u8)action;
 	pkt << (u16)getPlayerItem();
 
 	std::ostringstream tmp_os(std::ios::binary);
@@ -1195,7 +1188,7 @@ void Client::clearOutChatQueue()
 }
 
 void Client::sendChangePassword(const std::string &oldpassword,
-        const std::string &newpassword)
+	const std::string &newpassword)
 {
 	LocalPlayer *player = m_env.getLocalPlayer();
 	if (player == NULL)
--- a/src/network/serverpackethandler.cpp
+++ b/src/network/serverpackethandler.cpp
@@ -984,7 +984,7 @@ bool Server::checkInteractDistance(RemotePlayer *player, const f32 d, const std:
 	return true;
 }
 
-void Server::handleCommand_Interact(NetworkPacket* pkt)
+void Server::handleCommand_Interact(NetworkPacket *pkt)
 {
 	/*
 		[0] u16 command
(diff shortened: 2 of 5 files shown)


src/network/serverpackethandler.cpp
```diff
@@ -993,17 +993,16 @@ void Server::handleCommand_Interact(NetworkPacket* pkt)
...
-		4: use item
-		5: rightclick air ("activate")
 	*/
-	u8 action;
+
+	InteractAction action;
+	{
+		u8 temp;
+		*pkt >> temp;
+		action = static_cast<InteractAction>(temp);
```
sfan5: this is pretty ugly but I don't have a better solution.

ClobberXD: I couldn't think of anything better as well. Overloading `>>` for `InteractAction` is overkill, as the enum is passed to the operator only once in the whole code-base.

Also, `action = (u8)temp;` is more concise, but it's better to use `static_cast` instead of plain C-style casts wherever possible (I think).

DS-Minetest: Wouldn't `*pkt >> (u8)action;` work?

src/client/client.cpp
```diff
@@ -1195,7 +1188,7 @@ void Client::clearOutChatQueue()
 }
 
 void Client::sendChangePassword(const std::string &oldpassword,
-        const std::string &newpassword)
+		const std::string &newpassword)
```
DS-Minetest: https://dev.minetest.net/Code_style_guidelines#Function_declarations
>In case your function parameters don't fit within the defined line length use following style. Indention for follow up lines is **exactly** one tab.

src/network/serverpackethandler.cpp
ClobberXD: It doesn't. Remember, we aren't assigning `action` to `pkt` (i.e. `pkt = (u8)action`), so casting won't work. :)

Here's the error message, btw:
```
cannot bind non-const lvalue reference of type ‘irr::u8& {aka unsigned char&}’ to an rvalue of type ‘irr::u8 {aka unsigned char}’
  *pkt >> (u8)action;
          ^~~~~~~~~~
```

src/client/client.cpp
ClobberXD: Sorry, cap'n. Code-style's now fixed. O7 :medal_military:

src/network/serverpackethandler.cpp
DS-Minetest: Sorry, of course `*pkt >> (u8)action;` doesn't work, we have to cast to a reference: `*pkt >> (u8 &)action;` does work (I've tested).

DS-Minetest:
Tested, everything seems to work fine. :+1:

src/network/serverpackethandler.cpp
ClobberXD: Oh neat. I'll try it out.

ClobberXD: It works!

src/network/serverpackethandler.cpp
```diff
@@ -993,18 +993,14 @@ void Server::handleCommand_Interact(NetworkPacket* pkt)
...
-		5: rightclick air ("activate")
 	*/
-	u8 action;
+
+	InteractAction action;
+	*pkt >> (u8 &)action;
+
 	u16 item_i;
-	*pkt >> action;
 	*pkt >> item_i;
```
sfan5: since the complicated reading of action is no longer necessary I'd prefer it to look like before:
```
	InteractAction action;
	u16 item_i;
	*pkt >> (u8 &) action;
	*pkt >> item_i;
```

ClobberXD: Agreed. Done.

SmallJoker:
http://irc.minetest.net/minetest-dev/2019-08-07#i_5575033

SmallJoker:
e788ee2

PR #8749, opened 2019-08-04
Ignore near_plane setting on non-Android platforms

ClobberXD:
`near_plane` seems to be limited to 0.5 to prevent x-ray vision, but even 0.25 simulates x-ray vision across almost 25% of the screen. So this PR:

- Completely ignores this setting on non-Android devices.
- Changes the upper-bound of this setting from 0.5 to 0.25 on Android devices.
- Changes the default value of this setting from 0.1 to 0.05, because the default value would otherwise be the same as the upper-bound.

### To test

- Fire up one instance running the PR branch, and one instance running `master` (for comparison).
- Set `near_plane` to 0.25 in both instances.
- Observe that the x-ray effect is non-existent in the instance running the PR (as the setting is ignored).

Screenshot from 2019-08-04 14-27-55
(left: PR, right: `master`)

****

Needs testing on Android devices to ensure that 0.05 doesn't result in increased visual artifacts.

### DIFF
--- a/src/client/camera.cpp
+++ b/src/client/camera.cpp
@@ -571,7 +571,7 @@ void Camera::updateViewingRange()
 	m_cameranode->setNearValue(rangelim(
 		g_settings->getFloat("near_plane"), 0.0f, 0.25f) * BS);
 #else
-	m_cameranode->setNearValue(0.05f * BS);
+	m_cameranode->setNearValue(0.1f * BS);
 #endif
 
 	m_draw_control.wanted_range = std::fmin(adjustDist(viewing_range, getFovMax()), 4000);
--- a/src/client/camera.cpp
+++ b/src/client/camera.cpp
@@ -565,10 +565,16 @@ void Camera::update(LocalPlayer* player, f32 frametime, f32 busytime, f32 tool_r
 void Camera::updateViewingRange()
 {
 	f32 viewing_range = g_settings->getFloat("viewing_range");
-	f32 near_plane = g_settings->getFloat("near_plane");
+
+	// Ignore near_plane setting on non-Android platforms to prevent abuse
+#ifdef __ANDROID__
+	m_cameranode->setNearValue(rangelim(
+		g_settings->getFloat("near_plane"), 0.0f, 0.25f) * BS);
+#else
+	m_cameranode->setNearValue(0.05f * BS);
+#endif
 
 	m_draw_control.wanted_range = std::fmin(adjustDist(viewing_range, getFovMax()), 4000);
-	m_cameranode->setNearValue(rangelim(near_plane, 0.0f, 0.5f) * BS);
 	if (m_draw_control.range_all) {
 		m_cameranode->setFarValue(100000.0);
 		return;
@@ -596,7 +602,7 @@ void Camera::wield(const ItemStack &item)
 
 void Camera::drawWieldedTool(irr::core::matrix4* translation)
 {
-	// Clear Z buffer so that the wielded tool stay in front of world geometry
+	// Clear Z buffer so that the wielded tool stays in front of world geometry
 	m_wieldmgr->getVideoDriver()->clearZBuffer();
 
 	// Draw the wielded node (in a separate scene manager)
--- a/src/client/camera.cpp
+++ b/src/client/camera.cpp
@@ -20,6 +20,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include "camera.h"
 #include "debug.h"
 #include "client.h"
+#include "config.h"
 #include "map.h"
 #include "clientmap.h"     // MapDrawControl
 #include "player.h"
@@ -566,8 +567,8 @@ void Camera::updateViewingRange()
 {
 	f32 viewing_range = g_settings->getFloat("viewing_range");
 
-	// Ignore near_plane setting on non-Android platforms to prevent abuse
-#ifdef __ANDROID__
+	// Ignore near_plane setting on all other platforms to prevent abuse
+#if ENABLE_GLES
 	m_cameranode->setNearValue(rangelim(
 		g_settings->getFloat("near_plane"), 0.0f, 0.25f) * BS);
 #else
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -623,11 +623,11 @@ pause_on_lost_focus (Pause on lost window focus) bool false
 #    View distance in nodes.
 viewing_range (Viewing range) int 100 20 4000
 
-#   Camera 'near clipping plane' distance in nodes, between 0 and 0.5.
-#   Most users will not need to change this.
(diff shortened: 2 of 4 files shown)


src/client/camera.cpp
```diff
@@ -549,8 +549,13 @@ void Camera::updateViewingRange()
 	f32 viewing_range = g_settings->getFloat("viewing_range");
 	f32 near_plane = g_settings->getFloat("near_plane");
 
+#ifdef __ANDROID__
```
red-001: might be a good idea to leave a comment here on why it's different on android (and presumingly other mobile devices?)

Calinou:
Is there even a reason for this being a setting in minetest.conf? It could be increased automatically on Android and on platforms where the Z-buffer is only 16 bits (i.e. Intel graphics drivers when you don't explicitly request a 24-bit Z-buffer).

paramat:
No need to remove these changes but, there's no need to alter conf.example, as it is regularly auto-generated from settingtypes.txt, so is wasted work.
Just so you know forfuture PRs.

src/client/camera.cpp
ClobberXD: Good idea. Will do.

ClobberXD:
@Calinou I initially did contemplate removing this setting altogether, but dropped the idea because I _thought_ it'd be met with a lot of opposition. :)

Let me know if I should close this PR and _try_ to work on that instead.

@paramat Yes, I'm aware of that, but I did it manually nonetheless, because it's a trivial change, and the file would also be up to date right away.

SmallJoker:
> Is there even a reason for this being a setting in minetest.conf?

Good point. Higher values are only required on less performant hardware, such as the RPI or Android. I would like to see this setting to be changed to something like `increase_near_pane = false` which defaults to `true` on Android. The near pane value would then be 0.1, 0.25 respectively.

paramat:
> Is there even a reason for this being a setting in minetest.conf?

I agree this setting is probably not needed, and i agree with SmallJoker's suggestion (and will +1 it).

ClobberXD, unless you auto-generated conf.example for this PR (doesn't look so because other changes in it would appear) it is highly likely to already be out of date (because most PRs don't bother to update it because it is occasionally auto-generated), so changing it won't make it up to date, so it's wasted work.
It's only worth updating conf.example by auto-generating it, although i'm not requesting that.

ClobberXD:
> I would like to see this setting to be changed to something like `increase_near_plane = false` which defaults to true on Android. The near plane value would then be 0.1, 0.25 respectively.

But that means that anyone could just set it to true to get x-ray vision. So basically nothing would change (except for maybe the lowered upper-bound from 0.5 to 0.25). MT should either auto-detect and auto-configure `near_plane` based on the platform, or use different upper-bounds for different platforms. The latter is probably the easier of the two (not to mention, the PR's already ready :P), and could be merged to make it harder to cheat until the former is implemented.

> it is highly likely to already be out of date

Good point. At least the parts of the file relevant to this PR would be up-to-date, but nevertheless I'll stop updating conf.example in my PRs henceforth. :)

benrob0329:
I think he meant as a compile-time variable.

paramat:
ClobberXD, oh yeah, good point about the setting. I think i misunderstood what SmallJoker meant.
> MT should either auto-detect and auto-configure near_plane based on the platform, or use different upper-bounds for different platforms.

I agree.

ClobberXD:
> I think he meant as a compile-time variable.

I hope that's what he meant. :)

Then we could just detect if the platform is Android, and directly set `near_plane` to 0.25, and remove the setting altogether.

ClobberXD:
Update:

`near_plane` setting is completely ignored on non-Android platforms, and the camera's near-plane is set to a hard-coded value of 0.05. On Android, the behaviour is unchanged (from what the PR originally did).

Btw, shall I move `m_cameranode->setNearValue` to the constructor? I don't see why this is (re)set every time the viewing range is updated.

src/client/camera.cpp
ClobberXD: Added a one-line comment

paramat:
Related comment https://github.com/minetest/minetest/issues/8065#issuecomment-451771466

ClobberXD:
Regarding Calinou's suggestions (both in 8065 and in https://github.com/minetest/minetest/pull/8749#issuecomment-518020711), we can implement automatic near plane configuration later, once this PR has been merged. Note that this PR is sorta high-ish priority, as this setting is used openly to cheat on servers (esp. in survival and PvP modes).

ClobberXD:
Rebased. What's the status of this PR? Also, please note:

> Note that this PR is sorta high-ish priority, as this setting is used openly to cheat on servers (esp. in survival and PvP modes).

paramat:
Added to 5.2.0 milestone.

sfan5:
Since this PR is still in the backlog, here's my feedback:
The problem with the near plane generally applies to weaker GPUs, so it would make sense to replace the `#ifdef __ANDROID__` with `#if ENABLE_GLES`. Other than that the PR looks good to me.

paramat:
Although this is likely to be improved later, shall we merge this for 5.2.0? It is fairly high priority.

paramat:
Testing on desktop PC:
`near_plane = 0.5` is very cheaty, this clientside cheat should not be allowed:

screenshot_20200317_215319

With `near_plane = 0.25` and default FOV, no view inside nodes is possible, so seems a reasonable new upper limit for Android.
`near_plane = 0.05` causes no visual problems for me (Intel integrated GPU).

paramat:
With `near_plane = 0.1`, the current default, and an impractically extreme FOV of 140, viewing inside nodes is not possible. FOV 160 (the maximum FOV) allows an inside view but the world is very distorted so it is not of much use.
So we could safely leave the default at 0.1 for now at least, as there is a concern that 0.05 might cause problems on Android.

So my suggestion is: This PR is fine, but leave the default unchanged at 0.1. Then merge it for 5.2.0.

ClobberXD:
Roger that, will attend to the request in a week's time.

paramat:
A week might be a little late for release, we could merge this and alter it on merge.
I might test the PR, i cannot test on Android but that is not important as nothing really changes, Android is currently unavailable in the Play store anyway.

paramat:
If wanted, i can open a PR that contains ClobberXD's commit, i have created a PR branch https://github.com/paramat/minetest/tree/nearplane

sfan5:
Pushed your commit and the `ENABLE_GLES` change I suggested to the branch of this PR.

ClobberXD:
Oh, are we nearing the release/freeze for 5.2.0? Also, thanks for reverting the min value back to `0.1`.

paramat:
sfan5 thanks.
My branch (before without your commit) works fine, but i will try to test this PR.

paramat:
Tested PR on desktop.
:+1: 
Please can we merge this for 5.2.0?

PR #8750, opened 2019-08-04
Better F6 profiler

SmallJoker:
## Goal
 * Make the profiler easier to read
 * Rename profilers to they make more sense
 * Drop useless profilers or merge with others

## Result
 * Lower default update interval to 2 seconds
 * Show `ScopeProfiler` results in milliseconds
 * Many removed variables used for time measuring, or such that did not do anything

screenshot_20190804_172329

## To do

This PR is Ready for Review.

## How to test

1) Join a world
2) Same in-game performance (you'd notice that if there were accidental changes in the code)
3) Open the F6 profiler -> monospace text, nothing's cut off
4) Open the F5 graphs -> must all display values when walking around

### DIFF
--- a/src/client/activeobjectmgr.cpp
+++ b/src/client/activeobjectmgr.cpp
@@ -35,8 +35,7 @@ void ActiveObjectMgr::clear()
 void ActiveObjectMgr::step(
 		float dtime, const std::function<void(ClientActiveObject *)> &f)
 {
-	g_profiler->avg("Client::ActiveObjectMgr: num of objects",
-			m_active_objects.size());
+	g_profiler->avg("ActiveObjectMgr: SAO count [#]", m_active_objects.size());
 	for (auto &ao_it : m_active_objects) {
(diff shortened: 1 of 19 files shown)


paramat:
Does this do #4320 too? That would help a lot.

SmallJoker:
@paramat No it does not but I could make the colors based on the profiler name's hash.

paramat:
Ok a separate PR is fine, thank for considering.

SmallJoker:
I see a few thumbs up (for the concept?) but no feedbacks on the profiler names. Do they make sense? What could be improved?

SmallJoker:
Update.

grafik

ClobberXD:
I'm sorry, I like the concept but I'm unable to take a look at the changes and/or test.

SmallJoker:
@ClobberXD Thanks for the concept feedback. The profiler scopes can still be tweaked later if there are better suggestions :)

PR #8751, opened 2019-08-04
Do not add group values of zero to group lists.

shacknetisp:
This fixes an issue where when the engine looked up groups (for example, in ABM node names), NodeDefManager's m_group_to_items would contain nodes with a group value of zero, resulting in nodes with flammable = 0 being burned by a fire mod with a group:flammable checking ABM.

It brings consistency to the behaviour described in the api documentation, where zero and nil groups should be the same.

### DIFF
--- a/src/script/common/c_content.cpp
+++ b/src/script/common/c_content.cpp
@@ -1536,7 +1536,7 @@ void read_groups(lua_State *L, int index, ItemGroupList &result)
 		std::string name = luaL_checkstring(L, -2);
 		int rating = luaL_checkinteger(L, -1);
 		// zero rating indicates not in the group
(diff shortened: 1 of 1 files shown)


src/script/common/c_content.cpp
```diff
@@ -1535,7 +1535,9 @@ void read_groups(lua_State *L, int index, ItemGroupList &result)
 		// key at index -2 and value at index -1
 		std::string name = luaL_checkstring(L, -2);
 		int rating = luaL_checkinteger(L, -1);
-		result[name] = rating;
+		// zero rating indicates not in the group
+		if(rating != 0)
```
SmallJoker: Space after `if`

PR #8753, opened 2019-08-05
Clear old item groups when they are overridden.

shacknetisp:
This fixes overridden items keeping their old groups in the group to items mapping even after their groups have been changed in lua.
It also prevents a more widespread issue where overriding an item will add its content ID *twice* to the mapping, resulting in odd behaviour in features such as ABMs.

Testing code below. Give yourself a `test:node` and place it. In the ABMs, without this fix, `Bad!` will continue to print and `Good.` will print twice as often as it should.

```lua
minetest.register_node(":test:node", {
	groups = {bad = 1, good = 1},
})

minetest.override_item("test:node", {
	groups = {good = 1},
})

minetest.register_abm{
	nodenames = {"group:bad"},
	interval = 1,
	chance = 1,
	action = function()
		print("Bad!")
	end,
}

minetest.register_abm{
	nodenames = {"group:good"},
	interval = 1,
	chance = 1,
	action = function()
		print("Good.")
	end,
}
```

### DIFF
--- a/src/nodedef.cpp
+++ b/src/nodedef.cpp
@@ -1205,8 +1205,8 @@ inline void NodeDefManager::fixSelectionBoxIntUnion()
 void NodeDefManager::eraseIdFromGroups(content_t id)
 {
 	// For all groups in m_group_to_items...
-	for (std::unordered_map<std::string, std::vector<content_t>>::iterator iter_groups =
-			m_group_to_items.begin(); iter_groups != m_group_to_items.end();) {
+	for (auto iter_groups = m_group_to_items.begin();
+			iter_groups != m_group_to_items.end();) {
 		// Get the group items vector.
 		std::vector<content_t> &items = iter_groups->second;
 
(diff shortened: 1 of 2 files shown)


src/nodedef.h
```diff
@@ -668,6 +668,14 @@ class NodeDefManager {
 	 */
 	void addNameIdMapping(content_t i, std::string name);
 
+	/*!
+	 * Removes a content ID from all groups.
+	 * Erases content IDs from vectors in \ref m_group_to_items and
+	 * removes empty vectors.
+	 * @param i a content ID
```
SmallJoker: Self-explanatory and incorrect comment. Please correct to `@param id Content ID` or remove.

src/nodedef.cpp
```diff
@@ -1222,19 +1242,25 @@ content_t NodeDefManager::set(const std::string &name, const ContentFeatures &de
 		assert(id != CONTENT_IGNORE);
 		addNameIdMapping(id, name);
 	}
+
+	// If there is already ContentFeatures registered for this id, clear old groups
+	if (id < m_content_features.size()) {
```
SmallJoker: Brace removal possible

src/nodedef.cpp
```diff
@@ -1202,6 +1202,26 @@ inline void NodeDefManager::fixSelectionBoxIntUnion()
 }
 
 
+void NodeDefManager::eraseIdFromGroups(content_t id)
+{
+	// For all groups in m_group_to_items...
+	for (std::unordered_map<std::string, std::vector<content_t>>::iterator iter_groups =
```
SmallJoker: `for (auto iter_groups = ..` C++11 way, also shorter.

PR #8754, opened 2019-08-05
Update curl dependency in buildbot

sfan5:
curl 7.64.0 can(?) cause crashes when used with remote media servers, see #8690 

## To do


SmallJoker:
Well sure, if it fixes the issue feel free to merge :+1:

PR #8770, opened 2019-08-09
Auto delete MetaData when = 0

Lejo1:
We still automatic delete a key of a metadata or mod_storage data if it is = ""
This PR also auto deletes if it is = 0
As we basically also return 0 for no value this makes no compatibility problems
This saves memory when mods for example remove a player and set the var to default = 0

## How to test
``` Lua
local storage = minetest.get_mod_storage()

storage:set_string("placeholder", "make sure file is not null")

storage:set_int("int", 100)
storage:set_int("int", 0)
storage:set_float("float", 100.1)
storage:set_float("float", 0)

storage:set_float("float2", 100.1)
storage:set_float("float2", 0.0)

storage:set_string("string", "stringggggggggggg")
storage:set_string("string", "")
```
All the entries will disappear without the PR float, float2 and int will stay = 0

### DIFF
--- a/src/script/lua_api/l_metadata.cpp
+++ b/src/script/lua_api/l_metadata.cpp
@@ -153,7 +153,9 @@ int MetaDataRef::l_set_int(lua_State *L)
 	MetaDataRef *ref = checkobject(L, 1);
 	std::string name = luaL_checkstring(L, 2);
 	int a = luaL_checkint(L, 3);
-	std::string str = itos(a);
+	std::string str = std::string("");
+	if (a != 0)
+		str = itos(a);
 
 	Metadata *meta = ref->getmeta(true);
 	if (meta == NULL || str == meta->getString(name))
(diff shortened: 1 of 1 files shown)


src/script/lua_api/l_metadata.cpp
```diff
@@ -153,7 +153,9 @@ int MetaDataRef::l_set_int(lua_State *L)
 	MetaDataRef *ref = checkobject(L, 1);
 	std::string name = luaL_checkstring(L, 2);
 	int a = luaL_checkint(L, 3);
-	std::string str = itos(a);
+	std::string str = std::string("");
```
SmallJoker: ` = std::string("")` is superfluous.

src/script/lua_api/l_metadata.cpp
```diff
@@ -191,7 +193,9 @@ int MetaDataRef::l_set_float(lua_State *L)
 	MetaDataRef *ref = checkobject(L, 1);
 	std::string name = luaL_checkstring(L, 2);
 	float a = readParam<float>(L, 3);
-	std::string str = ftos(a);
+	std::string str = std::string("");
```
SmallJoker: Ditto

SmallJoker:
The only issue I can see is when you mix `MetaRef:get()` and `MetaRef:set_int()`. Latter will write `"0"`, but former will return `nil`. In terms of storage optimization this is totally fine.

Lejo1:
In my case it’s a money mod. Where I only use set_int and get_int.
When creating a function to fully reset a play including all data of all
mods I came up to this.

Moth-Tolias:
what's the status on this?

PR #8774, opened 2019-08-10
Fix breakage of non-GLES2 setups

osjc:
This pull request fixes a "gotcha" in src/CMakeLists.txt: When you move the OpenGL detection code to the place where the GLES2 detection code is, the build will fail because the OpenGL libraries will NOT be in the list of libraries to be linked in. What I suspect is happening is that find_package tries to add the libraries into the list and later the list is cleared out and filled with other libraries (git show 526a9e4b66abaf83eb6b1aaa3e93375acd87b830 shows the code at the wrong location, then a huge pile of $(BLABLA_LIBRAR[Y|IES]), then the code at the right location removed).

EDIT: It turns out it is not a "gotcha" but a typo. One of the new lines in the original commit says "elseif()" instead of "else()", causing the code relevant for the non-GLES2 setups to be permanently disabled.

EDIT2: Turned out the line most likely wanted to be "elseif(BUILD_CLIENT)" and the "BUILD_CLIENT" part got lost somewhere.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -104,15 +104,17 @@ endif()
 
 option(ENABLE_GLES "Use OpenGL ES instead of OpenGL" FALSE)
 mark_as_advanced(ENABLE_GLES)
-if(ENABLE_GLES)
-	find_package(OpenGLES2 REQUIRED)
-elseif()
-	if(NOT WIN32) # Unix probably
-		set(OPENGL_GL_PREFERENCE "LEGACY" CACHE STRING
-			"See CMake Policy CMP0072 for reference. GLVND is broken on some nvidia setups")
-		set(OpenGL_GL_PREFERENCE ${OPENGL_GL_PREFERENCE})
+if(BUILD_CLIENT)
+	if(ENABLE_GLES)
+		find_package(OpenGLES2 REQUIRED)
+	else()
+		if(NOT WIN32) # Unix probably
+			set(OPENGL_GL_PREFERENCE "LEGACY" CACHE STRING
+				"See CMake Policy CMP0072 for reference. GLVND is broken on some nvidia setups")
+			set(OpenGL_GL_PREFERENCE ${OPENGL_GL_PREFERENCE})
 
-		find_package(OpenGL REQUIRED)
+			find_package(OpenGL REQUIRED)
+		endif()
 	endif()
 endif()
 


osjc:
To be honest, I don't entirely understand what exactly is going on here. All I can tell for sure is that on my setup (Ubuntu 14 LTS, OpenGL with no GLES2) the code at the offending commit (526a9e4) does not build (the link phase fails with a ton of link errors) while either the code before the offending commit or the code with the offending commit and with this patch applied builds. Building with MAKE VERBOSE=1, redirecting the output to a file and then digging out and dissecting the link command revealed that the -lGL -lGLU arguments were missing when the offending commit was built.

osjc:
I think also that a non-GLES2 setup shall be added to Travis to catch breakages like this.

SmallJoker:
Non-OGLES build works fine on Ubuntu 18.04.

> I think also that a non-GLES2 setup shall be added to Travis to catch breakages like this.

None of these builds use the irrlicht ogl-es branch, hence the build would even if it were enabled. List of installed libraries:
https://github.com/minetest/minetest/blob/b19400aa74f30ef61ba0c1bb4a25a978e7cf4125/util/travis/common.sh#L23-L29

But it is surely an issue to look for OpenGL when only BUILD_SERVER is specified.

sfan5:
> because the OpenGL libraries will NOT be in the list of libraries to be linked in.

It should be linked in, though:
https://github.com/minetest/minetest/blob/b19400aa74f30ef61ba0c1bb4a25a978e7cf4125/src/CMakeLists.txt#L547-L558

osjc:
Yes, it should. But for some bizzare reason find_package(OpenGL) does not provide the libs when called from the GLES2 location while it does do that when called from the location where this PR is placing it. I will try to find what makes it break.

osjc:
Ok, I found the bug. Preparing a better commit ...

osjc:
So, here is the problem. Due to a bug in cmake (silently misinterpreting broken input instead of yelling loudly) whoever made the original commit, did not realize he wrote "elseif()" instead of "else()". cmake interprets "elseif()" as "elseif(false)", thus ignoring whatever code is in that branch. Therefore if GLES is not selected, then the OpenGL detection is skipped, ${OPENGL_LIBRARIES} expands to an empty string and the project is mislinked despite the lines 547 to 558 containing code that adds the OpenGL libraries to the list.

osjc:
Confirmed that this is the problem. After replacing the "elseif()" with "else()" the build works.

SmallJoker:
`elseif (BUILD_CLIENT)` would be better for server builds.

osjc:
Fixed.

osjc:
Works on my setup too.

src/CMakeLists.txt
```diff
@@ -106,7 +106,7 @@ option(ENABLE_GLES "Use OpenGL ES instead of OpenGL" FALSE)
 mark_as_advanced(ENABLE_GLES)
 if(ENABLE_GLES)
 	find_package(OpenGLES2 REQUIRED)
-elseif()
+elseif(BUILD_CLIENT)
```
sfan5: Just put the entire enable_gles stuff in an `if(BUILD_CLIENT)`

osjc: Right. Fixing ...

osjc: Fixed.

osjc:
(forgot to update the description of the commit)

PR #8783, opened 2019-08-12
Allow calling ItemStack with no arguments

pauloue:
Allows creating an ItemStack with `ItemStack()`. It creates an empy stack just like `ItemStack(nil)` and `ItemStack("")`.

~This also affects other places where `read_item` is used. E.g. `InvRef:add_item(listname)` will add an empty stack (note that `add_item(listname, nil)` is already possible).~

## How to test

`assert(ItemStack():to_string() == "")`

### DIFF
--- a/src/script/lua_api/l_item.cpp
+++ b/src/script/lua_api/l_item.cpp
@@ -404,7 +404,9 @@ ItemStack& LuaItemStack::getItem()
 int LuaItemStack::create_object(lua_State *L)
 {
 	NO_MAP_LOCK_REQUIRED;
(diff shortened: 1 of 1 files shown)


DS-Minetest:
A lua user expects `func()` and `func(nil)` being the same for every lua function `func`. :+1:

SmallJoker:
Even though `func()` does the same as `func(nil)` in Lua, the API functions should require explicit arguments. What's the use of `room_for_item("main")`?
More: http://irc.minetest.net/minetest-dev/2019-08-12#i_5576886

PR #8788, opened 2019-08-13
Fix unnecessary exception use in Map::isNodeUnderground

osjc:
Another exception usage nonsense to be removed. In this one the method in question throws an exception only to be caught one call level higher, slowing down the whole thing. Replaced the "one call level higher" code with call to the NoEx version and an "if" instead of try...catch.


osjc:
Squashed the 2 commits.

PR #8791, opened 2019-08-13
Fix unnecessary exception use in 3 more methods

osjc:
Found 3 more methods that mess around with exceptions unnecessarily.

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -2214,19 +2214,14 @@ void MMVManip::initialEmerge(v3s16 blockpos_min, v3s16 blockpos_max,
 			continue;
 
 		bool block_data_inexistent = false;
-		try
 		{
 			TimeTaker timer2("emerge load", &emerge_load_time);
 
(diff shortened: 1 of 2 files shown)


src/map.cpp
```diff
@@ -2229,20 +2229,15 @@ void MMVManip::initialEmerge(v3s16 blockpos_min, v3s16 blockpos_max,
...
 
 		bool block_data_inexistent = false;
-		try
 		{
 			TimeTaker timer2("emerge load", &emerge_load_time);
 
-			block = m_map->getBlockNoCreate(p);
-			if(block->isDummy())
+			block = m_map->getBlockNoCreateNoEx(p);
+			if (block == nullptr || block->isDummy())
```
SmallJoker: `!block` is faster to read and more common in our code base

osjc: Good to know. There are more of them. Going to fix them all.

PR #8792, opened 2019-08-13
Fix unnecessary exception use in Map::getSectorXXX

osjc:
The mapsector getters also mess up with exceptions unnecessarily. Get rid of the exception-full version and clean up the names of the two other methods that remain.


osjc:
"sector == nullptr" -> "!sector"

PR #8793, opened 2019-08-13
WIP: MapBlock: Remove unnecessary exceptions

osjc:
This time the private method MapBlock::getNodeRef uses exceptions unnecessarily.

### DIFF
--- a/src/mapblock.cpp
+++ b/src/mapblock.cpp
@@ -222,25 +222,17 @@ s16 MapBlock::getGroundLevel(v2s16 p2d)
 {
 	if(isDummy())
 		return -3;
-	try
-	{
-		s16 y = MAP_BLOCKSIZE-1;
-		for(; y>=0; y--)
-		{
-			MapNode n = getNodeRef(p2d.X, y, p2d.Y);
-			if (m_gamedef->ndef()->get(n).walkable) {
-				if(y == MAP_BLOCKSIZE-1)
-					return -2;
-
-				return y;
-			}
+	for (s16 y = MAP_BLOCKSIZE-1; y >= 0; y--) {
+		if (!isValidPosition(p2d.X, y, p2d.Y))
+			return -3;
+		MapNode n = getNodeRef(p2d.X, y, p2d.Y);
+		if (m_gamedef->ndef()->get(n).walkable) {
+			if (y == MAP_BLOCKSIZE-1)
+				return -2;
+			return y;
 		}
-		return -1;
 	}
-	catch(InvalidPositionException &e)
-	{
-		return -3;
-	}
+	return -1;
 }
 
 /*
--- a/src/mapblock.h
+++ b/src/mapblock.h
@@ -498,13 +498,11 @@ private:
 
 	/*
 		Used only internally, because changes can't be tracked
+		Before use check that isValidPosition(x, y. z) returns
+		"true".
 	*/
-
 	inline MapNode &getNodeRef(s16 x, s16 y, s16 z)
 	{
-		if (!isValidPosition(x, y, z))
-			throw InvalidPositionException();
-
 		return data[z * zstride + y * ystride + x];
 	}
 


src/mapblock.h
```diff
@@ -498,21 +498,14 @@ class MapBlock
...
 	inline MapNode &getNodeRef(s16 x, s16 y, s16 z)
 	{
-		if (!isValidPosition(x, y, z))
-			throw InvalidPositionException();
-
 		return data[z * zstride + y * ystride + x];
 	}
 
-	inline MapNode &getNodeRef(v3s16 &p)
-	{
```
sfan5: Just because it's unused <i>right now</i> does not mean this should be removed IMO

src/mapblock.h
```diff
@@ -498,21 +498,14 @@ class MapBlock
...
 	/*
 		Used only internally, because changes can't be tracked
+		Before use check that isValidPosition(x, y. z) returns
+		"true".
 	*/
-
 	inline MapNode &getNodeRef(s16 x, s16 y, s16 z)
 	{
-		if (!isValidPosition(x, y, z))
-			throw InvalidPositionException();
```
sfan5: have you checked whether there are any other callers?

osjc: Yes. There aren't any more.

osjc:
Reinstated the deleted overload.

src/mapblock.h
osjc: Reinstated.

src/mapblock.h
osjc: It is a private method so only mapblock.cpp needed to be checked. But the way I am doing this checking was to rename the method to getNodeRefX and see where else the build fails.

src/mapblock.cpp
```diff
@@ -222,25 +222,20 @@ s16 MapBlock::getGroundLevel(v2s16 p2d)
 {
 	if(isDummy())
 		return -3;
-	try
+	s16 y = MAP_BLOCKSIZE-1;
+	for(; y>=0; y--)
```
SmallJoker: In future changes, please correct the code style of each touched (/modified) line according to the rules specified here: https://dev.minetest.net/Code_style_guidelines

Thank you :)

osjc: Fixed.

pauloue: Can the initialization of `y` go in the `for`?
Also spaces around operators and after `for` and `if`.

osjc: I knew there was code that I touched and had messed up spacing. Thanks. Fixed.

src/mapblock.h
```diff
@@ -498,13 +498,11 @@ class MapBlock
 
 	/*
 		Used only internally, because changes can't be tracked
+		Before use check that isValidPosition(x, y. z) returns
+		"true".
 	*/
-
 	inline MapNode &getNodeRef(s16 x, s16 y, s16 z)
```
SmallJoker: Now there are no bound checks left over, and the function name should indicate that.
For example: `getNodeUnsafe`.

osjc:
Large scale code formatting cleanup prompts for this not to be merged yet.

src/mapblock.h
osjc: Will use "getNodeRefNoCheck" as this is the style for "unsafe" methods throughout the code.

src/mapblock.h
osjc: However our coding standards are based of the Linux Kernel coding standards and these state in chapter 21 "Conditional compilation" this:

> However, if a function or variable always goes unused, delete it.

And I agree with that. Dead code is subject to bitrot and I found some large blocks wrapped in #if 0 ... #endif and some of these have questionable application to the current code.

osjc:
I prefer to do code formatting changes in a different commit/PR than the actual code fix. Since currently I don't have much time, I closed this. Will reopen once I am ready for this.

Issue #8757, opened 2019-08-06
Possible denial of service attack (DoS)


thomasrudin:
##### Minetest version

```
5.0.1
```

##### OS / Hardware
Operating system: Linux
CPU: x86_64

##### Summary
This night i had several crashes over a time range of about 2 hours.
Every time before the server crashed (without warning) i had these lines in the logs:

```
minetest_1           | Duplicated seqnum 1 non matching packet detected:
minetest_1           | Old: seqnum: 00001 size: 0011, address: 37.228.235.x
minetest_1           | New: seqnum: 00001 size: 0025, address: 37.228.235.x
```

Sadly nothing in the debug logs or anything about that ip in the recent logs.
Output of `dmesg` showed lots of these too:

```
[582133.937849] ConnectionRecei[4399]: segfault at 8 ip 00005596f262da77 sp 00007fee8f7fce70 error 4 in minetestserver[5596f2523000+3d7000]
```

I blocked the IP with iptables but the underlying issue is still there, whatever it is.

##### Steps to reproduce
If i knew that i would have opened a PR instead.. :rofl:

MoNTE48:
I can only suggest  that this is a test of a modified client. Obviously failing for the Minetest server...
If someone wanted to do DDOS on your server, he would hardly have used this strategy.  In fact, Minetest has many other vulnerabilities that can help it “go away” for a long time...

HybridDog:
http://github.com/minetest/minetest/blob/master/src/network/connection.cpp#L341

thomasrudin:
> [http://github.com/minetest/minetest/blob/8da35c22d1c8933090330b2f3c44b4cf2c6e6760/src/network/connection.cpp#L341](https://github.com/minetest/minetest/blob/8da35c22d1c8933090330b2f3c44b4cf2c6e6760/src/network/connection.cpp#L341)

Yeah, that _should_ get caught here: https://github.com/minetest/minetest/blob/8da35c22d1c8933090330b2f3c44b4cf2c6e6760/src/network/connectionthreads.cpp#L1343

Parasitoid:
Hi! I have seen similar crash at different minetest server about 14 - 16 hours ago, and according to what admin of that server said, the log was very similar.

red-001:
removing the throw isn't really a full solution, you should be disconnecting the misbehaving peer.

Hume2:
I had the same issue 6 times.

sfan5:
[...]

PR #8801, opened 2019-08-15
network: Stricter handling of split packets

sfan5:
The reassembly code assumes that as soon `chunk_count` packets were received all chunk numbers `0 <= n < chunk_count` exist. Modify the checks to ensure that assumption can never be broken.

If the assumptions do end up broken the packet will just partially be filled with zeros, so this is not a vulnerability/DoS/crash-fix.
discovered thanks to https://github.com/minetest/minetest/issues/8757#issuecomment-521525539

## To do

This PR is Ready for Review.

### DIFF
--- a/src/network/connection.cpp
+++ b/src/network/connection.cpp
@@ -417,6 +417,11 @@ SharedBuffer<u8> IncomingSplitBuffer::insert(const BufferedPacket &p, bool relia
 			<< std::endl;
 		return SharedBuffer<u8>();
 	}
+	if (chunk_num >= chunk_count) {
+		errorstream << "IncomingSplitBuffer::insert(): chunk_num=" << chunk_num
+				<< " >= chunk_count=" << chunk_count << std::endl;
+		return SharedBuffer<u8>();
+	}
 
 	// Add if doesn't exist
 	if (m_buf.find(seqnum) == m_buf.end()) {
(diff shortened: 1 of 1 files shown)


src/network/connection.cpp
```diff
@@ -417,6 +417,11 @@ SharedBuffer<u8> IncomingSplitBuffer::insert(const BufferedPacket &p, bool relia
 			<< std::endl;
 		return SharedBuffer<u8>();
 	}
+	if (chunk_num >= chunk_count) {
+		errorstream << "IncomingSplitBuffer::insert(): chunk_num="<<chunk_num
+				<<" >= chunk_count="<<chunk_count<<std::endl;
```
SmallJoker: Spaces around binary operators.

src/network/connection.cpp
```diff
@@ -425,10 +430,11 @@ SharedBuffer<u8> IncomingSplitBuffer::insert(const BufferedPacket &p, bool relia
 
 	IncomingSplitPacket *sp = m_buf[seqnum];
 
-	if (chunk_count != sp->chunk_count)
-		LOG(derr_con<<"Connection: WARNING: chunk_count="<<chunk_count
-				<<" != sp->chunk_count="<<sp->chunk_count
-				<<std::endl);
+	if (chunk_count != sp->chunk_count) {
+		errorstream<<"IncomingSplitBuffer::insert(): chunk_count="<<chunk_count
+				<<" != sp->chunk_count="<<sp->chunk_count<<std::endl;
```
SmallJoker: Ditto.

PR #8804, opened 2019-08-15
Minor code refactoring in connection.cpp

sfan5:
## To do

This PR is Ready for Review.

## How to test

Verify that Minetest still works.

### DIFF
--- a/src/network/connection.cpp
+++ b/src/network/connection.cpp
@@ -169,6 +169,7 @@ void ReliablePacketBuffer::print()
 		index++;
 	}
 }
+
 bool ReliablePacketBuffer::empty()
 {
 	MutexAutoLock listlock(m_list_mutex);
@@ -177,7 +178,8 @@ bool ReliablePacketBuffer::empty()
 
 u32 ReliablePacketBuffer::size()
 {
-	return m_list_size;
+	MutexAutoLock listlock(m_list_mutex);
+	return m_list.size();
 }
 
 bool ReliablePacketBuffer::containsPacket(u16 seqnum)
@@ -198,17 +200,19 @@ RPBSearchResult ReliablePacketBuffer::findPacket(u16 seqnum)
 	}
 	return i;
 }
+
 RPBSearchResult ReliablePacketBuffer::notFound()
 {
 	return m_list.end();
 }
+
 bool ReliablePacketBuffer::getFirstSeqnum(u16& result)
 {
 	MutexAutoLock listlock(m_list_mutex);
 	if (m_list.empty())
 		return false;
-	BufferedPacket p = *m_list.begin();
-	result = readU16(&p.data[BASE_HEADER_SIZE+1]);
+	const BufferedPacket &p = *m_list.begin();
+	result = readU16(&p.data[BASE_HEADER_SIZE + 1]);
 	return true;
 }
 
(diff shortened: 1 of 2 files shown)


src/network/connection.cpp
```diff
@@ -384,6 +379,48 @@ std::list<BufferedPacket> ReliablePacketBuffer::getTimedOuts(float timeout,
 	return timed_outs;
 }
 
+/*
+	IncomingSplitPacket
+*/
+
+bool IncomingSplitPacket::addChunk(u32 chunk_num, SharedBuffer<u8> chunkdata)
```
SmallJoker: Copying once in `chunks[chunk_num] = chunkdata;` is enough. Please make it a const ref.

src/network/connection.cpp
```diff
@@ -249,15 +246,17 @@ BufferedPacket ReliablePacketBuffer::popSeqnum(u16 seqnum)
...
 
-	if (m_list_size == 0)
-	{ m_oldest_non_answered_ack = 0; }
-	else
-	{ m_oldest_non_answered_ack = readU16(&(*m_list.begin()).data[BASE_HEADER_SIZE+1]);	}
+	if (m_list.empty()) {
+		m_oldest_non_answered_ack = 0;
+	} else {
+		m_oldest_non_answered_ack =
+				readU16(&(*m_list.begin()).data[BASE_HEADER_SIZE+1]);
```
SmallJoker: Dereferencing could be replaced with 'pointer call' notation (or what the term for this is).
Same for `ReliablePacketBuffer::popFirst()`
Also: Missing spaces around `+`

src/network/connection.cpp
sfan5: `SharedBuffer` is a smart point and will not be implicitly copied

src/network/connection.cpp
```diff
@@ -219,16 +216,16 @@ BufferedPacket ReliablePacketBuffer::popFirst()
...
 	m_list.erase(m_list.begin());
-	--m_list_size;
 
-	if (m_list_size == 0) {
+	if (m_list.empty()) {
 		m_oldest_non_answered_ack = 0;
 	} else {
 		m_oldest_non_answered_ack =
-				readU16(&(*m_list.begin()).data[BASE_HEADER_SIZE+1]);
+				readU16(&m_list.begin()->data[BASE_HEADER_SIZE + 1]);
```
ClobberXD: Continuation lines needn't be indented by two tabs unless a one-tab indent aligns them with the following lines. (according to @DS-Minetest :P)

SmallJoker: LINT says 2 tabs for continued lines. However, I cannot find any rule about how many tabs to use for line continuation (only defined for functions).
Nitpicking.

ClobberXD: It's the code-style _guidelines_ that need fixing, then :D

DS-Minetest: It might indeed be useful if the code style guidelines were more exact at continuation lines.
(In the lua code style guidelines, for example, there are very clear rules: https://dev.minetest.net/Lua_code_style_guidelines#Continuation_lines)

[Kernel style](https://www.kernel.org/doc/html/latest/process/coding-style.html#breaking-long-lines-and-strings) isn't so exact either:
>Descendants are always substantially shorter than the parent and are placed substantially to the right.

In the first example of the [space paragraph](https://dev.minetest.net/Code_style_guidelines#Spaces) of the code style guidelines, there's indentation with one tab. However, there's no rule that clearly requires one to use either option.
This means that both, one tab and two tabs, are acceptable. (Theoretically one could even try to indent with three tabs.)

All in all, this doesn't really matter. The readability doesn't suffer much from the non-existence of such a rule. However, it might make sense to clearly state that one tab or two tabs can be used, to prevent confusion and useless discussions.

Btw. I can be less pedantic, if you want. :slightly_smiling_face: >_<

ClobberXD: > Btw. I can be less pedantic, if you want. :slightly_smiling_face: >_<

lol no. I was just joking :)

PR #8812, opened 2019-08-17
[NO SQUASH] Occlusion: Code cleanup and better checks using "light_propagates"

SmallJoker:
This PR includes:
 - Occlusion code cleanup (first commit)
 - Check against `light_propagates` that is more general than `NDT_NORMAL`, but still reliable
 - Make occlusion depend on `noclip` as well (not just `free_mode`)
 - Accurate mapblock bounds checks to skip some more occluded mapblocks
 - ~Show occluded blocks by blinking if the wireframe mode is active~

PR opened due to observations in https://github.com/BlockySurvival/issue-tracker/issues/125 to occlude blocks more precisely to get higher frame rates.

## To do

This PR is Ready for Review.

## How to test

1) Join your world
2) Explore the world including underground
3) Turn off noclip and turn on wireframe mode (F5)
4) Move into ground and above ground at full view range. Mapblocks will appear and vanish: those are occluded.

Note: The server does the same occlusion culling, thus you'll need to explore the underground as well.

### DIFF
--- a/src/client/clientmap.cpp
+++ b/src/client/clientmap.cpp
@@ -151,6 +151,11 @@ void ClientMap::updateDrawList()
 			occlusion_culling_enabled = false;
 	}
 
+	// Uncomment to debug occluded blocks in the wireframe mode
+	// TODO: Include this as a flag for an extended debugging setting
+	//if (occlusion_culling_enabled && m_control.show_wireframe)
+	//    occlusion_culling_enabled = porting::getTimeS() & 1;
+
 	for (const auto &sector_it : m_sectors) {
 		MapSector *sector = sector_it.second;
 		v2s16 sp = sector->getPos();
--- a/src/client/clientmap.cpp
+++ b/src/client/clientmap.cpp
@@ -144,7 +144,7 @@ void ClientMap::updateDrawList()
 	// No occlusion culling when free_move is on and camera is
 	// inside ground
 	bool occlusion_culling_enabled = true;
-	if (g_settings->getBool("free_move")) {
+	if (g_settings->getBool("free_move") && g_settings->getBool("noclip")) {
 		MapNode n = getNode(cam_pos_nodes);
 		if (n.getContent() == CONTENT_IGNORE ||
 				m_nodedef->get(n).solidness == 2)
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -1032,21 +1032,46 @@ void Map::removeNodeTimer(v3s16 p)
 	block->m_node_timers.remove(p_rel);
 }
 
-bool Map::isOccluded(v3s16 pos_origin, v3s16 pos_blockcenter, float step,
-	float stepfac, float offset, float end_offset, u32 needed_count)
+bool Map::isOccluded(const v3s16 &pos_camera, const v3s16 &pos_target,
+	const core::aabbox3d<s16> &block_bounds, float step, float stepfac, float offset,
+	u32 needed_count)
 {
-	float distance = BS * pos_origin.getDistanceFrom(pos_blockcenter);
-	v3f direction = intToFloat(pos_blockcenter - pos_origin, BS);
-	direction.normalize();
-	v3f pos_origin_f = intToFloat(pos_origin, BS);
+	// Worst-case safety distance to keep to the target position
+	// Anything smaller than the mapblock diagonal could result in in self-occlusion
+	// Diagonal = sqrt(1*1 + 1*1 + 1*1)
+	const static float BLOCK_DIAGONAL = BS * MAP_BLOCKSIZE * 1.732f;
 
+	v3f direction = intToFloat(pos_target - pos_camera, BS);
+	float distance = direction.getLength();
+
+	// Disable occlusion culling for near mapblocks in underground
+	if (distance < BLOCK_DIAGONAL)
+		return false;
+
+	// Normalize direction vector
+	if (distance > 0.0f)
+		direction /= distance;
+
+	v3f pos_origin_f = intToFloat(pos_camera, BS);
 	u32 count = 0;
-	for (; offset < distance + end_offset; offset += step){
-		v3f pf = pos_origin_f + direction * offset;
-		v3s16 pos_node = floatToInt(pf, BS);
-		MapNode node = getNode(pos_node);
-		if (m_nodedef->get(node).drawtype == NDT_NORMAL) {
-			// not transparent, see ContentFeature::updateTextures
+	bool is_valid_position;
+
+	for (; offset < distance; offset += step) {
+		v3f pos_node_f = pos_origin_f + direction * offset;
+		v3s16 pos_node = floatToInt(pos_node_f, BS);
+
+		if (offset > distance - BLOCK_DIAGONAL) {
+			// Do accurate position checks:
+			// Check whether the node is inside the current mapblock
+			if (block_bounds.isPointInside(pos_node))
+				return false;
+		}
+
+		MapNode node = getNode(pos_node, &is_valid_position);
+
+		if (is_valid_position &&
+				!m_nodedef->get(node).light_propagates) {
+			// Cannot see through light-blocking nodes --> occluded
 			count++;
 			if (count >= needed_count)
 				return true;
@@ -1058,8 +1083,11 @@ bool Map::isOccluded(v3s16 pos_origin, v3s16 pos_blockcenter, float step,
 
 bool Map::isBlockOccluded(MapBlock *block, v3s16 cam_pos_nodes)
 {
+	// Check occlusion for center and all 8 corners of the mapblock
+	// Overshoot a little for less flickering
 	static const s16 bs2 = MAP_BLOCKSIZE / 2 + 1;
-	static const v3s16 dir8[8] = {
+	static const v3s16 dir9[9] = {
+		v3s16( 0,  0,  0),
 		v3s16( 1,  1,  1) * bs2,
 		v3s16( 1,  1, -1) * bs2,
 		v3s16( 1, -1,  1) * bs2,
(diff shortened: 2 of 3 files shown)


SmallJoker:
Updated the step factor `stepfac` from `1.1` to `1.05`. This results in more node checks, but more accurate occlusion culling.
Old `step` 50m: 141m
New `step` 50m: 13.8m

src/map.h
```diff
@@ -310,8 +310,8 @@ class Map /*: public NodeContainer*/
 	// This stores the properties of the nodes on the map.
 	const NodeDefManager *m_nodedef;
 
-	bool isOccluded(v3s16 p0, v3s16 p1, float step, float stepfac,
-			float start_off, float end_off, u32 needed_count);
+	bool isOccluded(v3s16 pos_camera, v3s16 pos_blockcenter, float step,
+		float stepfac, float start_offset, float end_offset, u32 needed_count);
```
ClobberXD: Inconsistent naming: the declaration uses `start_offset` while the definition (incl. the body) uses `offset`. `offset` makes the code look cleaner, but `start_offset` is easier to understand.

paramat:
I think blinking is not a good idea, it's confusing and distracting when using wireframe mode for other purposes. If it is very useful to have please can it be switched by a setting?

src/map.h
SmallJoker: Intended. `isOccluded` just recycles the variable, thus no renaming there.

SmallJoker:
@paramat somehow I already seen this coming. Without full range there *should* be no blinking, and with full range it's useful to see what blocks could be drawn, but are occlusion culled.
Please suggest a setting name that fits for this blinking, if you need it to be a setting.

ClobberXD:
Tested.

- Toggling `noclip` seems to have no effect.
- I noticed quite a few times that a couple of map blocks were culled while in direct view.
  
  screenshot_20190817_221723

  Moving to make a larger portion of the blocks exposed made them visible again (couldn't word this better, sorry). This isn't the case in master - while there are a handful of incorrectly culled blocks, they're less obvious (more accurate occlusion culling algo?) and also way fewer.

SmallJoker:
> Toggling noclip seems to have no effect.

It only does when you're inside a solid node (or ignore) where the occlusion culling will be disabled so you can see the underground.

screenshot_20190818_091624

I hope it's better now.

ClobberXD:
> It only does when you're inside a solid node (or ignore) where the occlusion culling will be disabled so you can see the underground.

Ah ok, I misunderstood. I'll re-test today.

src/map.cpp
```diff
@@ -1032,69 +1032,80 @@ void Map::removeNodeTimer(v3s16 p)
...
-	v3s16 u0 = p1 - p0;
-	v3f uf = v3f(u0.X, u0.Y, u0.Z) * BS;
-	uf.normalize();
-	v3f p0f = v3f(p0.X, p0.Y, p0.Z) * BS;
+bool Map::isOccluded(v3s16 pos_camera, v3s16 pos_target, float step,
+	float stepfac, float offset, float end_offset, u32 needed_count)
+{
+	float distance = BS * pos_camera.getDistanceFrom(pos_target);
+	v3f direction = intToFloat(pos_target - pos_camera, 1.0f);
+	direction.normalize();
```
DS-Minetest: `getDistanceFrom` and `normalize` both get the length, suggestion:
```cpp
v3f direction = intToFloat(pos_target - pos_camera, 1.0f);
float distance = direction.getLength();
direction /= distance || 1.0f;
distance *= BS;
```
or
```cpp
v3f direction = intToFloat(pos_target - pos_camera, BS);
float distance = direction.getLength() * BS;
direction /= distance || 1.0f;
```
.

SmallJoker: Did you mean? (invalid `|| 1.0f` and double-applied `BS`)
```C++
v3f direction = intToFloat(pos_target - pos_camera, BS);
float distance = direction.getLength();
direction /= distance;
```

DS-Minetest: Oops, the `BS` was indeed double-applied. >_<

But the `|| 1.0f` is to avoid nan for distance being `0.0f`

DS-Minetest:
Tested.
It seems to work.
I haven't found any falsely culled blocks.
Using noclip to disable culling works, too.
I don't think that the blinking is irritating. (Imo settings don't hurt, but there are people who are against too many settings for small things.)

paramat:
> wireframe_blink_occluded

Seems good.

I resist adding unnecessary settings, but wireframe mode is not only for checking occlusion, for most uses of wireframe mode the blinking will be unnecessary and a problem. I'm -1 unless it is made optional.

SmallJoker:
[I finally fixed everything](https://filestore.community.support.microsoft.com/api/images/2a7248aa-ee94-4ad3-879a-c21caf70d365).

 - Uncommented the blinking effect because the settings mess is already big enough
 - Re-use of the `distance` variable to normalize the direction vector.

src/client/clientmap.cpp
```diff
@@ -144,13 +144,18 @@ void ClientMap::updateDrawList()
...
-	if (g_settings->getBool("free_move")) {
+	if (g_settings->getBool("free_move") && g_settings->getBool("noclip")) {
 		MapNode n = getNode(cam_pos_nodes);
 		if (n.getContent() == CONTENT_IGNORE ||
 				m_nodedef->get(n).solidness == 2)
 			occlusion_culling_enabled = false;
 	}
 
+	// TODO: Uncomment to debug occluded blocks in the wireframe mode
+	// Include this as a flag for an extended debugging setting
```
sfan5: the TODO should be here, not on the line above

src/map.cpp
```diff
@@ -1032,69 +1032,82 @@ void Map::removeNodeTimer(v3s16 p)
...
-		if(f.drawtype == NDT_NORMAL){
-			// not transparent, see ContentFeature::updateTextures
+	bool is_valid_position;
+
+	for (; offset < distance + end_offset; offset += step){
+		v3f pos_node_f = pos_origin_f + direction * offset;
+		MapNode node = getNode(floatToInt(pos_node_f, BS), &is_valid_position);
+
+		if (is_valid_position &&
+					!m_nodedef->get(node).light_propagates) {
```
sfan5: 2 indents?

src/map.cpp
```diff
@@ -1032,69 +1032,82 @@ void Map::removeNodeTimer(v3s16 p)
...
-	for(float s=start_off; s<d0+end_off; s+=step){
-		v3f pf = p0f + uf * s;
-		v3s16 p = floatToInt(pf, BS);
-		MapNode n = getNode(p);
-		const ContentFeatures &f = m_nodedef->get(n);
-		if(f.drawtype == NDT_NORMAL){
-			// not transparent, see ContentFeature::updateTextures
+	bool is_valid_position;
+
+	for (; offset < distance + end_offset; offset += step){
```
sfan5: `) {`

src/map.cpp
DS-Minetest: >When breaking conditionals, indent following lines of the conditional with **two** tabs and the statement body with one tab.

https://dev.minetest.net/Code_style_guidelines#Spaces

But here are 3 tabs, which is wrong.

src/map.cpp
```diff
@@ -1032,69 +1032,92 @@ void Map::removeNodeTimer(v3s16 p)
...
-	// The occlusion search of 'isOccluded()' must stop short of the target
-	// point by distance 'endoff' (end offset) to not enter the target mapblock.
-	// For the 8 mapblock corners 'endoff' must therefore be the maximum diagonal
-	// of a mapblock, because we must consider all view angles.
-	// sqrt(1^2 + 1^2 + 1^2) = 1.732
-	float endoff = -BS * MAP_BLOCKSIZE * 1.732050807569;
-	s16 bs2 = MAP_BLOCKSIZE / 2 + 1;
+bool Map::isBlockOccluded(MapBlock *block, v3s16 cam_pos_nodes)
+{
+	// Check occlusion for center and all 8 edges of the mapblock
```
DS-Minetest: A mapblock has 12 [edges](https://en.wikipedia.org/wiki/Edge_(geometry)).

I suggest calling them vertices or corners.

src/map.cpp
```diff
@@ -1032,69 +1032,92 @@ void Map::removeNodeTimer(v3s16 p)
...
+	block_bounds.second = block_bounds.first + (MAP_BLOCKSIZE - 1);
+
+	v3s16 pos_blockcenter = block_bounds.first + (MAP_BLOCKSIZE / 2);
+
+	// Starting step size, value between 1m and sqrt(3)m
+	float step = BS * 1.2f;
+	// Multiply step by each iteraction by 'stepfac' to reduce checks in distance
+	float stepfac = 1.05f;
+
+	float start_offset = BS * 1;
```
DS-Minetest:
```suggestion
	float start_offset = BS * 1.0f;
```

DS-Minetest:
Digging a 1*1 horizontal hole can cause false clipping:
screenshot_20190819_224142

paramat:
Good temporary solution for blinking, fine for me.

SmallJoker:
Re-tested.
 * `irr::core::vector3d<T>` comparator operators return `true` as soon one axis matches the condition
   * Where would this be useful?!
 * Use of `irr::core::aabbox3d<T>` which offers the correct function `isPointInside()`
 * Call to `MapBlock::getBox()` which returns the wanted absolute coordinates
   * The-certain-misleading-variable plus functions need renaming in a future PR

fluxionary:
Something still isn't right:
deepin-screen-recorder_Minetest_20190821204004

lhofhansl:
@fluxionary this code is not perfect. It checks the 8 corners of each map block (and now also the center), so if a mapblock is just situated so that you can see part of it, but not the corners the block is incorrectly culled. In the past I experimented sending more rays to also test the 8 points half-way between the center and the corners, that makes it better, but this system cannot be perfect in all cases.

That all said... As the last person who touched the code (made the server also do occlusion culling and refactored the code some in the process) I will say that this all is extremely fickle. The slightest and innocent-looking changes in some of the values and the behavior changes a lot.
I find it quite hard to convince myself that the new code behaves exactly as (or at least not worse than) the current code.

lhofhansl:
When does the blinking occur? When you go above and below ground? That would seem to be OK.
Let's *please* not add another setting that nobody understands.

sfan5:
@lhofhansl we [discussed](http://irc.minetest.net/minetest-dev/2019-08-21#i_5580367) the PR and some further ideas in IRC yesterday
if you haven't already, you might want to take a look

paramat:
Comment on some of the IRC discussion: Corridors that go 'diagonally down' are very common in the 3D noise tunnels, and no less common than axis-aligned tunnels. Dungeons also have diagonal-downward corridors.

sfan5:
I propose merging this first, so occlusion culling can be improved in an upcoming PR.

SmallJoker:
https://github.com/minetest/minetest/commit/6ada090bb05d9e6d5961a59f33ff6e104b44061a
https://github.com/minetest/minetest/commit/7d016b4efd279d2ac2bbb84b457bfe6e5d7f4a56

lhofhansl:
Just finally got to test this. By itself it makes a bunch of scenario much worse.
The refactoring is nice, but I propose we do this without functional changes.
(As I said above, touching this code is fickle, and I have reverted many a change when I tried to improve it.)

lhofhansl:
Let's revert or fix. #8844 is not the full story. I'd rather have (mostly) correct visuals than more culled blocks.

lhofhansl:
See #8850

lhofhansl:
Actually looking at the code again... I do not understand why it would occlude so many more blocks.
Before this we just stop a BLOCK_DIAGONAL before the target block, with this we precisely check that the target block does not self-occlude. I would have thought it to be roughly equivalent.

SmallJoker:
http://irc.minetest.net/minetest-dev/2019-08-23#i_5581805

sfan5's PR will solve that for corridors, and the re-added diagonal skipping will ensure nearby mapblocks are always shown even though there's only a few nodes visible.

PR #8813, opened 2019-08-17
Clean up craft replacements docs

pauloue:
This confused me. Craft replacement support for items with a stack max > 1 was added in https://github.com/minetest/minetest/commit/17ba584fe254eeaee3489cc20e03810a59f3ef9b.

### DIFF
--- a/src/craftdef.h
+++ b/src/craftdef.h
@@ -110,9 +110,6 @@ struct CraftOutput
 	Example: If ("bucket:bucket_water", "bucket:bucket_empty") is a
 	replacement pair, the crafting input slot that contained a water
 	bucket will contain an empty bucket after crafting.
-
-	Note: replacements only work correctly when stack_max of the item
-	to be replaced is 1. It is up to the mod writer to ensure this.
 */
 struct CraftReplacements
 {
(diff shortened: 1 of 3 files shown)


src/craftdef.h
```diff
@@ -407,10 +404,22 @@ class ICraftDefManager
...
+	 *
+	 * @param input The input grid.
+	 * @param output CraftOutput where the result is placed.
+	 * @param output_replacements A vector of ItemStacks where replacements are
+	 * placed if they cannot be placed in the input. Replacements can be placed
+	 * in the input if the stack of the replaced item has a count of 1.
+	 * @param decrementInput If true, consume or replace input items.
+	 * @param gamedef
+	 * @return true if a result was found, otherwise false.
+	 */
```
ClobberXD: Yay, documentation! :D

SmallJoker:
Sorry but I still don't get it.
1) Is `output.replacements` a list based on `replacements` that did not have any match in the input list?
2) Why is the replacement count for input limited to 1?
3) Shouldn't `decremented_input` rather show the new input (after replacements)?

pauloue:
1. No. Replacement items that can be placed in the input (because the stack of the replaced item is empty) are placed there. `output.replacements` is a list of the replacement items that could not be placed in the input.
2. Multiple replacements can be placed in the input, but they are not placed there is the stack of the replaced item is not empty.
3. That is the current behavior.

Example:
```lua
local out, decremented_in = minetest.get_craft_result({
	method = "fuel",
	width = 1,
	items = {"bucket:bucket_lava 3"}
})
```
```lua
print(dump(out))
```
```
{
	time = 60,
	replacements = {
		ItemStack("bucket:bucket_empty")
	},
	item = ItemStack("")
}
```
```lua
print(dump(decremented_in))
```
```
{
	width = 1,
	items = {
		ItemStack("bucket:bucket_lava 2")
	},
	method = "fuel"
}
```
For normal crafts, items in the replacements table are put in the player's inventory or dropped.

PR #8815, opened 2019-08-18
Code cleanup in serverpackethandler

HybridDog:
* call pkt->getPeerId() only once
* remove redundant vector int to float to int conversion
* some style changes

### DIFF
--- a/src/network/serverpackethandler.cpp
+++ b/src/network/serverpackethandler.cpp
@@ -998,32 +998,43 @@ void Server::handleCommand_Interact(NetworkPacket *pkt)
 	verbosestream << "TOSERVER_INTERACT: action=" << (int)action << ", item="
 			<< item_i << ", pointed=" << pointed.dump() << std::endl;
 
-	RemotePlayer *player = m_env->getPlayer(pkt->getPeerId());
+	session_t peer_id = pkt->getPeerId();
+	RemotePlayer *player = m_env->getPlayer(peer_id);
 
 	if (player == NULL) {
 		errorstream << "Server::ProcessData(): Canceling: "
-				"No player for peer_id=" << pkt->getPeerId()
+				"No player for peer_id=" << peer_id
 				<< " disconnecting peer!" << std::endl;
-		DisconnectPeer(pkt->getPeerId());
+		DisconnectPeer(peer_id);
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


HybridDog:
It works for me.

src/network/serverpackethandler.cpp
```diff
@@ -998,32 +998,43 @@ void Server::handleCommand_Interact(NetworkPacket *pkt)
...
 
+	// Get pointed to node, may be undefined
+	v3s16 p_under = pointed.node_undersurface;
+	v3s16 p_above = pointed.node_abovesurface;
+	/* Uncomment when players can place a node onto an object.
+	if (pointed.type == POINTEDTHING_OBJECT) {
+		p_under = floatToInt(pointed_object->getBasePosition(), BS);
+		p_above = p_above;
+	}
+	*/
```
sfan5: I don't think keeping "theoretical support code" like this around is useful.

src/network/serverpackethandler.cpp
```diff
@@ -998,32 +998,43 @@ void Server::handleCommand_Interact(NetworkPacket *pkt)
...
 		errorstream << "Server::ProcessData(): Canceling: "
-				"No player object for peer_id=" << pkt->getPeerId()
+				"No player object for peer_id=" << peer_id
 				<< " disconnecting peer!" << std::endl;
-		DisconnectPeer(pkt->getPeerId());
+		DisconnectPeer(peer_id);
 		return;
 	}
 
+	// Get pointed to node, may be undefined
```
sfan5: and instead of letting this be undefined, why not keep it at the player_pos by default?

SmallJoker:
No response from author in more than a month. Please ask to re-open if there are news.

src/network/serverpackethandler.cpp
HybridDog: I've moved this code.

HybridDog:
I've rebased it and added more style changes; please re-open.

SmallJoker:
Sorry, GitHub does not let me re-opening the PR. Weird.. this should be allowed.

grafik

HybridDog:
> Sorry, GitHub does not let me re-opening the PR. 

I've opened a new one.

PR #8844, opened 2019-08-22
Improve occlusion culling

sfan5:
depends on #8850, reviewers see 2nd commit

In comparison this PR is much better than before with straight corridors

(before)
screenshot_20190823_221658
(after)
screenshot_20190823_221637

## To do

This PR is Ready for Review.
## How to test

Dig a straight (and diagonal) tunnel and observe occlusion culling behaviour.

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -1032,6 +1032,70 @@ void Map::removeNodeTimer(v3s16 p)
 	block->m_node_timers.remove(p_rel);
 }
 
+bool Map::determineAdditionalOcclusionCheck(const v3s16 &pos_camera,
+	const core::aabbox3d<s16> &block_bounds, v3s16 &check)
+{
+	/*
+		This functions determines the node inside the target block that is
+		closest to the camera position. This increases the occlusion culling
+		accuracy in straight and diagonal corridors.
+		The returned position will be occlusion checked first in addition to the
+		others (8 corners + center).
+		No position is returned if
+		- the closest node is a corner, corners are checked anyway.
+		- the camera is inside the target block, it will never be occluded.
+	*/
+#define CLOSEST_EDGE(pos, bounds, axis) \
+	((pos).axis <= (bounds).MinEdge.axis) ? (bounds).MinEdge.axis : \
+	(bounds).MaxEdge.axis
+
+	bool x_inside = (block_bounds.MinEdge.X <= pos_camera.X) &&
+			(pos_camera.X <= block_bounds.MaxEdge.X);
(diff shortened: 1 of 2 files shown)


sfan5:
@SmallJoker this is the approach I described on Wednesday extended with 1-dimensional checks to also account for diagonals

lhofhansl:
Tried to apply this on top of #8812, but it does not apply. #8812 applies cleanly.

ClobberXD:
@lhofhansl This PR already contains the commits from #8812.

src/map.cpp
```diff
@@ -1032,69 +1032,173 @@ void Map::removeNodeTimer(v3s16 p)
...
+		return true;
+	}
+
+	// diagonal
+	if (x_inside) {
+		check = v3s16(pos_camera.X, 0, 0);
+		check.Y = (pos_camera.Y <= block_bounds.MinEdge.Y) ?
+				block_bounds.MinEdge.Y : block_bounds.MaxEdge.Y;
+		check.Z = (pos_camera.Z <= block_bounds.MinEdge.Z) ?
+				block_bounds.MinEdge.Z : block_bounds.MaxEdge.Z;
```
SmallJoker: The error sources would decrease if this were a macro or a generic function like in https://github.com/minetest/minetest/blob/master/src/client/mesh.cpp#L268.

src/map.cpp
```diff
@@ -1032,69 +1032,173 @@ void Map::removeNodeTimer(v3s16 p)
...
+				block_bounds.MinEdge.Y : block_bounds.MaxEdge.Y;
+		return true;
+	}
+
+	// diagonal
+	if (x_inside) {
+		check = v3s16(pos_camera.X, 0, 0);
+		check.Y = (pos_camera.Y <= block_bounds.MinEdge.Y) ?
+				block_bounds.MinEdge.Y : block_bounds.MaxEdge.Y;
+		check.Z = (pos_camera.Z <= block_bounds.MinEdge.Z) ?
```
SmallJoker: Shouldn't this be compared against the mapblock center?

sfan5: The camera pos is only ever outside of the bounds, so this wouldn't make a difference.

lhofhansl:
> This PR already contains the commits from #8812.

D'oh. :)

DS-Minetest:
Tested.
This works pretty well.

DS-Minetest:
Ringing the bell.

SmallJoker:
Forwards:
screenshot_20190824_165506

Straight diagonal down corridors also work for me.
screenshot_20190824_170405

:+1:

PR #8846, opened 2019-08-23
Fix default hand definition not using wieldhand.png

Wuzzy2:
- Goal of the PR: To fix the bug that `wieldhand.png` is not used by default.
    - The bug: If no hand is defined by any mod, no texture for the wieldhand is used, instead the hand looks and behaves just really buggy, with either a "fake" texture with a single pixel, or a huge white box, or other oddities
- How does the PR work? By adding `wield_image` to the default hand definition
- Does it resolve any reported issue? Yes, the issue that's (indirectly) reported here: https://github.com/minetest/minetest/pull/8845
- If not a bug fix, why is this PR needed? What usecases does it solve? N/A

## Justification

`wieldhand.png` is mentioned in texture docs. So it should be actually used by the engine.

## How to test

* Create an empty game (no mods)
* Start it
* Select the hand

Make an before/after comparison.

### DIFF
--- a/builtin/game/register.lua
+++ b/builtin/game/register.lua
@@ -373,6 +373,7 @@ core.register_node(":ignore", {
 -- The hand (bare definition)
 core.register_item(":", {
 	type = "none",
+	wield_image = "wieldhand.png",
 	groups = {not_in_creative_inventory=1},
 })
 


Panquesito7:
> - Does it resolve any reported issue? Yes, the issue that's (indirectly) reported here: #8845

You should type "Fixes #8845".

ClobberXD:
> You should type "Fixes #8845".

Words like "fixes" and "closes", [close the issue]((https://help.github.com/en/articles/closing-issues-using-keywords)) with the following number. Using this phraseology for PRs make no sense, and would just add to the confusion.

builtin/game/register.lua
```diff
@@ -373,6 +373,7 @@ core.register_node(":ignore", {
 -- The hand (bare definition)
 core.register_item(":", {
 	type = "none",
+	wield_image = "wieldhand.png",
 	groups = {not_in_creative_inventory=1},
```
ClobberXD: It'd be nice if you could fix the spacing in L376 too:

```suggestion
	groups = {not_in_creative_inventory = 1},
```

Wuzzy2:
IMO coding style changes are better suited in a separate PR. And more than just 1 line.
Just saying …

ClobberXD:
Alright, no issues. :)

SmallJoker:
https://github.com/minetest/minetest/commit/9c725609c3e2516c9dae8f4515f888d99de80a48

PR #8847, opened 2019-08-23
Add ItemStack:get_description() to get tooltip

### LINKED ISSUE #8080: Add function to get “real” description of itemstack
##### Issue type
 - Feature request
 
##### Minetest version
0.4.17.1

##### Summary
It's not obvious to get the “real” description (=tooltip) of an itemstack.
So far, the most common way is just by lookup in `minetest.registered_items[itemstring].description`.
However, this method is not reliable:
The description might be manipulated by the `description` item metadata, or it might have been falling back to the itemstring because the `description` field of the itemstring is empty. And it's possible that other ways to manipulate the tooltip / “real” description in future will appear. So manually parsing the description of an itemstack is not only tedious but also not future-proof.

Therefore, I suggest to add a function `get_description()` for itemstacks.

It returns the real user-visible description of the item, as displayed in the tooltip.
Color shall be preserved.

pauloue:
Closes #8080

## To do

This PR is Ready for Review.

## How to test

- Check that item tooltips in inventory lists are shown correctly.
```lua
print(ItemStack("default:pick_steel"):get_description()) --> Steel Pickaxe
print(ItemStack("foobar"):get_description()) --> Unknown Item
local stack = ItemStack("default:stone")
stack:get_meta():set_string("description", "Custom description")
print(stack:get_description()) --> Custom description
print(ItemStack("beds:bed_top"):get_description()) --> beds:bed_top
```

### DIFF
--- a/src/gui/guiFormSpecMenu.cpp
+++ b/src/gui/guiFormSpecMenu.cpp
@@ -2856,9 +2856,10 @@ void GUIFormSpecMenu::drawList(const ListDrawSpec &s, int layer,
 
 			// Draw tooltip
 			if (hovering && !m_selected_item) {
-				std::string tooltip = item.getDescription(
-						m_tooltip_append_itemname, m_client->idef());
+				std::string tooltip = item.getDescription(m_client->idef());
 				if (!tooltip.empty()) {
+					if (m_tooltip_append_itemname)
+						tooltip += "\n[" + item.name + "]";
 					showTooltip(utf8_to_wide(tooltip), m_default_tooltip_color,
 							m_default_tooltip_bgcolor);
 				}
--- a/src/gui/guiFormSpecMenu.cpp
+++ b/src/gui/guiFormSpecMenu.cpp
@@ -2844,20 +2844,17 @@ void GUIFormSpecMenu::drawList(const ListDrawSpec &s, int layer,
 		}
 
 		if (layer == 1) {
-			// Draw item stack
 			if (selected)
 				item.takeItem(m_selected_amount);
 
 			if (!item.empty()) {
+				// Draw item stack
 				drawItemStack(driver, m_font, item,
 					rect, &AbsoluteClippingRect, m_client,
 					rotation_kind);
-			}
-
-			// Draw tooltip
-			if (hovering && !m_selected_item) {
-				std::string tooltip = item.getDescription(m_client->idef());
-				if (!tooltip.empty()) {
+				// Draw tooltip
+				if (hovering && !m_selected_item) {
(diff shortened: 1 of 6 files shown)


src/inventory.cpp
```diff
@@ -246,6 +246,19 @@ std::string ItemStack::getItemString() const
...
+std::string ItemStack::getDescription(bool append_itemname,
+		IItemDefManager *itemdef) const
+{
+	std::string desc = metadata.getString("description");
+	if (desc.empty())
+		desc = getDefinition(itemdef).description;
+	if (desc.empty())
+		return name;
+	if (append_itemname)
+		desc += "\n[" + name + "]";
```
SmallJoker: I'd prefer if this client-only feature would added in `guiFormSpecMenu.cpp`. If you'd like to adapt that, it's also possible to set an input pointer to `true` just like `getNode` does for `is_valid_position`.

SmallJoker:
Would it be necessary to mention that the string might contain client-side translation marks? I don't think that many people would be aware of them, but if `getDescription` is copied to a text output it would work as expected.

src/inventory.cpp
pauloue: Moved it, but now the tooltip could have the itemname twice if there's no description. I think that's ok since `tooltip_append_itemname` is mostly for developers and items in the inventory will normally have a description.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -2855,26 +2855,15 @@ void GUIFormSpecMenu::drawList(const ListDrawSpec &s, int layer,
...
-				else
-					tooltip_text = utf8_to_wide(desc);
-
-				if (!item.name.empty()) {
-					if (tooltip_text.empty())
-						tooltip_text = utf8_to_wide(item.name);
-					else if (m_tooltip_append_itemname)
-						tooltip_text += utf8_to_wide("\n[" + item.name + "]");
+				std::string tooltip = item.getDescription(m_client->idef());
+				if (!tooltip.empty()) {
```
sfan5: this will always be true since getDescription never returns an empty description

src/inventory.cpp
sfan5: The new implementation isn't really better, also the behavior is different.
Might just be better to leave it with the parameter.

src/gui/guiFormSpecMenu.cpp
pauloue: getDescription returns an empty description when the stack is empty. I guess there's no need to get the tooltip in that case.

FreeLikeGNU:
The real gold here is how to use the itemstack in the example pauloue posted.

PR #8850, opened 2019-08-23
Restore approximate occlusion check

sfan5:
While less precise, it worked better which is what matters in the end.

## To do

This PR is Ready for Review.

## How to test

Observe that occlusion culling works better than before in narrow corridors.


lhofhansl:
Let's merge and improve the behavior in other PRs.

PR #8852, opened 2019-08-24
Improve performance of occlusion culling.

lhofhansl:
This should do both (slightly) more accurate and faster occlusion culling.

Please have a look. The main observations ar

1. If the minimum step size is slightly >= sqrt(3) a node cannot be touched more than one.
2. ~~We should start a bit back from the camera (to account for the fact that the camera is pane - not a point.) Use the rear_pane distance for that.~~
3. Now the needed_count hack is no longer needed.
4. The start_offset should also be remove (i.e. be 0), since otherwise the visibility is camera direction dependent, which leads to weird artifacts.

In a few tests I've done this is as accurate (does not cull block that should be culled), culls slightly more of the invisible blocks, and is more performant.

BUT... As the related issues before... This needs to be very carefully tested.
So consider this experimental, play around with it.

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -1130,8 +1130,7 @@ bool Map::isBlockOccluded(MapBlock *block, v3s16 cam_pos_nodes)
 	// Check occlusion for center and all 8 corners of the mapblock
 	// Overshoot a little for less flickering
 	static const s16 bs2 = MAP_BLOCKSIZE / 2 + 1;
-	static const v3s16 dir9[9] = {
-		v3s16( 0,  0,  0),
+	static const v3s16 dir8[8] = {
 		v3s16( 1,  1,  1) * bs2,
 		v3s16( 1,  1, -1) * bs2,
 		v3s16( 1, -1,  1) * bs2,
@@ -1167,7 +1166,13 @@ bool Map::isBlockOccluded(MapBlock *block, v3s16 cam_pos_nodes)
 			return false;
 	}
 
-	for (const v3s16 &dir : dir9) {
+	// Block center, end_offset can be halfed.
+	if (!isOccluded(cam_pos_nodes, pos_blockcenter, step, stepfac,
+			start_offset, end_offset / 2.0f))
+		return false;
+
+	// All the block's corners.
+	for (const v3s16 &dir : dir8) {
 		if (!isOccluded(cam_pos_nodes, pos_blockcenter + dir, step, stepfac,
 				start_offset, end_offset))
 			return false;
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -1097,7 +1097,7 @@ bool Map::determineAdditionalOcclusionCheck(const v3s16 &pos_camera,
 }
 
 bool Map::isOccluded(const v3s16 &pos_camera, const v3s16 &pos_target,
-	float step, float stepfac, float offset, float end_offset, u32 needed_count)
+	float step, float stepfac, float offset, float end_offset)
 {
 	v3f direction = intToFloat(pos_target - pos_camera, BS);
 	float distance = direction.getLength();
@@ -1107,7 +1107,6 @@ bool Map::isOccluded(const v3s16 &pos_camera, const v3s16 &pos_target,
 		direction /= distance;
 
 	v3f pos_origin_f = intToFloat(pos_camera, BS);
-	u32 count = 0;
 	bool is_valid_position;
 
 	for (; offset < distance + end_offset; offset += step) {
@@ -1119,9 +1118,7 @@ bool Map::isOccluded(const v3s16 &pos_camera, const v3s16 &pos_target,
 		if (is_valid_position &&
 				!m_nodedef->get(node).light_propagates) {
 			// Cannot see through light-blocking nodes --> occluded
-			count++;
-			if (count >= needed_count)
-				return true;
+			return true;
 		}
 		step *= stepfac;
 	}
@@ -1147,9 +1144,9 @@ bool Map::isBlockOccluded(MapBlock *block, v3s16 cam_pos_nodes)
 
 	v3s16 pos_blockcenter = block->getPosRelative() + (MAP_BLOCKSIZE / 2);
 
-	// Starting step size, slightly > sqrt(3),
+	// Starting step size is sqrt(3),
 	// so that each node is hit at most once.
-	float step = BS * 1.7321f;
+	float step = BS * 1.732f;
 	// Multiply step by each iteraction by 'stepfac' to reduce checks in distance
 	float stepfac = 1.05f;
 
@@ -1166,14 +1163,13 @@ bool Map::isBlockOccluded(MapBlock *block, v3s16 cam_pos_nodes)
 	v3s16 check;
 	if (determineAdditionalOcclusionCheck(cam_pos_nodes, block->getBox(), check)) {
 		// node is always on a side facing the camera, end_offset can be lower
-		if (!isOccluded(cam_pos_nodes, check, step, stepfac, start_offset,
-				-1.0f, 1))
+		if (!isOccluded(cam_pos_nodes, check, step, stepfac, start_offset, -1.0f))
 			return false;
 	}
 
 	for (const v3s16 &dir : dir9) {
 		if (!isOccluded(cam_pos_nodes, pos_blockcenter + dir, step, stepfac,
-				start_offset, end_offset, 1))
+				start_offset, end_offset))
 			return false;
 	}
 	return true;
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -1097,20 +1097,10 @@ bool Map::determineAdditionalOcclusionCheck(const v3s16 &pos_camera,
 }
 
 bool Map::isOccluded(const v3s16 &pos_camera, const v3s16 &pos_target,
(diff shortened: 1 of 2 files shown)


lhofhansl:
The behavior *is* slightly different. Better in some case, worse in other cases.
I think combined with #8844 this could be quite good.

lhofhansl:
Of course the offset for the camera has nothing to do with the near_plane, it's function of fov and the viewport (screen) position.

src/map.cpp
```diff
@@ -1097,14 +1100,9 @@ bool Map::isBlockOccluded(MapBlock *block, v3s16 cam_pos_nodes)
...
 
-	// to reduce the likelihood of falsely occluded blocks
-	// require at least two solid blocks
-	// this is a HACK, we should think of a more precise algorithm
-	u32 needed_count = 2;
-
 	for (const v3s16 &dir : dir9) {
 		if (!isOccluded(cam_pos_nodes, pos_blockcenter + dir, step, stepfac,
-				start_offset, end_offset, needed_count))
+				start_offset, end_offset, 1))
```
ClobberXD: If the new technique doesn't require `needed_count`, this extra param could just be removed altogether?

lhofhansl: Yep. I did not want to confuse the change too much.

src/map.cpp
```diff
@@ -1083,12 +1083,15 @@ bool Map::isBlockOccluded(MapBlock *block, v3s16 cam_pos_nodes)
...
+	// Starting step size, slightly > sqrt(3),
+	// so that each block is hit at most once.
+	float step = BS * 1.7321f;
 	// Multiply step by each iteraction by 'stepfac' to reduce checks in distance
 	float stepfac = 1.05f;
 
-	float start_offset = BS * 1.0f;
+	// Take the camera's front view plane into account, and start slighly back.
+	// (TODO: use near_plane config)
+	float start_offset = -BS * 0.1f;
```
lhofhansl: This is something I can help with. This should be the distance of the camera (the point) from the viewport (the screen). Is there a good way to calculate that?

lhofhansl:
I think another key observation is that the camera's viewpane is *not* a point. Rather we view through an area. One of the problems is that we only check from the center of the camera. We should from more points on the pane (perhaps the center and the four corners). It would be 5x the amount of work, but be far more accurate.
That is to say: A target might not be visible from the center of the screen, but might be from the side or a corner of the screen.

Now... What's the size of the screen in world-cordinates? Or what's the focal length of the camera, i.e. how far is the camera point from the actual view port through which the scene is watched?

lhofhansl:
Rebased and combined with @SmallJoker 's #8812 .
It still looks pretty accurate and the culling seems to be better.

Moth-Tolias:
> I think another key observation is that the camera's viewpane is _not_ a point. Rather we view through an area. One of the problems is that we only check from the center of the camera. We should from more points on the pane (perhaps the center and the four corners). It would be 5x the amount of work, but be far more accurate.

this could be achieved by using the point "behind" the viewpane where these intersect, right?

lhofhansl:
> this could be achieved by using the point "behind" the viewpane where these intersect, right?

I do not think so, since you're still just looking from that one point. An object might be occluded from there, but not from the side of the screen - this the problem I see the most often: You're just looking just around a corner, but the camera-point is still behind the corner but the side of the screen is is not.

lhofhansl:
> It still looks pretty accurate and the culling seems to be better.

Responding to myself... Nah, it's not ready. I see many bad artifact with the #8812 pulled in.

lhofhansl:
Removed #8812 again.
This code seems to be fairly equivalent to current master, but does much less work.
It starts with a larger step, large enough to make sure that the same block cannot be hit twice by a ray, and because of that it not need the counting hack.

In the end doing ray tracing here - as stated before - cannot be 100% accurate and fill require some tradeoffs between accuracy and performance.

src/map.cpp
lhofhansl: Sorry I meant I want to keep the change as small as possible. If this works out I'll remove the count nonsense completely.

lhofhansl:
Travis seems to have env issues.
Anybody willing to test this?

I'll do the camera offset points in another PR.

lhofhansl:
I have a prototype of ray casting from 5 points in the viewpane. With that I have not been able to see *any* false positives - although of course they are still possible. At all. :)

Needs some more testing, though. And of course it's quite a bunch of (5x) more work.

NathanSalapat:
I just built this, with no issues, how exactly would I go about testing it?

src/map.cpp
ClobberXD: Yay :)

src/map.cpp
```diff
@@ -1147,12 +1147,15 @@ bool Map::isBlockOccluded(MapBlock *block, v3s16 cam_pos_nodes)
 
 	v3s16 pos_blockcenter = block->getPosRelative() + (MAP_BLOCKSIZE / 2);
 
-	// Starting step size, value between 1m and sqrt(3)m
-	float step = BS * 1.2f;
+	// Starting step size, slightly > sqrt(3),
+	// so that each block is hit at most once.
```
sfan5: this initially confused me, it should say "so that each node is hit at most once"

src/map.cpp
sfan5: Isn't that the value returned by [`ICameraSceneNode::getNearValue()`](http://irrlicht.sourceforge.net/docu/classirr_1_1scene_1_1_i_camera_scene_node.html#aee5dfccee2ec0b0cbcdb1079a6430a25)?

Wuzzy once posted this illustration:

src/map.cpp
lhofhansl: Sorry, yes, each node. I'll update the comment in the next round.

lhofhansl:
@NathanSalapat Best testing is around long corridors (straight and diagonal) and see whether anything that should be shown is occluded.

The patch so far just reduces the work done (with hopefully equivalent behavior). I'm working on another change that adds a bunch of work again, but which should avoid false culling much better.

src/map.cpp
lhofhansl: Thanks @sfan5 . I dug out my trusted old "Computer Graphics - Principles and Practice" book (I love it when I get a reason to do that).
Looks like the near clipping place actually has nothing do to the projection pane, it's just there to clip near objects to avoid distortions.

Somewhere is the actual projection pane (i.e. your screen). If I can easily find its coordinates I can place the screen in world coordinates and then use rays from its corners.

In any case. It turns out that part of the change is not actually needed, I'll remove that soon.

lhofhansl:
Thinking about other options... One could also render nodes into a low-res software z-buffer. That way one could *precisely* determine which nodes (and hence blocks) are visible. Perhaps for nodes it's too expensive.
But it should definitely be possible for blocks, there we just would consider a block transparent if it has any transparent node (could make it better and pre-compute whether there are any transparent paths through the block). The we'd scan convert each block as just a pixel, keep track of the Z-coordinate and the block id. Then we'd have an accurate view of all blocks that a known visible... we might show too many blocks, but that's OK.
And perhaps blocks are too coarse, would have to try and see.

lhofhansl:
Turns out that even underground 80% of all maplbocks have at one transparent node in either of its 6 borders. I did not expect that. Here's how I calculate that (looks correct to me):
```
bool MapBlock::hasTransparentBorder()
{
	const NodeDefManager *nodemgr = m_gamedef->ndef();

	// z planes
	for (int x=0; x<MAP_BLOCKSIZE; x++)
	for (int y=0; y<MAP_BLOCKSIZE; y++)
	{
		for (int z=0; z<MAP_BLOCKSIZE; z+=MAP_BLOCKSIZE-1) {
			const MapNode &node = data[z * zstride + y * ystride + x];
			if (nodemgr->get(node).light_propagates) {
				return true;
			}
		}
	}
	// y planes
	for (int x=0; x<MAP_BLOCKSIZE; x++)
	for (int z=1; z<MAP_BLOCKSIZE-1; z++)
	{
		for (int y=0; y<MAP_BLOCKSIZE; y+=MAP_BLOCKSIZE-1) {
			const MapNode &node = data[z * zstride + y * ystride + x];
			if (nodemgr->get(node).light_propagates) {
				return true;
			}
		}
	}
	
	// x planes
	for (int y=1; y<MAP_BLOCKSIZE-1; y++)
	for (int z=1; z<MAP_BLOCKSIZE-1; z++)
	{
		for (int x=0; x<MAP_BLOCKSIZE; x+=MAP_BLOCKSIZE-1) {
			const MapNode &node = data[z * zstride + y * ystride + x];
			if (nodemgr->get(node).light_propagates) {
				return true;
			}
		}
	}
	return false;
}
```

lhofhansl:
Updated based on @sfan5 's comments.

lhofhansl:
In any case... I removed the needed_count. This code makes more sense, does less work, and produces (as far as I can see) the same results. Should be good.

lhofhansl:
Also restored the fact that for the ray to the center of the block we can only need an end_offset of sqrt(3)/2. That was lost with #8812. @sfan5

lhofhansl:
Yet one more change. start_offset should be zero. Otherwise the occlusion visibility become view direction dependent.

lhofhansl:
Upon reflection... Unless we're making a fundamental change to what we have it just has to be "good enough".

PR #8856, opened 2019-08-25
Send cumulated inventory changes only each step

SmallJoker:
This PR does further inventory sending optimizations. Previously `setInventoryModified` sent the inventory in many cases, such as:
 - Any change by Lua
 - Most interact actions

Now all changes *from Lua* will be collected and sent at the end of each server step. Here's a before/after comparison (without second commit):

<details>
<summary>Before</summary>

```
inv::serialize main
inv::serialize main
<^ snip, 16x the same>
changeItem craft
changeItem craft
<^ snip, 8x the same>
changeItem craftpreview
inv::serialize craft
inv::serialize craftpreview
```
</details>

<details>
<summary>After</summary>

```
changeItem craft
changeItem craft
<snip, 8x the same>
changeItem craftpreview
inv::serialize main
inv::serialize craft
inv::serialize craftpreview
```
</details>

## To do

This PR is Ready for Review.

- [x] Test whether this works well on laggy servers

## How to test

1) Keep the logging lines for testing
2) Start a world using a mod that does many `inv:remove_item` or `inv:add_item` calls
  * Example: unified_inventory's "copy recipe to crafting grid" function
3) Trigger the mod so that it does many inventory changes
4) See log output. Each `inv::serialize` means it's sent to the client

### DIFF
--- a/src/inventory.cpp
+++ b/src/inventory.cpp
@@ -569,7 +569,6 @@ ItemStack InventoryList::changeItem(u32 i, const ItemStack &newitem)
 	ItemStack olditem = m_items[i];
 	m_items[i] = newitem;
 	setModified();
-	std::cout << "changeItem " << m_name << std::endl;
 	return olditem;
 }
 
@@ -848,8 +847,6 @@ void Inventory::serialize(std::ostream &os, bool incremental) const
 		if (!incremental || list->checkModified()) {
 			os << "List " << list->getName() << " " << list->getSize() << "\n";
 			list->serialize(os, incremental);
-			if (incremental)
-				std::cout << "inv::serialize " << list->getName() << std::endl;
 		} else {
 			os << "KeepList " << list->getName() << "\n";
 		}
--- a/src/inventory.cpp
+++ b/src/inventory.cpp
@@ -569,6 +569,7 @@ ItemStack InventoryList::changeItem(u32 i, const ItemStack &newitem)
 	ItemStack olditem = m_items[i];
 	m_items[i] = newitem;
 	setModified();
+	std::cout << "changeItem " << m_name << std::endl;
 	return olditem;
 }
 
@@ -847,6 +848,8 @@ void Inventory::serialize(std::ostream &os, bool incremental) const
 		if (!incremental || list->checkModified()) {
 			os << "List " << list->getName() << " " << list->getSize() << "\n";
 			list->serialize(os, incremental);
+			if (incremental)
+				std::cout << "inv::serialize " << list->getName() << std::endl;
 		} else {
 			os << "KeepList " << list->getName() << "\n";
 		}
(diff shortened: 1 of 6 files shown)


lhofhansl:
Nice

thomasrudin:
As far as i see this is a server-only change, right?
So existing clients can benefit from an up-to-date server...?

SmallJoker:
@thomasrudin Whereas this is server-only, the players will still need a recent Minetest version in order to receive incrementally sent inventories. However, this PR also bundles multiple inventory-modifying calls from Lua into a single action, resulting is less data sent overall.

thomasrudin:
This is live on the pandorabox-server, i'll report back here if there are issues...

thomasrudin:
Having some issues with non-updated inventories in the `3d_armor` mod:


The boots are in there (hint: the preview on the right) but it does not show up right away, only if you add or remove some armor

SmallJoker:
@thomasrudin Those are detached inventories, and are unaffected by this change. Are you sure this bug does not occur in current master as well?

thomasrudin:
> @thomasrudin Those are detached inventories, and are unaffected by this change. Are you sure this bug does not occur in current master as well?

I'm switching later to https://github.com/minetest/minetest/commit/4291961e09db7851fbc940eac58a1b919668d935 again....
The bug must have been introduced after https://github.com/minetest/minetest/commit/bf22184d6e16f21b9001322aad1bf8f6dbaa372b, thats the last build i used and it worked there without problems...

SmallJoker:
Found a possible solution for this problem.
It's caused by sending detached inventories before they're actually modified. I wonder how this worked previously.

SmallJoker:
Update: tested the changes locally and the "ghost item stack" does no longer occur.

SmallJoker:
Did some tests with high server step values, but I could barely see the difference. Works well.

ClobberXD:
I understand that testing this in a remote-server + local-client is the way to go, but is there any other way I could help test this locally?

SmallJoker:
@ClobberXD You could test this one either by placing `std::cout`s in the inventory receiving function in clientpackethandler.cpp, and by simulating network lag using https://stackoverflow.com/questions/1094760/network-tools-that-simulate-slow-network-connection/1094797#1094797 .
I did not test latter, but the Lua step happens very shortly before sending all inventory changes.

I just noticed that crafting should also be way faster now. `ICraftAction::apply` called `setInventoryModified` (included inventory sending) for each crafting step. Instead of 10 individual inventory sends, there should only be one left.

SmallJoker:
fae6242d4ef6ca34d131054484acf5efc2fcba70

ClobberXD:
@SmallJoker Thanks for the reply, the lag simulation is especially new to me (that'd actually help a lot in testing our network-related PRs). Sorry I couldn't test this before merging.

PR #8857, opened 2019-08-25
Document world format version 28 (private nodemeta)

DS-Minetest:
- Goal of the PR: documentation
- How does the PR work? Well, there's more human-readable text in `doc/world_format.txt `.
- Does it resolve any reported issue? No. This change should probably already have happened in #5702.
- If not a bug fix, why is this PR needed? What usecases does it solve? I consider missing documentation as a bug.

## To do

## How to test

Read it and check whether the information is correct.

### DIFF
--- a/doc/world_format.txt
+++ b/doc/world_format.txt
@@ -365,7 +365,7 @@ if map format version <= 22:
     u16 content_size
     u8[content_size] content of metadata. Format depends on type_id, see below.
 if map format version >= 23:
-  u8 version (=1) -- Note the type is u8, while for map format version <= 22 it's u16
+  u8 version (=1 for map format version < 28, else =2) -- Note the type is u8, while for map format version <= 22 it's u16
   u16 count of metadata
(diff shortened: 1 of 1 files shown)


doc/world_format.txt
```diff
@@ -365,7 +365,7 @@ if map format version <= 22:
     u16 content_size
     u8[content_size] content of metadata. Format depends on type_id, see below.
 if map format version >= 23:
-  u8 version (=1) -- Note the type is u8, while for map format version <= 22 it's u16
+  u8 version (=1 for map format version < 28, else =2) -- Note the type is u8, while for map format version <= 22 it's u16
```
SmallJoker: More readable suggestion: (line too long)
```
  u8 version -- Was `u16` for map format version <= 22
    -- Value `1` for map format version < 28
    -- Value `2` since map format version 28
```

DS-Minetest: Fixed.

PR #8860, opened 2019-08-25
Force item entities out of solid nodes

### LINKED ISSUE #5297: Force dropped items out of solid nodes
An annoying issue is that dropped items can somehow end up inside solid nodes.
The most obvious scenario for this is when a falling node such a sand falls on dropped items.
But players can also just easily build on them as well.

For the player it seems the items are gone but they simply stay at the place. When you remove the block, the item becomes visible again.

The problem with this is also that many dropped items can end up pointlessly inside solid nodes. The players may be unaware of their existance,. This means that these dropped items just increase the number of entities in the world for no benefit, until they get either removed by the item entity TTL (which takes long) or a player find it again (which is unlikely).

So, my simple idea is to force items out of blocks when they are inside a solid node.
“Solid node” is here defined as any `walkable` node with a full cube collision box (1×1×1 collision box or larger), and a “free” node is any node that is not solid and not `ignore`.

When the item is in a solid node, it should look at its direct node neighbors: left, right, front, back and up. It first checks if and of the left, right, front or back nodes are free. If there is, it will start moving towards the node with the closest distance to “freedom”. If all 4 sideway nodes are not free, the item will move upwards, but only if the node above is not `ignore`. In that case, the item does not move until the area above is loaded. This is to prevent items floating into unloaded areas or even out of world boundaries.
The algorithm terminates as soon as the item's selection box is completely outside of any node's collision box, then the normal item physics apply (i.e. gravity).

This algorithm implies that items are still stuck inside a node if they are stuck in solid rock and are directly below `ignore`. If they are at the top of the world, they are stuck in the node forever
This also means items can not move out of map boundaries (e.g. the top of the world).

My algorithm may not be perfect. First, I have no idea how seriously this would affect performance if there are many dropped items stuck inside somewhere.
Second, it does not handle cases where items get stuck into stairs and slabs, but maybe you have an idea on how to handle this.

Wuzzy2:
This PR makes item entities getting pushed out of solid nodes. Closes #5297.

The algorithm is mostly as described in the issue, but I explain it again in brief:

* A “solid” node is defined as a node that is walkable and has a regular collision box
* A “free” node is any non-solid node that is also not `ignore`
* If item is inside solid node, disable traditional physics, and instead apply “force-out” physics
* The “force-out” physics: If item is inside a solid node, do:
    * Disable the classic item physics. `physical` is set to false (no collision), and the speed is completely controlled by the “force-out” physics now
    * Look at the 4 direct horizontal neighboring nodes
    * If any neighbor is free, push item to the closest free node
    * If all 4 neighbors are unfree, look at node above
    * If node above is not `ignore`, push item upwards (even if the node above is also solid)
    * Otherwise, do absolutely nothing
    * When item's collision box is completely outside of any solid node, the algorithm terminates and the standard item physics are enabled again

## Analysis
To illustrate what this will do:

Items that are lost deep in the underground tend to get pushed all the way to the surface, otherwise they just go sideways.

There's one special case in which even a stuck item refuses to move: When it's at the top of the world, and is surrounded by solid nodes. It won't go up because there's an `ignore`. This is deliberate to prevent destroying the item.

## Notes
The force-out does not apply for other solid nodes, such as complex node boxes like slabs. This is deliberate, to keep things reasonably simple.
In the original issue I specified “1×1×1 or larger” but I discarded the “or larger” for complexity reasons. It would require looking deep into the collisionbox definition, not sure if this might be a performance problem.

The algorithm has been copied from MineClone 2's mod `mcl_item_entity`. The code is mostly the same, the changes are mostly just variable name changes, but the algorithm is exactly the same.

This feature in MineClone 2 is in the game for a very long time, and I have not heard of any bug reports for item entities for ages now. So I'm pretty sure the algorithm is correct and the code is well tested. One possible source of bugs are bugs that appeared by my converting the code from MCL2 to builtin.

## How to test

* Start any game which does not override item entities (e.g. Minetest Game)
* Drop some items on the ground
* Place blocks inside the item

Test cases to test:
* Try different item positons (at the left, right front or back side of node boundaries)
* Try locking the item at the top of the world (below `ignore`)
* Try forcing the item out sideways
* Try forcing the item out upwards
* Drop item in deep 1×1 hole, then drop a huge sand pile into the hole
* Hint: Fly+noclip help a lot to find stuck items! :) Or just use glass

### DIFF
--- a/builtin/game/item_entity.lua
+++ b/builtin/game/item_entity.lua
@@ -178,12 +178,16 @@ core.register_entity(":__builtin:item", {
 		end
 
 		local node_this = core.get_node_or_nil(pos)
-		local def_this
+		local is_stuck = false
 		if node_this then
-			def_this = minetest.registered_nodes[node_this.name]
+			local sdef = minetest.registered_nodes[node_this.name]
+			is_stuck = (sdef.walkable == nil or sdef.walkable == true)
+				and (sdef.collision_box == nil or sdef.collision_box.type == "regular")
+				and (sdef.node_box == nil or sdef.node_box.type == "regular")
 		end
+
 		-- Push item out when stuck inside solid node
-		if def_this and def_this.walkable then
+		if is_stuck then
 			local shootdir
 			local cx = pos.x % 1
 			local cz = pos.z % 1
--- a/builtin/game/item_entity.lua
+++ b/builtin/game/item_entity.lua
@@ -27,6 +27,7 @@ local function enable_physics(object, luaentity)
 		object:set_acceleration({x=0,y=-gravity,z=0})
 	end
 end
+
 local function disable_physics(object, luaentity)
 	if luaentity.physical_state == true then
 		luaentity.physical_state = false
@@ -253,12 +254,19 @@ core.register_entity(":__builtin:item", {
 		-- It is responsible for making sure the entity is entirely outside the solid node
 		-- (with its full collision box), not just its center.
 		if self.force_out_timer > 0 then
-			local cbox = self.object:get_properties().collisionbox
+			local c = self.object:get_properties().collisionbox
+			local s = self.force_out_start
+			local f = self.force_out
 			local ok = false
-			if self.force_out.x > 0 and (pos.x > (self.force_out_start.x + 0.5 + (cbox[4] - cbox[1])/2)) then ok = true
-			elseif self.force_out.x < 0 and (pos.x < (self.force_out_start.x + 0.5 - (cbox[4] - cbox[1])/2)) then ok = true
-			elseif self.force_out.z > 0 and (pos.z > (self.force_out_start.z + 0.5 + (cbox[6] - cbox[3])/2)) then ok = true
-			elseif self.force_out.z < 0 and (pos.z < (self.force_out_start.z + 0.5 - (cbox[6] - cbox[3])/2)) then ok = true end
+			if f.x > 0 and (pos.x > (s.x + 0.5 + (c[4] - c[1])/2)) then
+				ok = true
+			elseif f.x < 0 and (pos.x < (s.x + 0.5 - (c[4] - c[1])/2)) then
+				ok = true
+			elseif f.z > 0 and (pos.z > (s.z + 0.5 + (c[6] - c[3])/2)) then
+				ok = true
+			elseif f.z < 0 and (pos.z < (s.z + 0.5 - (c[6] - c[3])/2)) then
+				ok = true
+			end
 			-- Item was successfully forced out. No more pushing
 			if ok then
 				self.force_out_timer = -1
--- a/builtin/game/item_entity.lua
+++ b/builtin/game/item_entity.lua
@@ -17,6 +17,26 @@ end
 local time_to_live = tonumber(core.settings:get("item_entity_ttl")) or 900
 local gravity = tonumber(core.settings:get("movement_gravity")) or 9.81
 
+local function enable_physics(object, luaentity)
+	if luaentity.physical_state == false then
+		luaentity.physical_state = true
+		object:set_properties({
+			physical = true
+		})
+		object:set_velocity({x=0,y=0,z=0})
+		object:set_acceleration({x=0,y=-gravity,z=0})
+	end
+end
+local function disable_physics(object, luaentity)
+	if luaentity.physical_state == true then
+		luaentity.physical_state = false
+		object:set_properties({
+			physical = false
+		})
+		object:set_velocity({x=0,y=0,z=0})
+		object:set_acceleration({x=0,y=0,z=0})
+	end
+end
 
 core.register_entity(":__builtin:item", {
 	initial_properties = {
@@ -35,7 +55,12 @@ core.register_entity(":__builtin:item", {
 	itemstring = "",
 	moving_state = true,
 	slippery_state = false,
+	physical_state = true,
 	age = 0,
+	-- Helper vars for pushing item out of solid nodes
+	force_out = nil,
+	force_out_start = nil,
+	force_out_timer = 0,
 
 	set_item = function(self, item)
 		local stack = ItemStack(item or self.itemstring)
@@ -152,6 +177,100 @@ core.register_entity(":__builtin:item", {
 			return
 		end
 
+		local node_this = core.get_node_or_nil(pos)
+		local def_this
+		if node_this then
+			def_this = minetest.registered_nodes[node_this.name]
+		end
+		-- Push item out when stuck inside solid node
+		if def_this and def_this.walkable then
+			local shootdir
+			local cx = pos.x % 1
+			local cz = pos.z % 1
+			local order = {}
+
+			-- First prepare the order in which the 4 sides are to be checked.
+			-- 1st: closest
+			-- 2nd: other direction
+			-- 3rd and 4th: other axis
+			local cxcz = function(o, cw, one, zero)
+				if cw > 0 then
+					table.insert(o, { [one]=1, y=0, [zero]=0 })
+					table.insert(o, { [one]=-1, y=0, [zero]=0 })
+				else
+					table.insert(o, { [one]=-1, y=0, [zero]=0 })
+					table.insert(o, { [one]=1, y=0, [zero]=0 })
+				end
+				return o
+			end
+			if math.abs(cx) > math.abs(cz) then
+				order = cxcz(order, cx, "x", "z")
+				order = cxcz(order, cz, "z", "x")
+			else
+				order = cxcz(order, cz, "z", "x")
+				order = cxcz(order, cx, "x", "z")
+			end
+
+			-- Check which one of the 4 sides is free
+			for o=1, #order do
+				local nn = minetest.get_node(vector.add(pos, order[o])).name
+				local def_side = minetest.registered_nodes[nn]
+				if def_side and def_side.walkable == false and nn ~= "ignore" then
+					shootdir = order[o]
+					break
+				end
+			end
+			-- If none of the 4 sides is free, shoot upwards
+			if shootdir == nil then
+				shootdir = { x=0, y=1, z=0 }
+				local nn = minetest.get_node(vector.add(pos, shootdir)).name
+				if nn == "ignore" then
+					-- Do not push into ignore
+					return
+				end
+			end
+
+			-- Set new item moving speed accordingly
+			local newv = vector.multiply(shootdir, 3)
+			self.object:set_acceleration({x = 0, y = 0, z = 0})
+			self.object:set_velocity(newv)
+
+			disable_physics(self.object, self)
+
+			if shootdir.y == 0 then
+				self.force_out = newv
+				self.force_out_start = vector.floor(pos)
+				self.force_out_timer = 1
+			end
+			return
+		end
+
+		-- This code is run after the entity got a push from above “push away” code.
+		-- It is responsible for making sure the entity is entirely outside the solid node
+		-- (with its full collision box), not just its center.
+		if self.force_out_timer > 0 then
+			local cbox = self.object:get_properties().collisionbox
+			local ok = false
+			if self.force_out.x > 0 and (pos.x > (self.force_out_start.x + 0.5 + (cbox[4] - cbox[1])/2)) then ok = true
+			elseif self.force_out.x < 0 and (pos.x < (self.force_out_start.x + 0.5 - (cbox[4] - cbox[1])/2)) then ok = true
+			elseif self.force_out.z > 0 and (pos.z > (self.force_out_start.z + 0.5 + (cbox[6] - cbox[3])/2)) then ok = true
+			elseif self.force_out.z < 0 and (pos.z < (self.force_out_start.z + 0.5 - (cbox[6] - cbox[3])/2)) then ok = true end
+			-- Item was successfully forced out. No more pushing
+			if ok then
+				self.force_out_timer = -1
+				self.force_out = nil
+				enable_physics(self.object, self)
+			else
+				self.force_out_timer = self.force_out_timer - dtime
+			end
+			return
+		elseif self.force_out or not self.physical_state then
+			self.force_out = nil
+			enable_physics(self.object, self)
+			return
+		end
+
+		-- Slide on slippery nodes
 		local vel = self.object:get_velocity()
 		local def = node and core.registered_nodes[node.name]
 		local is_moving = (def and not def.walkable) or
--- a/builtin/game/item_entity.lua
+++ b/builtin/game/item_entity.lua
@@ -190,8 +190,8 @@ core.register_entity(":__builtin:item", {
 		-- Push item out when stuck inside solid node
 		if is_stuck then
 			local shootdir
-			local cx = pos.x % 1
-			local cz = pos.z % 1
+			local cx = (pos.x % 1) - 0.5
+			local cz = (pos.z % 1) - 0.5
 			local order = {}
 
 			-- First prepare the order in which the 4 sides are to be checked.
@@ -199,7 +199,7 @@ core.register_entity(":__builtin:item", {
 			-- 2nd: other direction
 			-- 3rd and 4th: other axis
 			local cxcz = function(o, cw, one, zero)
-				if cw > 0 then
+				if cw < 0 then
 					table.insert(o, { [one]=1, y=0, [zero]=0 })
 					table.insert(o, { [one]=-1, y=0, [zero]=0 })
 				else
@@ -208,7 +208,7 @@ core.register_entity(":__builtin:item", {
 				end
 				return o
 			end
-			if math.abs(cx) > math.abs(cz) then
+			if math.abs(cx) < math.abs(cz) then
 				order = cxcz(order, cx, "x", "z")
 				order = cxcz(order, cz, "z", "x")
 			else


paramat:
The 'force out' behaviour seems good as descibed.

builtin/game/item_entity.lua
```diff
@@ -17,6 +17,26 @@ end
...
+	if luaentity.physical_state == false then
+		luaentity.physical_state = true
+		object:set_properties({
+			physical = true
+		})
+		object:set_velocity({x=0,y=0,z=0})
+		object:set_acceleration({x=0,y=-gravity,z=0})
+	end
+end
+local function disable_physics(object, luaentity)
```
paramat: separate with empty line

paramat:
> 		-- This code is run after the entity got a push from above “push away” code.
		-- It is responsible for making sure the entity is entirely outside the solid node
		-- (with its full collision box), not just its center.

Perhaps moving the object until the centre is outside 'solid' nodes is good enough, and this behaviour can be left out for simplicity?
The object will be properly lit and visible. Buried objects aren't very common so they don't need to be perfectly positioned.

Wuzzy2:
> Perhaps moving the object until the centre is outside 'solid' nodes is good enough, and this behaviour can be left out for simplicity?

No. Based on my experience, if I do just that, the items will still be stuck in the wall. It really must be pushed out in full so it can fall. This is there for a reason. If the item's collisionbox is still overlapping the solid node, it's technically still stuck. This is horrifying:
https://satoshiupload.com/images/IL9lJ4ZJOW.png

And frankly, if removed, I would consider it buggy behaviour.

If you want to argue on performance, please give us hard, real data, thanks. :-)

benrob0329:
I approve of this for two reasons:

1. Dropped Item Will Free Themselves
2. Item Elevators Will Work Without Extra Mod Or Game Code

paramat:
> It really must be pushed out in full so it can fall.

Good point i hadn't considered.
I was arguing on simplicity, not performance.

tacotexmex:
> Item Elevators Will Work Without Extra Mod Or Game Code

I came here to also say this. Maybe a setting for maximum item travel?

benrob0329:
I feel like that'd overcomplicate it as then the code would need to track how far it's traveled.

tenplus1:
Would this make it any easier, does exact same thing but very simple code, plus I already use it in my own fork of builtin_item: https://notabug.org/TenPlus1/builtin_item/src/master/init.lua#L346

tacotexmex:
@tenplus1 Does your code only ever move the item one node?

tenplus1:
@tacotexmex It searches for the first empty node surrounding it and moves it to there.

builtin/game/item_entity.lua
```diff
@@ -152,6 +177,104 @@ core.register_entity(":__builtin:item", {
...
+			return
+		end
+
+		-- This code is run after the entity got a push from above “push away” code.
+		-- It is responsible for making sure the entity is entirely outside the solid node
+		-- (with its full collision box), not just its center.
+		if self.force_out_timer > 0 then
+			local cbox = self.object:get_properties().collisionbox
+			local ok = false
+			if self.force_out.x > 0 and (pos.x > (self.force_out_start.x + 0.5 + (cbox[4] - cbox[1])/2)) then ok = true
```
ClobberXD: From https://dev.minetest.net/Lua_code_style_guidelines#Misc:
>      You can put conditionals / loops with small conditions and bodies on one line. This is discouraged for all but the smallest ones though. 
>
> Good:
>
> ```lua
> local f, err = io.open(filename, "r")
> if not f then return err end
> 
> if     foo then return foo
> elseif bar then return bar
> elseif qux then return qux
> end
> ```
>
> Bad:
>
> ```lua
> if not f and use_error then error(err) elseif not f then return err end
> ```

Wuzzy2:
@tenplus1: Your algorithm fails when a node is surrounded by torches because they are not air. Items that are surrounded by solid nodes will continue to be stuck. I can't take your code seriously. Also, I would appreciate it if you would refrain to hi-jack other people's PRs in future, thanks. :P

At all: I will handle the coding style stuff in a few days, preferably all at once (much easier for me).

> I came here to also say this. Maybe a setting for maximum item travel?

No. I don't know why you believe this is needed.

tacotexmex:
>No. I don't know why you believe this is needed.

Because players may then send items through the ceiling into, for example, protected areas higher above. Or am I missing something here?

SmallJoker:
Wouldn't it be easier if you could look out for the closest "free" node using a function like `minetest.find_nodes_in_area`?
Moving items up until they reach the surface looks rather like the last act of despair before the item dies after 5 minutes. Looking out and moving into nearby "free" nodes (if available) is fine, but I'm worried about the performance if there's some TNT blasts involved. The builtin item should be kept lightweight in terms of execution time.

Wuzzy2:
> Moving items up until they reach the surface looks rather like the last act of despair before the item dies after 5 minutes.

It's based on the assumption the item will surface eventually. The world usually has a surface after all. So this is usually a pretty safe bet. Another benefit of this approach is that it only looks at a few nodes (direct neighbors). Another benefit of this is that if a large pile of sand falls on items, the items will quickly surface.

No algorithm will guarantee the item will appear somewhere sane before it dies. There will always be items that are so hopelessly lost that no player could ever find them. This is no argument against this particular algorithm.

If you're afraid of performance, please provide real world data.
I am also not sure why a `minetest.find_node_near` approach should have any better performance.  This one needs to check for quite a number of nodes, doesn't it? Without any real world data to compare the two algorithms, I ignore any claims about performance.

To be clear: Sorry, I am not going to make major changes to my algorithm in this PR. This would just mean I have to throw away everything and start over. It's better to just start a new PR instead.

Also: The calls for “simplicity” are starting to really annoy me. How is my code not simple??? It's basically just looking at neighbors. Also, this is a PR, containing already working code, ready to use. It's not a propoal, or experimental, etc.

@SmallJoker: Please clarify: Are you against the forcing-out items in general or just this particular implementation?

tacotexmex:
>Unless item entities can now explode or something like that I don't think we really need to worry about it.

I was thinking of games where items may be used in different ways than default. For example, items with light properties. But now that you mention it, grenades and projectile mods will probably see their mechanic changed. 

I'm still in favor of this as a general solution though, it's a good improvement.

builtin/game/item_entity.lua
```diff
@@ -189,16 +190,16 @@ core.register_entity(":__builtin:item", {
...
 
 			-- First prepare the order in which the 4 sides are to be checked.
 			-- 1st: closest
 			-- 2nd: other direction
 			-- 3rd and 4th: other axis
 			local cxcz = function(o, cw, one, zero)
-				if cw > 0 then
+				if cw < 0 then
 					table.insert(o, { [one]=1, y=0, [zero]=0 })
 					table.insert(o, { [one]=-1, y=0, [zero]=0 })
```
ClobberXD: Spacing needed around `=`. Should've pointed this out earlier, sorry. :)

builtin/game/item_entity.lua
```diff
@@ -189,16 +190,16 @@ core.register_entity(":__builtin:item", {
...
-			local cz = pos.z % 1
+			local cx = (pos.x % 1) - 0.5
+			local cz = (pos.z % 1) - 0.5
 			local order = {}
 
 			-- First prepare the order in which the 4 sides are to be checked.
 			-- 1st: closest
 			-- 2nd: other direction
 			-- 3rd and 4th: other axis
 			local cxcz = function(o, cw, one, zero)
```
ClobberXD: Our code-style guidelines advise against using this form of function definition, unless defining a function within a table:

> Write function definitions of the form `function foo()` instead of the lambda form `foo = function()`, except when inserting functions in tables inline, where only the second form will work.

SmallJoker:
> The world usually has a surface after all. [...] Another benefit of this approach is that it only looks at a few nodes (direct neighbors).

As soon the item can go upwards (imagine -1000m mine), they will eventually merge with the upper unloaded mapblock. This results in an unknown expiry time since those mapblocks might be only generated, but never active. Not much of a concern, they'll just end up as unused garbage like the mapblock itself. This kind of also limits the time wasted in item movements.

> If you're afraid of performance, please provide real world data.

<details>
<summary>Timing tests</summary>

```
is_stuck:  115 us
is_stuck:  93 us
is_stuck:  126 us
is_stuck:  95 us
is_stuck:  33 us
is_stuck:  48 us
is_stuck:  24 us
is_stuck:  24 us
is_stuck:  24 us
is_stuck:  23 us
is_stuck:  26 us
is_stuck:  24 us
is_stuck:  25 us
is_stuck:  24 us
is_stuck:  24 us
is_stuck:  19 us
is_stuck:  24 us
is_stuck:  24 us
is_stuck:  24 us
is_stuck:  27 us
is_stuck:  26 us
is_stuck:  24 us
is_stuck:  25 us
is_stuck:  31 us
is_stuck:  25 us
is_stuck:  139 us
is_stuck:  29 us
is_stuck:  31 us
is_stuck:  25 us
is_stuck:  23 us
is_stuck:  27 us
is_stuck:  24 us
is_stuck:  25 us
is_stuck:  24 us
is_stuck:  21 us
is_stuck:  22 us
is_stuck:  20 us
force out:   0 us
force out:   0 us
force out:   0 us
force out:   0 us
slippery   7 us
slippery   6 us
slippery   6 us
slippery   6 us
slippery   6 us
slippery   6 us
slippery   6 us
slippery   7 us
slippery   7 us
slippery   6 us
slippery   6 us
slippery   5 us
slippery   6 us
slippery   5 us
slippery   8 us
slippery  64 us
slippery  88 us
slippery   6 us
slippery   4 us
slippery   5 us
```
</details>

Measured with two variables calling to `minetest.get_us_time()`, followed by `print`. Slippery (21 code lines) is for comparison.
The code is surprising fast, especially the node lookups. There's no sane way to make it even faster.

> I am also not sure why a minetest.find_node_near approach should have any better performance.

In this context I literally meant *easier*, not *faster*. The code for searching free neighbours is straight forward, with least possible node checks. The function I suggested would probably take a bit longer to execute, but is a more generic approach that could be used in other mods to get entities out of nodes.

> Also: The calls for “simplicity” are starting to really annoy me. 

Question is whether to re-invent the wheel or to make use of the existing functions. In this case the former was chosen to only check very few nodes - totally fine.

> Are you against the forcing-out items in general or just this particular implementation?

Stuck items are annoying, so I'm in favour to this approach. Personally I'd only check the neighbour nodes each few seconds, without moving them upwards. I am neither against the concept nor the implementation.

tenplus1:
@Wuzzy2 sorry if it seemed like I was hijacking your pull, I simply wanted to point out an easy code solution that we've been using for over a year on server.

builtin/game/item_entity.lua
paramat: I'm surprised that is allowed, i'd prefer it wasn't done as it is much less readable.
I think the code style guidelines should be changed.

So, please don't do this here.

builtin/game/item_entity.lua
paramat: Also, no spaces just inside curly brackets.

builtin/game/item_entity.lua
```diff
@@ -152,6 +178,111 @@ core.register_entity(":__builtin:item", {
...
+
+			if shootdir.y == 0 then
+				self.force_out = newv
+				self.force_out_start = vector.floor(pos)
+				self.force_out_timer = 1
+			end
+			return
+		end
+
+		-- This code is run after the entity got a push from above “push away” code.
```
paramat: I know this is ideal behaviour to force them out enough so that they drop, but this code is inside 'on step' and needs to be as a simple as possible.
Considering how rarely item entities get stuck inside nodes i don't think this level of perfectionism is needed. The entity will be exposed and visible, players can climb up to pick it up if it is stuck in a wall and won't drop (which again won't be often).

Please remove this to simplify.

builtin/game/item_entity.lua
```diff
@@ -152,6 +178,111 @@ core.register_entity(":__builtin:item", {
...
+		end
+
+		-- Push item out when stuck inside solid node
+		if is_stuck then
+			local shootdir
+			local cx = (pos.x % 1) - 0.5
+			local cz = (pos.z % 1) - 0.5
+			local order = {}
+
+			-- First prepare the order in which the 4 sides are to be checked.
```
paramat: This code block that prepares an order seems overkill complex inside 'on step', why not just search the surrounding 4 or 8 nodes? and why not search diagonals?

paramat:
I like the basic concept but there are 2 sections of unnecessarily complex and perfectionist code, see line comments.
I'm -1 unless this is simplified.

builtin/game/item_entity.lua
ClobberXD: I agree. The C++ code style guidelines is less lenient in this regard: https://dev.minetest.net/Code_style_guidelines#if_statements

DS-Minetest:
Related: #5482

builtin/game/item_entity.lua
DS-Minetest: As far as I can see, this code is not run on every step.

Without this checks there might still be something like this:

(Taken from #5409.)

The perfectionism _is_ needed.

DS-Minetest:
I wonder if it was better to use `move_to` for small movements instead of setting the velocity. :thinking:

builtin/game/item_entity.lua
DS-Minetest: A fixed order might look pretty bad.
Imho, this code block doesn't look that complicated. Much more detailed calculations would be possible.

builtin/game/item_entity.lua
```diff
@@ -152,6 +178,111 @@ core.register_entity(":__builtin:item", {
...
+					table.insert(o, { [one]=-1, y=0, [zero]=0 })
+				else
+					table.insert(o, { [one]=-1, y=0, [zero]=0 })
+					table.insert(o, { [one]=1, y=0, [zero]=0 })
+				end
+				return o
+			end
+			if math.abs(cx) < math.abs(cz) then
+				order = cxcz(order, cx, "x", "z")
+				order = cxcz(order, cz, "z", "x")
```
DS-Minetest: Isn't the closest face in the z direction in this case? `cz` is farther away from 0.

builtin/game/item_entity.lua
```diff
@@ -152,6 +178,111 @@ core.register_entity(":__builtin:item", {
...
+					table.insert(o, { [one]=1, y=0, [zero]=0 })
+					table.insert(o, { [one]=-1, y=0, [zero]=0 })
+				else
+					table.insert(o, { [one]=-1, y=0, [zero]=0 })
+					table.insert(o, { [one]=1, y=0, [zero]=0 })
+				end
+				return o
+			end
+			if math.abs(cx) < math.abs(cz) then
+				order = cxcz(order, cx, "x", "z")
```
DS-Minetest: You don't have to re-set `order`.

builtin/game/item_entity.lua
```diff
@@ -152,6 +178,111 @@ core.register_entity(":__builtin:item", {
...
+			local cxcz = function(o, cw, one, zero)
+				if cw < 0 then
+					table.insert(o, { [one]=1, y=0, [zero]=0 })
+					table.insert(o, { [one]=-1, y=0, [zero]=0 })
+				else
+					table.insert(o, { [one]=-1, y=0, [zero]=0 })
+					table.insert(o, { [one]=1, y=0, [zero]=0 })
+				end
+				return o
+			end
```
DS-Minetest: You could set the other axis to 2 and 3 and the other direction to 4:
```lua
local function cxcz(o, sign, axis, also_other) -- What does that cxcz, cw, cx and co. stand for btw?
	local v = vector.new()
	v[axis] = sign
	table.insert(o, dir)
	if also_other then
		cxcz(o, -sign, axis, false)
	end
end

local order = {}
if math.abs(cx) > math.abs(cz) then
	cxcz(order, math.sign(cx), "x", false)
	cxcz(order, math.sign(cz), "z", true)
	cxcz(order, math.sign(cx), "x", false)
else
	-- ...
end
```

builtin/game/item_entity.lua
```diff
@@ -152,6 +178,111 @@ core.register_entity(":__builtin:item", {
...
+			-- Check which one of the 4 sides is free
+			for o=1, #order do
+				local nn = minetest.get_node(vector.add(pos, order[o])).name
+				local def_side = minetest.registered_nodes[nn]
+				if def_side and def_side.walkable == false and nn ~= "ignore" then
+					shootdir = order[o]
+					break
+				end
+			end
+			-- If none of the 4 sides is free, shoot upwards
```
DS-Minetest: What if the item obj could fall out of the node in `-y` dir?
You could always add `(0,-1,0)` to `order`.

builtin/game/item_entity.lua
```diff
@@ -152,6 +178,111 @@ core.register_entity(":__builtin:item", {
...
+
+		-- This code is run after the entity got a push from above “push away” code.
+		-- It is responsible for making sure the entity is entirely outside the solid node
+		-- (with its full collision box), not just its center.
+		if self.force_out_timer > 0 then
+			local c = self.object:get_properties().collisionbox
+			local s = self.force_out_start
+			local f = self.force_out
+			local ok = false
+			if f.x > 0 and (pos.x > (s.x + 0.5 + (c[4] - c[1])/2)) then
```
DS-Minetest: This is overcomplicated and only works correctly for symmetric collisionboxes. Why do you do this?
```suggestion
			if f.x > 0 and pos.x > s.x + c[1] or
					f.x < 0 and pos.x < s.x + c[4] or
					... then
```

builtin/game/item_entity.lua
paramat: What's bad about a fixed order? Maybe i've missed something ...

This is also about reducing code maintenance in future. This code seems like unnecessary headache.

builtin/game/item_entity.lua
paramat: Seems good enough for this rare occurrence, the object was inside the node after all, this is consistent with that, the object is in the process of being squeezed out of the node and appears on the surface for collection.

builtin/game/item_entity.lua
DS-Minetest: Well, with a fixed order, placing nodes into dropped items could be used as compass.

Such things are bad for player experience and the players will think that minetest code is simply bad and not well pollished.

builtin/game/item_entity.lua
paramat: LIke i wrote above, this is ridiculously overkill for just creating an order for looking at the 4 neighbours. Please don't waste time on this.

builtin/game/item_entity.lua
paramat: Clouds, node textures and probably several other things can be used as compasses too. And ... very few players think of MT as bad due to that.

When objects are buried, players are usually not aware of their position, so they don't need to choose the closest 'empty' node.

Wuzzy2:
I give up.

tacotexmex:
This is very unfortunate. With all the sketchyness going on in the engine, _this_ can’t pass as quality code??

sfan5:
While comments like "Please don't waste time on this" sound unnecessarily dismissive, the concern of simplicity is warranted.
Forcing out items is a good feature, but iterating in a fixed order and assuming square collisionboxes would suffice IMO.

PR #8864, opened 2019-08-26
Document luaent pitch and roll in world format

DS-Minetest:
See https://github.com/minetest/minetest/pull/7395#issuecomment-524969730.

## To do

## How to test

Read it.

### DIFF
--- a/doc/world_format.txt
+++ b/doc/world_format.txt
@@ -510,7 +510,7 @@ Object types:
     u8[len] itemstring
 
 7: LuaEntity:
-  u8 version (=1)
+  u8 compatibility_byte (always 1)
   version 1:
     u16 len
     u8[len] entity name
--- a/doc/world_format.txt
+++ b/doc/world_format.txt
@@ -521,6 +521,10 @@ Object types:
     s32 velocity.y * 10000
     s32 velocity.z * 10000
     s32 yaw * 1000
+    if PROTOCOL_VERSION >= 37:
+      u8 version2 (=1)
+      s32 pitch * 1000
+      s32 roll * 1000
 
 Itemstring format
 ------------------
--- a/doc/world_format.txt
+++ b/doc/world_format.txt
@@ -510,7 +510,7 @@ Object types:
     u8[len] itemstring
 
 7: LuaEntity:
-  u8 version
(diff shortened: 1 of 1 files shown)


doc/world_format.txt
```diff
@@ -521,6 +521,10 @@ Object types:
     s32 velocity.y * 10000
     s32 velocity.z * 10000
     s32 yaw * 1000
+    if PROTOCOL_VERSION >= 37:
+      u8 version2 (=1)
```
sfan5: please also document that the version above is always kept at 1

DS-Minetest: Done.

SmallJoker: That's not *exactly* what he meant. `version` is and must always stay `1` for backwards compatibility.
Suggestion: `u8 compatibility_byte (always 1)` even though the names don't match.

DS-Minetest: But does `version` have to stay `1` in the world file? Saving backwards-compatible makes sense if you are sending the data to a client but not if you are saving to file.
I will just take your suggestion as I'm surely wrong.

SmallJoker: Network:
https://github.com/minetest/minetest/blob/86d7f84b899a507e979f1845f2057cce6f84e743/src/content_sao.cpp#L558
Map:
https://github.com/minetest/minetest/blob/86d7f84b899a507e979f1845f2057cce6f84e743/src/content_sao.cpp#L613-L620

`version(1)` was implemented in a forwards incompatible manner, resulting in unknown entities when loading the world from an old client. 5.x has to keep the map compatibility, hence a new version byte was needed to work around this issue.

doc/world_format.txt
```diff
@@ -510,7 +510,7 @@ Object types:
     u8[len] itemstring
 
 7: LuaEntity:
-  u8 version
+  u8 compatibility_byte (always 1)
   version 1:
```
sfan5: this line does not make sense now

DS-Minetest: Done.

DS-Minetest:
Related: #8868 
The changes of this PR might be wrong.

PR #8871, opened 2019-08-29
Move Quicktune code to util/

ClobberXD:
I very recently discovered what Quicktune actually is, but then I realised that the files were in the wrong place. `src/util/` is probably a more suitable location for these files, IMO.

Tested, code compiles without errors. This PR is ready for review.


DS-Minetest:
I think, you also have to change this for android:
https://github.com/minetest/minetest/blob/4291961e09db7851fbc940eac58a1b919668d935/build/android/jni/Android.mk#L244

And travis wants you to fix the code style.

ClobberXD:
> I think, you also have to change this for android

Thanks for letting me know. Done.

> And travis wants you to fix the code style.

The files were added to the clang format whitelist, but the whitelist entries became invalid since the files have been moved to a different directory. Fixed.

PR #8874, opened 2019-08-30
Send multiple occlusion rays from the camera viewpane.

lhofhansl:
Here's another attempt. Use the camera's view up vector (client) and the known camera rotation (server) to create spanning vectors on the camera pane (labeled x and y in the code). With that we can send multiple rays from the camera pane.
This takes care of the fact the view pane is not a point. Things might be hidden from the camera point and still be visible on the screen.

This includes most of the code from #8852 

(as before this cannot be perfect, there could be blocks that are visible from certain portions of the view plane only)

Please have a look and test. The extra effort (5x the rays) might not be worth it.

### DIFF
--- a/src/client/clientmap.cpp
+++ b/src/client/clientmap.cpp
@@ -125,6 +125,8 @@ void ClientMap::updateDrawList()
 	v3f camera_position = m_camera_position;
 	v3f camera_direction = m_camera_direction;
 	f32 camera_fov = m_camera_fov;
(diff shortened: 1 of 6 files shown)


lhofhansl:
Closing this one too. It was an experiment. I cannot convince myself that the extra work the CPU is doing is worth is. As stated somewhere else... Unless we're fundamentally changing how we do occlusion culling what we have is "good enough".

PR #8889, opened 2019-09-02
Non uniform fog with perlin-type noise.

lhofhansl:
Something I was playing with.
Just uses a free GLSL perlin noise function and applies it to the fog density.
Self adjusts for the fog range (so the distinct changes of fog density are nice at various view ranges).
Also has a time component so that fog slowly moves (along the z axis, like clouds and waves).

~~Experimental...~~ Have a look. It adds quite a lot of atmosphere, all handled in the shader.

If we like it, I can put it behind an option like the other shader options.

TODO:
~~1. Optional via shader option.~~ Done
~~2. Match cloud movement (which can be configured)...?~~ Decided not to do.
~~3. Special handling under water and in caves...?~~ Done.

### DIFF
--- a/builtin/mainmenu/tab_settings.lua
+++ b/builtin/mainmenu/tab_settings.lua
@@ -174,7 +174,7 @@ end
 
 local function formspec(tabview, name, tabdata)
 	local tab_string =
-		"box[0,0;3.75,4.5;#999999]" ..
+		"box[0,0;3.75,4.7;#999999]" ..
 		"checkbox[0.25,0;cb_smooth_lighting;" .. fgettext("Smooth Lighting") .. ";"
 				.. dump(core.settings:get_bool("smooth_lighting")) .. "]" ..
 		"checkbox[0.25,0.5;cb_particles;" .. fgettext("Particles") .. ";"
@@ -189,7 +189,7 @@ local function formspec(tabview, name, tabdata)
 				.. getSettingIndex.NodeHighlighting() .. "]" ..
 		"dropdown[0.25,3.6;3.5;dd_leaves_style;" .. dd_options.leaves[1] .. ";"
 				.. getSettingIndex.Leaves() .. "]" ..
-		"box[4,0;3.75,4.5;#999999]" ..
+		"box[4,0;3.75,4.7;#999999]" ..
 		"label[4.25,0.1;" .. fgettext("Texturing:") .. "]" ..
 		"dropdown[4.25,0.55;3.5;dd_filters;" .. dd_options.filters[1] .. ";"
 				.. getSettingIndex.Filter() .. "]" ..
@@ -201,7 +201,7 @@ local function formspec(tabview, name, tabdata)
 		"label[4.25,3.45;" .. fgettext("Screen:") .. "]" ..
 		"checkbox[4.25,3.6;cb_autosave_screensize;" .. fgettext("Autosave Screen Size") .. ";"
 				.. dump(core.settings:get_bool("autosave_screensize")) .. "]" ..
-		"box[8,0;3.75,4.5;#999999]"
+		"box[8,0;3.75,4.7;#999999]"
 
 	local video_driver = core.settings:get("video_driver")
 	local shaders_supported = video_driver == "opengl"
@@ -257,7 +257,9 @@ local function formspec(tabview, name, tabdata)
 			"checkbox[8.25,3;cb_waving_leaves;" .. fgettext("Waving Leaves") .. ";"
 					.. dump(core.settings:get_bool("enable_waving_leaves")) .. "]" ..
 			"checkbox[8.25,3.5;cb_waving_plants;" .. fgettext("Waving Plants") .. ";"
-					.. dump(core.settings:get_bool("enable_waving_plants")) .. "]"
+					.. dump(core.settings:get_bool("enable_waving_plants")) .. "]" ..
+			"checkbox[8.25,4.0;cb_fog_noise;" .. fgettext("Non-uniform Fog") .. ";"
+					.. dump(core.settings:get_bool("enable_fog_noise")) .. "]"
 	else
 		tab_string = tab_string ..
 			"label[8.38,0.7;" .. core.colorize("#888888",
@@ -273,7 +275,9 @@ local function formspec(tabview, name, tabdata)
 			"label[8.38,3.2;" .. core.colorize("#888888",
 					fgettext("Waving Leaves")) .. "]" ..
 			"label[8.38,3.7;" .. core.colorize("#888888",
-					fgettext("Waving Plants")) .. "]"
+					fgettext("Waving Plants")) .. "]" ..
+			"label[8.38,4.2;" .. core.colorize("#888888",
+					fgettext("Non-uniform Fog")) .. "]"
 	end
 
 	return tab_string
@@ -342,6 +346,10 @@ local function handle_settings_buttons(this, fields, tabname, tabdata)
 	end
 	if fields["cb_waving_water"] then
 		core.settings:set("enable_waving_water", fields["cb_waving_water"])
+		return true
+	end
+	if fields["cb_fog_noise"] then
+		core.settings:set("enable_fog_noise", fields["cb_fog_noise"])
 		return true
 	end
 	if fields["cb_waving_leaves"] then
--- a/client/shaders/nodes_shader/opengl_fragment.glsl
+++ b/client/shaders/nodes_shader/opengl_fragment.glsl
@@ -23,13 +23,48 @@ varying vec3 eyeVec;
 varying vec3 tsEyeVec;
 varying vec3 lightVec;
 varying vec3 tsLightVec;
+varying float nightRatio;
 
 bool normalTexturePresent = false;
 
 const float e = 2.718281828459;
 const float BS = 10.0;
 const float fogStart = FOG_START;
-const float fogShadingParameter = 1 / ( 1 - fogStart);
+float fogShadingParameter = 1 / ( 1 - fogStart);
+
+#ifdef ENABLE_FOG_NOISE
+//
+// Simple, fast noise function.
+// See: https://gist.github.com/patriciogonzalezvivo/670c22f3966e662d2f83
+//
+vec4 perm(vec4 x)
+{
+	return mod(((x * 34.0) + 1.0) * x, 289.0);
+}
+
+float snoise(vec3 p)
+{
+	vec3 a = floor(p);
+	vec3 d = p - a;
+	d = d * d * (3.0 - 2.0 * d);
+
+	vec4 b = a.xxyy + vec4(0.0, 1.0, 0.0, 1.0);
+	vec4 k1 = perm(b.xyxy);
+	vec4 k2 = perm(k1.xyxy + b.zzww);
+
+	vec4 c = k2 + a.zzzz;
+	vec4 k3 = perm(c);
+	vec4 k4 = perm(c + 1.0);
+
+	vec4 o1 = fract(k3 * (1.0 / 41.0));
+	vec4 o2 = fract(k4 * (1.0 / 41.0));
+
+	vec4 o3 = o2 * d.z + o1 * (1.0 - d.z);
+	vec2 o4 = o3.yw * d.x + o3.xz * (1.0 - d.x);
+
+	return o4.y * d.y + o4.x * (1.0 - d.y);
+}
+#endif
 
 #ifdef ENABLE_TONE_MAPPING
 
@@ -208,6 +243,20 @@ void main(void)
 	
 #ifdef ENABLE_TONE_MAPPING
 	col = applyToneMapping(col);
+#endif
+
+#ifdef ENABLE_FOG_NOISE
+	// Only do this with natural light (not in caves or under water.)
+	if (nightRatio < 0.99) {
+		// Add some noise to the haze/fog.
+		// Adjust the period to the fogDistance.
+		float noise = snoise((worldPosition + cameraOffset) /
+		                     (fogDistance / FOG_NOISE_FREQUENCY) +
+				     vec3(0,0,animationTimer * 10.0)) * FOG_NOISE_STRENGTH;
+
+		// 0.4 tested to roughly the same perceived fog distance.
+		fogShadingParameter += (noise - 0.4 * FOG_NOISE_STRENGTH) * (1 - nightRatio);
+	}
 #endif
 
 	// Due to a bug in some (older ?) graphics stacks (possibly in the glsl compiler ?),
--- a/src/client/shader.cpp
+++ b/src/client/shader.cpp
@@ -686,6 +686,16 @@ ShaderInfo generate_shader(const std::string &name, u8 material_type, u8 drawtyp
 	if (g_settings->getBool("enable_bumpmapping"))
 		shaders_header += "#define ENABLE_BUMPMAPPING\n";
 
+	if (g_settings->getBool("enable_fog_noise")) {
+		shaders_header += "#define ENABLE_FOG_NOISE\n";
+		shaders_header += "#define FOG_NOISE_FREQUENCY ";
+		shaders_header += itos(g_settings->getU16("fog_noise_frequency") / 10);
+		shaders_header += "\n";
+		shaders_header += "#define FOG_NOISE_STRENGTH ";
+		shaders_header += ftos(g_settings->getFloat("fog_noise_strength") * 4.0f);
+		shaders_header += "\n";
+	}
+
 	if (g_settings->getBool("enable_parallax_occlusion")){
 		int mode = g_settings->getFloat("parallax_occlusion_mode");
 		float scale = g_settings->getFloat("parallax_occlusion_scale");
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -589,6 +589,20 @@ enable_waving_leaves (Waving leaves) bool false
 #    Requires shaders to be enabled.
 enable_waving_plants (Waving plants) bool false
 
+[***Non-uniform Fog]
+#    Set to true to add a bit of noise to the fog.
+#    Requires shaders to be enabled.
+enable_fog_noise (Non-uniform Fog) bool false
+
+#    Frequency of the noise function.
+#    Larger values produce a finer and more
+#    visible pattern.
+fog_noise_frequency (Fog noise frequency) int 60 1 100
+
+#    The strenght of the fog effect
+#    Larger numbers produce a stronger effect.
+fog_noise_strength (Fog noise Strengh) float 0.25 0.0 1.0
+
 [**Advanced]
 
 #    Arm inertia, gives a more realistic movement of
--- a/client/shaders/nodes_shader/opengl_vertex.glsl
+++ b/client/shaders/nodes_shader/opengl_vertex.glsl
@@ -22,6 +22,7 @@ varying vec3 lightVec;
 varying vec3 tsEyeVec;
 varying vec3 tsLightVec;
 varying float area_enable_parallax;
+varying float nightRatio;
 
 // Color of the light emitted by the light sources.
 const vec3 artificialLight = vec3(1.04, 1.04, 1.04);
@@ -137,7 +138,7 @@ float disp_z;
 	// The pre-baked colors are halved to prevent overflow.
 	vec4 color;
 	// The alpha gives the ratio of sunlight in the incoming light.
-	float nightRatio = 1 - gl_Color.a;
+	nightRatio = 1 - gl_Color.a;
 	color.rgb = gl_Color.rgb * (gl_Color.a * dayLight.rgb + 
 		nightRatio * artificialLight.rgb) * 2;
 	color.a = 1;
--- a/src/defaultsettings.cpp
+++ b/src/defaultsettings.cpp
@@ -261,6 +261,9 @@ void set_default_settings(Settings *settings)
 	settings->setDefault("water_wave_speed", "5.0");
 	settings->setDefault("enable_waving_leaves", "false");
 	settings->setDefault("enable_waving_plants", "false");
+	settings->setDefault("enable_fog_noise", "false");
+	settings->setDefault("fog_noise_frequency", "60");
+	settings->setDefault("fog_noise_strength", "0.25");
 
 
 	// Input


lhofhansl:
Should probably be not applied underground and underwater.

client/shaders/nodes_shader/opengl_fragment.glsl
```diff
@@ -210,6 +288,10 @@ void main(void)
 	// As additions usually come for free following a multiplication, the new formula
 	// should be more efficient as well.
 	// Note: clarity = (1 - fogginess)
+	float noise = cnoise(vec3(worldPosition.x * 5 / fogDistance + animationTimer * 20.0,
+							  worldPosition.y * 5 / fogDistance,
+							  worldPosition.z * 5 / fogDistance));
```
DS-Minetest: Tab, then space alignment is bad for different tab sizes. I suggest to just indent once.

DS-Minetest:
Tested.
Dungeons feel so ghostly:
screenshot_20190902_132029
The landscape looks a bit patchy:
screenshot_20190902_132207
That one can only see the 3D fog where it traverses node surfaces makes this ugly.
And it's weird that the fog is moving perpendicular to the clouds.

lhofhansl:
Thanks for taking a look. That's exactly the feedback I'm looking for.
The patchiness can be changed. I will say that I like it bit patchy, that way it's visible.
Also I will make sure perlin fog is not used below the water line (i.e. under water and in dungeons).
I'll look into the cloud movement.

I should also say that this is a simple, lightweight thing. The fog is not modeled as a thing in the world, there's no ray-tracing to determine the fog level at each point, etc. This just changes the density at every point, which is as close to the "real thing" as one can with such a simple thing.

Calinou:
I like the idea, but I think the effect needs to be more subtle as @DS-Minetest said.

lhofhansl:
Do you guys want play with this and see what "patchiness" looks nice?
```
float noise = cnoise(vec3(worldPosition.x * 5 / fogDistance + animationTimer * 20.0,
				  worldPosition.y * 5 / fogDistance,
				  worldPosition.z * 5 / fogDistance));
```
The 5 is the "number of patches", if you decrease it you get fewer and bigger patches.
You can also change the fog_start in the settings. It defaults to 0.4, if you decrease it it is smoothed out a bit more.

Perhaps try fog_start = 0.2 in the settings, and change the 5 to 3 and play around with it.

I'll also do some tweaking.

(It turns out that on the client I do not know the water line. The server would have to send it. Perhaps it's easier to reduce the effect when the light is low.)

Lastly... I do not feel emotionally attached to this :) Just an experiment to see what's possible.

lhofhansl:
Made the patchiness constant.

With PR, viewrange 240:
screenshot_20190902_095408

Without PR, viewrange 240:
screenshot_20190902_095441

With PR, viewrange 100:
screenshot_20190902_095517

Without PR, viewrange 100:
screenshot_20190902_095446

Calinou:
The patchiness looks quite disturbing with high view ranges. It can make taking screenshots of large buildings much harder if you don't want fog to be in the view.

lhofhansl:
> To solve this, the patchiness should only be present in the area that would be fogged if the shader is disabled.

Not sure that's possible. The whole point of this is to have uneven fog. :)
I could go back and adjust the patchiness based on the fog range (as the original change had), though I'd think that'd be weird in other scenarios. I think in the end this should just be configurable as the other shader stuff.

(BTW. now that that I am thinking about this, the perlin noise could also be applied to waves and plant movement (removing that weird thing we have now). But that's for another PR)

paramat:
This is good that you are experimenting, but i don't think this makes sense, as this isn't actually 'fog', even through it may be called 'fog' in the code. It is an atmospheric haze due to view range distance, which is therefore uniform, and should be uniform to fade out the visible world just before the cutoff of world rendering. The haze gives the landscape a sense of distance, so varying it with noise leads to a distorted sense of distance.

I think any 'fog' feature has to be completely independent of view range haziness. 'Fog' is ideally implemented as low cloud, so should have cuboid volumes and be rendered the same way. The suitable implementation is to have a 2nd level of clouds, and controlled by the cloud API.

Any fog feature would be a fundamental feature so cannot be done by shaders, because many do not or cannot use shaders. For example, it obviously would not be right for clouds to depend on shaders.
Sorry, i don't like being critical like this, and i really appreciate that you are experienting.

lhofhansl:
I closed this PR... This will just lead to endless discussion and then not go anywhere. :)

Personally I think this *adds* depth to an otherwise perfectly sterile haze, the haze will still line up perfectly with the horizon and the fog still increases with distance as before (both because the density is adjusted, not the range), and haze is not uniform in nature either.

All good.

lhofhansl:
For posterity... The fog in motion.

vokoscreen-2019-09-02_14-52-14

lhofhansl:
Ahh shucks... Lemme leave it open so that others see it and comment. I do not plan to do anything else on this, though.

Regardless of this issue, I think MT is woefully falling behind the state of the art. Any change to the graphics is greeted with "Hmm. Nah, won't work on old machines", or "Hmm. Nah, we first need to fix XYZ.", or "Hmm. Nah, it looks different than it looked before.", "Hmm. But what if XYZ?!", etc, etc.

As stated I'm not specifically attached to this one, but generally to the ability and willingness to improve MT visually without perfectionism.

rubenwardy:
I like this, but I think it puts fog too close. Maybe a less extreme effect would be better

I would like MT to improve visually, but it would be nice to fix the broken use of OpenGL first

DevyHeavy:
I'm curious, is there any way there this could be built into Minetest's LUA api so that modders can choose how it looks? Keep the normal uniform fog that obscures the unloaded world but also let modders have this as an option.

I think it looks amazing. One thing that bothers me about Minetest is not being able to easily modify the 'atmosphere' of the game. Right now, all modders really have is clouds color/density and sky color (you can't even change the sun and the moon yet).

lhofhansl:
I think that would be great. Interacting with something as deep "down" as shaders in the engine from Lua is hard, though.

(Sometime like) This can easily be enabled or disabled in the settings, though. Just like waves, wavy plants, bump mapping, etc.
Do you think this is a modding thing, or something each player/client should enable/disable?

TumeniNodes:
I love this! Is there a performance hit?
The gif image would give a better visual ref if waving leaves was off, it throws it off due to the effect. Makes it look like the fog is throwing a weird "x-ray" glitch (which it is not but some may think it is when viewing)

*Off topic but, The fire in the background has me greatly intrigued, mind sharing somewhere (even via email?)

lhofhansl:
Had to check on the fire. (This is a random test world with all kinds of random old and new mods in it.)
I think what you see is a creature from the doom mod (I think) that leaves behind a particle trail:
fire

Regarding this. I got the math slightly wrong. Perlin varies from -1 to +1, so it can "move" the haze too close when it's negative. I'll play around with this as I find time.

lhofhansl:
I'm having fun with this... Here's what it would like if this was strictly less fog.
With PR (view range 240, fogstart 0.4):
screenshot_20190903_123816

Without PR:
screenshot_20190903_123932

View range 100, with PR:
screenshot_20190903_123847

And without
screenshot_20190903_123914

It's much more subtle this way.

lhofhansl:
And here with MT's fogStart parameter set to 0 (view range 240)
screenshot_20190903_124522
screenshot_20190903_124557

lhofhansl:
I found a simpler (and much faster) noise function. At higher resolutions I could see a perf defect before. I also made the effect much more subtle. Have a look.

With PR, view range 100:
screenshot_20190905_122245

Without PR:
screenshot_20190905_122356

With PR, view range 240:
screenshot_20190905_122300

Without PR:
screenshot_20190905_122342

I think if I add this as a shader option - just like waving plants - this should be good.
It's not changing the game play or anything, it's not adding new clouds or fog, it just enhances the mood a bit - just like waving plants or smooth lighting.

@paramat I agree the old effect looked too much like (fake) clouds. The effect now just makes the haze a little less regular, and hence it doesn't look so sterile anymore.

@rubenwardy What needs to be fixed in our use of OpenGL? Perhaps I'll find some time to look at that, too.

All, rereading my comment above it sounded more whiny than intended. :)

lhofhansl:
And made it configurable. Default off.

lhofhansl:
Added options to configure the effect. Also avoid the effect when it's dark (deep under water and in caves).

I do not plan to do anything more on this (until I get any more feedback).
My son and I will have this patch applied and default on :)

v-rob:
Honestly, this looks very good.  I would love some more GIFs, since I can only compile on a virtual machine and therefore get ~10 fps.  This is really is a much better atmospheric feeling than the current fog, which feels more like just a view range limiter.

DS-Minetest:
Tested again (only with default settings, I'm lazy).
With default settings (and non uniform fog on, of course) the effect is quite subtle, but looks somewhat nice.

There's a problem though.
Sometimes, also in older versions of this PR, I had the feeling that the moving fog jumps back.
Looking into code, I can see that the animation timer is reset every 100 seconds:
https://github.com/minetest/minetest/blob/0b4f424f414380b7a46270e1389e8aa0524d8fba/src/client/game.cpp#L496-L499

lhofhansl:
@DS-Minetest I noticed that too and came to the same conclusion. (The waves also jump sometimes.)
I haven't thought of a way around this, yet.

benrob0329:
Reverse the direction of the animation rather than resetting it (eased of course)? It wouldn't look great for waves (probably, if it's eased out a lot it might be fine) but it's definitely better than jumping back to a hard reset.

lhofhansl:
Actually the % operation is not even needed. 
```
-               u32 animation_timer = porting::getTimeMs() % 100000;
+               u32 animation_timer = porting::getTimeMs();
		float animation_timer_f = (float)animation_timer / 100000.f;
```
It's still a float, we just need a number changing with time. Continues to work with waves (just sine waves) and perlin noise as both have periodicity build in anyway.

We should just make this change.

paramat:
> Do you think this is a modding thing, or something each player/client should enable/disable?

I think it's best as it is, in shaders.
> It's not changing the game play or anything, it's not adding new clouds or fog, it just enhances the mood a bit - just like waving plants or smooth lighting.

Yes i agree.
> I agree the old effect looked too much like (fake) clouds. The effect now just makes the haze a little less regular

It does look much better now, being subtle and larger scale.

Since this is not, and cannot be considered biome fog (it can't be for many reasons i can explain if necessary) i would prefer it is not called 'fog' in the settings, maybe 'haze' or 'mist' or something? To not confuse with biome-fog that could come later, or can be done already using ground-level clouds, as in the Hidden Worlds game.

Does this still drift in a certain direction, or does it just vary with time in a non-directional way? The cloud API can alter the wind direction so a slow drift in a certain direction may look weird.

DS-Minetest:
The problem is that, without the modulo, the animation timer might become less precise. (`porting::getTimeMs()` returns a `u64`.)

lhofhansl:
Thanks @paramat. Agreed it's not biome fog, happy to call it haze (but then we already have fog_start, which was also added by me... Sorry :( Happy to change that too)

It's still moving in one direction (positive Z). I could add an API for that (perhaps overkill) or config options, or perhaps just take the cloud param values.

How does one alter the cloud direction? I find no configs for that. Can mods do that, only?

It would be nice if this would be just a simple mood-enhancing feature, like waves (which also do not follow any wind direction, then again it's different there). Perhaps it just shouldn't move? (the effect is much reduced when it's stationary, though).

lhofhansl:
@DS-Minetest Perhaps. A few moments ago we were at 1568061530474, so 15680615.30474 divided by 100.000. We're not interested in the size of the value, but it's change, the .30474 part, the precision of the mantissa is not changed from before, the larger value just moves into the exponent. So I'm fairly sure it's OK. In fact it works fine with the current value.
I can subtract 1568060000000 or something from the count to be safe, but I do not think it's needed.

paramat:
'haze' is fine for me, or maybe any other better word you can think of.

Yes i should quickly add, i'm not requesting the use of the cloud API wind direction. I also somewhat feel it's not worth doing, water waves and waving flora also use a fixed direction. Since it's subtly-changing ground haze it's probably not a big issue to differ from wind direction.

So i have no objection to the current movement. Just one request, the default cloud speed is 2 nodes per second, so i think the haze speed should be the same or lower, not faster.

Also to clarify, i don't officially disapprove of this PR, i'm neutral.

TumeniNodes:
"mist"

DS-Minetest:
>the larger value just moves into the exponent

What exactly do you mean with this?

>the precision of the mantissa is not changed from before

The bigger the number is, the more of the mantissa is used for the big values. There are less bits available for precise small changes.
If a client would be on a server for very very long time (or even if the computer that the client runs on is already on for a long time), the movement might be juddery or jumpy.

>the default cloud speed is 2 nodes per second, so i think the haze speed should be the same or lower, not faster.

This is the speed that the haze, fog, mist or whatever moves with: `animationTimer * 2 * BS` (Btw. does the `BS` really make sense here?)
`animationTimer` is in 100 seconds (see https://github.com/minetest/minetest/pull/8889#issuecomment-529498322), it increases with `1/100s`. Hence the speed is `20/100=0.2` nodes per second.

lhofhansl:
@DS-Minetest Yeah, BS makes no sense there. I had 20 there at some point before instead of 2 * BS.

Re: float, I just meant to say that the pure size of the number is of no significance, and that we just need something that changes. But you're right, we'd run into precision problems.

Another option is to increase the period from 100,000 to 1,000,000 (reset every 16 minutes) or even 10,000,000 (~ every 2h47min, and this is well within float precision). The divider (speed) would remain 100,000.

lhofhansl:
@paramat It's OK if you disapprove. I apologize for my reaction above.

Your math is correct @DS-Minetest , however, it looks like the haze is moving than the clouds.
It also depends on the periodicity of the noise function. Note that I'm dividing the world position to make the "blobs" a pleasing size. The divider comes to about 200 or so; in that case the fog would move at about 4n/s. I can half that to match the cloud speed.

Fixer-007:
Actually this fog could be the beginning of weather in vanilla minetest, it does not need to be persistent all day, but show up in some situations, like early morning, evening, at rain, etc

lhofhansl:
What do you folks think should happen at night? Looks like there about 5% daylight at night, which means the night_ration < 0.99 will be false it will generate dark haze patches at night. It's not incorrect.

@Fixer-007 It's just a mood-enhancer. I agree with @paramat that proper weather needs to be modeled as actual objects (rain, snow, fog objects, different clouds, etc). Of course the non-uniform haze could be integrated later.

lhofhansl:
On the naming of "fog" it's all over minetest already. F3 turns "fog" on an off, etc.
Perhaps if we wanted to change that it should be a different PR with (presumably) some amount of discussion.

I think this one is ready. It's optional, default off, and not that much code.

paramat:
Fixer-007
> Actually this fog could be the beginning of weather in vanilla minetest,

It can't be, as sort of explained earlier, as it is clientside and therefore world, biome and server-independent. Weather has to be server controlled and biome-dependent.
> On the naming of "fog" it's all over minetest already. F3 turns "fog" on an off, etc.

Yes i started to realise that too, so 'fog'' is ok, at least for now.

lhofhansl:
> Looks like there about 5% daylight at night

My statement doesn't make sense, of course. The night ratio is a feature of the position in the generated map and just indicated how little daylight *can* reach this place (not how much actually does).

client/shaders/nodes_shader/opengl_fragment.glsl
```diff
@@ -14,13 +15,43 @@ varying vec3 eyeVec;
...
 const float BS = 10.0;
 const float fogStart = FOG_START;
-const float fogShadingParameter = 1 / ( 1 - fogStart);
+float fogShadingParameter = 1 / ( 1 - fogStart);
+
+#ifdef ENABLE_FOG_NOISE
+
+// Simple, fast noise function.
+// See: https://gist.github.com/patriciogonzalezvivo/670c22f3966e662d2f83
+vec4 perm(vec4 x){return mod(((x * 34.0) + 1.0) * x, 289.0);}
```
ClobberXD:
```suggestion
vec4 perm(vec4 x) { return mod(((x * 34.0) + 1.0) * x, 289.0); }
```

DS-Minetest: All that copied code has horrible code style.

builtin/mainmenu/tab_settings.lua
```diff
@@ -257,7 +257,9 @@ local function formspec(tabview, name, tabdata)
 			"checkbox[8.25,3;cb_waving_leaves;" .. fgettext("Waving Leaves") .. ";"
 					.. dump(core.settings:get_bool("enable_waving_leaves")) .. "]" ..
 			"checkbox[8.25,3.5;cb_waving_plants;" .. fgettext("Waving Plants") .. ";"
-					.. dump(core.settings:get_bool("enable_waving_plants")) .. "]"
+					.. dump(core.settings:get_bool("enable_waving_plants")) .. "]" ..
+			"checkbox[8.25,4.0;cb_fog_noise;" .. fgettext("Non Uniform Fog") .. ";"
```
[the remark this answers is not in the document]
lhofhansl: Done

client/shaders/nodes_shader/opengl_fragment.glsl
lhofhansl: Fixed

lhofhansl:
@DS-Minetest The change is good as is, I think... Just for completeness I did more testing on floats. It looks like floats are accurate on whole number to exactly 2^24 or 16.777.216. Since we're dividing by 100.000. We have about 1677721600000 or about 53 years of precision.

lhofhansl:
Also. Does MT reset the worldPosition, i.e. the GL coordinates of the world? It certainly looks like. If I move around every now and then the fog jumps a bit. I ruled out the animation time for this. So it somehow must reset the world/model coordinates.

How do I get the actual node coordinates?

I think this is still, but it can be improved.

paramat:
> Does MT reset the worldPosition, i.e. the GL coordinates of the world?

Yes, every 200 nodes or so, it's a fix for visual bugs occuring when far from world centre.

lhofhansl:
Thanks @paramat . Hmm, so how does get actual world coordinates in the shader? I could pass in the offset, I suppose, but that would defeat the purpose.

paramat:
Not sure how.
I've also noticed jumps in shader behaviour, i assumed it was the animation timer resetting.

lhofhansl:
Is that also costly. Perhaps it explains some of the stuttering I see when moving around a lot.

Perhaps make this configurable? 1000 seems a better default anyway. Yes, some minor glitches might be visible at the extremes.

paramat:
As far as i know it isn't costly.
I also suggested 1000 years ago, but it was claimed that 200 was quite well chosen and apparently the rendering glitches occur at surprisingly small distances from 0,0,0. Haven't tested that though.

Stuttering when moving around is much more likely to be the load on mesh creation, that is what causes FPS stutter for me.

lhofhansl:
I see. Hmmm...

Any opposition to releasing this one as is? Or is it the "jumping" of the fog an unacceptable glitch?

lhofhansl:
OK... I have a change, which passes the current camera offset to the shaders. That way the shaders can decide what they want to do... If they want to use the actual coordinate, just add the passed cameraOffset to worldPosition. That's generally good to have!

For the fog the lack of precision works fine, because (a) it's fuzzy anyway, and (b) it's not doing any further transformation on it other using it for the noise function. I tested this at 20000 positions, and it all looks fine.

Should I update this PR with that?

lhofhansl:
Heck. Did it anyway. As I said the shaders might generally want to have a choice on whether to use the adjusted or original coordinates.

paramat:
> passes the current camera offset to the shaders. That way the shaders can decide what they want to do... If they want to use the actual coordinate, just add the passed cameraOffset to worldPosition. That's generally good to have!

This is useful, and good to see it wasn't complex to do.
So later we will be able to fix jumps in the other shaders.

lhofhansl:
Yep. Currently this is only used for the waves and non-uniform fog.

Good to merge this one?

lhofhansl:
Anything else I need to do here?
(Before merge I'd add a comment to the shaders explaining the meaning of worldPosition and cameraOffset.)

src/client/shader.cpp
```diff
@@ -686,6 +686,16 @@ ShaderInfo generate_shader(const std::string &name, u8 material_type, u8 drawtyp
 	if (g_settings->getBool("enable_bumpmapping"))
 		shaders_header += "#define ENABLE_BUMPMAPPING\n";
 
+	if (g_settings->getBool("enable_fog_noise")) {
+		shaders_header += "#define ENABLE_FOG_NOISE\n";
+		shaders_header += "#define FOG_NOISE_FREQUENCY ";
+		shaders_header += itos(g_settings->getU16("fog_noise_frequency") / 10);
+		shaders_header += "\n";
+		shaders_header += "#define FOG_NOISE_STRENGTH ";
+		shaders_header += ftos(g_settings->getFloat("fog_noise_strength") * 4.0);
```
paramat: 4.0f

paramat:
No further requests from me, once my line comment is attended to.

lhofhansl:
Thanks @paramat . Fixed the number format, and also added comments to the shaders explaining the use of both worldPosition and cameraOffset for future shader effects.

lhofhansl:
(When this is merged I have a proposal PR for Perlin waves that are configured with exactly the same parameters as the current waves, but instead of simple (and sterile) sine waves these use Perlin noise. No more work for the shader and look more realistic.)

But it needs some of the changes here.

lhofhansl:
Let's get this in. The change to pass the cameraOffset to the shaders is needed for other issues as well.

If need be I can make a *separate PR* just for that, but since I went through great length to make this configurable with small changes only we should just pull this in.

Let me know...

lhofhansl:
Hmm... I always mess things up when I do a rebase :(

lhofhansl:
So do we want this? :)
The only controversial behavior could potentially be in semi-dark places that are artificially lit, since I have nothing to go by other than the day-night ratio to decide whether this effect should happen or not.
(But even then it's only when this feature is enabled)

(@paramat , I'd love to think about how mods can control (at least enable/disable) shader effects in a future PR. This would go well with your new weather mod for example (and be totally lightweight on the server). The client would have to opt in of course.)

paramat:
I suspect core devs are busy with release, so hopefully this will get more attention after release.
I dislike the concept as explained above, but it is a harmless shader so i am not going to officially disapprove.

I think this shader is incompatible with my MTG weather intentions (wrong wind speed and direction for a start), and i would not want to have anything to do with it because i dislike the concept (sorry). Users can enable the shader if they want but that has to be the choice of the user as some users cannot use shaders or do not like them. So i think this shader being independently switched, as currently, is best, also simplest.

SmallJoker:
The fog effect is quite subtle, but might bring a little bit of variety to the game. No concept objections from my side.

lhofhansl:
Ping...

lhofhansl:
Hi all. Is there still interest in this?

Perhaps the effect is too subtle now? (it's configurable, I'd just have to change some default.
Please let me know. I'm also OK closing this.

paramat:
It seems liked by 2 core devs, rubenwardy posted 'i like this' above, and SmallJoker thumbed-up the first post.

hecktest:
What if this was also world space height based? That could look good.
Also consider #9895, generic effect APIs are more powerful than hardcoded effects, and the effort is better spent letting anyone make their own cool fog shader.

paramat:
The shader eye position bugfix should perhaps be split off from this into a new PR to fix bumpmapping issues. Bugfixes and features should ideally be separate commits anyway:
From https://github.com/minetest/minetest_game/issues/2473#issuecomment-534350169
https://github.com/minetest/minetest_game/issues/2473#issuecomment-534842743

> Edit: In fact it looks like the lightVec is miscalculated. The eyePosition passed to the shader is global position, whereas the worldPosition is relative to the camera offset. In minetest/minetest#8889 I pass the cameraOffset to the shaders so that real position can be calculated. We should either do that, or adjust the eyePosition with the cameraOffset before passing it to the shaders.

> We need the change from minetest/minetest#8889 for me to fix this, or alternatively I can create a engine PR just for passing the cameraOffset.

The delay in reviewing this PR is delaying the bugfix, and it seems the bugfix is fairly simple which means it will get reviewed quicker (i will review and test a bugfix PR, but i cannot approve this feature PR).

paramat:
Hmm but ...
See https://github.com/minetest/minetest/issues/9711#issuecomment-706549399
If we decide to improve the shader bumpmapping light source positioning, as needs to be done, that alters the same code, so we should decide on that first.

PR #8904, opened 2019-09-05
Move debug.txt after it grows too big

### LINKED ISSUE #4449: debug.txt slowly becomes huge

HybridDog:
Before opening the file for writing, its file size is tested. If it exceeds 50 MB, it is moved to debut.txt.1, otherwise the log is appended to the old messages. An old debut.txt.1 is removed if it already exists.
Additionally, information about the log file location and whether it is deleted is printed to the action log when starting minetest.

The function which opens the file is now called setFile because "open" sounds too simple for what it does.

Fixes #4449. 

## How to test

Apply the changes, adjust the `debug_log_level` setting if needed, and change `debug_log_size_max`.

### DIFF
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -1322,6 +1322,7 @@ debug_log_level (Debug log level) enum action ,none,error,warning,action,info,ve
 #    If the file size of debug.txt exceeds the number of bytes specified in
 #    this setting when it is opened, the file is moved to debug.txt.1,
 #    deleting an older debug.txt.1 if it exists.
+#    debug.txt is only moved if this setting is positive.
 debug_log_size_max (Debug log file size threshold) int 50000000
 
 #    IPv6 support.
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -1319,11 +1319,11 @@ language (Language) enum   ,be,ca,cs,da,de,dv,en,eo,es,et,fr,he,hu,id,it,ja,jbo,
 #    -    verbose
 debug_log_level (Debug log level) enum action ,none,error,warning,action,info,verbose
 
-#    If the file size of debug.txt exceeds the number of bytes specified in
+#    If the file size of debug.txt exceeds the number of megabytes specified in
 #    this setting when it is opened, the file is moved to debug.txt.1,
 #    deleting an older debug.txt.1 if it exists.
 #    debug.txt is only moved if this setting is positive.
-debug_log_size_max (Debug log file size threshold) int 50000000
+debug_log_size_max (Debug log file size threshold) int 50
 
 #    IPv6 support.
 enable_ipv6 (IPv6) bool true
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -1319,10 +1319,10 @@ language (Language) enum   ,be,ca,cs,da,de,dv,en,eo,es,et,fr,he,hu,id,it,ja,jbo,
 #    -    verbose
 debug_log_level (Debug log level) enum action ,none,error,warning,action,info,verbose
 
-#    If the file size of debug.txt exceeds the amount in bytes specified in
-#    this setting, the file's content is deleted when it is opened
-#    the next time.
-debug_log_size_max (Debug log file size threshold) int 1000000
+#    If the file size of debug.txt exceeds the number of bytes specified in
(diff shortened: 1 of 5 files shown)


src/log.cpp
```diff
@@ -310,16 +310,30 @@ void Logger::logToOutputs(LogLevel lev, const std::string &combined,
...
 ////
 
-void FileLogOutput::open(const std::string &filename)
+void FileLogOutput::setFile(const std::string &filename, u64 file_size_max)
 {
-	m_stream.open(filename.c_str(), std::ios::app | std::ios::ate);
+	actionstream << "Log messages are saved to " << filename << std::endl;
+
+	std::ifstream ifile(filename, std::ios::binary | std::ios::ate);
+	bool is_too_large = ifile.tellg() > static_cast<std::streamoff>(file_size_max);
```
sfan5: since `std::streamoff` is a signed type why not just make file_size_max an s64?

sfan5: I'd also like this mechanism to be optional by setting `debug_log_size_max = 0` (or -1)

src/log.cpp
```diff
@@ -310,16 +310,30 @@ void Logger::logToOutputs(LogLevel lev, const std::string &combined,
...
+	ifile.close();
+
+	if (is_too_large) {
+		std::string filename_secondary = filename + ".1";
+		actionstream << "The log file grew too big; it is moved to " <<
+			filename_secondary << std::endl;
+		std::remove(filename_secondary.c_str());
+		std::rename(filename.c_str(), filename_secondary.c_str());
```
sfan5: The existing code usually doesn't prefix C standard functions with `std::`

src/log.cpp
HybridDog: done

HybridDog:
> I'd also like this mechanism to be optional by setting debug_log_size_max = 0 (or -1)

I've added this feature. I'm not sure if it is very useful because you can simply set the maximum size to, for example, 1 TB.

src/log.cpp
HybridDog: I've changed it. Unfortunately there are no native C++ functions for file renaming and removal as far as I knot.

sfan5: There are, but they are C++17 and newer.
https://en.cppreference.com/w/cpp/filesystem/remove

src/log.cpp
```diff
@@ -310,16 +310,34 @@ void Logger::logToOutputs(LogLevel lev, const std::string &combined,
 //// *LogOutput methods
 ////
 
-void FileLogOutput::open(const std::string &filename)
+void FileLogOutput::setFile(const std::string &filename, s64 file_size_max)
 {
-	m_stream.open(filename.c_str(), std::ios::app | std::ios::ate);
+	actionstream << "Log messages are saved to " << filename << std::endl;
```
sfan5: This log message is annoying IMO.

builtin/settingtypes.txt
```diff
@@ -1319,6 +1319,12 @@ language (Language) enum   ,be,ca,cs,da,de,dv,en,eo,es,et,fr,he,hu,id,it,ja,jbo,
 #    -    verbose
 debug_log_level (Debug log level) enum action ,none,error,warning,action,info,verbose
 
+#    If the file size of debug.txt exceeds the number of bytes specified in
```
SmallJoker: Such large numbers are not user-friendly. The size in megabytes would be easier to understand for normal users. `50E6` would be possible here, but you cannot expect this notation from regular users.

SmallJoker:
Will approve as soon the comment above is handled.

builtin/settingtypes.txt
HybridDog: I've changed it to megabytes.

Issue #228, opened 2012-09-02
Rendering glitches when far from the center of the map


VanessaE:
When you're near the center (even a couple thousand meters from it), the game renders fine within previously-documented limits.  However, get really far from the center on both X and Z axes, say 30000m or so, and blocks start exhibiting severe rendering glitches.  Edges dance around and/or disappear into neighboring edges, signs and other 2d objects flicker in and out of existence, etc.  See this video for an example:

http://youtu.be/wBDPPJ6t0wA

Low framerate aside, this is typical of what I'm experiencing.

PR #8937, opened 2019-09-15
Make CAMERA_OFFSET configurable.

lhofhansl:
Title says it all.

I'd even vote for increasing the default to 1000, since only in fairly extreme situation would the glitches be noticeable. (But that's a different discussion)

Triggered in part by #8889

### DIFF
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -617,7 +617,7 @@ near_plane (Near clipping plane) float 0.1 0 0.5
 #   Rendering glitches are possible at large distances from the center of the map.
 #   This setting forces a reset of the map at the specified distance.
 #   The max setting disables the logic.
-camera_offset (Camera offset) int 200 100 31000
+camera_offset (Camera offset) int 512 100 31000
 
 #    Width component of the initial window size.
 screen_w (Screen width) int 1024
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -614,6 +614,11 @@ viewing_range (Viewing range) int 100 20 4000
 #   0.1 = Default, 0.25 = Good value for weaker tablets.
 near_plane (Near clipping plane) float 0.1 0 0.5
 
+#   Rendering glitches are possible at large distances from the center of the map.
+#   This setting forces a reset of the map at the specified distance.
+#   The max setting disables the logic.
+camera_offset (Camera offset) int 200 100 31000
+
 #    Width component of the initial window size.
 screen_w (Screen width) int 1024
 
--- a/src/client/camera.cpp
+++ b/src/client/camera.cpp
@@ -36,7 +36,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include "fontengine.h"
 #include "script/scripting_client.h"
 
-#define CAMERA_OFFSET_STEP 200
+static thread_local const s16 CAMERA_OFFSET_STEP = g_settings->getS16("camera_offset");
 #define WIELDMESH_OFFSET_X 55.0f
 #define WIELDMESH_OFFSET_Y -35.0f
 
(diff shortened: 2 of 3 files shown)


SmallJoker:
Since the attachment problems appear when the offset changes, how is this going to solve the issue? People around the 1000-borders will still experience the very same misbehavior.

paramat:
SmallJoker, i am assuming it is not intended to solve the issue.
I think i prefer to simply have a larger default rather than another setting. 200 does seem excessively small, i was told it was justified but i am sceptical somehow.
Maybe 512?

lhofhansl:
Yeah... It's not going to solve the problem - I don't know how to solve it for real.

paramat:
I remember one way we tested rendering glitches was to use a 'thinnest common nodebox' of 1/16th (or 1/32?) node thickness (modify default:snow), and use biome definitions to place this as 'biome_dust' on top of a contrasting colour node. Flying high above causes the underneath node to show though.

paramat:
See #1119 #1093 #228
Seems to occur even at 1000?: https://github.com/minetest/minetest/issues/1093#issuecomment-32012342
Thin snow nodeboxes on pines are a good way to test this (with opaque leaves).

paramat:
Ha look what i found #3694

lhofhansl:
But still a more rare occasion, no? Happy to leave the default at 200, though.

paramat:
Well, i like the idea of raising the default, i think testing and re-assessing is worth doing.

lhofhansl:
Cool... I added the testing-needed label as I'm not sure I'll will reliably know what to look for.
The only way I could see anything was at large X or Z (20000) and then watching a cloud going "through" a mountain... I could see the flickering when I set CAMERA_OFFSET to 31000.

ClobberXD:
What might be the potential implications of increasing camera offset? Also, if it isn't offtopic, would someone be able to help me understand why a camera offset is even needed? AFAICT, it's something to do with our camera position not syncing properly with Irrlicht's camera a little further away from 0,0,0. It'd nice if someone could shed more light on this though.

paramat:
My testing method described above is very sensitive to this effect, it's a good way to test.

lhofhansl:
@ClobberXD it's due to (float) precision as discussed in #1093. Instead of using actually large numbers for rendering, every now and then (200 nodes by default) we instead "move the world", so that the actual coordinates sent to the rendering pipeline are small (<= rendering distance + offset)

ClobberXD:
Oh I see. So we use floats to keep the size of data passed to the rendering pipeline small? Would using `double`s eliminate the need for this quirky re-adjustment?

lhofhansl:
Double would presumably fix it. But... even OpenGL uses only floats. I do not know how much of that is hardcoded. There's an ARB extension to switch to 64bit, but I do not know whether that's enough, and apparently it reduces performance a lot.

Also, do we know it's not zbuffer precision? I notice we set it to 24 bits, could try 32 bits instead.

paramat:
If it seems the distance cannot be increased to 512, 256 would be a nice number and only a slight increase.

lhofhansl:
Not sure that it needs to be a power of two. :)
We just reset the (visible) world every now and then so that it is centered around the current camera position.

From the code it looks like the cost of this is roughly proportional to the number of visible map blocks.

I solved the main problem in #8889, be simply passing the new center visible world (the camera offset) to the shader, so that they have a choice to use visible-world-coordinates, or use global coordinates by adding the offset.

I'm ok to close this one, assuming we get the other change in eventually.

paramat:
I can't yet see justification for a setting, although i think it's worth re-considering the hardcoded default.

jastevenson303:
And per-player, like zoom/FOV.

ClobberXD:
@jastevenson303 The concept of "per-player" doesn't exist here, because `CAMERA_OFFSET` is completely client-side. :)

jastevenson303:
I'm just used to being able to set my own preferred zoom FOV, independent
of the server or other players.  Thanks for the clarification!

lhofhansl:
Seems pointless now.

PR #8940, opened 2019-09-17
light.cpp: Range-limit all configurable lighting params

ClobberXD:
This PR actually enforces the ranges of all configurable lighting params. Without this PR, there's nothing stopping users from setting an exorbitantly high `lighting_alpha` value, for example. The upper and lower bounds of the lighting params are taken from `builtin/settingtypes.txt`.

This PR is Ready for Review.

### How to test

- Try to set a lighting param higher than the prescribed range, by manually editing `minetest.conf`.
- Observe that values higher than the upper bound have no effect, and the result is no different from the upper bound.

### DIFF
--- a/src/light.cpp
+++ b/src/light.cpp
@@ -55,16 +55,16 @@ float decode_light_f(float x)
 void set_light_table(float gamma)
 {
 // Lighting curve derivatives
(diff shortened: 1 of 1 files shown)


ClobberXD:
Replaced by 1f142ec06fc146464822a322fdf39bddb680bfe6 :)

PR #8941, opened 2019-09-17
Improve undersampling settings

### LINKED ISSUE #8939: Higher undersampling resolution
## Problem
I would like to be able to undersample more than four pixels

## Solutions

Make the undersample box a number input instead of a dropdown

Calinou:
The setting now accepts values between 1 and 8 in the Advanced Settings menu. Values 0 and 1 now behave the same way (setting it to 1 won't disable MSAA anymore), so there's no need to expose 0 as a value.

Very high undersampling values may be desired when playing on a 4K (or higher) display with slow graphics.

The setting description was also updated to be more detailed.

This closes #8939.

## To do

## How to test

Set `undersampling` to various values using the Advanced Settings menu, then start a local game or connect to a server.

### DIFF
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -508,10 +508,11 @@ texture_min_size (Minimum texture size) int 64
 #    when set to higher number than 0.
 fsaa (FSAA) enum 0 0,1,2,4,8,16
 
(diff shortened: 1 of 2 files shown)


SmallJoker:
05a7da627959afef2198f2036f4144e3d8abfbda

PR #8958, opened 2019-09-19
Allow grep-ing for on_grant and on_revoke

DS-Minetest:
Just two code comments are added.

- Goal of the PR: Adding code comments to make `grep`-ing on_grant and on_revoke.
- How does the PR work? It adds two code comments.
- Does it resolve any reported issue? No, but @Wuzzy2 complained about this on irc: http://irc.minetest.net/minetest-dev/2019-09-18#i_5593423
- why is this PR needed? It is useful for looking up the source code.

Here are the callbacks in doc:
https://github.com/minetest/minetest/blob/5c9983400fb085167bf11f8a0483b8e8f6dd8a24/doc/lua_api.txt#L7107-L7113
## To do

## How to test

See that code comments with "on_grant" and "on_revoke" in them are added.

### DIFF
--- a/builtin/game/chat.lua
+++ b/builtin/game/chat.lua
@@ -186,7 +186,7 @@ local function handle_grant_command(caller, grantname, grantprivstr)
 		return false, privs_unknown
 	end
 	for priv, _ in pairs(grantprivs) do
-		-- call the on_grant callback
+		-- call the on_grant callbacks
 		core.run_priv_callbacks(grantname, priv, caller, "grant")
 	end
 	core.set_player_privs(grantname, privs)
--- a/builtin/game/chat.lua
+++ b/builtin/game/chat.lua
@@ -259,7 +259,7 @@ core.register_chatcommand("revoke", {
 		end
 
 		for priv, _ in pairs(revoke_privs) do
-			-- call the on_revoke callback
+			-- call the on_revoke callbacks
 			core.run_priv_callbacks(revoke_name, priv, name, "revoke")
 		end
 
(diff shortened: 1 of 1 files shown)


builtin/game/chat.lua
```diff
@@ -186,6 +186,7 @@ local function handle_grant_command(caller, grantname, grantprivstr)
 		return false, privs_unknown
 	end
 	for priv, _ in pairs(grantprivs) do
+		-- call the on_grant callback
```
ClobberXD: callback**s**

```suggestion
		-- call the on_grant callbacks
```

builtin/game/chat.lua
```diff
@@ -258,6 +259,7 @@ core.register_chatcommand("revoke", {
 		end
 
 		for priv, _ in pairs(revoke_privs) do
+			-- call the on_revoke callback
```
ClobberXD:
```suggestion
			-- call the on_revoke callbacks
```

ClobberXD:
Neat, simple(st) idea (ever). But it'd work. :+1:

Even without `grep`, the comment is quite useful to document where exactly the callbacks are invoked, because the two functions have no comments in the code to document their function, AFAIK.

Wuzzy2:
Aahh, there it is called! Now I see it. Thanks. :-) I was just really confused. And of course, it's a string concatenation. Augh!

It was not really a complaint. I was just wondering. I even suspected the functions don't even exsit.

But thanks for adding code comments. ^_^

PR #8959, opened 2019-09-19
Punchwear (improved)

sfan5:
replaces #8862 
~~Sorry for not just reviewing and allowing the submitter to fix the PR, there were too many small issues that I could more easily just solve myself.~~

Note that the number of uses calculation is identical to the old PR.

## Discussion of the implementation

Existing tools automatically receive a value for `punch_attack_uses` by inserting the first suitable  groupcap into the formula `uses * 3^(maxlevel - 1)`.
This can of course be overriden or disabled by modders.

### The formula in practice

Resulting use counts for MTG tools (full punches):

tool | uses
---- | ----
default:sword_wood | 10
default:sword_stone | 20
default:sword_bronze | 75
default:sword_steel | 90
default:sword_mese | 270
default:sword_diamond | 360

## How to test

1. Punch another player and observe how the sword gets damaged.

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -6256,7 +6256,7 @@ Used by `minetest.register_node`, `minetest.register_craftitem`, and
 
         liquids_pointable = false,
 
-        -- See "Tools" section
+        -- See "Tools" section for an example including explanation
         tool_capabilities = {
             full_punch_interval = 1.0,
             max_drop_level = 0,
@@ -6266,6 +6266,14 @@ Used by `minetest.register_node`, `minetest.register_craftitem`, and
                          uses = 20, maxlevel = 2},
             },
             damage_groups = {groupname = damage},
+
+            punch_attack_uses = nil,
+            -- Amount of uses this tool has for attacking players and entities
+            -- by punching them (0 = infinite uses).
+            -- For compatibility, this is automatically set from the first
+            -- suitable groupcap using the forumla "uses * 3^(maxlevel - 1)".
+            -- It is recommend to set this explicitly instead of relying on the
+            -- fallback behavior.
         },
 
         node_placement_prediction = nil,
--- a/src/script/lua_api/l_object.cpp
+++ b/src/script/lua_api/l_object.cpp
@@ -187,7 +187,8 @@ int ObjectRef::l_punch(lua_State *L)
 	u16 dst_origin_hp = puncher->getHP();
 
 	// Do it
-	co->punch(dir, &toolcap, puncher, time_from_last_punch);
+	u16 wear = co->punch(dir, &toolcap, puncher, time_from_last_punch);
+	lua_pushnumber(L, wear);
 
 	// If the punched is a player, and its HP changed
 	if (src_original_hp != co->getHP() &&
@@ -202,7 +203,7 @@ int ObjectRef::l_punch(lua_State *L)
 		getServer(L)->SendPlayerHPOrDie((PlayerSAO *)puncher,
 				PlayerHPChangeReason(PlayerHPChangeReason::PLAYER_PUNCH, co));
 	}
-	return 0;
+	return 1;
 }
 
 // right_click(self, clicker); clicker = an another ObjectRef
--- a/builtin/game/register.lua
+++ b/builtin/game/register.lua
@@ -256,6 +256,18 @@ function core.register_tool(name, tooldef)
 	end
 	-- END Legacy stuff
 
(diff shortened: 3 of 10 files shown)


Wuzzy2:
> This makes it hard for modders to explicitly control how many uses a tool should have for damage purposes.

Is it possible to bypass or disable the punchwear?

sfan5:
Not really* with the current solution.


\* Assuming groupcap order is preserved you could add the below one, which would only apply to the wear calculation and result in a wear of zero:
`  a_dummy_groupcap = { times={0}, uses = 99999, maxlevel = 2},`

Wuzzy2:
Ooff, that's kind of awful. If this forces the wear system to tools with no sane way to disable it, it greatly lowers flexibility of mods. :-/

Moth-Tolias:
the gap between wood and stone / bronze and steel seems rather small.
+1 regarding the ability to customize and disable punchwear in the itemdef.

ClobberXD:
Tested by punching a player with a steel pick at varying speeds, in `minimal`. Works :+1:

doc/lua_api.txt
```diff
@@ -6266,6 +6266,14 @@ Used by `minetest.register_node`, `minetest.register_craftitem`, and
                          uses = 20, maxlevel = 2},
```
SmallJoker: Needs reference to `punch_attack_uses`

SmallJoker: Should also mention `punch_attack_uses`

src/script/lua_api/l_object.cpp
```diff
@@ -187,7 +187,8 @@ int ObjectRef::l_punch(lua_State *L)
 	u16 dst_origin_hp = puncher->getHP();
 
 	// Do it
-	co->punch(dir, &toolcap, puncher, time_from_last_punch);
+	u16 wear = co->punch(dir, &toolcap, puncher, time_from_last_punch);
+	lua_pushnumber(L, wear);
```
SmallJoker: Needs documentation.

sfan5: is already documented
https://github.com/minetest/minetest/blob/5fa614d97e13af64be490336392abe2a54fdcbc1/doc/lua_api.txt#L1811-L1813

doc/lua_api.txt
sfan5: > Should also mention punch_attack_uses

This example aims to explain how tools interact with nodes, so I chose not to include it here. (technically `damage_groups` doesn't need to be there either)

> Why do you need the part in register.lua if the group-specific parameter is never used?

The code in register.lua ensures existing tools get an useful `punch_attack_uses` value if they don't specify one themselves.

> group-specific

???
The parameter is part of tool_capabilities, not of one of the groupcaps.

src/script/lua_api/l_object.cpp
SmallJoker: Oh weird, this means it never worked before.

doc/lua_api.txt
SmallJoker: Sorry, my mistake. The Lua code actually modifies `toolcaps`, not `cap`, what I erroneously thought.

src/script/lua_api/l_object.cpp
sfan5: Indeed

Wuzzy2:
The title of this issue now includes the word “improved”.
Now what exactly has changed?

sfan5:
Modders can properly set a custom amount of uses:
https://github.com/minetest/minetest/blob/a4634d99063803c4b73fd5d7c6999000225ede53/doc/lua_api.txt#L6270-L6276

MoNTE48:
This breaks this (and not only) mod:
https://notabug.org/TenPlus1/mobs_redo/issues/27

1) Swords are not damaged when punching luaentity (non-player).
2) if you try to replace the wielditem of the kicking player, it will not happen. C++ will make it faster and the old reference to the wielditem will become invalid.

sfan5:
> 1. Swords are not damaged when punching luaentity (non-player).

Can't reproduce.

> 2. if you try to replace the wielditem of the kicking player, it will not happen. 

Yeah I can see why there's a problem.

MoNTE48:
1. Check this by installing the mobs_redo + mobs_animal. There will be no damage to the sword when you punch mob.

sfan5:
Since mobs_redo uses `set_wielded_item`, what you are seeing is the second bug.

MoNTE48:
@sfan5 Ok, I’ll test your PR and let you know. Thanks!

PR #8960, opened 2019-09-19
Move the code for meta formspec into Game::nodePlacement

DS-Minetest:
- Goal of the PR: Make the code more consistent and clearer.
- How does the PR work? The code that opens meta formspec is moved into `Game::nodePlacement`. It was basically just some cut and pasting.
- Related #8640 
- Another benefit of this PR is that the wieldhand is animated when the player opens a meta formspec.

## To do

## How to test

Open furnaces and co.
#8640 has testing code.

### DIFF
--- a/src/client/game.cpp
+++ b/src/client/game.cpp
@@ -802,7 +802,8 @@ private:
 	void updateChat(f32 dtime, const v2u32 &screensize);
 
 	bool nodePlacement(const ItemDefinition &selected_def, const ItemStack &selected_item,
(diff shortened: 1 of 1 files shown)


sfan5:
I'm not sure if moving all the code into `nodePlacement` makes anything more readable or consistent.
meh

DS-Minetest:
> I'm not sure if moving all the code into `nodePlacement` makes anything more readable or consistent.
> meh

Without this PR, the meta formspec is checked and opened outside `nodePlacement` while `rightclickable` is checked in that function. Ergo this PR creates more consistency.
Sadly the diffs are not very readable.

PR #8981, opened 2019-09-23
Fix some reference counters (memleak)

SmallJoker:
Fixes a few places where the reference counter was erroneously increased, resulting in a never dropped pointer.

Ready for Review.

## How to test

1) Join game
2) Leave game
3) No crash

### DIFF
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -306,6 +306,11 @@ Server::~Server()
 	for (auto &detached_inventory : m_detached_inventories) {
 		delete detached_inventory.second;
 	}
+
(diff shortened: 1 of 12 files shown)


src/server.cpp
```diff
@@ -306,6 +306,11 @@ Server::~Server()
 	for (auto &detached_inventory : m_detached_inventories) {
 		delete detached_inventory.second;
 	}
+
+	while (!m_unsent_map_edit_queue.empty()) {
```
SmallJoker: This fixes the memory leak, but the data lost here is negligible. The rest of the map changes are to improve allocation safety.

PR #8991, opened 2019-09-26
Simple shader fixes.

lhofhansl:
Since we're a bit slow on #8889 I factored out the shader fixes.

1. Pass current camera offset to shader, so shader have access to the global coordinates
2. Pass animation timer to fragment shader. C++ code is already there, just wasn't declared in the shader
3. Delay animation timer wrap-around (from 100s to about 16 minutes)

Let's merge these quickly. Should be uncontroversial and absolutely no behavioral change. Just simple changes that let us fix shader issues in the future and allow for new shader features.

See also #8889 and #8976 as well as minetest/minetest_game#2473

### DIFF
--- a/src/client/game.cpp
+++ b/src/client/game.cpp
@@ -413,6 +413,8 @@ class GameGlobalShaderConstantSetter : public IShaderConstantSetter
 	CachedPixelShaderSetting<float, 3> m_eye_position_pixel;
 	CachedVertexShaderSetting<float, 3> m_eye_position_vertex;
 	CachedPixelShaderSetting<float, 3> m_minimap_yaw;
+	CachedPixelShaderSetting<float, 3> m_camera_offset_pixel;
+	CachedPixelShaderSetting<float, 3> m_camera_offset_vertex;
 	CachedPixelShaderSetting<SamplerLayer_t> m_base_texture;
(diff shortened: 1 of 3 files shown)


src/client/game.cpp
```diff
@@ -493,7 +497,7 @@ class GameGlobalShaderConstantSetter : public IShaderConstantSetter
 			sunlight.b };
 		m_day_light.set(dnc, services);
 
-		u32 animation_timer = porting::getTimeMs() % 100000;
+		u32 animation_timer = porting::getTimeMs() % 1000000;
 		float animation_timer_f = (float)animation_timer / 100000.f;
```
sfan5: shouldn't this also be adjusted so animationTimer stays within `[0.0, 1.0[`?

lhofhansl: That would slow down any current animation (and hence would require a shader fix). And then when you speed up the timer there by 10x, you'd get the wrap exactly after 100s again.

The only way is to separate the wrap from the speed. The speed stays the same, but it wrap 10x slower.

Shader do not (and should not) rely on specific values.

sfan5: Makes sense.

lhofhansl:
Cool will merge soon (after announcing on IRC, assuming no concerns there)

ClobberXD:
It's great to see some shader work being done. Thanks!

PR #9014, opened 2019-10-03
Add dithering

HybridDog:
Goal:
Add dithering to reduce banding artifacts

How does it work: 
Dithering is disabled by default and can be enabled with a setting.
It is applied in the second-stage shader and uses the algorithm supplied by Calinou.

> Does it resolve any reported issue?

I don't think there's an issue about this.

> If not a bug fix, why is this PR needed? What usecases does it solve?

On some displays, [banding](https://en.wikipedia.org/wiki/Colour_banding) is easily noticeable in Minetest.
Dithering before the quantisation to 8 bits can prevent banding artifacts
and an implementation requires only a few lines of code.

## How to test

It may be difficult to see banding on some screens because 8x8x8 bit colours are already a lot possible values.

Disabled dithering (current behaviour):
dithering_off
Edited with GIMP to show the bands:
dithering_off_edited

Enabled dithering:
dithering_on
Edited with GIMP to show the dithering pattern:
dithering_on_edited

Please look at the pictures in 1:1 scaling or upscaled by an integer factor without interpolation, and ideally with a black background instead of the white Github page background.

### DIFF
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -585,11 +585,13 @@ exposure_compensation (Exposure compensation) float 0.0 -1.0 1.0
 #    screenshots and it works incorrectly if the display or operating system
 #    performs additional dithering or if the color channels are not quantized
 #    to 8 bits.
+#    If set to glsl_only, dithering is disabled only for OpenGL ES.
 #    With OpenGL ES, dithering only works if the shader supports high
-#    floating-point precision.
+#    floating-point precision and it may have a higher performance impact.
 #
 #    Requires: shaders
-debanding (Enable Debanding) bool true
+debanding (Debanding) enum glsl_only disable,enable,glsl_only
+
 
 [**Bloom]
 
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -580,6 +580,15 @@ enable_auto_exposure (Enable Automatic Exposure) bool false
 #    Requires: shaders, enable_auto_exposure
 exposure_compensation (Exposure compensation) float 0.0 -1.0 1.0
 
+#    Apply dithering to reduce colour banding artifacts.
+#    Dithering significantly increases the size of losslessly-compressed
+#    screenshots and it works incorrectly if the display or operating system
+#    performs additional dithering or if the colour channels are not quantized
+#    to 8 bits. Therefore it is disabled by default.
+#
+#    Requires: shaders
+dithering (Enable Dithering) bool false
+
 [**Bloom]
 
 #    Set to true to enable bloom effect.
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -584,10 +584,10 @@ exposure_compensation (Exposure compensation) float 0.0 -1.0 1.0
 #    Dithering significantly increases the size of losslessly-compressed
 #    screenshots and it works incorrectly if the display or operating system
 #    performs additional dithering or if the colour channels are not quantized
-#    to 8 bits. Therefore it is disabled by default.
+#    to 8 bits.
 #
 #    Requires: shaders
-dithering (Enable Dithering) bool false
+debanding (Enable Debanding) bool true
 
 [**Bloom]
 
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -585,13 +585,11 @@ exposure_compensation (Exposure compensation) float 0.0 -1.0 1.0
 #    screenshots and it works incorrectly if the display or operating system
 #    performs additional dithering or if the color channels are not quantized
 #    to 8 bits.
-#    If set to glsl_only, dithering is disabled only for OpenGL ES.
 #    With OpenGL ES, dithering only works if the shader supports high
 #    floating-point precision and it may have a higher performance impact.
 #
 #    Requires: shaders
-debanding (Debanding) enum glsl_only disable,enable,glsl_only
-
+debanding (Enable Debanding) bool true
 
 [**Bloom]
 
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -580,10 +580,10 @@ enable_auto_exposure (Enable Automatic Exposure) bool false
 #    Requires: shaders, enable_auto_exposure
 exposure_compensation (Exposure compensation) float 0.0 -1.0 1.0
 
-#    Apply dithering to reduce colour banding artifacts.
+#    Apply dithering to reduce color banding artifacts.
 #    Dithering significantly increases the size of losslessly-compressed
 #    screenshots and it works incorrectly if the display or operating system
-#    performs additional dithering or if the colour channels are not quantized
+#    performs additional dithering or if the color channels are not quantized
 #    to 8 bits.
 #
 #    Requires: shaders
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -585,6 +585,8 @@ exposure_compensation (Exposure compensation) float 0.0 -1.0 1.0
 #    screenshots and it works incorrectly if the display or operating system
 #    performs additional dithering or if the color channels are not quantized
 #    to 8 bits.
+#    With OpenGL ES, dithering only works if the shader supports high
+#    floating-point precision.
 #
 #    Requires: shaders
 debanding (Enable Debanding) bool true
--- a/client/shaders/second_stage/opengl_fragment.glsl
+++ b/client/shaders/second_stage/opengl_fragment.glsl
@@ -79,6 +79,20 @@ vec3 applySaturation(vec3 color, float factor)
 }
 #endif
 
+#ifdef ENABLE_DITHERING
+// From http://alex.vlachos.com/graphics/Alex_Vlachos_Advanced_VR_Rendering_GDC2015.pdf
+// and https://www.shadertoy.com/view/MslGR8 (5th one starting from the bottom)
+// NOTE: `frag_coord` is in pixels (i.e. not normalized UV).
+vec3 screen_space_dither(vec2 frag_coord) {
+	// Iestyn's RGB dither (7 asm instructions) from Portal 2 X360, slightly modified for VR.
+	vec3 dither = vec3(dot(vec2(171.0, 231.0), frag_coord));
+	dither.rgb = fract(dither.rgb / vec3(103.0, 71.0, 97.0));
+
+	// Subtract 0.5 to avoid slightly brightening the whole viewport.
+	return (dither.rgb - 0.5) / 255.0;
+}
+#endif
+
 void main(void)
 {
 	vec2 uv = varTexCoord.st;
@@ -124,6 +138,11 @@ void main(void)
 
 	// return to sRGB colorspace (approximate)
 	color.rgb = pow(color.rgb, vec3(1.0 / 2.2));
+
+#ifdef ENABLE_DITHERING
+	// Apply dithering just before quantisation
+	color.rgb += screen_space_dither(gl_FragCoord.xy);
+#endif
 
 	gl_FragColor = vec4(color.rgb, 1.0); // force full alpha to avoid holes in the image.
 }
--- a/client/shaders/second_stage/opengl_fragment.glsl
+++ b/client/shaders/second_stage/opengl_fragment.glsl
@@ -1,6 +1,13 @@
 #define rendered texture0
 #define bloom texture1
 
+#ifdef GL_ES
+// Dithering requires sufficient floating-point precision
+#if GL_FRAGMENT_PRECISION_HIGH != 1
+#undef ENABLE_DITHERING
+#endif
+#endif
+
 struct ExposureParams {
 	float compensationFactor;
 };
@@ -83,9 +90,9 @@ vec3 applySaturation(vec3 color, float factor)
 // From http://alex.vlachos.com/graphics/Alex_Vlachos_Advanced_VR_Rendering_GDC2015.pdf
 // and https://www.shadertoy.com/view/MslGR8 (5th one starting from the bottom)
 // NOTE: `frag_coord` is in pixels (i.e. not normalized UV).
-vec3 screen_space_dither(vec2 frag_coord) {
+vec3 screen_space_dither(highp vec2 frag_coord) {
 	// Iestyn's RGB dither (7 asm instructions) from Portal 2 X360, slightly modified for VR.
-	vec3 dither = vec3(dot(vec2(171.0, 231.0), frag_coord));
+	highp vec3 dither = vec3(dot(vec2(171.0, 231.0), frag_coord));
 	dither.rgb = fract(dither.rgb / vec3(103.0, 71.0, 97.0));
 
 	// Subtract 0.5 to avoid slightly brightening the whole viewport.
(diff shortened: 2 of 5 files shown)


sfan5:
This is just a static dither pattern so I wouldn't call it debanding (though it does have that effect).

HybridDog:
> This is just a static dither pattern so I wouldn't call it debanding (though it does have that effect).

The dithering is caused by rounding after the values are changed in a chessboard pattern. Do you have a suggestion on how I could call the setting?

> The effect is also very subtle, so I'm not sure if this PR is worth it.

In my experience, banding is easy to see when the screen brightness is very high and you look at a stone node in a dim cave.
The dithering is optional and fairly simply (not very much code).

Calinou:
I don't remember seeing banding in Minetest, but the implementation looks simple enough. Out of curiosity, what's the performance impact like?

HybridDog:
> Out of curiosity, what's the performance impact like?

I don't know. How can I measure it?

Calinou:
@HybridDog Measure the FPS in a complex scene before and after enabling it :slightly_smiling_face:

paramat:
I am not convinced it is worth adding, seeing banding is apparently rare (i never have) and when you do see it is probably very subtle and not a problem.
:-1:

HybridDog:
> Measure the FPS in a complex scene before and after enabling it slightly_smiling_face

I've tested it and didn't notice any difference. It was about 40 FPS with and without the dithering.

paramat:
But it also adds to review, testing, bugfixing and maintenance workload, and to code complexity, for near-zero benefit. I think this is what sfan5 means by 'worth it'.

HybridDog:
> But it also adds to review, testing, bugfixing and maintenance workload, and to code complexity, for near-zero benefit. 

The code for this is very simple, so I think there's not much to review and maintenance, and if there's a bug, you can simply disable the setting or remove the feature. Testing it also shouldn't be difficult.

benrob0329:
My issue is that if it significantly increases screenshot size, then it probably also increases recording file size. So I'll never use this, and more than likely neither will other content creators because recording MT already produces huge files when recording.

HybridDog:
> My issue is that if it significantly increases screenshot size, then it probably also increases video file size.

It should only affect lossless compression.

> Then, if it's also disabled by default, and a large portion of our user-base either don't look at settings, or disable shaders anyways, why add it in the first place?

It is just a small feature, so there's not much to be added. 
Since the shaders path can be configured, it is alternatively possible to add the dithering to a custom shader fragment code.

Calinou:
If anyone is interested in reviving this PR, note that you can use a more effective algorithm which is still pretty cheap: https://github.com/godotengine/godot/pull/42942

```glsl
// From http://alex.vlachos.com/graphics/Alex_Vlachos_Advanced_VR_Rendering_GDC2015.pdf
// and https://www.shadertoy.com/view/MslGR8 (5th one starting from the bottom)
// NOTE: `frag_coord` is in pixels (i.e. not normalized UV).
vec3 screen_space_dither(vec2 frag_coord) {
	// Iestyn's RGB dither (7 asm instructions) from Portal 2 X360, slightly modified for VR.
	vec3 dither = vec3(dot(vec2(171.0, 231.0), frag_coord));
	dither.rgb = fract(dither.rgb / vec3(103.0, 71.0, 97.0));

	// Subtract 0.5 to avoid slightly brightening the whole viewport.
	return (dither.rgb - 0.5) / 255.0;
}

// ...

#ifdef USE_DEBANDING
	// For best results, debanding should be done before tonemapping.
	// Otherwise, we're adding noise to an already-quantized image.
	color += screen_space_dither(gl_FragCoord.xy);
#endif
```

HybridDog:
I have reopened this and applied Calinou's suggestion, which works a lot better than a simple checkerboard pattern.

Calinou:
Tested locally, it works as expected.

debanding webp

Cropped section of the image with **Colors > Auto > Equalize** in GIMP and 3× nearest-neighbor scaling:

debanding_crop webp

It may be worth enabling this setting by default, as it'll help a lot once https://github.com/minetest/minetest/pull/13881 are implemented (you can notice banding on the screenshots I posted there).

HybridDog:
> It may be worth enabling this setting by default, as it'll help a lot once #13881 are implemented (you can notice banding on the screenshots I posted there).

Done. I assume that on OLED screens, which have almost infinite contrast, banding may be very visible in Minetest when playing in a dark cave, but I haven't tested it.

Dithering may also reduce hypothetical subtle screen flickering caused by rounding and numerical precision limits, for example a background colour alternating between #B9EBE9 and #B9ECE9. A similar but, as far as I know, stronger flickering happened in the past (#8416).

> Also, I suggest renaming the setting to debanding as dithering may be confused for a setting that intentionally makes the game look worse (e.g. palletization shaders that aim for a PS1-like look).

I've renamed it. I think debanding is a more generic term and it includes not only dithering but also, for example, algorithms which remove banding after the quantisation, such as [AdaDeband](https://arxiv.org/pdf/2009.10804.pdf).

Calinou:
> Done. I assume that on OLED screens, which have almost infinite contrast, banding may be very visible in Minetest when playing in a dark cave, but I haven't tested it.

I'm using a LG C2 42" which is an OLED display :slightly_smiling_face:

builtin/settingtypes.txt
```diff
@@ -580,6 +580,15 @@ enable_auto_exposure (Enable Automatic Exposure) bool false
 #    Requires: shaders, enable_auto_exposure
 exposure_compensation (Exposure compensation) float 0.0 -1.0 1.0
 
+#    Apply dithering to reduce colour banding artifacts.
+#    Dithering significantly increases the size of losslessly-compressed
+#    screenshots and it works incorrectly if the display or operating system
+#    performs additional dithering or if the colour channels are not quantized
+#    to 8 bits.
```
Desour, line 587: We use the american one.
```suggestion
#    Apply dithering to reduce color banding artifacts.
#    Dithering significantly increases the size of losslessly-compressed
#    screenshots and it works incorrectly if the display or operating system
#    performs additional dithering or if the color channels are not quantized
#    to 8 bits.
```

Desour:
Tested, works.
Left is without dithering, right is with:
screenshot_20231025_171321

Interestingly, this also in a way improves screenshots. This is how the screenshot looks like if I look at it with an image viewer that transforms the colors to my screen's color space (which is not srgb):
2023-10-25-173059_1920x1080_scrot

--------

Considering how simple this change is, I support it.

Code looks fine.

@Calinou Is it actually fine to use the code from those slides, copyright-wise? (IANAL)

Calinou:
> @Calinou Is it actually fine to use the code from those slides, copyright-wise? (IANAL)

The code is short enough and the formula is trivial enough, so yes.

grorp (commented):
According to Calinou's comment, debanding is best applied before tonemapping:

```
// For best results, debanding should be done before tonemapping.
// Otherwise, we're adding noise to an already-quantized image.
```

However, in this PR it is applied at the very end of the post-processing shader, even after the conversion to sRGB. Please either change or explain this.

HybridDog:
> According to Calinou's comment, debanding is best applied before tonemapping:
> 
> ```
> // For best results, debanding should be done before tonemapping.
> // Otherwise, we're adding noise to an already-quantized image.
> ```
> 
> However, in this PR it is applied at the very end of the post-processing shader, even after the conversion to sRGB. Please either change or explain this.

I don't understand the second line of the comment. As far as I understand that shader code, `apply_tonemapping` does not quantise (in other words, round) the values, so the dithering may be applied in the wrong place.

The dithering algorithm adds a value $x$ in $[\frac{-0.5}{255}, \frac{0.5}{255}]^3$ to the colour and the quantisation happens automatically after the shader is executed on the resulting output colour.

I'm not sure how to explain it easily.
If $x$ is added right before quantisation, the resulting colour is the next brighter or darker colour closest to the unquantized undithered colour.
If, on the other hand, $x$ is added earlier, for example before a gamma2.2 EOTF, the resulting colour is far away from the unquantized undithered colour if it is dark, and dithering does not work for for bright unquantized undithered colours.

Dithering applied early, before the EOTF:
dither_before_eotf
dither_before_eotf_brightened



Dithering applied right before quantisation:
dither_before_quantisation
dither_before_quantisation_brightened

Calinou:
The comment may have been relevant in Godot's context (which is what I originally wrote it for), but perhaps it means nothing in the Minetest shaders.

HybridDog:
I've found a webpage about average brightness preservation when dithering in sRGB: http://www.thetenthplanet.de/archives/5367
I don't implement that because it is worse for performance and probably has no noticeable advantage in Minetest.

Desour (approved):
LGTM, and works.

grorp (commented):
Thank you for the explanation in https://github.com/minetest/minetest/pull/9014#issuecomment-1781586509.

I don't usually notice banding while playing Minetest, but if I look for it, it is very obvious. This PR fixes the obvious cases of banding. In a direct before/after comparison, it also improves the image quality in scenes where I didn't notice any banding.

-------------------

The debanding doesn't work on my Android phone ("ogles2" Irrlicht video driver). The reason seems to be a lack of precision: it works if I add `precision highp float;` at the beginning of the shader. For testing, I took two screenshots with 50x dithering applied:

Screenshot of broken 50x dithering (this PR)
Screenshot of broken 50x dithering

Screenshot of working 50x dithering (this PR with `precision highp float;`)
Screenshot of working 50x dithering

I see two solutions:

1. Disable debanding by default on Android.

    This might also be a good idea because Godot had a problem with this debanding code on mobile, which apparently still hasn't been solved:

    godotengine/godot#49989
    godotengine/godot#56443
    godotengine/godot#45366

2. On OpenGL ES, use `precision highp float;` if available (`#ifdef GL_FRAGMENT_PRECISION_HIGH`). Disable debanding otherwise.

Calinou:
> This might also be a good idea because Godot had a problem with this debanding code on mobile, which apparently still hasn't been solved:

We could probably add `highp` where required to fix it. This means debanding will run slower, but it's better to have it run slower than to be broken.

I'm not aware of a high-quality `mediump`-friendly debanding shader.

HybridDog:
Thanks for testing. 
I have added the check for `GL_FRAGMENT_PRECISION_HIGH` and `highp` qualifiers but I don't know if it works.

> I'm not aware of a high-quality mediump-friendly debanding shader.

On the lower left corner the dithering doesn't look broken (there are slight differences due to camera movement):
lower_left_borders

The reason is probably that `gl_FragCoord` is smaller there.
It may be possible to make the dithering work with low floating-point precision by adding modulo operations, e.g.
```glsl
	frag_coord = vec2(mod(frag_coord.x, 33.0), mod(frag_coord.y, 33.0));
```
However, I don't know which effect this has on performance and quality.

client/shaders/second_stage/opengl_fragment.glsl
```diff
@@ -1,6 +1,13 @@
 #define rendered texture0
 #define bloom texture1
 
+#ifdef GL_ES
+// Dithering requires sufficient floating-point precision
+#if GL_FRAGMENT_PRECISION_HIGH != 1
```
grorp, line 6: > The built-in macro GL_FRAGMENT_PRECISION_HIGH [...] is not defined on systems not supporting highp precision in the fragment language.
> https://registry.khronos.org/OpenGL/specs/es/2.0/GLSL_ES_Specification_1.00.pdf

Since I have no device which doesn't support `GL_FRAGMENT_PRECISION_HIGH`, I tested this by replacing `GL_FRAGMENT_PRECISION_HIGH` with `MY_UNDEFINED_MACRO`. And apparently such use of an undefined macro results in an error:

```
ERROR[Main]: Irrlicht: GLSL shader failed to compile
ERROR[Main]: Irrlicht: ERROR: 0:12: '' :     GLSL compile error:  unresolved identifier
ERROR: 0:12: '' :     GLSL compile error:  unexpected tokens following an #if or #elif.
ERROR: 2 compilation errors.  No code generated.
ERROR[Main]: generate_shader(): failed to generate "second_stage", addHighLevelShaderMaterial failed.
```

So I think we have to use this instead:

```suggestion
#ifndef GL_FRAGMENT_PRECISION_HIGH
```

minetest.conf.example
```diff
@@ -508,6 +508,15 @@
...
 
+#    Apply dithering to reduce colour banding artifacts.
+#    Dithering significantly increases the size of losslessly-compressed
+#    screenshots and it works incorrectly if the display or operating system
+#    performs additional dithering or if the colour channels are not quantized
+#    to 8 bits.
+#    type: bool
+# debanding = true
+
+
```
grorp, line 519: Please don't update `minetest.conf.example`, it is regenerated automatically.

grorp (commented):
Thanks for the changes. The debanding now works on my Android device (and it also still works on my desktop computer).

> The reason is probably that `gl_FragCoord` is smaller there. It may be possible to make the dithering work with low floating-point precision by adding modulo operations, e.g.
> 
> ```glsl
> 	frag_coord = vec2(mod(frag_coord.x, 33.0), mod(frag_coord.y, 33.0));
> ```
> 
> However, I don't know which effect this has on performance and quality.

That doesn't seem to be worth it.

> We can still choose debanding to be disabled by default on mobile for performance reasons.

I am in favor of that.

client/shaders/second_stage/opengl_fragment.glsl
HybridDog: I agree. I've overlooked this.

HybridDog:
I have changed the setting now so that dithering can be enabled on OpenGL only (and not OpenGL ES).

grorp:
> I have changed the setting now so that dithering can be enabled on OpenGL only (and not OpenGL ES).

Why is the enum value called `glsl_only`? Shouldn't it be `opengl_only`?

Anyway, I'd prefer if the setting stayed a boolean and you just added `settings->setDefault("debanding", "false");` to the "Altered settings for Android" section of `defaultsettings.cpp`.

HybridDog:
> Why is the enum value called glsl_only? Shouldn't it be opengl_only?

If I remember correctly, I mistakenly thought that OpenGL is a broad term which includes OpenGL ES, whereas GLSL is not a broad term and thus does not include the OpenGL ES Shading Language. I think `opengl_only` sounds better.

> Anyway, I'd prefer if the setting stayed a boolean and you just added settings->setDefault("debanding", "false"); to the "Altered settings for Android" section of defaultsettings.cpp (reasons: 1. simplicity, 2. consistency with other graphics settings).

I also prefer a simple bool setting but I think it's not possible to set a different default value in settingtypes.txt on Android, so if I understand it correctly, either the desktop or the Android settings menu would show the wrong default value. 
I have changed it back to bool nonetheless and added the line to defaultsettings.cpp since other settings also only have the desktop default in settingtypes.txt.

grorp:
> so if I understand it correctly, either the desktop or the Android settings menu would show the wrong default value.

No, luckily the settings menu (implicitly) gets its default values from `defaultsettings.cpp` and only uses `settingtypes.txt` as a fallback. (This is because `Settings:get` also returns default values from `defaultsettings.cpp`.)

HybridDog:
No, I haven't looked at the settings code thoroughly and accidentally assumed that it would show the default value from `settingtypes.txt`.

grorp (approved):
Thanks for your incredible patience!

PR #9015, opened 2019-10-03
[NO SQUASH] Translation update

SmallJoker:
Final update before 5.1.0 (hopefully).


sfan5:
IRC discussion about what's wrong with the Weblate import: http://irc.minetest.net/minetest-dev/2019-10-03#i_5599128

SmallJoker:
Waiting for response of est31 and ShadowNinja who are both online, but did not answer yet.
Alternatively the new translation files could be downloaded via the interface which seems to provide an up to date file, but that's not the goal.

PR #9018, opened 2019-10-05
Revert "Fix the bgcolor formspec element (#8716)"

SmallJoker:
This reverts commit 1db3d252cff9e8d61fecf1052d7497813851da51.
Temporary solution until a compatible solution is found to define both - formspec and fullscreen backgrounds.

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2095,8 +2095,8 @@ Elements
 ### `bgcolor[<color>;<fullscreen>]`
 
 * Sets background color of formspec as `ColorString`
-* If `true`, the background color is drawn fullscreen (does not affect the size
-  of the formspec).
+* If `true`, a fullscreen background is drawn and the color is ignored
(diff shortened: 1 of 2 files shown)


doc/lua_api.txt
```diff
@@ -2092,15 +2092,11 @@ Elements
...
-
-* Sets background color of formspec as `ColorString`
-
 ### `bgcolor[<color>;<fullscreen>]`
 
-* If `color` is a valid `ColorString`, the fullscreen background color
-  is set to `color`.
-* If `fullscreen` is a true value, the fullscreen background color is drawn.
+* Sets background color of formspec as `ColorString`
+* If `true`, the background color is drawn fullscreen (does not affect the size
```
sfan5: one thing though: The docs here should be corrected
`* If `true`, a fullscreen background is drawn and the color is ignored (does not affect the size of the formspec)`

PR #6559, opened 2017-10-26
Set vertex color on upright sprites.

basicer:
This sets the vertex color on upright sprites so that they respect the colors object property similar to model.

PR #9021, opened 2019-10-06
Fix upright_sprite lighting when colors are set

### LINKED ISSUE #9020: upright_sprite: lighting is reset when texture changes
##### Minetest version
```
5.1.0-dev
```
@ https://github.com/minetest/minetest/commit/dde062897657d43161b33739a79481cbd6b0baad

##### OS / Hardware
Operating system: Arch Linux
CPU: Intel(R) Core(TM) i5-3210M CPU @ 2.50GHz

GPU model: Intel HD 4000
OpenGL version: 4.2

##### Summary
I've been experiencing a lighting bug/regression with MT 5.0, where the texture of an entity using `visual = "upright_sprite"` is always being rendered at full brightness after it has been changed/updated. Lighting is only applied correctly again after the light level in the entity's surroundings have changed. 

I first noticed this behavior with mods like https://github.com/pyrollo/display_modpack and https://github.com/minetest-mods/painting, which use an entity to display a dynamically generated texture in front of a node. 

After some bisecting I found  https://github.com/minetest/minetest/commit/787cd15c14b65f50b06c6b9ccecf28e937318430 to be responsible for this. Before this commit, everything worked as expected (at least for me).

##### Steps to reproduce
Update the texture of any upright_sprite in a dark environment. It will be rendered at full brightness, regardless of the surrounding light level.

An easy way to test this with MTG is to use player_api to set the player model to the default upright_sprite periodically, eg:

`minetest.register_globalstep(function()
	for _, p in ipairs(minetest.get_connected_players()) do
		player_api.set_model(p, nil)
	end
end)`

Then run around in third person view at night.

sfan5:
fixes #9020, bug was introduced in #6559
I don't know <i>[why](https://github.com/minetest/minetest_game/blob/master/mods/player_api/api.lua#L54-L59)</i> the player has a color set, but it doesn't make sense to apply it to the mesh either way since it conflicts with lighting.

## To do

This PR is Ready for Review.
## How to test

see linked issue

### DIFF
--- a/src/client/content_cao.cpp
+++ b/src/client/content_cao.cpp
@@ -1264,18 +1264,19 @@ void GenericCAO::updateTextures(std::string mod)
 					buf->getMaterial().AmbientColor = m_prop.colors[1];
 					buf->getMaterial().DiffuseColor = m_prop.colors[1];
 					buf->getMaterial().SpecularColor = m_prop.colors[1];
-					setMeshColor(mesh, m_prop.colors[1]);
 				} else if (!m_prop.colors.empty()) {
 					buf->getMaterial().AmbientColor = m_prop.colors[0];
 					buf->getMaterial().DiffuseColor = m_prop.colors[0];
 					buf->getMaterial().SpecularColor = m_prop.colors[0];
-					setMeshColor(mesh, m_prop.colors[0]);
 				}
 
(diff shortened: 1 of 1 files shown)


tuedel:
Just tested, this fixes the issue for me. :+1: 

Thanks!

PR #9025, opened 2019-10-07
Revert Client::sendPlayerPos optimization (part of 81c7f0a)

### LINKED ISSUE #9022: Using 3d_armor mod with "Enable Damage" inactive makes the world mostly refuse to load
##### Minetest version

```
I did a bisect and found the first commit where this bug appeared:

81c7f0a ("Send ActiveObjects once right after Init2", 2019-08-25)

```

##### OS / Hardware

Operating system: Ubuntu 18.04
CPU: AMD x86 64

GPU model:
OpenGL version:

##### Summary

If you switch to Damage not enabled, the world will mostly refuse to load.

##### Steps to reproduce

Create a new world with MTG and then enable "3d_armor" mod. Uncheck the "Enable Damage" option. "Play Game" and notice how it mostly doesn't load.

ClobberXD:
While #9022 is apparently a mod issue, I think we should consider this optimisation carefully, due to the potential breakage in mods. Even if an affected mod is actively maintained, this optimisation would force users and servers to update the mod. So this PR comments out the `if` statement that prevents sending `TOSERVER_PLAYERPOS` if the player is dead, as this causes a breakage in `3D_Armor`. We'll look at this again, once 5.1.0 has been released.

Tested, works. Fixes #9022. To test, follow the testing instructions in #9022.

### DIFF
--- a/src/client/client.cpp
+++ b/src/client/client.cpp
@@ -1245,8 +1245,9 @@ void Client::sendPlayerPos()
 	// Save bandwidth by only updating position when
 	// player is not dead and something changed
 
-	if (m_activeobjects_received && player->isDead())
-		return;
+	// FIXME: This part causes breakages in mods like 3d_armor, and has been commented for now
+	// if (m_activeobjects_received && player->isDead())
+	//	return;
 
 	if (
 			player->last_position     == player->getPosition() &&


style-nine:
Works for me.

ClobberXD:
Awesome, thanks.

Enrikoo:
While I was compiling this commit, I got those errors:
https://pasteboard.co/IASs8f9.png
How can I fix it?

lhofhansl:
I'll try a bit later today, but looks good.

lhofhansl:
I can confirm that this fixes the problem I have seen!

PR #9029, opened 2019-10-08
Correction for alpha blending issues in texture mod compositing

Warr1024:
Minetest's texture modifier operators (e.g. `^` and `[combine`) are used to overlay images atop one another, but they handle the alpha component incorrectly.  This leads to visuals being noticeably incorrect when either alpha blending is enabled on a surface using the combined texture, or it's then combined with other textures.

The test case below combines 2 obvious circles (an opaque white one and a translucent black one).  The left one is merged using gimp and saved as a single pre-composited texture, while the right one is merged via in-game ^ texture combining.  Each is drawn as an upright_sprite entity in front of the checkered node background.

### Before (MT master):
broken

### After (with this patch):
fixed

Remaining discrepancies between the gimp reference image and MT's rendering are subtle, and probably due to perceptual/nonlinear magic that gimp does that is probably beyond what's necessary for MT.

### DIFF
--- a/src/client/tile.cpp
+++ b/src/client/tile.cpp
@@ -1812,6 +1812,23 @@ bool TextureSource::generateImagePart(std::string part_of_name,
 }
 
 /*
+	Calculate the color of a single pixel drawn on top of another pixel.
+
+	This is a little more complicated than just video::SColor::getInterpolated
+	because getInterpolated does not handle alpha correctly.  For example, a
+	pixel with alpha=64 drawn atop a pixel with alpha=128 should yield a
+	pixel with alpha=160, while getInterpolated would yield alpha=96.
+*/
+static inline video::SColor blitPixel(const video::SColor &src_c, const video::SColor &dst_c, u32 ratio)
+{
+	if (dst_c.getAlpha() == 0)
+		return src_c;
+	video::SColor out_c = src_c.getInterpolated(dst_c, (float)ratio / 255.0f);
+	out_c.setAlpha(dst_c.getAlpha() + (255 - dst_c.getAlpha()) * src_c.getAlpha() / 255 * ratio / 255);
+	return out_c;
+}
+
+/*
 	Draw an image on top of an another one, using the alpha channel of the
 	source image
 
@@ -1830,7 +1847,7 @@ static void blit_with_alpha(video::IImage *src, video::IImage *dst,
 		s32 dst_y = dst_pos.Y + y0;
 		video::SColor src_c = src->getPixel(src_x, src_y);
 		video::SColor dst_c = dst->getPixel(dst_x, dst_y);
-		dst_c = src_c.getInterpolated(dst_c, (float)src_c.getAlpha()/255.0f);
+		dst_c = blitPixel(src_c, dst_c, src_c.getAlpha());
 		dst->setPixel(dst_x, dst_y, dst_c);
 	}
 }
@@ -1853,7 +1870,7 @@ static void blit_with_alpha_overlay(video::IImage *src, video::IImage *dst,
 		video::SColor dst_c = dst->getPixel(dst_x, dst_y);
 		if (dst_c.getAlpha() == 255 && src_c.getAlpha() != 0)
 		{
-			dst_c = src_c.getInterpolated(dst_c, (float)src_c.getAlpha()/255.0f);
+			dst_c = blitPixel(src_c, dst_c, src_c.getAlpha());
 			dst->setPixel(dst_x, dst_y, dst_c);
 		}
 	}
--- a/src/client/tile.cpp
+++ b/src/client/tile.cpp
@@ -1824,7 +1824,8 @@ static inline video::SColor blitPixel(const video::SColor &src_c, const video::S
 	if (dst_c.getAlpha() == 0)
 		return src_c;
 	video::SColor out_c = src_c.getInterpolated(dst_c, (float)ratio / 255.0f);
-	out_c.setAlpha(dst_c.getAlpha() + (255 - dst_c.getAlpha()) * src_c.getAlpha() * ratio / (255 * 255));
+	out_c.setAlpha(dst_c.getAlpha() + (255 - dst_c.getAlpha()) *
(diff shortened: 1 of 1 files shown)


src/client/tile.cpp
```diff
@@ -1811,6 +1811,23 @@ bool TextureSource::generateImagePart(std::string part_of_name,
...
 
+/*
+	Calculate the color of a single pixel drawn on top of another pixel.
+
+	This is a little more complicated than just video::SColor::getInterpolated
+	because getInterpolated does not handle alpha correctly.  For example, a
+	pixel with alpha=64 drawn atop a pixel with alpha=128 should yield a
+	pixel with alpha=160, while getInterpolated would yield alpha=96.
+*/
+static video::SColor blit_pixel(video::SColor src_c, video::SColor dst_c, u32 ratio)
```
sfan5: should be `inline` and pass the SColor arguments by reference

src/client/tile.cpp
```diff
@@ -1811,6 +1811,23 @@ bool TextureSource::generateImagePart(std::string part_of_name,
...
+{
+	if(dst_c.getAlpha() == 0)
```
ClobberXD:
```suggestion
	if (dst_c.getAlpha() == 0)
```

src/client/tile.cpp
```diff
@@ -1811,6 +1811,23 @@ bool TextureSource::generateImagePart(std::string part_of_name,
...
+		return src_c;
+	video::SColor out_c = src_c.getInterpolated(dst_c, (float)ratio/255.0f);
```
ClobberXD:
```suggestion
	video::SColor out_c = src_c.getInterpolated(dst_c, (float)ratio / 255.0f);
```

src/client/tile.cpp
```diff
@@ -1811,6 +1811,23 @@ bool TextureSource::generateImagePart(std::string part_of_name,
...
 
+/*
+	Calculate the color of a single pixel drawn on top of another pixel.
+
+	This is a little more complicated than just video::SColor::getInterpolated
+	because getInterpolated does not handle alpha correctly.  For example, a
+	pixel with alpha=64 drawn atop a pixel with alpha=128 should yield a
+	pixel with alpha=160, while getInterpolated would yield alpha=96.
+*/
+static video::SColor blit_pixel(const video::SColor &src_c, const video::SColor &dst_c, u32 ratio)
```
ClobberXD: Minetest (largely) follows the `camelCase` naming convention for functions. It'd be nice if this could is renamed to smth. like `blitPixel` instead.

ClobberXD: I'm aware that the other related functions in this file use the underscore convention, but I think it's better if we don't continue this style and use the correct naming convention (`camelCase`, in this case) instead.

src/client/tile.cpp
```diff
@@ -1921,7 +1938,7 @@ static void apply_colorize(video::IImage *dst, v2u32 dst_pos, v2u32 size,
 		for (u32 x = dst_pos.X; x < dst_pos.X + size.X; x++) {
 			dst_c = dst->getPixel(x, y);
 			if (dst_c.getAlpha() > 0) {
-				dst_c = color.getInterpolated(dst_c, interp);
+				dst_c = blit_pixel(color, dst_c, interp);
```
sfan5: this is incorrect, blit_pixel will divide interp by / 255 another time

src/client/tile.cpp
sfan5: still no `inline` keyword

src/client/tile.cpp
Warr1024: I ended up backing out the getInterpolated -> blitPixel change in the `colorize` and the dead-code `blit_with_interpolate_overlay` areas, since it looks like the documentation for them suggests that getInterpolated was actually the more correct behavior.

I'm actually a little confused about the `blit_with_alpha_overlay` behavior.  It looks like it's only used to draw the crack texture, and it only applies the texture to pixels with alpha = 255.  This makes the behavior obviously correct for fully transparent or fully opaque pixels, but what it should do for middle-alpha pixels is ill-defined.

src/client/tile.cpp
SmallJoker: FYI: SColor is `u32`, hence references are OK but not really needed.

src/client/tile.cpp
```diff
@@ -1811,6 +1811,23 @@ bool TextureSource::generateImagePart(std::string part_of_name,
...
+	because getInterpolated does not handle alpha correctly.  For example, a
+	pixel with alpha=64 drawn atop a pixel with alpha=128 should yield a
+	pixel with alpha=160, while getInterpolated would yield alpha=96.
+*/
+static inline video::SColor blitPixel(const video::SColor &src_c, const video::SColor &dst_c, u32 ratio)
+{
+	if (dst_c.getAlpha() == 0)
+		return src_c;
+	video::SColor out_c = src_c.getInterpolated(dst_c, (float)ratio / 255.0f);
+	out_c.setAlpha(dst_c.getAlpha() + (255 - dst_c.getAlpha()) * src_c.getAlpha() / 255 * ratio / 255);
```
SmallJoker: Are you sure that you want to multiply and divide by entire integers? Especially on low values this gets inaccurate way too fast.

SmallJoker: That, or convert it all into floats. i.e. `(float)ratio / 255.0f`

Warr1024: Is there any specific case where doing the math in floats yields a different answer than just doing the multiplications first before the division?  Upcasting and then downcasting multiple values for each pixel seems like overkill to achieve the same final answer, and probably not great to do inside an inner loop like this.

SmallJoker: For this case your approach of multiplying before dividing is better. Whereas fixed/floating point casting takes negligible time, the decimals will be lost anyway.
This should work just fine:
```C++
out_c.setAlpha(dst_c.getAlpha() + (255 - dst_c.getAlpha()) * src_c.getAlpha() * ratio / (255 * 255));
```

PR #9059, opened 2019-10-21
Remove legacy flat-file map code

### LINKED ISSUE #9051: Abandon file based maps
I open this issue to discuss about the merit of keeping old code for opening file based maps.

In map.cpp, there is still a fallback code for searching mapblocks in files if they are not found:
https://github.com/minetest/minetest/blob/master/src/map.cpp#L2175

Is this still really needed ? 

I'm working on map.cpp for Kidscode, trying to improve map management, adding possibility of threading. I'm first doing some simplifications.

Before proposing any PR I'd like to have some advice about this old code.

random-geek:
Closes #9051.
I wasn't sure what to do with `world_format.txt`, so further changes may be needed.


doc/world_format.txt
```diff
@@ -569,9 +563,9 @@ EndInventoryList
 EndInventory
 ---
 
-==============================================
-Minetest World Format used as of 2011-05 or so
-==============================================
+===========================================================
```
sfan5: You can just remove everything below here IMO.

PR #9073, opened 2019-10-26
(#3075) Hackish but simple fix for the ~5yo waving water bug

### LINKED ISSUE #3075: Interruption of water by waving water shaderㅤ

rexim:
Resolves #3075

1. Extract the code that determines the tile index by the face direction so it can be reused in other places. Specifically in the `face_contents` function which determines which face of the adjacent nodes to hide based on their real and visual solidness.

2. Extend `face_contents` to accept tile index which determines the direction of the face and force showing the top liquid face when the waving water is enabled and the face has a more solid block on top of it.

Not the best solution but it kinda works. Please let me know what you think about it. Also any ideas on how to improve the solution are really appreciated.

## To do

## How to test

See Steps to Reproduce in https://github.com/minetest/minetest/issues/3075#issuecomment-545681201

### DIFF
--- a/src/client/mapblock_mesh.cpp
+++ b/src/client/mapblock_mesh.cpp
@@ -664,7 +664,7 @@ static u8 face_contents(content_t m1, content_t m2, bool *equivalent,
 	const NodeDefManager *ndef, u16 tid1,
 	int s1, int s2)
 {
-	static thread_local const float enable_waving_water =
+	static thread_local const bool enable_waving_water =
 		g_settings->getBool("enable_waving_water");
 
 	*equivalent = false;
--- a/src/client/mapblock_mesh.cpp
+++ b/src/client/mapblock_mesh.cpp
@@ -661,7 +661,8 @@ static void makeFastFace(const TileSpec &tile, u16 li0, u16 li1, u16 li2, u16 li
 	TODO: Add 3: Both faces drawn with backface culling, remove equivalent
 */
 static u8 face_contents(content_t m1, content_t m2, bool *equivalent,
-	const NodeDefManager *ndef, u16 tid1)
+	const NodeDefManager *ndef, u16 tid1,
+	int s1, int s2)
 {
 	static thread_local const float enable_waving_water =
 		g_settings->getBool("enable_waving_water");
@@ -699,12 +700,12 @@ static u8 face_contents(content_t m1, content_t m2, bool *equivalent,
 	}
 
 	if (c1 > c2) {
-		if (enable_waving_water && f2.isLiquid() && tid1 == 1)
+		if (s1 == 0 && enable_waving_water && f2.isLiquid() && tid1 == 1)
 			return 2;
 		return 1;
 	}
 
-	if (enable_waving_water && f1.isLiquid() && tid1 == 0)
+	if (s2 == 0 && enable_waving_water && f1.isLiquid() && tid1 == 0)
 		return 1;
 	return 2;
 }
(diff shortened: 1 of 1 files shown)


src/client/mapblock_mesh.cpp
```diff
@@ -661,8 +661,11 @@ static void makeFastFace(const TileSpec &tile, u16 li0, u16 li1, u16 li2, u16 li
 	TODO: Add 3: Both faces drawn with backface culling, remove equivalent
 */
 static u8 face_contents(content_t m1, content_t m2, bool *equivalent,
-	const NodeDefManager *ndef)
+                        const NodeDefManager *ndef, u16 tid1)
```
SmallJoker: Don't listen to the code checked. The old indents were correct.

src/client/mapblock_mesh.cpp
```diff
@@ -661,8 +661,11 @@ static void makeFastFace(const TileSpec &tile, u16 li0, u16 li1, u16 li2, u16 li
...
 {
+    static thread_local const float enable_waving_water =
+        g_settings->getBool("enable_waving_water");
```
SmallJoker: Please use tabs to indent.

src/client/mapblock_mesh.cpp
```diff
@@ -695,9 +698,16 @@ static u8 face_contents(content_t m1, content_t m2, bool *equivalent,
 			return 2;
 	}
 
-	if (c1 > c2)
+	if (c1 > c2) {
+        if (enable_waving_water && f2.isLiquid() && tid1 == 1) {
+            return 2;
+        }
```
SmallJoker: > Do not unnecessarily use braces where a single statement will do.

https://www.kernel.org/doc/Documentation/process/coding-style.rst

src/client/mapblock_mesh.cpp
paramat: I agree.

src/client/mapblock_mesh.cpp
paramat: The 3 lines above seem to have weird indentation, only use tabs not spaces.

src/client/mapblock_mesh.cpp
```diff
@@ -695,9 +698,16 @@ static u8 face_contents(content_t m1, content_t m2, bool *equivalent,
...
 		return 1;
+    }
 
+    if (enable_waving_water && f1.isLiquid() && tid1 == 0) {
+        return 1;
+    }
 	return 2;
```
paramat: Weird indentation above.

src/client/mapblock_mesh.cpp
```diff
@@ -721,13 +731,43 @@ void getNodeTileN(MapNode mn, const v3s16 &p, u8 tileindex, MeshMakeData *data,
...
 
-/*
-	Gets node tile given a face direction.
-*/
-void getNodeTile(MapNode mn, const v3s16 &p, const v3s16 &dir, MeshMakeData *data, TileSpec &tile)
+static const u16 dir_to_tile[24 * 16] =
 {
-	const NodeDefManager *ndef = data->m_client->ndef();
+    // 0     +X    +Y    +Z           -Z    -Y    -X   ->   value=tile,rotation
+    0,0,  2,0 , 0,0 , 4,0 ,  0,0,  5,0 , 1,0 , 3,0 ,  // rotate around y+ 0 - 3
```
paramat: Indentation seems wrong, use 1 tab.

src/client/mapblock_mesh.cpp
```diff
@@ -743,46 +783,22 @@ void getNodeTile(MapNode mn, const v3s16 &p, const v3s16 &dir, MeshMakeData *dat
...
-		   0,0,  3,2 , 1,2 , 4,2 ,  0,0,  5,2 , 0,2 , 2,2 ,  // rotate around y- 20 - 23
-		   0,0,  5,2 , 1,3 , 3,2 ,  0,0,  2,2 , 0,1 , 4,2 ,
-		   0,0,  2,2 , 1,0 , 5,2 ,  0,0,  4,2 , 0,0 , 3,2 ,
-		   0,0,  4,2 , 1,1 , 2,2 ,  0,0,  3,2 , 0,3 , 5,2
-
-	};
-	u16 tile_index = facedir * 16 + dir_i;
-	getNodeTileN(mn, p, dir_to_tile[tile_index], data, tile);
-	tile.rotation = tile.world_aligned ? 0 : dir_to_tile[tile_index + 1];
+    const u16 tindex = tile_index(dir, facedir);
```
paramat: Indentation.

src/client/mapblock_mesh.cpp
```diff
@@ -820,7 +836,8 @@ static void getTileInfo(
 	// This is hackish
 	bool equivalent = false;
 	u8 mf = face_contents(n0.getContent(), n1.getContent(),
-			&equivalent, ndef);
+                          &equivalent, ndef,
+                          dir_to_tile[tile_index(face_dir, 0)]);
```
paramat: No need to indent so much, 1 or 2 tabs is enough.

paramat:
Note that roughly 90% of the 'Clang format' build check suggestions are very bad suggestions that make code unreadable.
Only check the report for valid suggestions, then, add the file to the clang format whitelist to silence the check https://github.com/minetest/minetest/blob/master/util/travis/clang-format-whitelist.txt

SmallJoker:
Same issue, but now underwater: (does not happen in master)
image

srifqi:
The bottom part of the node (which is above water) is missing. The bug pointed by @SmallJoker has been fixed, though.
bottom part is still missing

rexim:
Ok, I solved all of the mentioned problems, but I don't think this is a good solution and I'm not even sure what would be a good solution that doesn't make the code even worse.

Sorry, for wasting your time.

paramat:
No problem =) thanks for the contribution.

PR #9092, opened 2019-11-08
Some CSM changes

sfan5:
<b>DONT SQUASH</b>

* `Load client mods into memory before execution.`
  As the commit says, server-sent CSM will eventually need this since we will not want to write mod files to disk first.
* `Introduce get_modpath() for CSM`
  The "modname:path/to/file.lua" thing is an implementation detail. We may want to change it in the future, so give mods the opportunity to be (and stay) portable.
* `Corrections to client_lua_api.txt`
  Modpacks do already work in fact. max_hear_distance is a purely server-side optimization and does not make sense in the client API.
* `[CSM] Remove non-functional minetest.get_day_count()`
  The client does not know when the world was created, so remove this API (it always returned 0).

## To do

## How to test

Verify that the preview mod still works.

### DIFF
--- a/src/script/cpp_api/s_base.cpp
+++ b/src/script/cpp_api/s_base.cpp
@@ -197,18 +197,22 @@ void ScriptApiBase::loadModFromMemory(const std::string &mod_name)
 {
 	ModNameStorer mod_name_storer(getStack(), mod_name);
 
-	const std::string *init_filename = getClient()->getModFile(mod_name + ":init.lua");
-	const std::string display_filename = mod_name + ":init.lua";
-	if(init_filename == NULL)
-		throw ModError("Mod:\"" + mod_name + "\" lacks init.lua");
+	sanity_check(m_type == ScriptingType::Client);
 
-	verbosestream << "Loading and running script " << display_filename << std::endl;
+	const std::string init_filename = mod_name + ":init.lua";
+	const std::string chunk_name = "@" + init_filename;
+
+	const std::string *contents = getClient()->getModFile(init_filename);
+	if (!contents)
+		throw ModError("Mod \"" + mod_name + "\" lacks init.lua");
+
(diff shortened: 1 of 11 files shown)


src/script/cpp_api/s_base.cpp
```diff
@@ -197,18 +197,22 @@ void ScriptApiBase::loadModFromMemory(const std::string &mod_name)
...
-	if(init_filename == NULL)
-		throw ModError("Mod:\"" + mod_name + "\" lacks init.lua");
+	sanity_check(m_type == ScriptingType::Client);
 
-	verbosestream << "Loading and running script " << display_filename << std::endl;
+	const std::string init_filename = mod_name + ":init.lua";
+	const std::string chunk_name = "@" + init_filename;
+
+	const std::string *contents = getClient()->getModFile(init_filename);
+	if(contents == nullptr)
```
nerzhul: missing space

ClobberXD:
```suggestion
	if (!contents)
```

src/script/cpp_api/s_security.cpp
```diff
@@ -406,68 +408,49 @@ bool ScriptApiSecurity::safeLoadFile(lua_State *L, const char *path, const char
...
-		return false;
-	}
-
 	// Read the file
 	int ret = std::fseek(fp, 0, SEEK_END);
 	if (ret) {
 		lua_pushfstring(L, "%s: %s", path, strerror(errno));
-		std::fclose(fp);
 		if (path) {
+			std::fclose(fp);
```
SmallJoker: What's the reason of moving this here?

sfan5: Consistency. In normal operation fclose would only be called inside an `if(path)` [line 449 in the old code] but the error handlers would always call fclose.

PR #9101, opened 2019-11-10
Add scroll_container formspec element (redo)

### LINKED ISSUE #8576: Formspec: scrollable container

Desour:
Does what is described in #8576 and closes #8576.
This is a redo of #8591 (rebasing would probably have been more difficult). Many things are copied from that PR.
Note that #8740 has made this PR much easier.

- There's the new class `GUIScrollContainer`.
- The scroll_container consists of two elements. One clips every child, the other moves them. The clipper is the parent of the mover.
- The scroll container mover (of class `GUIScrollContainer`) is moved on event.

## To do

## How to test

Use `/formspec` in minimal.
old:
<details><summary>Here's a formspec that I've used for testing.</summary>

```lua
local my_formspec =
	"formspec_version[3]"..
	"size[10,8]"..
	--~ "real_coordinates[true]"..
	--~ "position[0.2,0.7]"..
	--~ "anchor[0,1]"..
	"button[8,1;1,1;bla;Bla]"..
	--~ "scrollbaroptions[max=100]"..
	"box[1,1;8,6;#00aa]"..
	"scroll_container[1,1;8,6;scrbar;down]"..
		"button[0,1;1,1;baum;Baum]"..
		--~ "container[0,0]"..
		"button[0,10;1,1;planze;Pflanze]"..
		"pwdfield[2,2;1,1;baum2;Baum]"..
		--~ "container_end[]"..
		"list[current_player;main;4,4;1,5;]"..
		"box[2,5;3,2;#ffff00]"..
		"image[1,10;3,2;bubble.png]"..
		"image[3,1;bubble.png]"..
		"item_image[2,6;3,2;default:mese]"..
		"label[2,15;bla baum\nfoo bar]"..
		--~ "textarea[2,10;5,6;txtarea;Baum;bla]"..
		"item_image_button[2,3;1,1;default:dirt_with_grass;itemimagebutton;ItemImageButton]"..
		--~ "tooltip[planze;Blatt;#f00;#000]"..
		"tooltip[0,11;3,2;Blatt;#f00;#000]"..
		"box[0,11;3,2;#00ff00]"..
		--~ "checkbox[2,25;chb;ChB;false]",
		"container[0,18]"..
			"box[1,2;3,2;#0a0a]"..
			"scroll_container[1,2;3,2;scrbar2;right;0.06]"..
				"label[10,0;nest]"..
				"button[0,0;6,1;butnest;Nest]"..
				--~ "bgcolor[#aa0a]"..
			"scroll_container_end[]"..
			"scrollbar[1,0;3.5,0.3;horizontal;scrbar2;0]"..
		"container_end[]"..
		--~ "tablecolumns[text,bla]"..
		--~ "table[1,23;3,1;tbl;a,b,c,d;2]"..
		--~ "textlist[0,5;2,0.5;txtlst;a,b,c]"..
		"dropdown[0,6;2;hmdrpdwn;apfel,birne;1]"..
		"image_button[0,4;2,2;bubble.png;bubblebutton;bbbbtt;false;true;heart.png]"..
		--~ "tabheader[-1,9;mytabheader;a,b,c;2;false;false]"..
		--~ "background[2,4;2,2;default_wood.png;true]"..
		--~ "bgcolor[#00aa]"..
		"box[1,22.5;4,1;#a00a]"..
	"scroll_container_end[]"..
	"scrollbaroptions[max=170]".. -- lowest seen pos is: 0.1*170+6=23 (factor*max+height)
	"scrollbar[7.5,0;0.3,4;vertical;scrbar;0]"..
	"scrollbar[8,0;0.3,4;vertical;scrbarhmmm;0]"..
	"dropdown[0,6;2;hmdrpdwnnn;apfel,birne;1]"..
	--~ "background[5,5;1,1;testbg.png;true;10]"..
	--~ "textarea[5,0.5;1,1;scrtxtarea;Baum;blablaajk adfkjgn \n akfjb]"..
	--~ "bgcolor[;true]"..
	--~ "bgcolor[#0000]"..
	""
```
</details>

Here's an (old) demo video: [scroll_container_show.m4v.zip](https://github.com/minetest/minetest/files/3273388/scroll_container_show.m4v.zip)

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2108,8 +2108,8 @@ Elements
   * be additionally moved by the current value of the scrollbar with the name
     `scrollbar name` times `scroll factor` along the orientation `orientation` and
   * be clipped to the rectangle defined by `X`, `Y`, `W` and `H`.
-* Possible values for `orientation` are `vertical` and `horizontal`.
-* `scroll factor` is optional and defaults to `0.1`.
+* `orientation`: possible values are `vertical` and `horizontal`.
+* `scroll factor`: optional, defaults to `0.1`.
 * Nesting is possible.
 * Some elements might work a little different if they are in a scroll_container.
 * Note: If you want the scroll_container to actually work, you also need to add a
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2101,6 +2101,26 @@ Elements
 * End of a container, following elements are no longer relative to this
   container.
 
+### `scroll_container[<X>,<Y>;<W>,<H>;<scrollbar name>;<orientation>;<scroll factor>]`
+
+* Start of a scroll_container block. All contained elements will ...
+  * take the scroll_container coordinate as position origin,
+  * be additionally moved by the current value of the scrollbar with the name
+    `scrollbar name` times `scroll factor` along the orientation `orientation` and
+  * be clipped to the rectangle defined by `X`, `Y`, `W` and `H`.
+* Possible values for `orientation` are `vertical` and `horizontal`.
+* `scroll factor` is optional and defaults to `0.1`.
+* Nesting is possible.
+* Some elements might work a little different if they are in a scroll_container.
+* Note: If you want the scroll_container to actually work, you also need to add a
+  scrollbar element with the specified name. Furthermore, it is highly recommended
+  to use a scrollbaroptions element on this scrollbar.
+
+### `scroll_container_end[]`
+
+* End of a scroll_container, following elements are no longer bound to this
+  container.
+
 ### `list[<inventory location>;<list name>;<X>,<Y>;<W>,<H>;]`
 
 * Show an inventory list if it has been sent to the client. Nothing will
--- a/src/gui/guiFormSpecMenu.cpp
+++ b/src/gui/guiFormSpecMenu.cpp
@@ -379,19 +379,6 @@ void GUIFormSpecMenu::parseScrollContainer(parserData *data, const std::string &
 	v2s32 pos = getRealCoordinateBasePos(v_pos);
 	v2s32 geom = getRealCoordinateGeometry(v_geom);
 
-	// old parent (at first: this)
-	// ^ is parent of clipper
-	// ^ is parent of mover
-	// ^ is parent of other elements
-
-	// make clipper
-	FieldSpec spec_clipper(
-		"",
-		L"",
-		L"",
-		258 + m_fields.size()
-	);
-
 	if (orientation == "vertical")
 		scroll_factor *= -imgsize.Y;
 	else if (orientation == "horizontal")
@@ -401,12 +388,16 @@ void GUIFormSpecMenu::parseScrollContainer(parserData *data, const std::string &
 				<< "Invalid scroll_container orientation: " << orientation
 				<< std::endl;
 
+	// old parent (at first: this)
+	// ^ is parent of clipper
+	// ^ is parent of mover
+	// ^ is parent of other elements
+
+	// make clipper
 	core::rect<s32> rect_clipper = core::rect<s32>(pos, pos + geom);
 
 	gui::IGUIElement *clipper = new gui::IGUIElement(EGUIET_ELEMENT, Environment,
-			data->current_parent, spec_clipper.fid, rect_clipper);
-
-	m_fields.push_back(spec_clipper);
+			data->current_parent, 0, rect_clipper);
 
 	// make mover
 	FieldSpec spec_mover(
--- a/src/gui/guiFormSpecMenu.cpp
+++ b/src/gui/guiFormSpecMenu.cpp
@@ -65,6 +65,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include "guiInventoryList.h"
 #include "guiItemImage.h"
 #include "guiScrollBar.h"
+#include "guiScrollContainer.h"
 #include "guiTable.h"
 #include "intlGUIEditBox.h"
 #include "guiHyperText.h"
@@ -143,6 +144,8 @@ GUIFormSpecMenu::~GUIFormSpecMenu()
 		tooltip_rect_it.first->drop();
 	for (auto &clickthrough_it : m_clickthrough_elements)
 		clickthrough_it->drop();
+	for (auto &scroll_container_it : m_scroll_containers)
+		scroll_container_it.second->drop();
 
 	delete m_selected_item;
 	delete m_form_src;
@@ -351,6 +354,111 @@ void GUIFormSpecMenu::parseContainerEnd(parserData* data)
 	}
 }
 
+void GUIFormSpecMenu::parseScrollContainer(parserData *data, const std::string &element)
+{
+	std::vector<std::string> parts = split(element, ';');
+
+	if (parts.size() < 4 ||
+			(parts.size() > 5 && m_formspec_version <= FORMSPEC_API_VERSION)) {
+		errorstream << "Invalid scroll_container start element (" << parts.size()
+				<< "): '" << element << "'" << std::endl;
+		return;
+	}
+
+	std::vector<std::string> v_pos  = split(parts[0], ',');
+	std::vector<std::string> v_geom = split(parts[1], ',');
+	std::string scrollbar_name = parts[2];
+	std::string orientation = parts[3];
+	f32 scroll_factor = 0.1f;
+	if (parts.size() >= 5 && !parts[4].empty())
+		scroll_factor = stof(parts[4]);
+
+	MY_CHECKPOS("scroll_container", 0);
+	MY_CHECKGEOM("scroll_container", 1);
+
+	v2s32 pos = getRealCoordinateBasePos(v_pos);
+	v2s32 geom = getRealCoordinateGeometry(v_geom);
+
+	// old parent (at first: this)
+	// ^ is parent of clipper
+	// ^ is parent of mover
+	// ^ is parent of other elements
+
+	// make clipper
+	FieldSpec spec_clipper(
+		"",
+		L"",
+		L"",
+		258 + m_fields.size()
+	);
+
+	if (orientation == "vertical")
+		scroll_factor *= -imgsize.Y;
+	else if (orientation == "horizontal")
+		scroll_factor *= -imgsize.X;
+	else
+		warningstream << "GUIFormSpecMenu::parseScrollContainer(): "
+				<< "Invalid scroll_container orientation: " << orientation
+				<< std::endl;
+
+	core::rect<s32> rect_clipper = core::rect<s32>(pos, pos + geom);
+
+	gui::IGUIElement *clipper = new gui::IGUIElement(EGUIET_ELEMENT, Environment,
+			data->current_parent, spec_clipper.fid, rect_clipper);
+
+	m_fields.push_back(spec_clipper);
+
+	// make mover
+	FieldSpec spec_mover(
+		"",
+		L"",
+		L"",
+		258 + m_fields.size()
+	);
+
+	core::rect<s32> rect_mover = core::rect<s32>(0, 0, geom.X, geom.Y);
+
+	GUIScrollContainer *mover = new GUIScrollContainer(Environment,
+			clipper, spec_mover.fid, rect_mover, orientation, scroll_factor);
+
+	data->current_parent = mover;
+
+	m_scroll_containers.emplace_back(scrollbar_name, mover);
+
+	m_fields.push_back(spec_mover);
+
+	clipper->drop();
+
+	// remove interferring offset of normal containers
+	container_stack.push(pos_offset);
+	pos_offset.X = 0.0f;
+	pos_offset.Y = 0.0f;
+}
+
+void GUIFormSpecMenu::parseScrollContainerEnd(parserData *data)
+{
+	if (data->current_parent == this || data->current_parent->getParent() == this ||
+			container_stack.empty()) {
+		errorstream << "Invalid scroll_container end element, "
+				<< "no matching scroll_container start element" << std::endl;
+		return;
+	}
+
+	if (pos_offset.getLengthSQ() != 0.0f) {
+		// pos_offset is only set by containers and scroll_containers.
+		// scroll_containers always set it to 0,0 which means that if it is
+		// not 0,0, it is a normal container that was opened last, not a
+		// scroll_container
+		errorstream << "Invalid scroll_container end element, "
+				<< "an inner container was left open" << std::endl;
+		return;
+	}
+
+	data->current_parent = data->current_parent->getParent()->getParent();
+	pos_offset = container_stack.top();
+	container_stack.pop();
+}
+
 void GUIFormSpecMenu::parseList(parserData *data, const std::string &element)
 {
 	if (m_client == 0) {
@@ -443,9 +551,9 @@ void GUIFormSpecMenu::parseList(parserData *data, const std::string &element)
 				pos.X + (geom.X - 1) * slot_spacing.X + imgsize.X,
 				pos.Y + (geom.Y - 1) * slot_spacing.Y + imgsize.Y);
 
-		GUIInventoryList *e = new GUIInventoryList(Environment, this, spec.fid,
-				rect, m_invmgr, loc, listname, geom, start_i, imgsize, slot_spacing,
-				this, data->inventorylist_options, m_font);
+		GUIInventoryList *e = new GUIInventoryList(Environment, data->current_parent,
+				spec.fid, rect, m_invmgr, loc, listname, geom, start_i, imgsize,
+				slot_spacing, this, data->inventorylist_options, m_font);
 
 		m_inventorylists.push_back(e);
 		m_fields.push_back(spec);
@@ -550,8 +658,8 @@ void GUIFormSpecMenu::parseCheckbox(parserData* data, const std::string &element
 
 		spec.ftype = f_CheckBox;
 
-		gui::IGUICheckBox *e = Environment->addCheckBox(fselected, rect, this,
-					spec.fid, spec.flabel.c_str());
+		gui::IGUICheckBox *e = Environment->addCheckBox(fselected, rect,
+				data->current_parent, spec.fid, spec.flabel.c_str());
 
 		auto style = getDefaultStyleForElement("checkbox", name);
 		e->setNotClipped(style.getBool(StyleSpec::NOCLIP, false));
@@ -610,8 +718,8 @@ void GUIFormSpecMenu::parseScrollBar(parserData* data, const std::string &elemen
 
 		spec.ftype = f_ScrollBar;
 		spec.send  = true;
-		GUIScrollBar *e = new GUIScrollBar(Environment, this, spec.fid, rect,
-				is_horizontal, true);
+		GUIScrollBar *e = new GUIScrollBar(Environment, data->current_parent,
+				spec.fid, rect, is_horizontal, true);
 
 		auto style = getDefaultStyleForElement("scrollbar", name);
 		e->setNotClipped(style.getBool(StyleSpec::NOCLIP, false));
@@ -737,7 +845,8 @@ void GUIFormSpecMenu::parseImage(parserData* data, const std::string &element)
 			1
 		);
 		core::rect<s32> rect(pos, pos + geom);
-		gui::IGUIImage *e = Environment->addImage(rect, this, spec.fid, 0, true);
+		gui::IGUIImage *e = Environment->addImage(rect, data->current_parent,
+				spec.fid, 0, true);
 		e->setImage(texture);
 		e->setScaleImage(true);
 		auto style = getDefaultStyleForElement("image", spec.fname);
@@ -774,8 +883,8 @@ void GUIFormSpecMenu::parseImage(parserData* data, const std::string &element)
 			L"",
 			258 + m_fields.size()
 		);
-		gui::IGUIImage *e = Environment->addImage(texture, pos, true, this,
-				spec.fid, 0);
+		gui::IGUIImage *e = Environment->addImage(texture, pos, true,
+				data->current_parent, spec.fid, 0);
 		auto style = getDefaultStyleForElement("image", spec.fname);
 		e->setNotClipped(style.getBool(StyleSpec::NOCLIP, m_formspec_version < 3));
 		m_fields.push_back(spec);
@@ -886,7 +995,7 @@ void GUIFormSpecMenu::parseItemImage(parserData* data, const std::string &elemen
 		);
 		spec.ftype = f_ItemImage;
 
-		GUIItemImage *e = new GUIItemImage(Environment, this, spec.fid,
+		GUIItemImage *e = new GUIItemImage(Environment, data->current_parent, spec.fid,
 				core::rect<s32>(pos, pos + geom), name, m_font, m_client);
 		auto style = getDefaultStyleForElement("item_image", spec.fname);
 		e->setNotClipped(style.getBool(StyleSpec::NOCLIP, false));
@@ -949,8 +1058,8 @@ void GUIFormSpecMenu::parseButton(parserData* data, const std::string &element,
 		if(type == "button_exit")
 			spec.is_exit = true;
 
-		GUIButton *e = GUIButton::addButton(Environment, rect, m_tsrc, this,
-				spec.fid, spec.flabel.c_str());
+		GUIButton *e = GUIButton::addButton(Environment, rect, m_tsrc,
+				data->current_parent, spec.fid, spec.flabel.c_str());
 
 		auto style = getStyleForElement(type, name, (type != "button") ? "button" : "");
 		e->setStyles(style);
@@ -1141,7 +1250,8 @@ void GUIFormSpecMenu::parseTable(parserData* data, const std::string &element)
 		}
 
 		//now really show table
-		GUITable *e = new GUITable(Environment, this, spec.fid, rect, m_tsrc);
+		GUITable *e = new GUITable(Environment, data->current_parent, spec.fid,
+				rect, m_tsrc);
 
 		if (spec.fname == data->focused_fieldname) {
 			Environment->setFocus(e);
@@ -1217,7 +1327,8 @@ void GUIFormSpecMenu::parseTextList(parserData* data, const std::string &element
 		}
 
 		//now really show list
-		GUITable *e = new GUITable(Environment, this, spec.fid, rect, m_tsrc);
+		GUITable *e = new GUITable(Environment, data->current_parent, spec.fid,
+				rect, m_tsrc);
 
 		if (spec.fname == data->focused_fieldname) {
 			Environment->setFocus(e);
@@ -1293,7 +1404,8 @@ void GUIFormSpecMenu::parseDropDown(parserData* data, const std::string &element
 		spec.send = true;
 
 		//now really show list
-		gui::IGUIComboBox *e = Environment->addComboBox(rect, this, spec.fid);
+		gui::IGUIComboBox *e = Environment->addComboBox(rect, data->current_parent,
+				spec.fid);
 
 		if (spec.fname == data->focused_fieldname) {
 			Environment->setFocus(e);
@@ -1379,7 +1491,8 @@ void GUIFormSpecMenu::parsePwdField(parserData* data, const std::string &element
 			);
 
 		spec.send = true;
-		gui::IGUIEditBox * e = Environment->addEditBox(0, rect, true, this, spec.fid);
+		gui::IGUIEditBox *e = Environment->addEditBox(0, rect, true,
+				data->current_parent, spec.fid);
 
 		if (spec.fname == data->focused_fieldname) {
 			Environment->setFocus(e);
@@ -1390,7 +1503,7 @@ void GUIFormSpecMenu::parsePwdField(parserData* data, const std::string &element
 			rect.UpperLeftCorner.Y -= font_height;
 			rect.LowerRightCorner.Y = rect.UpperLeftCorner.Y + font_height;
 			gui::StaticText::add(Environment, spec.flabel.c_str(), rect, false, true,
-				this, 0);
+				data->current_parent, 0);
 		}
 
 		e->setPasswordBox(true,L'*');
@@ -1425,7 +1538,7 @@ void GUIFormSpecMenu::createTextField(parserData *data, FieldSpec &spec,
 	if (!is_editable && !is_multiline) {
 		// spec field id to 0, this stops submit searching for a value that isn't there
 		gui::StaticText::add(Environment, spec.flabel.c_str(), rect, false, true,
-				this, 0);
+				data->current_parent, 0);
 		return;
 	}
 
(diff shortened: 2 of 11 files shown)


DS-Minetest:
Rebased.

Hm, currently I'm unsure what to do with `bgcolor`. The old PR changed it so that a box is added as background to the `scroll_container` if `bgcolor` is used in a `scroll_container`, but now I don't know what benefits this would have over just adding a box element before the `scroll_container` in the formspec.

v-rob:
`bgcolor` useless to add since boxes have proper stacking, so I'm in favor of less code for a useless thing.

rubenwardy:
I'd avoid adding unessential features like that

This PR is next on the formspec review list, would appreciate an update :)

DS-Minetest:
Ok, now it should be rebased correctly.
(Btw. github doesn't make notifications when labels are added.)

DS-Minetest:
Would someone like to remove the `WIP` and `Rebase needed` labels btw? They don't apply anymore.

doc/lua_api.txt
```diff
@@ -2037,6 +2037,24 @@ Elements
 * End of a container, following elements are no longer relative to this
   container.
 
+### `scroll_container[<X>,<Y>;<W>,<H>;<scrollbar name>;<direction>;<scroll factor>]`
+
+* Start of a scroll_container block, all elements in a scroll_container block will
```
```suggestion
* Start of a scroll_container block. All contained elements will ...
```

doc/lua_api.txt
```diff
@@ -2037,6 +2037,24 @@ Elements
...
+  * take the block's coordinate as position origin,
```
```suggestion
  * take the scroll container coordinate as position origin
```

src/gui/guiFormSpecMenu.cpp
```diff
@@ -347,6 +348,99 @@ void GUIFormSpecMenu::parseContainerEnd(parserData* data)
...
+	pos_offset.Y = 0.0f;
+}
+
+void GUIFormSpecMenu::parseScrollContainerEnd(parserData *data)
+{
+	if (data->current_parent == this || data->current_parent->getParent() == this ||
+			container_stack.empty()) {
+		errorstream << "Invalid scroll_container end element, "
+				<< "no matching scroll_container start element" << std::endl;
+	} else if (pos_offset.getLengthSQ() != 0.0f) {
```
SmallJoker: When does this happen?
What if the scroll container is at `0,0` ?

src/gui/guiFormSpecMenu.cpp
```diff
@@ -347,6 +348,99 @@ void GUIFormSpecMenu::parseContainerEnd(parserData* data)
...
+	pos_offset.X = 0.0f;
+	pos_offset.Y = 0.0f;
+}
+
+void GUIFormSpecMenu::parseScrollContainerEnd(parserData *data)
+{
+	if (data->current_parent == this || data->current_parent->getParent() == this ||
+			container_stack.empty()) {
+		errorstream << "Invalid scroll_container end element, "
+				<< "no matching scroll_container start element" << std::endl;
```
SmallJoker: Please add an early return here and drop `else` below to clarify the execution structure. Nothing may happen if there's no valid container to close.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -738,10 +832,12 @@ void GUIFormSpecMenu::parseImage(parserData* data, const std::string &element)
...
 		);
 		core::rect<s32> rect(pos, pos + geom);
-		gui::IGUIImage *e = Environment->addImage(rect, this, spec.fid, 0, true);
+		gui::IGUIImage *e = Environment->addImage(rect, data->current_parent,
+				spec.fid, 0, true);
 		e->setImage(texture);
 		e->setScaleImage(true);
-		e->setNotClipped(true);
+		if (data->current_parent == this)
+			e->setNotClipped(true);
```
SmallJoker: How about styles? Probably needs rebase?

src/gui/guiScrollContainer.h
```diff
@@ -0,0 +1,49 @@
...
+		if (caller == m_scrollbar)
+			updateScrolling();
+	}
+
+	inline void setScrollBar(GUIScrollBar *scrollbar) { m_scrollbar = scrollbar; }
+
+private:
+	GUIScrollBar *m_scrollbar;
+	std::string m_direction;
+	f32 m_scrollfactor;
```
SmallJoker: You could replace both `m_scrollfactor` and `m_direction` with `v2f32` which points either down or right.

doc/lua_api.txt
DS-Minetest: Why should the comma be removed?

doc/lua_api.txt

src/gui/guiFormSpecMenu.cpp
DS-Minetest: `pos_offset` is only used by normal containers. `scroll_container`s set the `pos_offset` to (0,0) so that their children aren't affected by this offset. If the offset length is not 0 here, there was an unclosed normal container before.
I'll add a code comment.

doc/lua_api.txt
SmallJoker: No particular reason. Probably happened by re-writing the line.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Done.

src/gui/guiScrollContainer.h
DS-Minetest: But then I would have to check the components of that vector. I can't simply add a vector to `rect.UpperLeftCorner` as the scroll position is given absolute.

I've replaced the string with an enum here.

DS-Minetest:
>Additionally: similar to textarea[], scroll_container[] should best add the scroll bar automatically. Manual placement is annoying and forgetting it upsets Android players (no scroll wheel).

- `scroll_container`s don't work without a scrollbar, so there's no reason to worry about android players.
- Automatically added scrollbars would give less freedom.
- Now that we have real coordinates, placement should be easy.

doc/lua_api.txt
DS-Minetest: I'll mark this as resolved.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: I've rebased, there were no conflicts (and I couldn't find any new place where `this` used as parent).
#9262 will probably cause a conflict here.
:thinking: This `if` here was probably a bad idea. When #9262 is merged, the not clipping will only be the default for old formspec versions (which don't support `scroll_container`s anyway). This will make the change here unneeded, so I'll remove it.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -2091,7 +2088,7 @@ void GUIFormSpecMenu::parseTabHeader(parserData* data, const std::string &elemen
 		}
 
 		auto style = getStyleForElement("tabheader", name);
-		e->setNotClipped(style.getBool(StyleSpec::NOCLIP, data->current_parent == this));
+		e->setNotClipped(style.getBool(StyleSpec::NOCLIP, true));
```
DS-Minetest: I someone really wants to use a tabheader element in a `scroll_container`, they should step it clipped themselves.

DS-Minetest:
Thanks for reviewing!
Rebased and fixed comments as good as I could.

DS-Minetest:
Rebased again.

src/gui/guiScrollContainer.h
SmallJoker: Alright, I hoped it would be simpler than that solution. Good enough I guess :man_shrugging:

ClobberXD:
Nice work! Upon testing (with the formspec provided in the first post) I noticed a couple of things:

- Some formspec elements like buttons, lists, and dropdowns are clipped while images and boxes aren't:
  screenshot_20200117_151128
  screenshot_20200117_151120
  screenshot_20200117_151112
- Both the horizontal and vertical scrollbars allow scrolling _way_ beyond the actual content.

I'm not sure if either of these are unintended or even related to this PR, but I just wanted to seek clarification regarding these.

Df458:
> Some formspec elements like buttons, lists, and dropdowns are clipped while images and boxes aren't

I'm not the author, but I believe that is to be expected. This is an issue that applies to those elements normally, one of my PRs (#9262) just fixed this and I expect that the code isn't rebased on that yet. It would be interesting to cherry-pick the fix locally for testing, however.

ClobberXD:
> > Some formspec elements like buttons, lists, and dropdowns are clipped while images and boxes aren't
> 
> I'm not the author, but I believe that is to be expected. This is an issue that applies to those elements normally, one of my PRs (#9262) just fixed this and I expect that the code isn't rebased on that yet. It would be interesting to cherry-pick the fix locally for testing, however.

Ah, nice. Thanks for clarifying.

DS-Minetest:
Thanks for testing, @ClobberXD!

> Some formspec elements like buttons, lists, and dropdowns are clipped while images and boxes aren't:

You have to use `formspec_version[3]` (as @Df458 already noted) or enable clipping for these elements with style.
I'll update the test code in the first post.

> Both the horizontal and vertical scrollbars allow scrolling _way_ beyond the actual content.

That's the expected behaviour. `scrollbaroptions` should be used here. I'll add a note in the doc.

ClobberXD:
> You have to use formspec_version[3] (as @Df458 already noted) or enable clipping for these elements with style.

Yes, I'll test this again after tweaking the test script and let you know.

> That's the expected behaviour. scrollbaroptions should be used here. I'll add a note in the doc.

Righty-ho, good to know. Thanks!

ClobberXD:
Cherry-picked 680aaa1 into this branch. The boxes and images clip properly now, but I notice that an image of a meseblock is stuck to the bottom of the formspec, and doesn't scroll with the rest of the container. Also, only the very top of the vertical scrollbar is clickable, and clicking any further down the scrollbar does not allow me to interact with it.

screenshot_20200118_083822

I only added `formspec_version[3]` to the start of the formspec string, and didn't modify anything else in the test formspec from the first post. It's to be noted that there were a couple of merge conflicts when cherry-picking the aforementioned commit, and I probably messed up some of the code while resolving the conflicts (I don't understand much of the GUI code). Do let me know if I'm mistaken.

DS-Minetest:
> Cherry-picked [680aaa1](https://github.com/minetest/minetest/commit/680aaa1bd55fd4f0162ba31ca47d7e5b3563f2a1) into this branch.

That shouldn't be needed, I've rebased after that PR was merged.

> The boxes and images clip properly now, but I notice that an image of a meseblock is stuck to the bottom of the formspec, and doesn't scroll with the rest of the container.

Can't reproduce.

> Also, only the very top of the vertical scrollbar is clickable, and clicking any further down the scrollbar does not allow me to interact with it.

That's because in the test formspec the scrollbar was behind the `scroll_container`, sorry.
I've updated the test formspec. (There's also a max now.)

ClobberXD:
> That shouldn't be needed, I've rebased after that PR was merged.

Oh lol. That explains the oddities. Will retest again. :)

ClobberXD:
Merge conflict. I think I'm the reason for this. Should be pretty trivial to resolve though.

DS-Minetest:
Rebased.

rubenwardy:
Rebased here: https://github.com/rubenwardy/minetest/tree/scroll_container2

src/gui/guiScrollContainer.cpp
```diff
@@ -0,0 +1,58 @@
+/*
+Minetest
+Copyright (C) 2013 celeron55, Perttu Ahola <celeron55@gmail.com>
```
rubenwardy: wrong copyright

src/gui/guiScrollContainer.h
```diff
@@ -0,0 +1,56 @@
+/*
+Minetest
+Copyright (C) 2013 celeron55, Perttu Ahola <celeron55@gmail.com>
```
rubenwardy: wrong copyright

src/gui/guiFormSpecMenu.cpp
```diff
@@ -347,6 +348,111 @@ void GUIFormSpecMenu::parseContainerEnd(parserData* data)
...
+		return;
+	}
+
+	if (pos_offset.getLengthSQ() != 0.0f) {
+		// pos_offset is only set by containers and scroll_containers.
+		// scroll_containers allways set it to 0,0 which means that if it is
+		// not 0,0, it is a normal container that was opened last, not a
+		// scroll_container
+		errorstream << "Invalid scroll_container end element, "
+				<< "try to close normal container start element" << std::endl;
```
rubenwardy: an inner container was left open

src/gui/guiFormSpecMenu.cpp
```diff
@@ -347,6 +348,111 @@ void GUIFormSpecMenu::parseContainerEnd(parserData* data)
...
+
+	// remove interferring offset of normal containers
+	container_stack.push(pos_offset);
+	pos_offset.X = 0.0f;
+	pos_offset.Y = 0.0f;
+}
+
+void GUIFormSpecMenu::parseScrollContainerEnd(parserData *data)
+{
+	if (data->current_parent == this || data->current_parent->getParent() == this ||
```
rubenwardy: I'm not a fan of how this works, I think it would be better to make the container_stack take a struct rather than positions, which contains meta information on the container types. However, I'm fine with this given there's only one special container type currently

src/gui/guiFormSpecMenu.cpp
```diff
@@ -347,6 +348,111 @@ void GUIFormSpecMenu::parseContainerEnd(parserData* data)
...
+	v2s32 pos = getRealCoordinateBasePos(v_pos);
+	v2s32 geom = getRealCoordinateGeometry(v_geom);
+
+	// old parent (at first: this)
+	// ^ is parent of clipper
+	// ^ is parent of mover
+	// ^ is parent of other elements
+
+	// make clipper
+	FieldSpec spec_clipper(
```
rubenwardy: Why do you need to allocate an ID for the clipper and mover?

src/gui/guiFormSpecMenu.cpp
```diff
@@ -971,22 +1079,34 @@ void GUIFormSpecMenu::parseBackground(parserData* data, const std::string &eleme
 			258 + m_fields.size()
 		);
 
-		core::rect<s32> rect;
+		GUIBackgroundImage *e = nullptr;
 		if (!clip) {
 			// no auto_clip => position like normal image
-			rect = core::rect<s32>(pos, pos + geom);
-		} else {
+			core::rect<s32> rect(pos, pos + geom);
```
rubenwardy: Please change this to use the old design, where you set variables then only create the GUI element in one place. This makes it *a lot* easier to read and reason about. You can set the clip variable in these if-statements

rubenwardy:
Please also add a new tab to the formspec shown using `/formspec` and defined in `games/minimal/test/formspec.lua`

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Done.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Done.

src/gui/guiScrollContainer.cpp
DS-Minetest: I'm not good at such things. Feel free to suggest something.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Afaik, this is done for every element.
What id should I use instead for the clipper?

DS-Minetest:
Rebased and addressed most of the comments.
I'm sorry that I don't really have time for minetest at the moment.

src/gui/guiScrollContainer.cpp
rubenwardy: >  Copyright (C) 2020 DS-Minetest

src/gui/guiScrollContainer.h
```diff
@@ -0,0 +1,56 @@
...
+		if (caller == m_scrollbar)
+			updateScrolling();
+	}
+
+	inline void setScrollBar(GUIScrollBar *scrollbar) { m_scrollbar = scrollbar; }
+
+private:
+	enum DirectionEnum
+	{
+		DOWN,
```
rubenwardy: please use `VERTICAL` and `HORIZONTAL` to be consistent with the scrollbar element

doc/lua_api.txt
```diff
@@ -2071,6 +2071,27 @@ Elements
 * End of a container, following elements are no longer relative to this
   container.
 
+### `scroll_container[<X>,<Y>;<W>,<H>;<scrollbar name>;<direction>;<scroll factor>]`
```
rubenwardy: call it `<orientation>`

doc/lua_api.txt
```diff
@@ -2071,6 +2071,27 @@ Elements
...
+
+* Start of a scroll_container block. All contained elements will ...
+  * take the scroll_container coordinate as position origin,
+  * be additionally moved by the current value of the scrollbar with the name
+    `scrollbar name` times `scroll factor` in the direction `direction` and
+  * be clipped to the rectangle defined by `X`, `Y`, `W` and `H`.
+* Possible values for `direction` are `down` and `right`.
```
```suggestion
* Possible values for `orientation` are `vertical` and `horizontal`.
```

src/gui/guiFormSpecMenu.cpp
rubenwardy: I admit that the scope of fieldspecs confuse me. To clarify - is there a reason why the id isn't -1 with no fieldspec?

src/gui/guiFormSpecMenu.cpp
```diff
@@ -349,6 +350,111 @@ void GUIFormSpecMenu::parseContainerEnd(parserData* data)
...
+	if (data->current_parent == this || data->current_parent->getParent() == this ||
+			container_stack.empty()) {
+		errorstream << "Invalid scroll_container end element, "
+				<< "no matching scroll_container start element" << std::endl;
+		return;
+	}
+
+	if (pos_offset.getLengthSQ() != 0.0f) {
+		// pos_offset is only set by containers and scroll_containers.
+		// scroll_containers allways set it to 0,0 which means that if it is
```
rubenwardy:
```suggestion
		// scroll_containers always set it to 0,0 which means that if it is
```

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Hm, I don't think, anything would speak against -1.
Should I use -1?

DS-Minetest:
I've renamed the directions, as requested. (I guess it is clear, that it scrolls down respectively to the right.)

rubenwardy:
Bug: if you start scrolling whilst an item slot is hovered, the tooltip will stay until it is replaced by another tooltip

DS-Minetest:
You're right. Buttons also stay hovered after scrolling.
I'll try to fix this.

DS-Minetest:
I've tried saving the last mouse move event in `GUIFormspecMenu` (`m_last_mouse_move_event`) and re-sending it in (the end of) `GUIFormspecMenu::OnEvent` like this:
```cpp
if (scroll_container_was_moved_in_this_OnEvent) {
		Environment->postEventFromUser(m_last_mouse_move_event); // needed for buttons (env->getHovered()) to be updated
		Environment->getRootGUIElement()
				->getElementFromPoint(
						core::position2d<s32>(m_last_mouse_move_event.MouseInput.X,
								m_last_mouse_move_event.MouseInput.Y))
				->OnEvent(m_last_mouse_move_event); // needed for GUIInventorLists to update their hovered_i (postEventFromUser doesn't seem to call OnEvent)
}
```
But then scrolling using the scrollbar slider didn't work anymore.

Hence I don't know how to fix this properly.
Imo the bug is ugly, but acceptable, it doesn't really hurt.

Desour:
:/ `background[]`s to scroll containers don't seem to work nicely. I'll simply remove the changes to `background[]` as an `image[]` element is much better here and ordering elements is possible now.

Desour:
Everything should be fine now.
Apart from the comment that -1 could be used as ID for the clipper, I can't decide here.

kilbith:
Question: what happens if a `scroll_container` is inside another one?

rubenwardy:
I can confirm that your changes fix the named and area tooltips, but not the item tooltips.

To fix this cleanly, we could do with a refactor of how tooltips are set. It should be done using one method - GUIFormspecMenu should iterate through elements under the cursor, calling a `getTooltip()` method. The first one which returns a non-empty string wins

Desour:
I didn't push any changes regarding the tooltip bug.

The area and named tooltips are implemented in GUIFormspecMenu by checking the mouse pos on every draw. The tooltip in InventoryLists (= the index of the selected item) and whether a button is hovered is determined when the mouse moves (there are also events (guievents) for when the cursor leaves or enters the rect of an element).

Firing a mouse event was what I was trying to do, but as written before that resulted in problems (scrolling by pulling the scrollbar slider somehow didn't work anymore).
(Updating where the mouse is relative to other elements should also be done because of other reasons, eg. hovered and pressed buttons stay hovered respectively pressed.)

Calling something like `getTooltip()` on every draw from the elements at the mousepos (in a particular order) is indeed a good idea. There's also a [member function in IGUIElement](http://irrlicht.sourceforge.net/docu/classirr_1_1gui_1_1_i_g_u_i_element.html#aaaac5a17bb0f0d7dcdc8709a800a6ae0) (with wide strings). Lists currently do it the other way around: https://github.com/minetest/minetest/blob/0e88176db8ce9f3fb631feb56bdd29d4ada054f5/src/gui/guiInventoryList.cpp#L147
However, I don't think that belongs into this issue.

rubenwardy:
This is gross, but works:

```cpp.diff
diff --git a/src/gui/guiScrollContainer.cpp b/src/gui/guiScrollContainer.cpp
index f58161e1a..cef64588d 100644
--- a/src/gui/guiScrollContainer.cpp
+++ b/src/gui/guiScrollContainer.cpp
@@ -36,9 +36,30 @@ GUIScrollContainer::GUIScrollContainer(gui::IGUIEnvironment *env,
 bool GUIScrollContainer::OnEvent(const SEvent &event)
 {
        if (event.EventType == EET_MOUSE_INPUT_EVENT &&
+                       event.MouseInput.Event == EMIE_MOUSE_MOVED) {
+               lastMousePosition.X = event.MouseInput.X;
+               lastMousePosition.Y = event.MouseInput.Y;
+       } else if (event.EventType == EET_MOUSE_INPUT_EVENT &&
                        event.MouseInput.Event == EMIE_MOUSE_WHEEL && m_scrollbar) {
                Environment->setFocus(m_scrollbar);
-               return m_scrollbar->OnEvent(event);
+
+               if (!m_scrollbar->OnEvent(event)) {
+                       return false;
+               }
+
+               // HACK: this is gross, but is needed to allow children to update hovers
+               irr::SEvent evt{};
+               evt.EventType = EET_MOUSE_INPUT_EVENT;
+               evt.MouseInput = irr::SEvent::SMouseInput{};
+               evt.MouseInput.Event = irr::EMOUSE_INPUT_EVENT::EMIE_MOUSE_MOVED;
+               evt.MouseInput.X = lastMousePosition.X;
+               evt.MouseInput.Y = lastMousePosition.Y;
+               for (auto child : this->getChildren()) {
+                       if (child->OnEvent(evt)) {
+                               return true;
+                       }
+               }
+               return true;
        }
 
        return IGUIElement::OnEvent(event);
diff --git a/src/gui/guiScrollContainer.h b/src/gui/guiScrollContainer.h
index 835a661bf..e50621a8c 100644
--- a/src/gui/guiScrollContainer.h
+++ b/src/gui/guiScrollContainer.h
@@ -51,6 +51,7 @@ class GUIScrollContainer : public gui::IGUIElement
        GUIScrollBar *m_scrollbar;
        OrientationEnum m_orientation;
        f32 m_scrollfactor;
+       irr::core::vector2di lastMousePosition {};
 
        void updateScrolling();
 };
```

rubenwardy:
> there are also events (guievents) for when the cursor leaves or enters the rect of an element

Would it be possible to change inventory slots to use this system if inventory slots were elements in their own right?

Desour:
Fixed the bug mostly.
The only problem that still seems to persists is that you can press a button, scroll and it will stay in place. I guess this is because the focus gets changed.

A bit of irc log here: http://irc.minetest.net/minetest-dev/2020-03-08#i_5649322

Desour:
>The only problem that still seems to persists is that you can press a button, scroll and it will stay in place.

Fixed.

Desour:
Squashed and rebased.

(And also changed copyright name back to "DS" and added drop calls to `m_scroll_containers` in the GUIFormspecMenu destructor (I wonder why it was missing).)

Desour:
Would someone mind removing the `Rebase needed` label?

rubenwardy:
Sorry lol!

Desour:
Rebased again.

rubenwardy:
Code looks good to me.

Hypertext elements aren't clipped. There's also weird cursor behaviour when hovering over an anchor outside the scroll container - it flickers between clickable and not

image

```
diff --git a/games/minimal/mods/test/formspec.lua b/games/minimal/mods/test/formspec.lua
index b96c5b13d..2059e6f33 100644
--- a/games/minimal/mods/test/formspec.lua
+++ b/games/minimal/mods/test/formspec.lua
@@ -1,5 +1,34 @@
 local color = minetest.colorize
 
+local hypertext = minetest.formspec_escape([[^M
+       <global halign=justify valign=center background=#CCF color=#444 actioncolor=darkblue margin=10>^M
+<center><bigger>Furnace</bigger></center>^M
+<style color=black>Furnaces</style> are <action name="crafting">crafted</action> and used by the player for the purpose of cooking food and <action name="smelting">smelting</action> various items.^M
+<item name=default:furnace float=right width=128 rotate=yes>^M
+<style color=black>Type:</style> Solid block^M
+<style color=black>Drops:</style> Itself^M
+<style color=black>Physics:</style> No^M
+<style color=black>Luminance:</style> Inactive:No Active:Yes (8)^M
+<style color=black>Flammable:</style> No^M
+<style color=black>Generated:</style> No^M
+<style color=black>Renewable:</style> Yes^M
+<style color=black>Stackable:</style> Yes (99)^M
+<style color=black>Itemstring:</style> default:furnace default:furnace_active^M
+<big>Usage</big>^M
+The furnace menu can be accessed by <action name="using">using</action> the furnace.^M
+The furnace has 3 <action name="inventory">inventories</action>: An input slot, a fuel slot and 4 output slots. The fire in the furnace will automatically start when there is a smeltable item in the input slot and a fuel in the fuel slot.^M
+As long as the fire is on, the furnace will smelt any smeltable item in the input slot, one by one, until it is empty. When the fire goes off, it will smelt the next item until there are no smeltable items and no fuel items left.^M
+The current stage of cooking can be seen by <action name="pointing">pointing</action> the furnace or by viewing the furnace menu. In the furnace menu, the flame symbol roughly shows the remaining burning time. The arrow symbol shows the progress of the current smelting process.^M
+<big>Renewing</big>^M
+Furnaces can be crafted from e.g. <action name="default:cobblestone">cobblestone</action>, a renewable resource.^M
+<big>Crafting</big>^M
+Sorry no way to display crafting yet in formspec pages.^M
+<big>Fuel</big>^M
+See <action name="smelting">Smelting</action> for a list of furnace fuels.^M
+<big>Recipes</big>^M
+See the <action name="smelting">Smelting</action> page.^M
+]])^M
+^M
 local clip_fs = [[
        style_type[label,button,image_button,item_image_button,
                        tabheader,scrollbar,table,animated_image
@@ -207,6 +236,7 @@ Number]
                        "item_image_button[2,3;1,1;default:dirt_with_grass;itemimagebutton;ItemImageButton]"..
                        "tooltip[0,11;3,2;Buz;#f00;#000]"..
                        "box[0,11;3,2;#00ff00]"..
+                       "hypertext[1,1;3,3;;" .. hypertext .. "]" ..^M
                        "container[0,18]"..
                                "box[1,2;3,2;#0a0a]"..
                                "scroll_container[1,2;3,2;scrbar2;horizontal;0.06]"..
```

Desour:
Fixed hypertext element not being clipped, as reported above. Scrolling hypertext also works now in scroll containers (the mousewheel event is handled, hence return true). Also added the test hypertext formspec provided above into minimal test formspec.

src/gui/guiFormSpecMenu.cpp
SmallJoker: `FieldSpec` entries are only needed if the element....

* has a name (for styling, sending fields or tooltips)
* or needs special legacy sorting
* ~or has clipped text/label that is shown entirely using a tooltip~
   * ^ possible use-case in the future, not implemented.

All other `m_fields` entries are superfluous because they neither contain information nor provide any special functionality.

doc/lua_api.txt
```diff
@@ -2101,6 +2101,26 @@ Elements
...
   container.
 
+### `scroll_container[<X>,<Y>;<W>,<H>;<scrollbar name>;<orientation>;<scroll factor>]`
+
+* Start of a scroll_container block. All contained elements will ...
+  * take the scroll_container coordinate as position origin,
+  * be additionally moved by the current value of the scrollbar with the name
+    `scrollbar name` times `scroll factor` along the orientation `orientation` and
+  * be clipped to the rectangle defined by `X`, `Y`, `W` and `H`.
+* Possible values for `orientation` are `vertical` and `horizontal`.
```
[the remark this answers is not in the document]
Desour: Done.

Wuzzy2:
I see there have been changes to `minimal` …

May I suggest for the future to keep all PRs that touch `minimal` on hold until DevTest (#9450) is merged? It's not fun when the code bases rapidly diverge …

Issue #8584, opened 2019-06-07
Server freeze/"zombie mode"


thomasrudin:
##### Issue type
 - Bug report
 
##### Minetest version
```
5.1.0-dev (happened with 0.4.17.1 too)
568694122cb662d6c6cb071338c9d30721372b50
```

##### OS / Hardware
Operating system: Linux
CPU: x86_64

##### Summary

A couple times a week my server goes into "zombie mode":
* clients are online but the actions don't get through (no map-loading either)
* in the background (debug.txt) the loaded areas are still working (pipeworks, mesecons messages)

This can go on for minutes (about 20 minutes average)
In this time the "min-lag" (min-time between steps) increases to something above 100 ms

Monitoring image with elevated "min-lag" (yellow on the right):
<img src="https://i.imgur.com/5JXEEkw.png"></img>

I mitigated the effects of this a bit with my "watchdog" mod:
* https://git.rudin.io/minetest/watchdog/src/master/init.lua

It just measures the "min-time" between steps and if it is elevated for a while it restarts the server so everybody can play again (this is what happend in the above image).
But this should only be a temporary mitigation, i hope the root-cause can be found someday.

The server-log is filled with lots of Pathfinder messages while that happens:
```
... lots more of those below ...
2019-06-07 17:52:42: WARNING[Server]: Pathfinder: failed to update cost map
2019-06-07 17:52:42: WARNING[Server]: Pathfinder: failed to update cost map
2019-06-07 17:52:42: WARNING[Server]: Watchdog triggered!
```

Installed (relevant) mods:
* mobs_redo (monsters, animals, horse, npc)
* advtrains

Relevant links:
* monitoring code for "min-lag": https://github.com/thomasrudin-mt/monitoring/blob/master/builtin/lag.lua#L36
* Watchdog mod: https://git.rudin.io/minetest/watchdog/src/master/init.lua
* Graph: https://i.imgur.com/5JXEEkw.png (https://monitoring.minetest.land/d/YUpouLmWk/overview?orgId=1&from=1559928314152&to=1559930391076)
* My issue: https://github.com/pandorabox-io/pandorabox.io/issues/127
* Related issue: #6771

##### Steps to reproduce
* no clue, happens sporadically

SmallJoker:
IhrFussel reported a similar issue, where the mapblocks were no longer sent to the clients until a certain player left. I believe there it's caused by either a high packet loss or wrong calculation for the mapblock re-sending delay.
[...]

PR #9114, opened 2019-11-14
Rework packet receiving in ServerThread

### LINKED ISSUE #8207: Packets time out in singleplayer
##### Issue type

 - Bug report
 
##### Minetest version

```
0.4.17.1
5.0.0dev
```

##### OS / Hardware

Operating system: Windows 10
CPU: Eight-core processor, 3.00 GHz

##### Summary

Sometimes, for seemingly no reason, singleplayer will suddenly become unplayably laggy. Playing on a server is also affected. By laggy, I mean mapblocks take a ridiculously long time to load, commands take minutes to execute, and opening the inventory is not even usually an option. This is the worst-case scenario; usually it starts out mild and slowly grows to unplayability. To be clear, I have a good, fast computer, a reliable network connection, and not many other open processes. Rebooting the computer seems to fix the problem temporarily, but it still happens frequently enough for it to be annoying. This happens on 0.4.xx as well as the latest development build.
I noticed these types of error messages in mass quantity when I turned on verbose logging, perhaps they have something to do with it:

```
VERBOSE[ConnectionSend]: con(10056/1)RE-SENDING timed-out RELIABLE to 127.0.0.1(t/o=3): from_peer_id=1, channel=2, seqnum=19
VERBOSE[ConnectionSend]: con(10056/1)RE-SENDING timed-out RELIABLE to 127.0.0.1(t/o=3): from_peer_id=1, channel=2, seqnum=40
VERBOSE[ConnectionReceive]: con(10112/2)WARNING: ACKed packet not in outgoing queue
VERBOSE[ConnectionSend]: con(10056/1)RE-SENDING timed-out RELIABLE to 127.0.0.1(t/o=3): from_peer_id=1, channel=0, seqnum=150
VERBOSE[ConnectionSend]: con(10056/1)RE-SENDING timed-out RELIABLE to 127.0.0.1(t/o=3): from_peer_id=1, channel=2, seqnum=26
VERBOSE[ConnectionReceive]: con(10112/2)WARNING: ACKed packet not in outgoing queue
VERBOSE[ConnectionReceive]: con(10112/2)WARNING: ACKed packet not in outgoing queue
VERBOSE[ConnectionReceive]: con(10112/2)WARNING: ACKed packet not in outgoing queue
VERBOSE[ConnectionSend]: con(10112/2)RE-SENDING timed-out RELIABLE to 127.0.0.1(t/o=3): from_peer_id=2, channel=2, seqnum=5
VERBOSE[ConnectionSend]: con(10112/2)RE-SENDING timed-out RELIABLE to 127.0.0.1(t/o=3): from_peer_id=2, channel=2, seqnum=6
VERBOSE[ConnectionReceive]: con(10112/2)WARNING: ACKed packet not in outgoing queue
VERBOSE[ConnectionSend]: con(10056/1)RE-SENDING timed-out RELIABLE to 127.0.0.1(t/o=3): from_peer_id=1, channel=2, seqnum=20
VERBOSE[ConnectionSend]: con(10056/1)RE-SENDING timed-out RELIABLE to 127.0.0.1(t/o=3): from_peer_id=1, channel=2, seqnum=21
VERBOSE[ConnectionSend]: con(10056/1)RE-SENDING timed-out RELIABLE to 127.0.0.1(t/o=3): from_peer_id=1, channel=2, seqnum=27
VERBOSE[ConnectionReceive]: con(10112/2)WARNING: ACKed packet not in outgoing queue
VERBOSE[ConnectionSend]: con(10112/2)RE-SENDING timed-out RELIABLE to 127.0.0.1(t/o=3): from_peer_id=2, channel=2, seqnum=7
VERBOSE[ConnectionSend]: con(10056/1)RE-SENDING timed-out RELIABLE to 127.0.0.1(t/o=3): from_peer_id=1, channel=2, seqnum=41
VERBOSE[ConnectionReceive]: con(10112/2)WARNING: ACKed packet not in outgoing queue
```

I doubt this is caused by 968ce9af598024ec71e9ffb2d15c3997a13ad754 as it seems to happen on any version. I'm guessing this will be marked as an unconfirmed bug since it is essentially un-reproducible, but I thought I would report it anyways.

### LINKED ISSUE #9111: [performance] UDP Receive exhaustion
##### Minetest version
```
0.4.16 - 5.1.0
```

##### Summary
Currently, in the `ServerThread::run()` method the `Receive()` method is only called once after the main Lua loop (server step):

https://github.com/minetest/minetest/blob/master/src/server.cpp#L100
```cpp
	while (!stopRequested()) {
		try {
			m_server->AsyncRunStep();

			m_server->Receive();

		}
```

The `Receive()` method collects the incoming packets from the receive thread itself to process it in the main thread.

The problem is, that if the server is under heavy load (serverstep over 100ms for example) the method gets only called once after the current server-step finished.
This is due to the `dtime` check here:

https://github.com/minetest/minetest/blob/master/src/server.cpp#L490
```cpp
	if((dtime < 0.001) && !initial_step)
		return;
```

If this dtime is not below 0.001 seconds the `AsyncRunStep` does not return early and the `Receive` method has no chance to keep up with the inflow of packets/events.

This caused some very weird bugs on my server, for example:
* All mobs run around but there is no way to interact, eat or move inventory
* IRC Chat works from the server to the client but the clients can't send chat messages back

Some of those "interruptions" can go up to a couple minutes!
**The fun thing is:** that this apparent lag is not noticeable in the server logs or in `/status` because it is caused by the packets arriving too late.

##### Related issues
* #8584 lockup due to Pathfinder increased globalstep time
* I'm sure #8207 is caused by some of this
* #8183
* #7418
* #6009 (@VanessaE 's servers)

##### Mitigation

I started to mitigate this effect by calling the `Receive()` method for a fixed amount every serverstep: https://github.com/pandorabox-io/minetest/commit/af6956e7d2c5fae3151ddf5d769d52b81f9a7270

But i think there has to be a better solution than this.. :confounded: 

##### Steps to reproduce
* Add lots of technic/automation mods to the server (`mesecons`, `technic`, `pipeworks`)
* Invite 10 to 20 players to play and build
* Observe degraded serverstep and the above described effect

sfan5:
should fix #9111, <i>might</i> fix #8207 
see also #8584, #6009

## To do

### DIFF
--- a/src/network/connection.cpp
+++ b/src/network/connection.cpp
@@ -1323,16 +1323,21 @@ void Connection::Disconnect()
 	putCommand(c);
 }
 
-void Connection::Receive(NetworkPacket* pkt)
+bool Connection::Receive(NetworkPacket *pkt, u32 timeout)
 {
+	/*
+		Note that this function can potentially wait infinitely if non-data
+		events keep happening before the timeout expires.
+		This is not considered to be a problem (is it?)
+	*/
 	for(;;) {
-		ConnectionEvent e = waitEvent(m_bc_receive_timeout);
+		ConnectionEvent e = waitEvent(timeout);
 		if (e.type != CONNEVENT_NONE)
 			LOG(dout_con << getDesc() << ": Receive: got event: "
 					<< e.describe() << std::endl);
 		switch(e.type) {
 		case CONNEVENT_NONE:
-			throw NoIncomingDataException("No incoming data");
+			return false;
 		case CONNEVENT_DATA_RECEIVED:
 			// Data size is lesser than command size, ignoring packet
 			if (e.data.getSize() < 2) {
@@ -1340,7 +1345,7 @@ void Connection::Receive(NetworkPacket* pkt)
 			}
 
 			pkt->putRawPacket(*e.data, e.data.getSize(), e.peer_id);
-			return;
+			return true;
 		case CONNEVENT_PEER_ADDED: {
 			UDPPeer tmp(e.peer_id, e.address, this);
 			if (m_bc_peerhandler)
@@ -1358,7 +1363,19 @@ void Connection::Receive(NetworkPacket* pkt)
 					"(port already in use?)");
 		}
 	}
-	throw NoIncomingDataException("No incoming data");
+	return false;
+}
+
+void Connection::Receive(NetworkPacket *pkt)
+{
+	bool any = Receive(pkt, m_bc_receive_timeout);
+	if (!any)
+		throw NoIncomingDataException("No incoming data");
+}
+
+bool Connection::TryReceive(NetworkPacket *pkt)
+{
+	return Receive(pkt, 0);
 }
 
 void Connection::Send(session_t peer_id, u8 channelnum,
--- a/src/network/connection.h
+++ b/src/network/connection.h
@@ -771,6 +771,7 @@ public:
 	bool Connected();
 	void Disconnect();
 	void Receive(NetworkPacket* pkt);
+	bool TryReceive(NetworkPacket *pkt);
 	void Send(session_t peer_id, u8 channelnum, NetworkPacket *pkt, bool reliable);
 	session_t GetPeerID() const { return m_peer_id; }
 	Address GetPeerAddress(session_t peer_id);
@@ -802,6 +803,8 @@ protected:
 
 	UDPSocket m_udpSocket;
 	MutexedQueue<ConnectionCommand> m_command_queue;
+
+	bool Receive(NetworkPacket *pkt, u32 timeout);
 
 	void putEvent(ConnectionEvent &e);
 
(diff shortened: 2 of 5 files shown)


src/network/connection.cpp
```diff
@@ -1358,7 +1363,19 @@ void Connection::Receive(NetworkPacket* pkt)
...
+}
+
+void Connection::Receive(NetworkPacket* pkt)
+{
+	bool any = Receive(pkt, m_bc_receive_timeout);
+	if (!any)
+		throw NoIncomingDataException("No incoming data");
+}
+
+bool Connection::TryReceive(NetworkPacket* pkt)
```
SmallJoker: Would you please be so nice to correct the pointers in all touched lines?

thomasrudin:
this patch is now live on the pandorabox server, i'll keep you posted if there are any issues... :+1:

thomasrudin:
Thanks @sfan5 for that patch, works without problems since 2 days with 10+ players :+1:

PR #9121, opened 2019-11-17
Formspec: Fix priorities for version < 3

### LINKED ISSUE #9103: Issues with formspec ordering
`git bisect` confirmed that 5506e97ed897dde2d4820 causes at least two regressions:

- The formspec ordering is enforced regardless of the `formspec_version` (supposed to take effect only for >= 3).
- The default tooltips associated with the `item_image_button` aren't showing anymore.

You can test with `craftguide`: https://github.com/minetest-mods/craftguide

In this example, the `image` is declared before the `item_image_button` element:  https://github.com/minetest-mods/craftguide/blob/master/init.lua#L848-L856

`formspec_version` ain't specified in `craftguide` so it's assumed to be `1`. Regardless of that, `image` is drawn on top of `item_image_button`.

**Cc:** @DS-Minetest @SmallJoker

SmallJoker:
Fixes #9103 by assigning the draw priorities within `FieldSpec`.

## To do

This PR is Ready for Review.

## How to test

1. Get a build prior commit 5506e97
2. Get craftguide: https://github.com/minetest-mods/craftguide/
3. Compare before (old build) and after (master + this PR)
4. It MUST look exactly identical

Other test code:
```Lua

minetest.register_chatcommand("fstest", {
	func = function(name)
		minetest.show_formspec(name, "lab:testy",
			-- SWITCH BETWEEN THESE TWO
			"formspec_version[2]"..
			--"formspec_version[3]"..

			"size[10,8]"..
			"no_prepend[]"..
			"button_exit[0,0;1,1;quit;X1]"..
				"button[1,4;1,1;RB;Foo]"..
				"item_image_button[1,2;1,1;default:dirt;a;Bar]"..
			"tooltip[RB;Some regular button]"..
			"bgcolor[#F005;false]".. -- > 2: Cover, < 3: behind everything
				"item_image_button[3,2;1,1;default:snow;b;Baz]"..
				"button[3,4;1,1;d;Zat]"..
			"image_button_exit[2,0;1,1;default_wood.png;quit2;X2]"..

			"image_button[2,3;1,1;default_stone.png;IB;Stone]"..
			"item_image[4,4;1,1;default:cobble]"..
			"image[4.5,4.5;1,1;default_brick.png]".. -- > 2: Cover, < 3: behind item image
			"dropdown[6,2;2;d;A,B,C,D;2]"..
			"tooltip[IB;Some image button]"..
			"box[-1,-1;8,7;#0FFA]"
		)
	end
})
```

### DIFF
--- a/src/gui/guiFormSpecMenu.cpp
+++ b/src/gui/guiFormSpecMenu.cpp
@@ -21,8 +21,8 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include <cstdlib>
 #include <algorithm>
 #include <iterator>
-#include <sstream>
 #include <limits>
+#include <sstream>
 #include "guiFormSpecMenu.h"
 #include "constants.h"
 #include "gamedef.h"
@@ -427,7 +427,8 @@ void GUIFormSpecMenu::parseList(parserData* data, const std::string &element)
 			"",
 			L"",
 			L"",
-			258 + m_fields.size()
+			258 + m_fields.size(),
+			3
 		);
 
 		v2s32 pos;
@@ -671,7 +672,8 @@ void GUIFormSpecMenu::parseImage(parserData* data, const std::string &element)
 			name,
 			L"",
 			L"",
-			258 + m_fields.size()
+			258 + m_fields.size(),
+			1
 		);
 		core::rect<s32> rect(pos, pos + geom);
 		gui::IGUIImage *e = Environment->addImage(rect, this, spec.fid, 0, true);
@@ -750,7 +752,8 @@ void GUIFormSpecMenu::parseItemImage(parserData* data, const std::string &elemen
 			"",
 			L"",
 			L"",
-			258 + m_fields.size()
+			258 + m_fields.size(),
+			2
 		);
 		spec.ftype = f_ItemImage;
 
@@ -1654,7 +1657,8 @@ void GUIFormSpecMenu::parseLabel(parserData* data, const std::string &element)
 				"",
 				wlabel_colors,
 				L"",
-				258 + m_fields.size()
+				258 + m_fields.size(),
+				4
 			);
 			gui::IGUIStaticText *e = gui::StaticText::add(Environment,
 					spec.flabel.c_str(), rect, false, false, this, spec.fid);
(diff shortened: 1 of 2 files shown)


SmallJoker:
For comparison, see https://github.com/minetest/minetest/issues/9103#issuecomment-553024197

@kilbith, @DS-Minetest

src/gui/guiFormSpecMenu.cpp
```diff
@@ -2987,59 +2994,42 @@ void GUIFormSpecMenu::regenerateGui(v2u32 screensize)
...
-			i++;
-		}
+	// 1: Copy into a sortable container
+	std::vector<IGUIElement *> elements;
+	for (auto it = from; it != to; ++it)
+		elements.emplace_back(*it);
+
+	// 2: Sort the container
+	std::sort(elements.begin(), elements.end(),
+			[=] (const IGUIElement *a, const IGUIElement *b) -> bool {
```
DS-Minetest: Is the capture-default needed?
```suggestion
			[] (const IGUIElement *a, const IGUIElement *b) -> bool {
```

SmallJoker: Yes. `[]` would isolate it entirely from the `guiFormSpec` class, meaning it has no access to `getSpecByID`.

DS-Minetest:
I've tested a bit (with current master, this PR and 5.1 stable using craftguide master).
The only difference I noticed (apart from features that weren't implemented in 5.1 stable, i.e. button highlighting) is that the search image buttons are a bit stretched in 5.1 stable:
This PR:
2019-11-17-162320_1920x1080_scrot
5.1-stable:
2019-11-17-162316_1920x1080_scrot

SmallJoker:
@DS-Minetest Indeed, the scaling seems to have changed somewhere between 3fcf3f1 (26 Sept) and 13 Oct (no valid commit).
Aside from that it seems to work as intended, right?

DS-Minetest:
>Aside from that it seems to work as intended, right?

Yes.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -4403,13 +4393,13 @@ std::string GUIFormSpecMenu::getNameByID(s32 id)
 }
 
 
-FormspecFieldType GUIFormSpecMenu::getTypeByID(s32 id)
+GUIFormSpecMenu::FieldSpec *GUIFormSpecMenu::getSpecByID(s32 id)
```
sfan5: should return const pointer

src/gui/guiFormSpecMenu.cpp
sfan5: does `[this] (const IGUIElement *a, const IGUIElement *b) -> bool` work here?

SmallJoker: Yes.

PR #9122, opened 2019-11-17
Allow relative directories for `screenshot_path`, tweak default path

Calinou:
This will likely be more intuitive for users and should play better with sandboxed distributions such as Flatpak.

In addition, the screenshot directory will now be created if it doesn't exist already.

PS: Should I update `minetest.conf.example`? If not, its header should be changed to denote that it's a generated file (it's not clear right now).

## To do

## How to test

Build it, start Minetest and take screenshots :slightly_smiling_face:

### DIFF
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -902,8 +902,9 @@ fallback_font_shadow_alpha (Fallback font shadow alpha) int 128 0 255
 #    This font will be used for certain languages or if the default font is unavailable.
 fallback_font_path (Fallback font path) filepath fonts/DroidSansFallbackFull.ttf
 
-#    Path to save screenshots at.
-screenshot_path (Screenshot folder) path
+#    Path to save screenshots at. Can be an absolute or relative path.
+#    The folder will be created if it doesn't already exist.
+screenshot_path (Screenshot folder) path screenshots
 
 #    Format of screenshots.
 screenshot_format (Screenshot format) enum png png,jpg,bmp,pcx,ppm,tga
--- a/src/client/client.cpp
+++ b/src/client/client.cpp
@@ -1774,12 +1774,23 @@ void Client::makeScreenshot()
 	char timetstamp_c[64];
 	strftime(timetstamp_c, sizeof(timetstamp_c), "%Y%m%d_%H%M%S", tm);
 
-	std::string filename_base = g_settings->get("screenshot_path")
+	std::string screenshot_dir;
+
+	if (fs::IsPathAbsolute(g_settings->get("screenshot_path")))
+		screenshot_dir = g_settings->get("screenshot_path");
+	else
+		screenshot_dir = porting::path_user + DIR_DELIM + g_settings->get("screenshot_path");
+
+	std::string filename_base = screenshot_dir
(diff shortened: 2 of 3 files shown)


builtin/settingtypes.txt
```diff
@@ -903,6 +903,9 @@ fallback_font_shadow_alpha (Fallback font shadow alpha) int 128 0 255
 fallback_font_path (Fallback font path) filepath fonts/DroidSansFallbackFull.ttf
 
 #    Path to save screenshots at.
+#    If starting with a slash, it will be considered as an absolute path.
```
SmallJoker: Windows be like: `C:\path\to\minetest\`
Does not start with a slash but is an absolute path.

builtin/settingtypes.txt
```diff
@@ -902,7 +902,8 @@ fallback_font_shadow_alpha (Fallback font shadow alpha) int 128 0 255
 #    This font will be used for certain languages or if the default font is unavailable.
 fallback_font_path (Fallback font path) filepath fonts/DroidSansFallbackFull.ttf
 
-#    Path to save screenshots at.
+#    Path to save screenshots at. Can be an absolute or relative path.
+#    The folder will be created if it doesn't already exist.
 screenshot_path (Screenshot folder) path
```
SmallJoker: Needs update according to `defaultsettings.cpp` changes.

src/client/client.cpp
```diff
@@ -1774,13 +1774,25 @@ void Client::makeScreenshot()
 	char timetstamp_c[64];
 	strftime(timetstamp_c, sizeof(timetstamp_c), "%Y%m%d_%H%M%S", tm);
 
-	std::string filename_base = g_settings->get("screenshot_path")
+	std::string screenshot_dir;
+
+	if (fs::IsPathAbsolute(g_settings->get("screenshot_path"))) {
+		screenshot_dir = g_settings->get("screenshot_path");
+	} else {
```
SmallJoker: Brackets can be omitted.

src/client/client.cpp
```diff
@@ -1774,13 +1774,25 @@ void Client::makeScreenshot()
...
+		screenshot_dir = porting::path_user + DIR_DELIM + g_settings->get("screenshot_path");
+	}
+
+	std::string filename_base = screenshot_dir
 			+ DIR_DELIM
```
SmallJoker: What happens if this becomes `C:\path\to\screenshots\\`, or `/path/to/screenshots//`?

Calinou: Is there a "plus_file" method of sorts that can be used to avoid issues with trailing slashes?

SmallJoker: At least Linux removes duplicated DIR_DELIMs automatically. I hope Windows does the same.

Calinou:
Bump :slightly_smiling_face:

src/client/client.cpp
v-rob: Windows deletes trailing slashes and backslashes, but it does not delete multiple slashes in the middle, e.g. `path/to/dir//`,`path\to\dir\\`, and `path/to\dir/\` all work, but `path//to/dir` doesn't.

PR #9123, opened 2019-11-18
Add PostgreSQL Auth Backend

rndevuk:
This commit enables the ability to use PostgreSQL as a Authentication backend.

### DIFF
--- a/src/database/database-postgresql.cpp
+++ b/src/database/database-postgresql.cpp
@@ -74,6 +74,20 @@ void Database_PostgreSQL::connectToDatabase()
 	m_pgversion = PQserverVersion(m_conn);
 
 	/*
+	* We are using CITEXT extension from PostgreSQL 8.4 where possible.
+	* It must be added to each database that wishes to use it and in most
+	* cases it is usually not done by default, so we do it here.
+	*/
+	if (m_pgversion >= 80400) {
+		std::string ext_query = "CREATE extension if not exists citext;";
+		checkResults(PQexec(m_conn, ext_query.c_str()));
+	} else {
+		warningstream << "Your PostgreSQL server lacks CITEXT "
+			<< "support. Use version 8.4 or better if possible."
+			<< std::endl;
+	}
+
+	/*
 	* We are using UPSERT feature from PostgreSQL 9.5
 	* to have the better performance where possible.
 	*/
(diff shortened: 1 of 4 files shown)


crazyR-MT:
I'm not a c++ programmer so if there is anything i can do to improve this then feel free to let me know below and i will do my best to accommodate.

I couldn't find a reference for database connection strings in the `world_format.txt` so i didn't add it into there.

The connection string is same format as other PostgreSQL  Features `pgsql_auth_connection = host=<host> user=<user> password=<password> dbname=<dbname>`

MoNTE48:
Very well.
But also code for migration would be useful

crazyR-MT:
migration is already in and didn't need amending.

src/database/database-postgresql.cpp
```diff
@@ -631,4 +645,159 @@ void PlayerDatabasePostgreSQL::listPlayers(std::vector<std::string> &res)
...
+}
+
+void AuthDatabasePostgreSQL::createDatabase()
+{
+	if (getPGVersion() < 80400) {
+		createTableIfNotExists("auth",
+			"CREATE TABLE auth ("
+				"id SERIAL PRIMARY KEY,"
+				"name TEXT UNIQUE,"
+				"password VARCHAR(512),"
```
nerzhul: please use TEXT not VARCHAR.

src/database/database-postgresql.cpp
```diff
@@ -631,4 +645,159 @@ void PlayerDatabasePostgreSQL::listPlayers(std::vector<std::string> &res)
...
+				"privileges TEXT,"
```
nerzhul: please use a dedicated proper table to store privileges

src/database/database-postgresql.cpp
```diff
@@ -631,4 +645,159 @@ void PlayerDatabasePostgreSQL::listPlayers(std::vector<std::string> &res)
...
+				"last_login INT"
```
nerzhul: instead of using a INT, use a timestamp

src/database/database-postgresql.cpp
nerzhul: also add NOT NULL constraint on required fields

crazyR-MT:
Thanks for the feedback @nerzhul. i will start work on these changes over the weekend.

PR #9133, opened 2019-11-19
Use a safer implementation of gsub in core.chat_format_message

### LINKED ISSUE #9132: Crash when certain character sequences are entered in chat
##### Minetest version
```
5.1.0 through dev-9776460
```

##### OS / Hardware
Linux Server

##### Summary
If a player enters e.g. `%2` in chat, the game crashes with the following trace:

```
minetest5/src/script/cpp_api/s_base.cpp:89: static int ScriptApiBase::luaPanic(lua_State*): A fatal error occurred: LUA PANIC: unprotected error in call to Lua API (...est/minetest_exec/minetest5/bin/../builtin/game/chat.lua:18: invalid capture index)
```

Entering `%1` results in `@message`. Basically, `%#` is being interpreted as a capture index, and if the index is greater than the number of captures, it crashes. The solution is probably to escape `%`s.  However, as this doesn't affect Minetest 5.0, I expect someone will want to figure out why. 

##### Steps to reproduce
Create a fresh server instance w/ the affected versions, launch it, type `%2` in chat.

ClobberXD:
This PR fixes the following:
- Crash, when a chat-message contains `%n`, where `n > 1`
- Replacing `@timestamp` with the timestamp, if present in the actual chat message

Fixes #9132.

Tested; works. To test, enter the following messages in chat, and observe that all of them are displayed exactly as you entered them:

- `blah %10 blah`
- `blah % blah`
- `blah 50% blah`
- `blah %d blah`
- `blah blah @timestamp`

### DIFF
--- a/builtin/game/chat.lua
+++ b/builtin/game/chat.lua
@@ -1,26 +1,40 @@
 -- Minetest: builtin/game/chat.lua
 
+-- Helper function that implements search and replace without pattern matching
+-- Returns the string and a boolean indicating whether or not the string was modified
+local function safe_gsub(s, replace, with)
+	local i1, i2 = s:find(replace, 1, true)
+	if not i1 then
+		return s, false
+	end
+
+	return s:sub(1, i1 - 1) .. with .. s:sub(i2 + 1), true
+end
+
 --
 -- Chat message formatter
 --
 
 -- Implemented in Lua to allow redefinition
 function core.format_chat_message(name, message)
-	local str = core.settings:get("chat_message_format")
 	local error_str = "Invalid chat message format - missing %s"
-	local i
+	local str = core.settings:get("chat_message_format")
+	local replaced
 
-	str, i = str:gsub("@name", name, 1)
-	if i == 0 then
+	-- Name
+	str, replaced = safe_gsub(str, "@name", name)
+	if not replaced then
 		error(error_str:format("@name"), 2)
 	end
 
-	str, i = str:gsub("@message", message, 1)
-	if i == 0 then
+	-- Timestamp
+	str = safe_gsub(str, "@timestamp", os.date("%H:%M:%S", os.time()))
+
+	-- Insert the message into the string only after finishing all other processing
+	str, replaced = safe_gsub(str, "@message", message)
+	if not replaced then
 		error(error_str:format("@message"), 2)
 	end
-
-	str = str:gsub("@timestamp", os.date("%H:%M:%S", os.time()), 1)
 
 	return str
 end


sfan5:
I would prefer if this was fixed in such a way that user input is not passed to gsub, even if escaped. (by reimplementing a safe variant)

ClobberXD:
> I would prefer if this was fixed in such a way that user input is not passed to gsub, even if escaped.

`gsub` is used because the message can end up anywhere in the resulting string, based on the chat-message format, and so `..` can't be used. Is there any other way to do this?

> (by reimplementing a safe variant)

I don't understand. Could you please elaborate?

sfan5:
> Could you please elaborate?

```lua
function safe_gsub(s, replace, with)
    local i1, i2 = s:find(replace, 1, true)
    if i1 == nil then return s end
    return s:sub(1, i1 - 1) .. with .. s:sub(i2 + 1)
end
```

sfan5:
Also there's another issue that should be solved here:
If you enter "@timestamp" in chat, the current time is displayed in chat.

ClobberXD:
> If you enter "@timestamp" in chat, the current time is displayed in chat.

Bug? Or a nifty feature to allow players to display the server time in chat? :D
Seriously though, that's indeed an issue.

But how is the "safe" version of `gsub` different from the original `gsub`?

fluxionary:
> But how is the "safe" version of `gsub` different from the original `gsub`?

Because printf is a treacherous game that many players have lost in the past.

ClobberXD:
Ah, so `gsub` uses `printf` (or `sprintf`?) internally. Thanks for the info.

ClobberXD:
- This PR now uses @sfan5's `safe_gsub` implementation (exposed as a `minetest.` function for use in mods) in favour of `string.gsub`.
- Timestamp, although optional, is handled before the message is inserted, to prevent `@timestamp` in messages from being tokenised. Now, the message is inserted into the format string only after all other processing has been done.
- The `message:gsub("%%", "%%%%")` has been removed, as it has become redundant.

Tested - `%` doesn't cause weird behaviour or crashes, and `@timestamp` is displayed as-is. Note that I'm yet to add documentation for `core.safe_gsub`. This PR is otherwise ready for review.

sfan5:
> But how is the "safe" version of gsub different from the original gsub?

It searches for the string using pattern matching disabled (not important in this case) and it replaces it without interpreting the replacement string in any way.

builtin/game/chat.lua
```diff
@@ -1,27 +1,40 @@
 -- Minetest: builtin/game/chat.lua
 
+-- Helper function made global for use in mods
```
sfan5: this needs to be documented if kept global
but I don't think it makes much sense to expose anyway

builtin/game/chat.lua
```diff
@@ -1,27 +1,40 @@
...
-	local str = core.settings:get("chat_message_format")
 	local error_str = "Invalid chat message format - missing %s"
-	local i
+	local str = core.settings:get("chat_message_format")
+	local old_str = str
 
-	str, i = str:gsub("@name", name, 1)
-	if i == 0 then
+	str = core.safe_gsub(old_str, "@name", name)
+	if old_str == str then
```
sfan5: surely it would have been easier to extend `safe_gsub` to return 0 or 1 instead of doing this string comparison stuff

ClobberXD: Yes, indeed...

builtin/game/chat.lua
ClobberXD: > this needs to be documented if kept global

> Note that I'm yet to add documentation for core.safe_gsub. This PR is otherwise ready for review.

> but I don't think it makes much sense to expose anyway

Wouldn't mods benefit from this function when faced with a similar to use this for similar purposes? We do expose many other useful functions for mods this way, like `table.indexof` and the like.

sfan5: Our `safe_gsub` only does one substitution at a time. Modders may want behaviour that is closer to the original instead.

ClobberXD: You're right, I'll make the function local.

builtin/game/chat.lua
ClobberXD: Done

PR #9187, opened 2019-12-11
StaticText/EnrichedString: Styling support

SmallJoker:
`StaticText` was not capable of styling, thus `GUIButton` does not work either any more (since 9284313).
This PR adds styling support to `EnrichedString` and makes use of its properties in `StaticText`.

Changes in `EnrichedString`:
 * "Override color" (from Styling) is used as default color until the color changes (escapes)
 * The length of the default color is kept and extended on concat operations
   * There's yet no "reset color" escape sequence, thus a single integer is enough for now.

Changes in `StaticText`:
 * Get and set color properties directly using `EnrichedString`
   * "duplicated" variables were removed
 * Unused features and broken features (bgcolor, override bgcolor) trimmed/simplified
 * Remove duplicated text due to `WordWrap`. Text format unified in `updateText()`

## To do

This PR is Ready for Review.

- [x] Testing using the minimal `/formspec` "unittest"
- [x] Testing using a custom formspec (textareas, labels, buttons)
- [x] Testing using craftguide
- [ ] ?? find more useful test formspecs

## How to test

1) Get an older Minetest build (5.1.0, or before formspec_version 3)
2) Get craftguide, minimal game, custom testing formspecs
3) Open the formspecs
4) Compare PR/old -> buttons and colors must look the same or more correct

### DIFF
--- a/games/minimal/mods/test/formspec.lua
+++ b/games/minimal/mods/test/formspec.lua
@@ -1,3 +1,5 @@
+local color = minetest.colorize
+
 local clip_fs = [[
 	style_type[label;noclip=%c]
 	style_type[button;noclip=%c]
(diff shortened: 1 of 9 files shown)


kilbith:
It's worth noting that the default background color of the tooltip changed from opaque green to translucent white, which isn't bad IMO.

Otherwise works :+1:

SmallJoker:
@kilbith Thank you for this report. It turned out that `tooltip[` was broken because `setStaticText` cleared the override settings. Fixed that by converting the tooltip string to an `EnrichedString` directly.

Tested tooltips with `freetype = true` (not shown)
 * Foreground: yellow - OK
 * Background: red - OK
And `freetype = false` (see image below)
 * Foreground: not colorized (unsupported) OK
 * Background: red - OK

grafik

Styling also works (freetype enabled)
grafik

For additional safety I added a new unittest function for EnrichedString (passed).

sfan5 (approved):
Works, code looks okay too
:+1: for unittests

PR #9208, opened 2019-12-16
Correct documentation around minetest.item_place

sfan5:
## To do

This PR is Ready for Review.

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -4851,10 +4851,8 @@ Rollback
     * Revert latest actions of someone
     * `actor`: `"player:<name>"`, also `"liquid"`.
 
-Defaults for the `on_*` item definition functions
--------------------------------------------------
-
-These functions return the leftover itemstack.
+Defaults for the `on_place` and `on_drop` item definition functions
+-------------------------------------------------------------------
 
 * `minetest.item_place_node(itemstack, placer, pointed_thing[, param2, prevent_after_place])`
     * Place item as a node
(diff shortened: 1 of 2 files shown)


doc/lua_api.txt
```diff
@@ -4864,20 +4862,22 @@ These functions return the leftover itemstack.
     * returns `itemstack, success`
 * `minetest.item_place_object(itemstack, placer, pointed_thing)`
     * Place item as-is
-* `minetest.item_place(itemstack, placer, pointed_thing, param2)`
-    * Use one of the above based on what the item is.
+    * returns the leftover itemstack
+    * **Note**: This is never called and exists solely for compatibility.
```
SmallJoker: How about: `**Note**: This function is deprecated.`
`never called` is not so accurate, and "deprecated" would kind of indicate that it's only there for compatibility reasons.

doc/lua_api.txt
```diff
@@ -4864,20 +4862,22 @@ These functions return the leftover itemstack.
...
-    * `param2` overrides `facedir` and wallmounted `param2`
+    * `param2` overrides facedir and wallmounted `param2`
     * returns `itemstack, success`
 * `minetest.item_drop(itemstack, dropper, pos)`
     * Drop the item
-* `minetest.item_eat(hp_change, replace_with_item)`
-    * Eat the item.
+    * returns the leftover itemstack
+* `minetest.item_eat(hp_change[, replace_with_item])`
+    * Returns a function that calls `minetest.do_item_eat`
```
SmallJoker: Suggestion: `Returns 'function(itemstack, user, pointed_thing)' as a function wrapper for 'minetest.do_item_eat'.` (' -> `)

doc/lua_api.txt
sfan5: "Deprecated" however does not imply that the function is never called

PR #9211, opened 2019-12-16
Camera: Smoother collision using Raycast

### LINKED ISSUE #2276: 3rd person camera shake in confined spaces

SmallJoker:
Fixes #2276. Currently the F7 camera "collision detection" is very rough and does not look nice at all. This PR makes it really smooth.
Code taken from my previous PR #8015

## To do

This PR is Ready for Review.

## How to test

1) F7
2) Compare smoothness of the camera collision

### DIFF
--- a/src/client/camera.cpp
+++ b/src/client/camera.cpp
@@ -430,37 +430,41 @@ void Camera::update(LocalPlayer* player, f32 frametime, f32 busytime, f32 tool_r
 				break; // POINTEDTHING_OBJECT is unavailable, skip
 
 			MapNode n = map.getNode(pointed.node_real_undersurface);
-			if (nodemgr->get(n).walkable) {
-				if (pointed.distanceSq < 0.001f)
-					break; // Head in node
+			if (!nodemgr->get(n).walkable)
+				continue;
 
-				my_cp = pointed.intersection_point;
+			if (pointed.distanceSq < 0.001f)
+				break; // Head in node
 
-				float distance = std::sqrt(pointed.distanceSq);
-				if (!near_offset && distance < near_distance) {
-					// Collision is too near to the head: Start Raycast above head
-					near_offset = true;
-					shootline_end = my_cp - m_camera_direction * near_distance;
-					ray = RaycastState(core::line3d<f32>(m_camera_position +
-						v3f(0.0f, 0.4f * (near_distance - distance), 0.0f), shootline_end),
-						false, false);
-				} else {
-					// Calculate intersection angle to clip "propery" with walls
-					float collision_d = m_camera_direction.dotProduct(
-						intToFloat(pointed.intersection_normal, 1.0f));
-					collision_d = std::acos(collision_d) / M_PI; // -0.5 ... 0.5
-					collision_d = std::abs(collision_d) * distance + 0.1f * BS;
+			// First best camera position
+			my_cp = pointed.intersection_point;
 
-					if (collision_d > distance)
-						my_cp = m_camera_position;
-					else if (distance - collision_d > shootline_d)
-						my_cp = shootline_end; // Outside of shootline_d
-					else
-						my_cp += m_camera_direction * collision_d;
-
-					break;
-				}
+			float distance = std::sqrt(pointed.distanceSq);
+			if (!near_offset && distance < near_distance) {
+				// Collision is too near to the head: Start Raycast above head
+				near_offset = true;
+				shootline_end = my_cp - m_camera_direction * near_distance;
+				ray = RaycastState(core::line3d<f32>(m_camera_position +
+					v3f(0.0f, 0.4f * (near_distance - distance), 0.0f), shootline_end),
+					false, false);
+				continue;
 			}
+
+			// Final node collision
+			// Calculate intersection angle to clip "propery" with walls
+			float collision_d = m_camera_direction.dotProduct(
+				intToFloat(pointed.intersection_normal, 1.0f));
+			collision_d = std::acos(collision_d) / M_PI; // -0.5 ... 0.5
+			collision_d = std::abs(collision_d) * distance + 0.1f * BS;
+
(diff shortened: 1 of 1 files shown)


sfan5:
Standing in a 2-node space, previously you'd get this view:
screenshot_20191216_212204

It now gives this view:
screenshot_20191216_212123
or worse this one:
screenshot_20191216_212339

paramat:
There seem to be performance problems with using raycast, as it does some intensive object stuff that is unnecessary for the camera. See https://github.com/minetest/minetest/pull/8015#issuecomment-450529888

"i'm -1 for using raycast, as it is far more complex than the current code, and this is being used in camera update which is run every frame, so can be 60+ FPS.

Unlike the previous code, continueRaycast() detects objects, so calls getSelectedActiveObjects() which loops through all loaded objects to get those on the shootline, gets their selectionboxes, tests for a collision between the shootline and the selectionboxes and even calculates the exact intersection of the shootline with the selectionboxes. Very thorough but complete overkill for this purpose.
This is done every camera update.

Although it is thorough to account for objects it isn't needed, we have no issues with simply testing for 'walkable' and the third person camera being inside objects isn't a problem. 
The current code just tests for walkable and ignores objects."

rubenwardy:
Our raycast is super efficient and is unlikely to cause issues with a low range limit. I suggest showing profiler results to prove this to paramat

SmallJoker:
@paramat FYI: Raycast was added because the old code was _exponentially_ slower the higher `range` values a tool had (used in `game.cpp` IIRC). Now it's likely the best you can get in terms of speed.
I will do some profiling when I found a solution to avoid the camera to be inside the head. From what I can tell: the execution time will be about the same.

paramat:
> Our raycast is super efficient (with only nodes)

I may have misunderstood (no one responded to my comment in the original PR), is it possible to use raycast and disable the object detection? (which is not needed for the camera).
> Raycast was added because the old code was exponentially slower the higher range values a tool had (used in game.cpp IIRC). Now it's likely the best you can get in terms of speed.

I do not doubt this, but this is not in comparison to current camera code, which may be very simple and lightweight.

My concern is that the camera collision code should perhaps be specifically written and optimised for the purpose, it may be able to be much simpler than applying a generalised raycast.
However, i have not studied this in depth.
I would appreciate some profiling, thanks. if it is not significantly more intensive i have no objection.

rubenwardy:
> I may have misunderstood (no one responded to my comment in the original PR), is it possible to use raycast and disable the object detection? (which is not needed for the camera).

Yes. And if it were not, then it would be very easy to add such a feature

https://github.com/minetest/minetest/blob/master/src/raycast.h#L40-L41

https://github.com/minetest/minetest/blob/master/doc/lua_api.txt#L4605-L4610

paramat:
Thanks, i also just found
https://github.com/minetest/minetest/blob/b298b0339c79db7f5b3873e73ff9ea0130f05a8a/src/raycast.h#L40

src/client/camera.cpp
```diff
@@ -403,34 +404,37 @@ void Camera::update(LocalPlayer* player, f32 frametime, f32 busytime, f32 tool_r
...
-			if (features.walkable) {
-				my_cp.X += m_camera_direction.X*-1*-BS/2;
-				my_cp.Z += m_camera_direction.Z*-1*-BS/2;
-				my_cp.Y += m_camera_direction.Y*-1*-BS/2;
-				abort = true;
+		v3f shootline_end = my_cp - m_camera_direction * 2.75f * BS;
+		v3f shootline_extended = shootline_end - m_camera_direction * BS;
+
+		RaycastState ray(core::line3d<f32>(m_camera_position, shootline_extended),
+			true, false);
```
paramat: So, how about `false, false` to remove object checking (which current code does not do)? I guess this depends on profiling.

paramat: In a situation where objects are moving around near the player, perhaps it is better those objects do not move the camera about?

paramat:
> More complex things can (and usually do) perform better than naive approaches

Yes, i should have referred only to intensiveness, not complexity/simplicity.

paramat:
I cannot yet think of any problems caused by having the camera not collide with objects, but there is a (possibly large) performance advantage, and it seems better for objects to not move the camera around (which is current behaviour).
So i suggest that object collision is disabled, we can always enable it later if serious problems arise.

As for the change in behaviour shown by sfan5, that does not seem a big problem, as the view angle can be lowered to prevent that when the player is in a 2 node high space. I would prefer this behaviour if it measn we can use a simple raycast.

rubenwardy:
From a quick glance, the current code doesn't check and collide with entities either

paramat:
Function:
https://github.com/minetest/minetest/blob/b298b0339c79db7f5b3873e73ff9ea0130f05a8a/src/raycast.h#L40-L41
Call in PR:
```
RaycastState ray(core::line3d<f32>(m_camera_position, shootline_extended),
			true, false);
```
?

SmallJoker:
Before: 5 us maximal (3 us average)
After: 1365 us maximal (600 us average)

Is this the death of this PR?

paramat:
Wow i didn't expect that much difference. I feel a little better about making a fuss now =)
Can you confirm that object collisions are enabled? They seem to be looking at the bools in the call, if so does disabling them make a difference?
Certainly, the performance drop outweighs the benefits.

SmallJoker:
@paramat Object detection is already disabled in commit 2/3 and 3/3.

paramat:
Ah sorry, i did not see the new commits just before your profiling.

SmallJoker:
There's likely no future for this PR.

PR #9223, opened 2019-12-19
Add chat_log_level setting

### LINKED ISSUE #9199: Option to display more debug log output in chat
## Problem

I'm always frustrated when warnings don't appear in chat. Currently, only errors appear here. In development, I find it inconvenient to be forced to always look back in my console/log to find warnings, and I frequently overlook warnings that might be important to me, especially after a long testing session (switching back and forth between game and console (the real console, not the in-game one!) isn't fun).

If the game would tell me directly in my face all those warnings, it would make things easier.

## Solutions

Add a setting to configure which debug log messages are allowed to appear in chat. Basically the same as `debug_log_level`, but for chat.

In this setting, you can then choose the minimum log level to put stuff into chat (error, warning, action, info, etc.). E.g. if you select "warning", both errors and warnings will appear.

Also, I suggest to use a yellow font when a warning appears in chat.

## Alternatives

Don't add a setting, but instead show warnings in chat by default. Simpler solution, but might annoy players.

## Additional context

A very long time ago, Minetest displayed warnings in the chat as well, but this feature was removed because players complained (games had a lot of warnings).

Well, that's not really a good fix. The warnings were never fixed, just hidden from view. And IMHO many warnings do expose real problems that should not be ignored.
As a developer, I definitely WANT that Minetest screams right into my face that something went bad, for quality assurance. But on the other hand, I understand the rationale why warnings are no longer in chat. Therefore, I suggested a setting as a compromise.

SmallJoker:
Closes #9199.

 * New setting `chat_log_level` to log to the chat
 * Colored similar to the terminal output (color codes somewhat trivial, debatable)
 * `log.h` cleanup -> move to source file for compiling time

Preview

## To do

This PR is Ready for Review.

## How to test

1) Heat your room by recompiling the entire Minetest source code
2) minetest.conf:
```
chat_log_level = info
```
3) Get spammed.

### DIFF
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -375,6 +375,20 @@ void StreamLogOutput::logRaw(LogLevel lev, const std::string &line)
 	}
 }
 
+void LogOutputBuffer::updateLogLevel()
+{
+	const std::string &conf_loglev = g_settings->get("chat_log_level");
+	LogLevel log_level = Logger::stringToLevel(conf_loglev);
+	if (log_level == LL_MAX) {
+		warningstream << "Supplied unrecognized chat_log_level; "
+			"showing none." << std::endl;
+		log_level = LL_NONE;
+	}
+
+	m_logger.removeOutput(this);
+	m_logger.addOutputMaxLevel(this, log_level);
+}
+
 void LogOutputBuffer::logRaw(LogLevel lev, const std::string &line)
 {
 	std::string color;
@@ -397,7 +411,7 @@ void LogOutputBuffer::logRaw(LogLevel lev, const std::string &line)
 		}
 	}
 
- 	m_buffer.push(color.append(line));
+	m_buffer.push(color.append(line));
 }
 
 ////
--- a/src/log.cpp
+++ b/src/log.cpp
@@ -23,6 +23,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include "debug.h"
 #include "gettime.h"
 #include "porting.h"
+#include "settings.h"
 #include "config.h"
 #include "exceptions.h"
 #include "util/numeric.h"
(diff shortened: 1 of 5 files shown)


src/log.cpp
```diff
@@ -338,7 +339,65 @@ void FileLogOutput::setFile(const std::string &filename, s64 file_size_max)
 		"-------------\n" << std::endl;
 }
 
+void StreamLogOutput::logRaw(LogLevel lev, const std::string &line)
+{
+	bool colored_message = (Logger::color_mode == LOG_COLOR_ALWAYS) ||
+		(Logger::color_mode == LOG_COLOR_AUTO && is_tty);
+	if (colored_message) {
```
SmallJoker: No functional changes. Added braces here according to the code style guidelines.

SmallJoker:
Pinging @Wuzzy2 because he requested this feature.

Wuzzy2:
Thanks, I have tested this for a bit now. So, one problem I found is that Minetest does not guarantee the log message actually appears in chat if that log message appears early at startup time.

I can reliably reproduce this with the following steps:

1. Create a dummy mod with a single line: `minetest.log("warning", "THIS IS A TEST WARNING!")`
2. Set `chat_log_level` to `warning`
3. (Re-)Start Minetest
4. Enable the dummy mod
5. Start Minimal Development Test

The warning should be visible in chat. But it won't be visible. When I restart Minimal Development Test (WITHOUT restarting Minetest!), the warning will then reliably appear on every subsequent restart.
Apparently the log message fails to appear at the startup of game if you started up a game for the first time in the current Minetest session. At least that's how it looks to me. There might something else be going on as well. If you restart Minetest, you will get the same results.

Now some random nitpicks:

> #    Like "Debug log level" but shows log messages in the chat.
> #    Restart Minetest to apply changes.

I don't like descriptions that are written in the form “it works like setting XYZ”. Better write a full standalone description instead, even if it's a little redundant. Why? Because in Minetest, the user will only see one setting description at a time, and it's just inconvenient when you have to hunt down the other setting just to find out what `chat_log_level` does.

There's a leading space before the tab in line 400 of `src/log.cpp`.

SmallJoker:
@Wuzzy2 Thanks for testing. I fixed your reported issues in d8119d592.

Wuzzy2:
Works fine for me now.

Issue #9198, opened 2019-12-14
Display CJK characters correctly in non-fallback languages


zaoqi:
## Problem

cannot display CJK characters

## Solutions

use another font?

## Additional context

(partially) fix https://github.com/minetest/minetest/issues/4638

SmallJoker:
Is Irrlicht capable of reading your inputs? Its input system is broken, thus I wonder whether you could really type non-latin characters.

zaoqi:
I renamed a world to a non-latin name

Enrikoo:
What if someone doesn't like the font and wants a refactor of arimo?

zaoqi:
@Enrikoo it can change font

zaoqi:
@Enrikoo
> What if someone doesn't like the font and wants a refactor of arimo?

If someone doesn't like the default English or CJK font, he can change font files.

zaoqi:
@SmallJoker I found that minetest already has a CJK font, but only enabled when `needs_fallback_font` is yes. I want minetest able to display CJK characters when the language is en.

SmallJoker:
Translator's fault. Will be fixed in next release unless someone reverts the translation change.

zaoqi:
@SmallJoker 
The problem is that minetest cannot display CJK characters when language is en (`needs_fallback_font` is not `yes`)
should I open an new issue?

v-rob:
This shouldn't have a "Bug" label.  Many fonts (Minetest's normal and monospace fonts included) don't support CJK characters.  It really can't be considered a bug since no font has every character in Unicode.  However, one can make it better by doing `font_path = fonts/DroidSansFallbackFull.ttf`.

However, Minetest could try to better handle these problems by doing the following: if an unknown character is encountered, attempt to draw the character with the fallback font.  If that fails, draw a question mark or empty box to indicate that it couldn't be drawn.  However, such a thing is non-trivial since fonts are drawn all over Minetest, such as in the GUI, HUD, nametags, debug text, etc.  In addition, Unicode has the ability to combine characters, so there has to be a way to detect character combinations and render them all with the fallback font together (ugh, Unicode is so complicated, even more so with UTF-8 to worry about).

In other words, this can't reasonably done without reworking the font system completely and making everything use that font system.

sfan5:
I disagree that this is non-trivial. We have a central class (`CGUITTFont`) that already *is* used everywhere, surely it can be made to handle this.

v-rob:
[...]

PR #9227, opened 2019-12-20
Add Noto CJK font

zaoqi:

fix https://github.com/minetest/minetest/issues/9198

- Goal of the PR: correctly disaply CJK characters
- How does the PR work? add font from https://github.com/koreader/koreader-fonts/tree/master/noto
- Does it resolve any reported issue? https://github.com/minetest/minetest/issues/9198 https://github.com/minetest/minetest/issues/4638

## To do

- [ ] relevant C ++ codes


zaoqi:
Noto CJK fonts as fallback fonts only seems to crash or no effect.

SmallJoker:
You may remove `DroidSansFallbackFull.ttf` and all its references as well. It's no longer needed from what I can see.

zaoqi:
@SmallJoker It as fallback font just doesn't change anything.And combining Armio and Noto CJK will break LICENSE

PR #9228, opened 2019-12-21
lua_api.txt: Improve privs_to_string and string_to_privs documentation

ClobberXD:
This documentation PR primarily adds info on the earlier-missing optional param `delim` in both `minetest.string_to_privs` and `minetest.privs_to_string`. Inputs and suggestions welcome.

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -4275,9 +4275,13 @@ Setting-related
 Authentication
 --------------
 
-* `minetest.string_to_privs(str)`: returns `{priv1=true,...}`
-* `minetest.privs_to_string(privs)`: returns `"priv1,priv2,..."`
-    * Convert between two privilege representations
+* `minetest.string_to_privs(str[, delim])`:
+    * Converts string representation of privs into table form
+    * `delim`: String separating the privs. Defaults to `","`.
+    * Returns `{ priv1 = true, ... }`
+* `minetest.privs_to_string(privs[, delim])`:
+    * Returns the string representation of `privs`
+    * `delim`: String to delimit privs. Defaults to `","`.
 * `minetest.get_player_privs(name) -> {priv1=true,...}`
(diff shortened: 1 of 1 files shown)


doc/lua_api.txt
```diff
@@ -4275,9 +4275,14 @@ Setting-related
...
-* `minetest.privs_to_string(privs)`: returns `"priv1,priv2,..."`
-    * Convert between two privilege representations
+* `minetest.string_to_privs(str, delim)`:
+    * Converts string representation of privs into table form
+    * `delim`: String separating the privs. Defaults to `","`.
+    * Returns `{ priv1 = true, ... }`
+* `minetest.privs_to_string(privs, delim)`:
+    * Convert table form of privs into string representation
+    * `delim`: String to delimit privs. Defaults to `","`.
+    * Returns `"priv1,priv2,..."`
```
ClobberXD: This would be correct only if `delim` was omitted or set to `","`. Should I replace this with a description of the return value instead?

SmallJoker: If it's optional this is enough:
```
* `minetest.privs_to_string(privs[, delim])`
...
    * `delim`: Optional string to delimit privileges. Defaults to `","`.
```

ClobberXD: Do you mean the `Returns ...` line can be omitted?

SmallJoker: Not that, but a generic description would be more correct.
Maybe something like `Returns the privileges concatenated to a string`? The delimiter is already explained a line above, so that's self-explaining.

ClobberXD: Hmm, but that's already mentioned in the `Convert table form into ...` line. I think we can combine the `Returns ...` line with this line like so:

```markdown
Returns the string representation of `privs`
```

Thoughts?

ClobberXD:
@SmallJoker Merge? :)

PR #9234, opened 2019-12-22
[wip] rename `core` to `minetest`

zaoqi:
Because `core` and` minetest` are identical objects, there is no need to have two.

this pr depends on https://github.com/minetest/minetest_game/pull/2552


SmallJoker:
No, this is intended.
Minetest builtin Lua code -> `core`
Mods -> `minetest`

PR #9237, opened 2019-12-23
Tweak default textures

Wuzzy2:
This PR changes some default textures.

First, it fixes a problem with `heart.png` and `bubble.png`. These images had size 16×16, but the statbars expect 24×24. Which meant Minetest would resize the 16×16 images by a non-power of 2, which is not allowed in pixel art. The new images are 12×12.

Second, `player.png`, `player_back.png` are changed to be a bit more recognizable as player images. They are still ugly, however, which is completely intentional. This is to force game authors to replace these images or use a player model.

Third, documentation is updated to reflect these changes.

## To do

This PR is a ready for review.

## How to test

Start an empty game. Or just look at the images directly. :P


Df458:
This PR reminds me of #8886...

Personally I would prefer to see integer UI scaling, but barring that this is a decent alternative. 24x24 is the resolution I see on a standard 1080p display, and I think that's a decent baseline for textures.

As for the player textures, I'm indifferent.

LoneWolfHT:
10/10 for the green wieldhand lol

paramat:
I support the improvement of heart and bubble, and MTG should probably do something similar.
However, please can we keep the classic player textures? The new ones are even more ugly but are also no longer the classic textures that are so much a part of MT history. :-1: for these new player textures.

Wuzzy2:
It is completely intentional that the player textures are very ugly. We generally should want to force game authors to overwrite these (if they don't use 3D players anyway). :-)

My issue with the old textures was not that they are ugly, but more like that it's not really clear what these strange green pixels are supposed to mean. The new textures are at least clear on that.

“Part of MT history” … Umm, I don't really care.

benrob0329:
The 2D player textures have sentimental value to many in the community, I'd prefer them left unchanged as it is inconsequential to anyone but developers _or_ games in the early stages of development.

Wuzzy2:
Player textures restored. :-(

sfan5:
Why is the bubble scaled up to 24px while the heart is scaled down to 12px?
Also, did you run the textures through `optipng`?

paramat:
I should clarify, i am not attached to the green thing, i just think that if we change the default player sprite it should be something completely new instead of an awkward alteration of the 'sacred green thing'.

I do not think that ugly textures are ever acceptable. If the intention is to encourage modders to replace the default player sprite then that can be done with a non-ugly but very simple placeholder texture that is symbolic of a player.

I think the green thing is obviously a humanoid, and the mysterious square of dark pixels suggests a 'front', the front will be clear from movement anyway.

Thanks.

paramat:
For example, if we used a new texture, i would do something like:

new_player_sprite256

As it matches the commonly used player model.
Maybe not green though, this example is just meant to be in the style of the classic green thing.

Wuzzy2:
I get your point about simplicity. But I suggest to *at least* add a simple outline so that it's still visible against green-ish backgrounds. :)

doc/texture_packs.txt
```diff
@@ -102,6 +104,9 @@ by texture packs. All existing fallback textures can be found in the directory
 
 * `wieldhand.png`: texture of the wieldhand
 
+Note: The default textures of `player.png`, `player_back.png` and `wieldhand.png`
```
paramat: Needs removing

doc/texture_packs.txt
```diff
@@ -102,6 +104,9 @@ by texture packs. All existing fallback textures can be found in the directory
...
+are intentionally ugly. You are supposed to replace these in the game.
```
paramat: Please remove the `are intentionally ugly` part, unnecessary and these are classic MT textures.

Wuzzy2:
But all of this is true, right? Those textures are ugly, intentionally look like they look like, and meant to be overwritten by game (unless they go 3D, of course).

SmallJoker:
It's true that they are not aesthetically appealing, but the documentation should not state it like this. More appropriate would be an explanation like this:
`These textures are placeholders to be overwritten by the game.`

paramat:
I very much doubt celeron55 designed them to be intentionally ugly, because that was the MT player texture for a few years.

SmallJoker:
@paramat 3rd person view didn't exist back then, thus nobody cared about the texture except those who played on one of the few multiplayer servers.

paramat:
But other players were visible surely? =)

Wuzzy2, why is the heart 12x12 pixels? It is stated to be 24x24 in the documents.
Was your intention that the textures are 12x12 designs scaled to 24x24 pixels?

Wuzzy2:
Because I was to lazy to draw 24×24 versions of the images and I'm not a really good pixel artist.

And because 12×12 gets upscaled to 24×24 just fine. =)

I did not use optipng, zopflipng or any similar thing. Seems a bit overkill if you ask me … I guess this will save me a few bytes at best. \*shrug\*

paramat:
Ok, in that case, the bubble should be 12x12 pixels too. I agree that 12x12 is plenty enough resolution for such simple shapes.

sfan5:
> Because I was to lazy to draw 24×24 versions of the images

Then why is the bubble texture 24x24? Did you scale it up?
As paramat said, it should be 12x12 like the other texture.

Wuzzy2:
Oops. Bubble was not intentional. Reduced size to 12×12.

I ran both textures thru zopflipng now and it saved me a whopping 223 bytes! Woohoo!

doc/texture_packs.txt
```diff
@@ -74,6 +75,7 @@ by texture packs. All existing fallback textures can be found in the directory
 * `halo.png`: used for the node highlighting mesh
 
 * `heart.png`: used to display the health points of the player
+               (default size: 24×24)
```
paramat: This and the similar line above should be
`(default size: 24×24)`
?

Wuzzy2:
Hmm, actualy, when I think about it … Yeah, if the images are 12×12, kinda makes sense to change it in the docs, too. Minetest supports both 12×12 and 24×24 anyway, so …

paramat:
Does the heart need to be RGBA? It does not seem to have semi-transparency. How about indexed-alpha instead?
+1 for the PR if that is done.

Wuzzy2:
Woah, what's all with that perfectionism with PNGs lately? Is it about to save another 20 bytes? o_O

SmallJoker:
@paramat That would maybe make sense for larger images. But here it's enough to pass them to `optipng`, `pngcrush` (or similar). Optimizing too much with certain tools also arise color profile warnings because libpng does not like them. It's good enough.

paramat:
It is not perfectionism. We always make sure that textures are in the proper format, and we always ask for the proper format, and should do when a texture is first added, because optimising later is a pain. We have even discussed how we need to make an effort to get textures right first time for this reason.
The small number of bytes saved is irrelevant, otherwise we would never bother optimising textures at all.

I do not know why some, including a core dev, are now giving the impression that this is unnecessary perfectionism or that these texture format requests are something new.
The point is to get a texture right first time, in the proper format and with no metadata or comments added, so that we do not have to endlessly bulk re-optimise, as we seem to do.

So, due to the sloppy attitudes i have to be :-1: until the correct format is used.

Wuzzy2:
I just have no idea what you want me to do.

Moth-Tolias:
you can convert the heart to indexed colour using gimp.
there are also a number of very similar colours that could be combined, but i don't think anyone will mind

Wuzzy2:
Done.

But a question: Is it anywhere documented what the “correct” PNG format is?

paramat:
Sorry if i was not clear, 'indexed alpha' is an image format that results in Gimp when you take a RGBA image and 'convert to indexed colour'.
Thanks =)

Moth-Tolias:
ah, you misunderstand - i was suggesting that perhaps the similar colours _should_ be combined, so as to lower the filesize. my bad, i should have used a different wording

paramat:
Ahh sorry, i see. Actually you were quite clear, my misunderstanding.

Issue #9241, opened 2019-12-24
Bumpmapping shader has incorrect behaviour and the lighting needs a new implementation


HybridDog:
##### Minetest version
```
Minetest 5.2.0-dev-37f771a9-dirty (Linux)
Using Irrlicht 1.8.4
BUILD_TYPE=Release
RUN_IN_PLACE=0
USE_GETTEXT=1
USE_SOUND=1
USE_CURL=1
USE_FREETYPE=1
USE_LUAJIT=1
STATIC_SHAREDIR="/usr/local/share/minetest"
```

##### OS / Hardware
Operating system: linux
CPU: intel

GPU model: intel
OpenGL version: 4.5

##### Summary

I cannot find an explanation about what the currently implemented bumpmapping calculates.
If I test it with normal vectors corresponding to a pyramid, it looks like this:
![2](https://user-images.githubusercontent.com/3192173/71368662-ca7acb80-25a8-11ea-9468-a9969a4b43be.png)
I am not sure if it should look like this or if it is a bug.

By the way:
For testing, I have changed code to get correct normal vectors in Tangent Space in the fragment shader.
The default lighting in minetest uses the surface normal vectors and calculates a directional ambient light with them.
I have moved the code to the fragment shader and used the normal vectors from the normalmap instead of only the surface normal: https://github.com/HybridDog/minetest/tree/m_bumpmapping_fixes
The result looks like this:
![3](https://user-images.githubusercontent.com/3192173/71369316-a61fee80-25aa-11ea-800a-92494afe5404.jpg)
There is no specular light because there are no light sources in directional ambient lighting,
so it looks quite unspectacular.

##### Steps to reproduce

Play around with bumpmapping, and the corresponding shader code

paramat:
Good question! =D it has always looked wrong to me.

I think @lhofhansl has discovered a bug in bumpmapping, discussed in another issue somewhere.

hecktest:
If there are no light sources to get a vector from, perhaps the vector can be guessed from the light gradient?

HybridDog:
> perhaps the vector can be guessed from the light gradient?

The light gradient does not have information about the light source distance along the surface's normal vector. If you made an approximation, e.g. assume that this distance is zero, the lighting would probably look wrong.

It is possible to calculate a light vector to the sun, see #9272.
[...]

PR #9242, opened 2019-12-24
Add bumpmapping and parallax occlusion testing nodes to the devtest game

HybridDog:
Goal: Simple nodes for testing bumpmapping and parallax occlusion code changes

How does it work:
The alpha opacity in the normalmap texture is the heightmap for Parallax Occlusion.
The sRGB values in the normalmap store the normal vector, e.g. #8080ff is (2*(128/256-0.5), 2*(128/256-0.5), 2*(255/256-0.5)) = (0,0,1) (the default surface normal).

Does it resolve any reported issue: It could indirectly help to solve related issues.

If not a bug fix, why is this PR needed? What usecases does it solve: The height pyramid normals have 45° rotation, so they can be used to validate if bumpmapping, or a changed bumpmapping implementation, works correctly.

I do not understand what bumpmapping in minetest does:

Relief mapping (one of the Parallax Occlusion methods):

## To do

This PR is Ready for Review.

## How to test

Enable Bumpmapping and/or Parallax Occlusion, place the nodes and see what happens.


SmallJoker:
#9450 looks quite promising, thus I think your changes should to into the new minimal game.

Wuzzy2:
I should probably cannibalize this for DevTest. Or is this not good enough yet? :D

Tbh, I would really appreciate if DevTest (#9450) is merged sooner than later. It's annoying for me if I have to manually copy every change that you make in legacy [minimal].

Wuzzy2:
To clarify: If requested by core devs, I will include this test into my DevTest (#9450) and update my PR. *Only* on explicit request.

I have no idea if the tests added in this PR are actually correct; I mean if the textures are created in a way that they should.

Also, these special `_normal.png` textures are not defined anywhere in `lua_api.txt`. I would like it more if we actually know what we're actually testing here. ;-)

Df458:
> Also, these special _normal.png textures are not defined anywhere in lua_api.txt. I would like it more if we actually know what we're actually testing here. ;-)

Looks like they're not in `texture_packs.txt` either. I was aware of the normal textures from _somewhere_, but it seems like a strange oversight that they're not documented.

Wuzzy2:
Bugs that have been revealed by these nodes (assuming they are correctly defined, which I still do not know):

* Inventory image of "bumpmapping and parallax occlusion" nodes does not match the node. It appears as a white texture with none of the bump-mapping applied
* Wield image of the same node also sometimes does not match the node, depending on your view angle. Sometimes the “pyramid” appears, sometimes it's gone

HybridDog:
> I think your changes should to into the new minimal game.

I can easily copy-paste the changes when DevTest is merged. 

> I have no idea if the tests added in this PR are actually correct; I mean if the textures are created in a way that they should.
> […] I would like it more if we actually know what we're actually testing here. ;-)

The bumpmapping and parallax occlusion code is in the nodes fragment shader:
https://github.com/minetest/minetest/blob/6cf15cf872cc9ce76990d83d380ca3c4b7485eb1/client/shaders/nodes_shader/opengl_fragment.glsl
Explanations about how bumpmapping and parallax occlusion work can be found in the internet, example: https://learnopengl.com/Advanced-Lighting/Normal-Mapping
I didn't know that the alpha channel defines a height map until I've looked at the fragment shader code.

In minetest, the bumpmap and heightmap are both stored in the `_normal.png` image in the RGB (for the normal vector) and alpha (for the height) channels. 
The height pyramid has a simple white (albedo) texture so that it can be used to test how shading with bumpmapping works. The normal map contains four vectors which point diagonally away from the surface (e.g. `(0.7, 0.7, 0)`), and the heights in the height map linearly increase towards the centre, so that the surface corresponds to a simple pyramid. I made this special normal map because I think with this it is easy to see if the cube surface normal vectors are rotated correctly (they are not in minetest as far as I can see). The white albedo texture has small black indicators which can be used to see how it is transformed ingame (and thus see if there's rotation around the normal vector). 
The parallax occlusion tester has no special normal vectors (it's a flat surface) but a complex heightmap and albedo texture. The yellow scrawl shows where the heightmap has high values, i.e. it should be extruded (see the screenshot).
I didn't know where to properly document what those testers mean and whether this information is important enough to be documented (if you look at the textures and know how bumpmapping works, it could be obvious).

Wuzzy2:
Bump. Devtest is merged. Please add labels.

HybridDog:
rebased; the nodes are now in the testnodes mod

Wuzzy2:
Please add code comments to document the intent of these nodes. You don't need to explain what bumpmapping is, just what these nodes are *supposed* to do (and how to use them) and at which details to direct attention to. 1-2 sentences for each node are probably enough.

I think you should use `dig_immediate=3`, like most of the other testnodes.

The `description` of the slabs and stairs need updating (consistent writing style).

HybridDog:
> I suggest to redo this texture completely and replace it with something that pronounces the effect more strongly, so that it's more obvious.

Do you mean a high-resolution texture?

This is how it currently looks like:
fot_2020-05-30-13:08:45_701304666

Wuzzy2:
Oh. OK, forget it, the contrast is actually not bad at all. Strange ...

HybridDog:
I've changed the contrast in the latest commit.

rubenwardy:
Parallax tests don't work: 
image

HybridDog:
It works for me:
fot_2020-08-29-20:22:03_784213031
I have tested it after rebasing, with and without a fresh minetest.conf.

rubenwardy:
Recompiled, still no parallax effect. I have the following error:

```
ERROR[Main]: Irrlicht: Invalid size of image for OpenGL Texture.
```

I've tried clearing my settings, no help. I suspect this may a bug not caused by this PR

```
deprecated_lua_api_handling = log
name = rubenwardy
enable_server = true
devtest_unittests_autostart = true
autosave_screensize = false
enable_waving_plants = true
anisotropic_filter = false
trilinear_filter = false
bilinear_filter = false
leaves_style = fancy
enable_waving_leaves = true
enable_waving_water = true
tone_mapping = true
creative_mode = true
mip_map = false
fsaa = 0
menu_last_game = devtest
enable_damage = true
enable_parallax_occlusion = true
node_highlighting = box
server_announce = false
enable_bumpmapping = true
```

rubenwardy:
Shouldn't parallax mapping use a different map to the normal map? I'm not really sure how it's supposed to work

image

HybridDog:
> Recompiled, still no parallax effect. I have the following error: […]

I don't get this error. You could try the `--verbose` argument to find the texture which causes this message. 
Maybe you have an old or special GPU or driver where the alpha component of the normalmap image is not preserved, or some other problem happens.
I have tested it with an intel GPU and OpenGL version 4.5.

> Shouldn't parallax mapping use a different map to the normal map? I'm not really sure how it's supposed to work

The [rgb components (in glsl they are the same as xyz)](https://github.com/minetest/minetest/blob/c18dbadcb898db6d1f68b72afec56c599af317ec/client/shaders/nodes_shader/opengl_fragment.glsl#L87) define the [normal vector for bumpmapping](https://github.com/minetest/minetest/blob/c18dbadcb898db6d1f68b72afec56c599af317ec/client/shaders/nodes_shader/opengl_fragment.glsl#L205) and the [alpha component defines the height](https://github.com/minetest/minetest/blob/c18dbadcb898db6d1f68b72afec56c599af317ec/client/shaders/nodes_shader/opengl_fragment.glsl#L98) for parallax occlusion. Parallax occlusion and bump mapping use the same image file, so only one instead of two textures need to be loaded from disk, decoded and sent to the GPU.

rubenwardy:
I no longer have the error, randomly

My specs:

```
CPU: Intel® Core™ i7-8550U Processor @ 1.8 - 4.0GHz
GPU: Intel® UHD Graphics 620 
OpenGL: 4.6 (Compatibility Profile) Mesa 20.1.6
```

lhofhansl:
The bumpmapping code is code is wrong. The diffuse light only depends on the direction of the surface normal relative to the lightsource (yet the eye-vector is used there), only specular light depends on the position and if the viewer.

I tried to fix it on and off. And found the following problems:
- The direction of the sunlight is not right
- Since the nodes are already lit/shaded by MT light logic the nodes are too dark when diffuse lighting is applied. It seems we either have to do it all in the shader or nothing.

I'd like to dig deeper but I have very little time ATM.

HybridDog:
> I'd like to dig deeper but I have very little time ATM.

There is an open issue about what bumpmapping does: #9241
I have fixed bumpmapping in Minetest, see the commits at: https://github.com/HybridDog/minetest/tree/m_bumpmapping_fixes
The sun is the only known light source in the shaders, so I changed the code in the last commit to calculate the directional ambience light for the normal vectors instead of doing Phong shading.

paramat:
Tested these nodes, they appear as intended:

screenshot_20201009_184439

screenshot_20201009_184905

:+1: 
See https://github.com/minetest/minetest/pull/9242#issuecomment-683424739 for rubenwardy's approval on testing, this can now be merged.

PR #9247, opened 2019-12-25
Fix MSAA stripes

HybridDog:
This only works when shaders are enabled.
The centroid varying avoids that the textures (which repeat themselves out of bounds) are sampled out of bounds in MSAA.
If MSAA (called FSAA in minetest) is disabled, the centroid keyword does nothing.

Related issue: #6860
An older related PR: #8123

### DIFF
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -517,8 +517,8 @@ texture_min_size (Minimum texture size) int 64
 #    This algorithm smooths out the 3D viewport while keeping the image sharp,
 #    but it doesn't affect the insides of textures
 #    (which is especially noticeable with transparent textures).
-#    This option is experimental and might cause visible spaces between blocks
-#    when set above 0.
+#    Visible spaces appear between nodes when shaders are disabled.
+#    If set to 0, MSAA is disabled.
 #    A restart is required after changing this option.
 fsaa (FSAA) enum 0 0,1,2,4,8,16
 
--- a/client/shaders/nodes_shader/opengl_fragment.glsl
+++ b/client/shaders/nodes_shader/opengl_fragment.glsl
@@ -16,7 +16,7 @@ varying vec3 vPosition;
 // precision must be considered).
 varying vec3 worldPosition;
 varying lowp vec4 varColor;
-varying mediump vec2 varTexCoord;
+centroid varying mediump vec2 varTexCoord;
 varying vec3 eyeVec;
 
 const float fogStart = FOG_START;
@@ -46,7 +46,7 @@ vec4 applyToneMapping(vec4 color)
 	const float gamma = 1.6;
 	const float exposureBias = 5.5;
 	color.rgb = uncharted2Tonemap(exposureBias * color.rgb);
-	// Precalculated white_scale from 
+	// Precalculated white_scale from
 	//vec3 whiteScale = 1.0 / uncharted2Tonemap(vec3(W));
 	vec3 whiteScale = vec3(1.036015346);
 	color.rgb *= whiteScale;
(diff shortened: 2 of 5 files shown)


numberZero:
1. Does this work on older systems? (I mean those not supporting MSAA as well, they should work as before).

2. Is there any noticeable performance penalty? As you wrote it, it applies to everything including even solid nodes; that might fix a more subtle bug I’ve seen some time ago, though.

3. Does it work well together with mipmapping, anisotropic filtering, etc.?
> The shading language specification considers derivatives derived from centroid varings to be so fraught with inaccuracy that it was resolved they are simply undefined.
 — https://www.opengl.org/pipeline/article/vol003_6/

P.S. MSAA (multisample anti-aliasing) is a kind of FSAA (full-scene anti-aliasing) used in Minetest.

HybridDog:
> Does this work on older systems? 

It does nothing when shaders are disabled. varying centroid is explained in the glsl 1.2 (which is related to opengl 2.1) specification. The centroid keyword is simply ignored when MSAA is disabled.

> Is there any noticeable performance penalty?

There should not be any penalty if MSAA is disabled. 
In my opinion, MSAA with small nearest-neighbour sampled textures looks worse than disabled MSAA if centroid is not used.

> As you wrote it, it applies to everything including even solid nodes; […]

I've only changed the nodes shader; maybe the wield shader should be changed, too.

> Does it work well together with mipmapping, anisotropic filtering, etc.?

Something similar to the derivation is only required for parallax occlusion I think. Anyways, the wrong-border-colour artifacts are probably much more disturbing than artifacts caused by the centroid offset.

HybridDog:
> Maybe a comment could explain why it's centroid instead.

Done. I hope the comment is not too long.

sfan5:
This causes pretty severe issues with mipmapping enabled on rails:
grafik

HybridDog:
> This causes pretty severe issues with mipmapping enabled on rails

I don't see the problem. Mipmapping with partially transparent textures works wrong anyway.

#### MSAA disabled: I can't see a difference:

master:
master

this branch:
this_branch

#### MSAA enabled: 

master:
master_msaa

this branch:
this_branch_msaa

sfan5:
> I don't see the problem. Mipmapping with partially transparent textures works wrong anyway.

Sure, but that doesn't mean it should look worse than before.

HybridDog:
> Sure, but that doesn't mean it should look worse than before.

In my opinion it is negligible in comparison to the visible MSAA stripes.
I think that the disappearing railway sleepers are much more noticeably than those "pretty severe" issues.

HybridDog:
rebased.

SmallJoker (approved):
Tested with OpenGL 4.6.0. Looks good, although I can barely find any difference.

sfan5:
This breaks GLES:

<pre>
2020-12-15 22:29:24: ERROR[Main]: Irrlicht: GLSL shader failed to compile
2020-12-15 22:29:24: ERROR[Main]: Irrlicht: 0:56(1): error: syntax error, unexpected NEW_IDENTIFIER, expecting end of file

[...]
2020-12-15 22:29:24: WARNING[Main]: 55: varying lowp vec4 varColor;
2020-12-15 22:29:24: WARNING[Main]: 56: centroid varying mediump vec2 varTexCoord;
2020-12-15 22:29:24: WARNING[Main]: 57: 
2020-12-15 22:29:24: WARNING[Main]: 58: varying vec3 eyeVec;
[...]
</pre>

numberZero:
@sfan5 I see 3 possible fixes:
1. `#define centroid` in the preamble, just like `#define lowp` for non-GLES
2. `#if GL_ES` / `#ifdef GL_ES` in the shader itself
3. bump GLSL ES version (requires other changes)

sfan5:
My preference order is 2 1 3 here.

PR #9254, opened 2019-12-27
Add mapgen settings in world creation dialog

Wuzzy2:

- Goal of the PR: Adding more mapgen settings (mostly mapgen flags) in the world creation menu to change basic mapgen settings. Settings that affect the world the most are used.

- How does the PR work? More widgets (mostly checkboxes) are added in the world creation menu that mostly affect mapgen flags. This includes all `_spflags`, a selection of `mg_flags` and the fractal type (`mgfractal_fractal`). Also, a new setting for `game.conf` is added: `disabled_settings`. This is similar for `disabled_mapgens`, except it hides entire mapgen settings.
- Does it resolve any reported issue? No.
- If not a bug fix, why is this PR needed? What usecases does it solve? Quality-of-life improvement. Mapgen settings are made available where they make most sense, and not only deep in advanced settings. This makes it much faster to play around with mapgen settings. Currently, you are forced to go to the advanced settings menu for every mapgen settings change you make.

## To do

Needs to be tested by someone else than me.
Code might be a mess, has rough edges. But it works fine, to my knowledge.
UI design could be improved, but judge yourself.

## How to test

1. Start Minetest
2. Open world creation dialog
3. Choose mapgen type

### DIFF
--- a/builtin/mainmenu/dlg_create_world.lua
+++ b/builtin/mainmenu/dlg_create_world.lua
@@ -17,12 +17,108 @@
 
 local worldname = ""
 
+local function table_to_flags(ftable)
+	-- Convert e.g. { jungles = true, caves = false } to "jungles,nocaves"
+	local str = {}
+	for flag, is_set in pairs(ftable) do
+		str[#str + 1] = is_set and flag or ("no" .. flag)
+	end
+	return table.concat(str, ",")
+end
+
+-- Same as check_flag but returns a string
+local function strflag(flags, flag)
+	return (flags[flag] == true) and "true" or "false"
+end
+
+local cb_caverns = { "caverns", fgettext("Caverns"), "caverns",
+	fgettext("Very large caverns deep in the underground") }
+local tt_sea_rivers = fgettext("Sea level rivers")
+
+local flag_checkboxes = {
+	v5 = {
+		cb_caverns,
+	},
+	v7 = {
+		cb_caverns,
+		{ "ridges", fgettext("Rivers"), "ridges", tt_sea_rivers },
+		{ "mountains", fgettext("Mountains"), "mountains" },
+		{ "floatlands", fgettext("Floatlands (experimental)"), "floatlands",
+		fgettext("Floating landmasses in the sky") },
+	},
+	carpathian = {
+		cb_caverns,
+		{ "rivers", fgettext("Rivers"), "rivers", tt_sea_rivers },
+	},
+	valleys = {
+		{ "altitude-chill", fgettext("Altitude chill"), "altitude_chill",
+		fgettext("Reduces heat with altitude") },
+		{ "altitude-dry", fgettext("Altitude dry"), "altitude_dry",
+		fgettext("Reduces humidity with altitude") },
+		{ "humid-rivers", fgettext("Humid rivers"), "humid_rivers",
+		fgettext("Increases humidity around rivers") },
+		{ "vary-river-depth", fgettext("Vary river depth"), "vary_river_depth",
+		fgettext("Low humidity and high heat causes shallow or dry rivers") },
+	},
+	flat = {
+		{ "hills", fgettext("Hills"), "hills" },
+		{ "lakes", fgettext("Lakes"), "lakes" },
+	},
+	fractal = {
+		{ "terrain", fgettext("Terrain"), "terrain",
+		fgettext("Generate non-fractal terrain: Oceans and underground") },
+	},
+	v6 = {
+		{ "trees", fgettext("Trees and jungle grass"), "trees" },
+		{ "flat", fgettext("Flat terrain"), "flat" },
+		{ "mudflow", fgettext("Mud flow"), "mudflow",
+		fgettext("Terrain surface erosion") },
+		-- Biome settings are in mgv6_biomes below
+	},
+}
+
+local mgv6_biomes = {
+	{
+		fgettext("Temperate, Desert, Jungle, Tundra, Taiga"),
+		{jungles = true, snowbiomes = true}
+	},
+	{
+		fgettext("Temperate, Desert, Jungle"),
+		{jungles = true, snowbiomes = false}
+	},
+	{
+		fgettext("Temperate, Desert"),
+		{jungles = false, snowbiomes = false}
+	},
+}
+
 local function create_world_formspec(dialogdata)
+
+	-- Error out when no games found
+	if #pkgmgr.games == 0 then
+		return "size[12.25,3,true]" ..
+			"box[0,0;12,2;#ff8800]" ..
+			"textarea[0.3,0;11.7,2;;;"..
+			fgettext("You have no games installed.") .. "\n" ..
+			fgettext("Download one from minetest.net") .. "]" ..
+			"button[4.75,2.5;3,0.5;world_create_cancel;" .. fgettext("Cancel") .. "]"
+	end
+
 	local mapgens = core.get_mapgen_names()
 
 	local current_seed = core.settings:get("fixed_map_seed") or ""
 	local current_mg   = core.settings:get("mg_name")
 	local gameid = core.settings:get("menu_last_game")
+
+	local flags = {}
+	flags.main = core.settings:get_flags("mg_flags")
+	flags.v5 = core.settings:get_flags("mgv5_spflags")
+	flags.v6 = core.settings:get_flags("mgv6_spflags")
+	flags.v7 = core.settings:get_flags("mgv7_spflags")
+	flags.fractal = core.settings:get_flags("mgfractal_spflags")
+	flags.carpathian = core.settings:get_flags("mgcarpathian_spflags")
+	flags.valleys = core.settings:get_flags("mgvalleys_spflags")
+	flags.flat = core.settings:get_flags("mgflat_spflags")
 
 	local gameidx = 0
 	if gameid ~= nil then
@@ -35,6 +131,7 @@ local function create_world_formspec(dialogdata)
 	end
 
 	local game_by_gameidx = core.get_game(gameidx)
+	local disallowed_mapgen_flags = {}
 	if game_by_gameidx ~= nil then
 		local gamepath = game_by_gameidx.path
 		local gameconfig = Settings(gamepath.."/game.conf")
@@ -64,48 +161,193 @@ local function create_world_formspec(dialogdata)
 				end
 			end
 		end
+
+		local ds = (gameconfig:get("disallowed_mapgen_flags") or ""):split()
+		for _, value in pairs(ds) do
+			disallowed_mapgen_flags[value:trim()] = true
+		end
 	end
 
 	local mglist = ""
-	local selindex = 1
+	local selindex
 	local i = 1
+	local first_mg
 	for k,v in pairs(mapgens) do
+		if not first_mg then
+			first_mg = v
+		end
 		if current_mg == v then
 			selindex = i
 		end
 		i = i + 1
 		mglist = mglist .. v .. ","
 	end
+	if not selindex then
+		selindex = 1
+		current_mg = first_mg
+	end
 	mglist = mglist:sub(1, -2)
 
-	current_seed = core.formspec_escape(current_seed)
-	local retval =
-		"size[11.5,6.5,true]" ..
-		"label[2,0;" .. fgettext("World name") .. "]"..
-		"field[4.5,0.4;6,0.5;te_world_name;;" .. minetest.formspec_escape(worldname) .. "]" ..
+	local mg_main_flags = function(mapgen, y)
+		if mapgen == "singlenode" then
+			return "", y
+		end
+		if disallowed_mapgen_flags["mg_flags"] then
+			return "", y
+		end
 
-		"label[2,1;" .. fgettext("Seed") .. "]"..
-		"field[4.5,1.4;6,0.5;te_seed;;".. current_seed .. "]" ..
+		local form = "checkbox[0," .. y .. ";flag_mg_caves;" ..
+			fgettext("Caves") .. ";"..strflag(flags.main, "caves").."]"
+		y = y + 0.5
 
-		"label[2,2;" .. fgettext("Mapgen") .. "]"..
-		"dropdown[4.2,2;6.3;dd_mapgen;" .. mglist .. ";" .. selindex .. "]" ..
+		form = form .. "checkbox[0,"..y..";flag_mg_dungeons;" ..
+			fgettext("Dungeons") .. ";"..strflag(flags.main, "dungeons").."]"
+		y = y + 0.5
 
-		"label[2,3;" .. fgettext("Game") .. "]"..
-		"textlist[4.2,3;5.8,2.3;games;" .. pkgmgr.gamelist() ..
-		";" .. gameidx .. ";true]" ..
+		local d_name, d_tt
+		if mapgen == "v6" then
+			d_name = fgettext("Decorations (no effect on trees/jungle grass)")
+			d_tt = fgettext("Structures appearing on the terrain")
+		else
+			d_name = fgettext("Decorations")
+			d_tt = fgettext("Structures appearing on the terrain, typically trees and plants")
+		end
+		form = form .. "checkbox[0,"..y..";flag_mg_decorations;" ..
+			d_name .. ";" ..
+			strflag(flags.main, "decorations").."]" ..
+			"tooltip[flag_mg_decorations;" ..
+			d_tt ..
+			"]"
+		y = y + 0.5
 
-		"button[3.25,6;2.5,0.5;world_create_confirm;" .. fgettext("Create") .. "]" ..
-		"button[5.75,6;2.5,0.5;world_create_cancel;" .. fgettext("Cancel") .. "]"
-
-	if #pkgmgr.games == 0 then
-		retval = retval .. "box[2,4;8,1;#ff8800]label[2.25,4;" ..
-				fgettext("You have no games installed.") .. "]label[2.25,4.4;" ..
-				fgettext("Download one from minetest.net") .. "]"
-	elseif #pkgmgr.games == 1 and pkgmgr.games[1].id == "minimal" then
-		retval = retval .. "box[1.75,4;8.7,1;#ff8800]label[2,4;" ..
-				fgettext("Warning: The minimal development test is meant for developers.") .. "]label[2,4.4;" ..
-				fgettext("Download a game, such as Minetest Game, from minetest.net") .. "]"
+		form = form .. "tooltip[flag_mg_caves;" ..
+		fgettext("Network of tunnels and caves")
+		.. "]"
+		return form, y
 	end
+
+	local mg_specific_flags = function(mapgen, y)
+		if not flag_checkboxes[mapgen] then
+			return "", y
+		end
+		if disallowed_mapgen_flags["mg"..mapgen.."_spflags"] then
+			return "", y
+		end
+		local form = ""
+		for _,tab in pairs(flag_checkboxes[mapgen]) do
+			local id = "flag_mg"..mapgen.."_"..tab[1]
+			form = form .. ("checkbox[0,%f;%s;%s;%s]"):
+				format(y, id, tab[2], strflag(flags[mapgen], tab[3]))
+
+			if tab[4] then
+				form = form .. "tooltip["..id..";"..tab[4].."]"
+			end
+			y = y + 0.5
+		end
+
+		if mapgen ~= "v6" then
+			-- No special treatment
+			return form, y
+		end
+		-- Special treatment for v6 (add biome widgets)
+
+		-- Biome type (jungles, snowbiomes)
+		local biometype
+		if flags.v6.snowbiomes == true then
+			biometype = 1
+		elseif flags.v6.jungles == true  then
+			biometype = 2
+		else
+			biometype = 3
+		end
+		y = y + 0.3
+
+		form = form .. "label[0,"..(y+0.1)..";" .. fgettext("Biomes") .. "]"
+		y = y + 0.6
+
+		form = form .. "dropdown[0,"..y..";6.3;mgv6_biomes;"
+		for b=1, #mgv6_biomes do
+			form = form .. mgv6_biomes[b][1]
+			if b < #mgv6_biomes then
+				form = form .. ","
+			end
+		end
+		form = form .. ";" .. biometype.. "]"
+
+		-- biomeblend
+		y = y + 0.55
+		form = form .. "checkbox[0,"..y..";flag_mgv6_biomeblend;" ..
+			fgettext("Biome blend") .. ";"..strflag(flags.v6, "biomeblend").."]" ..
+			"tooltip[flag_mgv6_biomeblend;" ..
+			fgettext("Smooth transition between biomes") .. "]"
+
+		return form, y
+	end
+
+	current_seed = core.formspec_escape(current_seed)
+
+	local y_start = 0.0
+	local y = y_start
+	local str_flags, str_spflags
+	local label_flags, label_spflags = "", ""
+	y = y + 0.3
+	str_flags, y = mg_main_flags(current_mg, y)
+	if str_flags ~= "" then
+		label_flags = "label[0,"..y_start..";" .. fgettext("Mapgen flags") .. "]"
+		y_start = y + 0.4
+	else
+		y_start = 0.0
+	end
+	y = y_start + 0.3
+	str_spflags = mg_specific_flags(current_mg, y)
+	if str_spflags ~= "" then
+		label_spflags = "label[0,"..y_start..";" .. fgettext("Mapgen-specific flags") .. "]"
+	end
+
+	-- Warning if only minimal is installed
+	local minimal_only = ""
+	local gamelist_height = 2.3
+	if #pkgmgr.games == 1 and pkgmgr.games[1].id == "minimal" then
+		minimal_only = "box[0,0;5.8,1.7;#ff8800]" ..
+				"textarea[0.3,0;6,1.8;;;"..
+				fgettext("Warning: The minimal development test is meant for developers.") .. "\n" ..
+				fgettext("Download a game, such as Minetest Game, from minetest.net") .. "]"
+		gamelist_height = 0.5
+	end
+
+	local retval =
+		"size[12.25,7,true]" ..
+
+		-- Left side
+		"container[0,0]"..
+		"field[0.3,0.6;6,0.5;te_world_name;" ..
+		fgettext("World name") ..
+		";" .. core.formspec_escape(worldname) .. "]" ..
+
+		"field[0.3,1.7;6,0.5;te_seed;" ..
+		fgettext("Seed") ..
+		";".. current_seed .. "]" ..
+
+		"label[0,2;" .. fgettext("Mapgen") .. "]"..
+		"dropdown[0,2.5;6.3;dd_mapgen;" .. mglist .. ";" .. selindex .. "]" ..
+
+		"label[0,3.35;" .. fgettext("Game") .. "]"..
+		"textlist[0,3.85;5.8,"..gamelist_height..";games;" ..
+		pkgmgr.gamelist() .. ";" .. gameidx .. ";false]" ..
+		"container[0,4.5]" ..
+		minimal_only ..
+		"container_end[]" ..
+		"container_end[]" ..
+
+		-- Right side
+		"container[6.2,0]"..
+		label_flags .. str_flags ..
+		label_spflags .. str_spflags ..
+		"container_end[]"..
+
+		-- Menu buttons
+		"button[3.25,6.5;3,0.5;world_create_confirm;" .. fgettext("Create") .. "]" ..
+		"button[6.25,6.5;3,0.5;world_create_cancel;" .. fgettext("Cancel") .. "]"
 
 	return retval
 
@@ -163,8 +405,50 @@ local function create_world_buttonhandler(this, fields)
 		return true
 	end
 
+	for k,v in pairs(fields) do
+		local split = string.split(k, "_", nil, 3)
+		if split and split[1] == "flag" then
+			local setting
+			if split[2] == "mg" then
+				setting = "mg_flags"
+			else
+				setting = split[2].."_spflags"
+			end
+			-- We replaced the underscore of flag names with a dash.
+			local flag = string.gsub(split[3], "-", "_")
+			local ftable = core.settings:get_flags(setting)
+			if v == "true" then
+				ftable[flag] = true
+			else
+				ftable[flag] = false
+			end
+			local flags = table_to_flags(ftable)
+			core.settings:set(setting, flags)
+			return true
+		end
+	end
+
 	if fields["world_create_cancel"] then
 		this:delete()
+		return true
+	end
+
+	if fields["mgv6_biomes"] then
+		local entry = minetest.formspec_escape(fields["mgv6_biomes"])
+		for b=1, #mgv6_biomes do
+			if entry == mgv6_biomes[b][1] then
+				local ftable = core.settings:get_flags("mgv6_spflags")
+				ftable.jungles = mgv6_biomes[b][2].jungles
+				ftable.snowbiomes = mgv6_biomes[b][2].snowbiomes
+				local flags = table_to_flags(ftable)
+				core.settings:set("mgv6_spflags", flags)
+				return true
+			end
+		end
+	end
+
+	if fields["dd_mapgen"] then
+		core.settings:set("mg_name", fields["dd_mapgen"])
 		return true
 	end
 
--- a/builtin/mainmenu/dlg_create_world.lua
+++ b/builtin/mainmenu/dlg_create_world.lua
@@ -110,15 +110,16 @@ local function create_world_formspec(dialogdata)
 	local current_mg   = core.settings:get("mg_name")
 	local gameid = core.settings:get("menu_last_game")
 
-	local flags = {}
-	flags.main = core.settings:get_flags("mg_flags")
-	flags.v5 = core.settings:get_flags("mgv5_spflags")
-	flags.v6 = core.settings:get_flags("mgv6_spflags")
-	flags.v7 = core.settings:get_flags("mgv7_spflags")
-	flags.fractal = core.settings:get_flags("mgfractal_spflags")
-	flags.carpathian = core.settings:get_flags("mgcarpathian_spflags")
-	flags.valleys = core.settings:get_flags("mgvalleys_spflags")
-	flags.flat = core.settings:get_flags("mgflat_spflags")
+	local flags = {
+		main = core.settings:get_flags("mg_flags"),
+		v5 = core.settings:get_flags("mgv5_spflags"),
+		v6 = core.settings:get_flags("mgv6_spflags"),
+		v7 = core.settings:get_flags("mgv7_spflags"),
+		fractal = core.settings:get_flags("mgfractal_spflags"),
+		carpathian = core.settings:get_flags("mgcarpathian_spflags"),
+		valleys = core.settings:get_flags("mgvalleys_spflags"),
+		flat = core.settings:get_flags("mgflat_spflags"),
+	}
 
 	local gameidx = 0
 	if gameid ~= nil then
--- a/builtin/mainmenu/dlg_create_world.lua
+++ b/builtin/mainmenu/dlg_create_world.lua
@@ -65,7 +65,7 @@ local flag_checkboxes = {
 		{ "lakes", fgettext("Lakes"), "lakes" },
 	},
 	fractal = {
-		{ "terrain", fgettext("Terrain"), "terrain",
+		{ "terrain", fgettext("Additional terrain"), "terrain",
 		fgettext("Generate non-fractal terrain: Oceans and underground") },
 	},
 	v6 = {
@@ -205,12 +205,11 @@ local function create_world_formspec(dialogdata)
 			fgettext("Dungeons") .. ";"..strflag(flags.main, "dungeons").."]"
 		y = y + 0.5
 
-		local d_name, d_tt
+		local d_name = fgettext("Decorations")
+		local d_tt
 		if mapgen == "v6" then
-			d_name = fgettext("Decorations (no effect on trees/jungle grass)")
-			d_tt = fgettext("Structures appearing on the terrain")
+			d_tt = fgettext("Structures appearing on the terrain (no effect on trees and jungle grass created by v6)")
 		else
-			d_name = fgettext("Decorations")
 			d_tt = fgettext("Structures appearing on the terrain, typically trees and plants")
 		end
 		form = form .. "checkbox[0,"..y..";flag_mg_decorations;" ..
@@ -278,7 +277,7 @@ local function create_world_formspec(dialogdata)
 		-- biomeblend
 		y = y + 0.55
 		form = form .. "checkbox[0,"..y..";flag_mgv6_biomeblend;" ..
-			fgettext("Biome blend") .. ";"..strflag(flags.v6, "biomeblend").."]" ..
+			fgettext("Biome blending") .. ";"..strflag(flags.v6, "biomeblend").."]" ..
 			"tooltip[flag_mgv6_biomeblend;" ..
 			fgettext("Smooth transition between biomes") .. "]"
 
--- a/builtin/mainmenu/dlg_create_world.lua
+++ b/builtin/mainmenu/dlg_create_world.lua
@@ -132,7 +132,7 @@ local function create_world_formspec(dialogdata)
 	end
 
 	local game_by_gameidx = core.get_game(gameidx)
-	local disallowed_mapgen_flags = {}
+	local disallowed_mapgen_settings = {}
 	if game_by_gameidx ~= nil then
 		local gamepath = game_by_gameidx.path
 		local gameconfig = Settings(gamepath.."/game.conf")
@@ -163,9 +163,9 @@ local function create_world_formspec(dialogdata)
 			end
 		end
 
-		local ds = (gameconfig:get("disallowed_mapgen_flags") or ""):split()
+		local ds = (gameconfig:get("disallowed_mapgen_settings") or ""):split()
 		for _, value in pairs(ds) do
-			disallowed_mapgen_flags[value:trim()] = true
+			disallowed_mapgen_settings[value:trim()] = true
 		end
 	end
 
@@ -193,7 +193,7 @@ local function create_world_formspec(dialogdata)
 		if mapgen == "singlenode" then
 			return "", y
 		end
-		if disallowed_mapgen_flags["mg_flags"] then
+		if disallowed_mapgen_settings["mg_flags"] then
 			return "", y
 		end
 
@@ -231,7 +231,7 @@ local function create_world_formspec(dialogdata)
 		if not flag_checkboxes[mapgen] then
 			return "", y
 		end
-		if disallowed_mapgen_flags["mg"..mapgen.."_spflags"] then
+		if disallowed_mapgen_settings["mg"..mapgen.."_spflags"] then
 			return "", y
 		end
 		local form = ""
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -75,8 +75,8 @@ The game directory can contain the following files:
       When both `allowed_mapgens` and `disallowed_mapgens` are
       specified, `allowed_mapgens` is applied before
       `disallowed_mapgens`.
-    * `disallowed_mapgen_flags = <comma-separated mapgen flags>`
-      e.g. `disallowed_mapgen_flags = mgv5_spflags`
+    * `disallowed_mapgen_settings= <comma-separated mapgen settings>`
+      e.g. `disallowed_mapgen_settings = mgv5_spflags`
       These settings are hidden for this game in the world creation
       dialog and game start menu.
 * `minetest.conf`:
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -75,6 +75,10 @@ The game directory can contain the following files:
       When both `allowed_mapgens` and `disallowed_mapgens` are
       specified, `allowed_mapgens` is applied before
       `disallowed_mapgens`.
+    * `disallowed_mapgen_flags = <comma-separated mapgen flags>`
+      e.g. `disallowed_mapgen_flags = mgv5_spflags`
+      These settings are hidden for this game in the world creation
+      dialog and game start menu.
 * `minetest.conf`:
   Used to set default settings when running this game.
 * `settingtypes.txt`:


paramat:
I would have preferred this was discussed first. However, maybe this was a valuable coding experience and you are ok with some of this being disapproved, in which case, no problem.

There have been suggestions to do this before, but there were concerns because this makes any changes or additions to mapgen-specific flags much more complex and require irritating formspec redesigns each time.
The advantage of leaving these to 'All Settings' is how that menu automatically formats any new or changed flags. The mapgen flags are easy to find and adjust in 'All Settings'.
However, i can see the advantage of setting mapgen flags in world creation, as long as these only apply to the world then created and are not written to .conf (is this the case?).

I am fine with the global mapgen flags being set in world creation, as they are few in number, rarely change and apply to all mapgens. I wil support that.

I am concerned about mapgen-specific flags for the reasons above, but do not disapprove currently.

I disapprove of fractal type being set in world creation, as that is neither a global mapgen flag or and mapgen-specific flag, it is a mapgen-specific parameter so should not be settable in world creation. It is also fairly advanced and best left to the 'All Settings' menu alongside all the other advanced fractal parameters.
Altering fractal type does not give good results unless you also understand and alter the other fractal parameters, so it is actually bad to have fractal type in world creation as it will result in disappointment and a bad impression.
:-1: for fractal type being included in this.

> Also, a new setting for game.conf is added: disabled_settings. This is similar for disabled_mapgens, except it hides entire mapgen settings.

Please could you explain why this is useful or needed? To be clear, i do not object,

> deep in advanced settings

The mapgen flags are not 'deep' in 'All Settings' at all, they are easy to find and alter:
Settings tab -> All Settings -> Mapgen -> mg_flags
Settings tab -> All Settings -> Mapgen -> Mapgen Name -> mgname_spflags

> you are forced to go to the advanced settings menu

A quick note: It is called 'All Settings' now, to try to get away from the idea these are 'advanced' settings, most are not advanced at all. Also to get away from the idea (often stated to try to justify certain arguments) that this is a very difficult to use advanced menu. The only problem is finding certain settings, this would still be a problem if all settings were present in a designed formspec, because there would still need to be navigation through submenus.

Wuzzy2:
> Altering fractal type does not give good results unless you also understand and alter the other fractal parameters, so it is actually bad to have fractal type in world creation as it will result in disappointment and a bad impression.

That's a good argument. I didn't really realize that. But that's also a huge bummer, as this means getting good mapgen results for fractals is harder than I thought. :(

> There have been suggestions to do this before, but there were concerns because this makes any changes or additions to mapgen-specific flags much more complex and require irritating formspec redesigns each time.

The addition of mapgen flags (including specific ones) is rare enough that I don't think this will be a huge problem. And I don't see formspec changes to be complex.

But, oh well, I could at least simplify the code to make additions of new checkboxes simpler to minimize the amount of code you need to touch.

Also note the intention is not to expose ALL mapgen flags, only the relevant ones. Obscure flags such as “light” are intentionally left out. Also note this PR is NOT meant to replace All Settings.

Also note that nothing will break if you add a new mapgen flag and don't immediately add it to world creation. It will always be changable in All Settings. Whether a setting is “promoted” to world creation needs to be decided on a case-to-case basis, based on the setting's importance/usefulness.

> However, i can see the advantage of setting mapgen flags in world creation, as long as these only apply to the world then created and are not written to .conf (is this the case?).

Well, they are directly written to `minetest.conf`. However, as far I understand, all mapgen settings are ONLY relevant when you create a new world. Once the world is created, they are never read again. This is why I think it's so awkward you have to cancel world creation to change a single flag.

Note this is also how the dialog currently works with regards to `mg_name` and seed.

> The mapgen flags are not 'deep' in 'All Settings' at all, they are easy to find and alter:
> Settings tab -> All Settings -> Mapgen -> mg_flags
> Settings tab -> All Settings -> Mapgen -> Mapgen Name -> mgname_spflags

4 layers IS deep, and there's a lot of steps required to change a single flag for what should only require a single click. This is what this PR does.

> Please could you explain why this is useful or needed [`disabled_settings`]?

Well, same reason as `disabled_mapgens`, basically. Except that it's not for the mapgen itself, but for mapgen settings. If a game wants to force its mapgen settings, those should be hidden from the UI as irrelevant. Note this still does not affect All Settings.

Wuzzy2:
* Fractal type removed
* Refactored code to make adding new flags much easier. Flags are stored in new table `flag_checkboxes`, offsets are calculated on the fly

paramat:
Nice update, thanks. No objections to any of this currently, i will review soonish.
> Refactored code to make adding new flags much easier. Flags are stored in new table flag_checkboxes, offsets are calculated on the fly

I think the relevant page in All Settings does this, so it is good that this happens in world creation too.

Mapgen Fractal is quite unstable, i am thinking of ways to improve it, so yes best to leave fractal type out of this PR.
> Well, they are directly written to minetest.conf

Ok, this may be unavoidable, so is probably fine.
Actually, i think i am wrong that choices made in world creation are not written to .conf, so, ignore my original comment =)
> as far I understand, all mapgen settings are ONLY relevant when you create a new world. Once the world is created, they are never read again

Anything mapgen setting added to .conf (either manually or automatically) persists and will be applied to any new world created.

Not 4 layers, starting from the All Settings menu, flags are at the 2nd or 3rd layer. However i agree it is good that flags are settable at world creation.

paramat:
See the failing Luacheck build check for some errors.
Also, many lines are too long. Luacheck uses 120 columns but the MT limit is 90 (with tab size 4).

paramat:
Have you forgotten the 'biomes' global mapgen flag? Cannot see it in the code.

Wuzzy2:
The goal is not to include all flags, only the useful ones. “light” flag intentionally excluded for being hardly useful to most. I also intentionally excluded “biomes” since you rarely want stone-only worlds. But it'd be fairly trivial to add it back, if you insist.
All other flags from all mapgens are included.

So, the idea is that when someone adds a new mapgen flag, it might or might not be added to world creation. Needs a manual decision, based on the importance/usefulness of the flag. But with this PR code, this can be now done with a single line of code. Future coders just need to be aware of this whenever they change mapgen flags. I hope it's not too much to ask to change 1 line of code. :-)

> Mapgen Fractal is quite unstable, i am thinking of ways to improve it, so yes best to leave fractal type out of this PR.

`terrain` flag is still included, which doesn't seem to hurt.

Wuzzy2:
I find it very hard to shorten the lines below 90 characters, especially without violating the other rules. I tried my best, and it made the code actually more ugly (imho).

For the text-heavy lines it seems impossible. Am I allowed to line-break strings fed into `fgettext`? I'm afraid the string collection script might not be smart enough for this, that it only supports when the string is in the full line. :-( Correct me if I'm wrong.

paramat:
>  I also intentionally excluded “biomes” since you rarely want stone-only worlds

Good point, it is rarely wanted, so it is ok to leave it out. I agree with leaving 'light' out.
> terrain flag is still included, which doesn't seem to hurt.

That is fine and desirable.
> I find it very hard to shorten the lines below 90 characters, especially without violating the other rules

The only issue is breaking fgettext() lines, for now i think we can make that an exception to the line length rule.
It looks much better now. Long lines run off the edge of Github pages meaning scrolling back and forth must be used to read the lines, which makes review very difficult.

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -17,13 +17,115 @@
...
+local tt_caverns = fgettext("Large cavities will be carved in the ground")
+
+local flag_checkboxes = {
+	v5 = {
+		{ "caverns", fgettext("Caverns"), "caverns", tt_caverns },
+	},
+	v7 = {
+		{ "caverns", fgettext("Caverns"), "caverns", tt_caverns },
+		{ "ridges", fgettext("Rivers"), "ridges",
+		fgettext("Rivers will be aggressively carved through terrain at sea level") },
```
paramat: No need for 'aggressively', it conveys no essential additional information.

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -17,13 +17,115 @@
...
+		{ "hills", fgettext("Hills"), "hills" },
+		{ "lakes", fgettext("Lakes"), "lakes" },
+	},
+	fractal = {
+		{ "terrain", fgettext("Terrain"), "terrain",
+		fgettext("Enables generation of non-fractal terrain: Ocean, islands and underground") },
+	},
+	v6 = {
+		{ "trees", fgettext("Trees and jungle grass"), "trees",
+		fgettext("Generate trees and jungle grass, v6 style") },
```
paramat: No need for 'v6 style'.

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -51,46 +154,172 @@ local function create_world_formspec(dialogdata)
...
+		end
+
+		local form = "checkbox[4.2," .. y .. ";flag_mg_caves;" ..
+			fgettext("Caves") .. ";"..chflag(flags.main, "caves").."]"
+		y = y + 0.5
+		form = form .. "checkbox[4.2,"..y..";flag_mg_dungeons;" ..
+			fgettext("Dungeons") .. ";"..chflag(flags.main, "dungeons").."]"
+		y = y + 0.5
+
+		if mapgen == "v6" then
```
paramat: No need to split this into 2 and treat v6 specially due to trees and junglegrass being controlled by flags. The flags already show that trees and junglegrass are controlled by flags.
V6 is only a legacy and old world supporting mapgen, so it is better to keep this code simple and focussed on the non-v6 mapgens.

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -51,46 +154,172 @@ local function create_world_formspec(dialogdata)
...
+			fgettext("Decorations are special structures provided by the game that appear on the surface. Note: This does not affect trees and jungle grass generated by the v6 mapgen")
+			.. "]"
+
+			y = y + 0.5
+		else
+			form = form .. "checkbox[4.2,"..y..";flag_mg_decorations;" ..
+			fgettext("Decorations") .. ";" ..
+			chflag(flags.main, "decorations").."]" ..
+			"tooltip[flag_mg_decorations;" ..
+			fgettext("Decorations are special structures that generate on the surface includes things such as plants, trees and other small extras")
```
paramat: This tooltip is unnecessarily long and over-explaining.

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -51,46 +154,172 @@ local function create_world_formspec(dialogdata)
...
+			.. "]"
+
+			y = y + 0.5
+		end
+		form = form .. "tooltip[flag_mg_caves;" ..
+		fgettext("Carve a large, interconnected network of tunnels and caves in the terrain")
```
paramat: Unnecessary tooltip, people know what caves are.

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -51,46 +154,172 @@ local function create_world_formspec(dialogdata)
...
+				local id = "flag_mg"..mapgen.."_"..tab[1]
+				form = form .. "checkbox[4.2," .. y .. ";" .. id .. ";" .. tab[2] .. ";" ..
+					chflag(flags[mapgen], tab[3]).."]"
+				if tab[4] then
+					form = form .. "tooltip["..tab[1]..";"..tab[4].."]"
+				end
+				y = y + 0.5
+			end
+
+			if mapgen == "v6" then
```
paramat: For reasons explained in a line comment above, i disapprove of adding so much perfectionist complexity just for v6. Users can read documentation to see how the biome flags work if they need to.
We only keep V6 around to support old worlds. V6 is at the end of the mapgen list to discourage use, new worlds are rarely started using v6. So this amount of complex specialisation for it is not justified.

Which biomes result from the flags is somewhat obvious already, so there is little justification for this special treatment. It seems to be for tooltips to show an exact list of biomes, which is unnecessary.

So please have v6 taken care of by the simple non-mgv6 code.

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -150,11 +379,60 @@ local function create_world_buttonhandler(this, fields)
...
+			return true
+		end
+	end
+
 	if fields["world_create_cancel"] then
 		this:delete()
 		return true
 	end
 
+	if fields["mgv6_biomes"] then
```
paramat: No separate treatment for v6 please.

paramat: Docs already explain how v6 flags are interpreted:

#    The 'snowbiomes' flag enables the new 5 biome system.
#    When the 'snowbiomes' flag is enabled jungles are automatically enabled and
#    the 'jungles' flag is ignored.

Wuzzy2:
Paramat, I'm quite unhappy how you're often so quick to shut down many usability-improving concepts because you believe they are “unneccessary”, there's “no need”, that they are “useless”, etc. It's a general theme I have seen in so many of your remark throughout the years. And I disagree a lot.
I have come to a point that I feel, to be frank, quite insulted by such remarks; you are making it way too easy for yourself. The main fallacy you're suffering most of the time (and in this issue) is that you're viewing the UI only through your own eyes and only think about *your own* use case and assume that every other user thinks like you. I want you to understand that **you are not the user** (well, not the only one, to be precise). I hereby assign you a reading list:

https://uxmyths.com/post/715988395/myth-you-are-like-your-users
https://www.nngroup.com/articles/computer-skill-levels/

---

> No need for 'aggressively', it conveys no essential additional information.

Well, I think it is quite accurate tho. If the terrain grows to extreme heights, the rivers still cut through them all the way down to sea level, which frequently leads to giant cliffs. I am not sure if I really should remove this word (but I admit it sounds a bit dramatic. But hey, it's a game after all (OK, a game *engine*, but still! :D )). I wait for a second opinion.

> The flags already show that trees and junglegrass are controlled by flags.

Do they? It doesn't seem obvious tbh. `decorations` has that weird exception that it doesn't affect v6 decorations. If I would just write “Decorations” for v6, then it's misleading as it doesn't affect v6 stuff. I'm waiting for a second opinion.

> > (tooltip for `trees`)
> No need for 'v6 style'.

Then the tooltip would be basically empty. ;-) Maybe I can think of something else or I just remove the tooltip here.

> > (tooltip for `caves`)
> Unnecessary tooltip, people know what caves are.

I know what you think. It sounds dumb and redundant at first glance but it's actually not. It's basically there to distinguish `caves` from `caverns`, two flags which are subtly different.

> > fgettext("Decorations are special structures that generate on the surface includes things such as plants, trees and other small extras") 
> This tooltip is unnecessarily long and over-explaining.

Well, the thing is, the flag is called “decorations”, which is a fairly technical term, so definitely needs explaining. Length is an argument however. I'm not sure how to rewrite it yet.

> We only keep V6 around to support old worlds. V6 is at the end of the mapgen list to discourage use, new worlds are rarely started using v6. So this amount of complex specialisation for it is not justified.

I think it's a mistake to view v6 as a “legacy mapgen” at this point. It's actively and **willingly** supported by many games, and some games even openly embrace it. Voxelgarden is the best example, but also Minetest Hades / Hades Revisited use it. I know you want to replace v6 at some point, but we are simply not there yet. Your claim that “new worlds are rarely started using v6” is completely false.

I think you are also grossly overexaggerating the complexity of the code. A few `if` statements to handle special cases are hardly complex.

> So this amount of complex specialisation for it is not justified.
> 
> Which biomes result from the flags is somewhat obvious already, so there is little justification for this special treatment.
> 
> So please have v6 taken care of by the simple non-mgv6 code.

With regards to v6, you're basically asking me to make the UI design deliberately worse for some weird sense of “code purity”. I said already that I don't think this code is very complex. Also, since this is v6, this code is likely not required to be touched anytime soon, so I absolutely don't see a problem.

The way how v6 selects biomes is a bit weird if you look at how the flags work: `snowbiomes` implies `jungles`, which means that `snowbiomes,jungles` gives the same result as `snowbiomes,nojungles`, which in turn means if you do this in checkboxes (as you seem to suggest), the `jungles` checkbox becomes pointless when `snowbiomes` is enabled. It's very confusing to have a checkbox that does nothing under certain circumstances.
So although 2 checkboxes can have 4 states together, there would be only 3 *meaningful* combinations with `snowbiomes` and `jungles`. This is all a bit confusing so I simply abstracted all of that away from the user point of view. I think this is a good thing, and removing that for the sake of more “pure” code will make the UI worse.

> Docs already explain how v6 flags are interpreted:

That's true, but completely missing the point of this PR. This text won't be visible in world creation dialog and the fewest people proactively read `settingtypes.txt`.

rubenwardy:
I suggest hiding this behind a collapsing section named "Mapgen Settings" to avoid spamming the user with options. It would also be nice if it could calculate the required dialog height

One good way to handle both of those things is to separate the logic for the GUI code - the formspec code should just be given a list of settings and options to generate from, and so can work out how much space it needs.

Wuzzy2:
This sounds nice in theory, but there's a catch: Where's the widget for collapsing sections?

rubenwardy:
button[]/checkbox[] and if statements

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -17,13 +17,115 @@
...
+		else
+			str = str .. "no" .. flag
+		end
+		first = false
+	end
+	return str
+end
+
+local function chflag(flags, flag)
+	if flags[flag] == true then
```
SmallJoker: Suggestion: `return (flags[flag] == true) and "true" or "false"`

At least to me it is not clear what this function does. What do you think about `flag_value_str` or `flag_to_str`?

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -17,13 +17,115 @@
...
+	for flag, is_set in pairs(ftable) do
+		if not first then
+			str = str .. ","
+		end
+		if is_set then
+			str = str .. flag
+		else
+			str = str .. "no" .. flag
+		end
+		first = false
```
SmallJoker: You could use a table here.
1) Takes care of tailing comma separation
2) Is faster than string concat
3) Is shorter
```Lua
-- Convert { jungles = true, caves = false } to "jungles,nocaves"
local str = {}
for flag, is_set in pairs(ftable) do
	str[#str + 1] = is_set and flag or ("no" .. flag)
end
return table.concat(str, ",")

File: builtin/mainmenu/dlg_create_world.lua [hunk 10]
```diff
@@ -17,13 +17,115 @@
...
+
+local function chflag(flags, flag)
+	if flags[flag] == true then
+		return "true"
+	else
+		return "false"
+	end
+end
+
+local function chflag2(flags, flag)
```
SmallJoker: Function name suggestion: `check_flag`

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -17,13 +17,115 @@
...
+	return flags[flag] == true
+end
+
+local tt_caverns = fgettext("Large cavities will be carved in the ground")
```
SmallJoker: Suggestion: Put the entire cavern entry into this variable. It'll be the same anyway.

builtin/mainmenu/dlg_create_world.lua
SmallJoker: Only after reading @Wuzzy2's comment I understood what's meant by "aggressively". It makes sense, but does not provide helpful information what's exactly meant here.

The only suggestion I have here is to remove "aggressively" (as well) because "carved" may already indicate that it'll happen no matter the elevation of the surrounding terrain.

builtin/mainmenu/dlg_create_world.lua
SmallJoker: Self-explaining. Drop the tooltip.

doc/lua_api.txt
```diff
@@ -67,6 +67,10 @@ The game directory can contain the following files:
     * `disallowed_mapgens = <comma-separated mapgens>`
       e.g. `disallowed_mapgens = v5,v6,flat`
       These mapgens are removed from the list of mapgens for the game.
+    * `disallowed_settings = <comma-separated mapgens>`
+      e.g. `disallowed_settings = mgv5_spflags`
```
SmallJoker: Q: Could the game's `minetest.conf` set the value of `mgv5_spflags` to - say `nocaverns` - so that it is used in the generated world?

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -150,11 +379,60 @@ local function create_world_buttonhandler(this, fields)
...
+				ftable.snowbiomes = false
+			end
+			local flags = table_to_flags(ftable)
+			core.settings:set("mgv6_spflags", flags)
+			return true
+		end
+	end
+
+	if fields["dd_mapgen"] then
+		core.settings:set("mg_name",fields["dd_mapgen"])
```
SmallJoker: Space after comma

builtin/mainmenu/dlg_create_world.lua
SmallJoker: This generic GUI is to make it intuitive for any user so they get frustrated when one combination does not work. 
Special treatment is totally fine to me.

However, _please_ combine the dropdown entries and the setting overrides. Example:
```Lua
local mgv6_biomes = {
	{
		fgettext("Temperate, Desert, Jungle, Tundra, Taiga"),
		{jungles = true, snowbiomes = true}
	},
	{
		fgettext("Temperate, Desert, Jungle"),
		{jungles = true, snowbiomes = false}
	},
	{
		fgettext("Temperate, Desert"),
		{jungles = false, snowbiomes = false}
	}
}
```
1) `dropdown` return values can be indexed directly
2) No `string.split` hack
3) Setting overrides are specified directly with the description
4) Allows simple iterating

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -51,46 +154,172 @@ local function create_world_formspec(dialogdata)
...
+		return form, y
+	end
+
+	local mg_specific_flags = function(mapgen, y)
+		if flag_checkboxes[mapgen] then
+			if disallowed_settings["mg"..mapgen.."_spflags"] then return "" end
+			local form = ""
+			for _,tab in pairs(flag_checkboxes[mapgen]) do
+				local id = "flag_mg"..mapgen.."_"..tab[1]
+				form = form .. "checkbox[4.2," .. y .. ";" .. id .. ";" .. tab[2] .. ";" ..
```
SmallJoker: Suggestion:
```Lua
form = form .. ("checkbox[4.2,%f;%s;%s;%s]"):
	format(y, id, tab[2], chflag(flags[mapgen], tab[3]))
```

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -51,46 +154,172 @@ local function create_world_formspec(dialogdata)
...
+				"3: " .. fgettext("Temperate, Desert") ..
+
+				-- There's no entry for to "nojungles,snowbiomes" because
+				-- "snowbiomes" force-enables "jungles".
+
+				";" .. biometype.. "]"
+
+				y = y + 0.55
+				form = form .. "checkbox[4.2,"..y..";flag_mgv6_biomeblend;" ..
+				fgettext("Biome blend") .. ";"..chflag(flags.v6, "biomeblend").."]" ..
```
SmallJoker: Missing indents for continued lines.

builtin/mainmenu/dlg_create_world.lua
SmallJoker: This is fine to me if all mvg6 information is together in one table. Comment below.

Please return here like so to reduce indents:
```Lua
if mapgen ~= "v6" then
	-- No special treatment
	return "", y
end
```

Wuzzy2:
> Q: Could the game's minetest.conf set the value of mgv5_spflags to - say nocaverns - so that it is used in the generated world?

Yes. But that's not the point here. This setting `disallowed_settings` is only a GUI thing. It is analogous to `disallowed_mapgens` which existed a while ago. It does not touch the setting at all, it is just hidden. The game is supposed/expected to force-set their flags when they hide them. The reason I added this is to hide unneccessary settings for games that like to override things, to avoid giving the user the illusion of control (it's pointless to display the `mgv5_flags` settings if it's overwritten by game anyway).

> Setting overrides are specified directly with the description

Downside: It depends on the translation, which is not predictable which is why I used this weird numbering. Although it's very unlikely in this case that the translation for 2 of these 3 strings will be the same. IMO it was a major design flaw that `receive_formspec_fields` returns a string instead of the index.

Wuzzy2:
@rubenwardy: Hmm, when I think about it, I am not so sure if it's a good idea to hide the additional settings in collapsible menu anymore. The idea not only has upsides. It would increase interaction cost. I don't think we make use of such collapsible menus anywhere else in Minetest, right?

Also, all the settings I chose are pretty high-level.

https://www.nngroup.com/articles/accordions-complex-content/

The key point in this above article is this:

> Hiding content behind navigation diminishes people’s awareness of it.

So it's not a black-and-white matter here.

Another problem I see that by hiding the settings, it does not mean they are not effective. For example, if you left on the `flat` flag for v6, and then after a year you return to Minetest and create a v6 world (and the settings are collapsed), you might be surprised why the v6 mapgen is flat for “no reason”.

builtin/mainmenu/dlg_create_world.lua
paramat: Or alternatively i mean: Why have an over-explaining tooltip for caves but nothing for dungeons?
If dungeons are obvious and need no tooltip, then caves are too, i do not understand the inconsistency.

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -51,46 +159,165 @@ local function create_world_formspec(dialogdata)
...
+			fgettext("Caves") .. ";"..strflag(flags.main, "caves").."]"
+		y = y + 0.5
+
+		form = form .. "checkbox[4.2,"..y..";flag_mg_dungeons;" ..
+			fgettext("Dungeons") .. ";"..strflag(flags.main, "dungeons").."]"
+		y = y + 0.5
+
+		local d_name
+		if mapgen == "v6" then
+			d_name = fgettext("Decorations (without trees and jungle grass)")
```
paramat: Thanks, this section looks simpler now.
This text might be better as something like:
"Decorations (no effect on trees and jungle grass)"
The current text is a little unclear.

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -51,46 +159,165 @@ local function create_world_formspec(dialogdata)
...
+		else
+			d_name = fgettext("Decorations")
+		end
+		form = form .. "checkbox[4.2,"..y..";flag_mg_decorations;" ..
+		d_name .. ";" ..
+		strflag(flags.main, "decorations").."]" ..
+		"tooltip[flag_mg_decorations;" ..
+		fgettext("Decorations are structures created by the game that appear on the surface") ..
```
paramat: We have underground decorations now too, that appear on the floors or ceilings of underground voids.

builtin/mainmenu/dlg_create_world.lua
paramat: Nevermind, sorry, answered in thread:
"It's basically there to distinguish caves from caverns, two flags which are subtly different."
FIne for me.

paramat:
I expected to be overruled by the other core devs on the special treatment for V6 =) I do not like it but I will not disapprove the PR due to that.
MT has a lot of 'pain in the ass' stuff for supporting V6, i just do not like adding more when it is not essential.

Wuzzy2, concerning your unhappiness at the start of your post above:

We disagree, that is fine. None of this is an insult and should not be taken as such. I make decisions purely based on what i feel is best for MT, this unavoidably sometimes makes contributors unhappy.
How quickly i make decisions is irrelevant, often i know very quickly that something is, in my opinion, wrong. Sometimes i get it wrong and change my mind and apologise.
We have to work quickly due to limited core dev time. Sometimes this can come across as strictness, abruptness or harshness. I do not have time to write 'in my opinon' in every sentence, that should be obvious anyway.

> The main fallacy you're suffering most of the time (and in this issue) is that you're viewing the UI only through your own eyes and only think about your own use case and assume that every other user thinks like you.

This is obviously ridiculous and foundationless. It is a cheap and common accusation used by those who disagree with a core dev's decision and try to demonise the core dev instead of accepting a difference of opinion. I will not read your linked articles.

> I know you want to replace v6 at some point,

I would personally love to remove it but i do not think it will be possible for a long time, there are servers dependent on V6 remaining as it is. A modernised replacement would be significally different and servers would not be happy with that. So i have given up on the idea of a modernised version.

As for V6 biome special treatment, you did good and thorough work, i would just prefer simplicity, standard checkboxes and the odd behaviour explained by text as it is in All Settings:
```
#    The 'snowbiomes' flag enables the new 5 biome system.
#    When the 'snowbiomes' flag is enabled jungles are automatically enabled and
#    the 'jungles' flag is ignored.
```

Please can the disallowed_settings feature be left to another PR? I might support it, but i am not yet convinced it is justified and i do not think it should be decided upon by this PR which is obviously desired.

Otherwise, i currently have no objections to this PR.

SmallJoker:
* :ballot_box_with_check: Changed checkbox options are applied as expected
* :negative_squared_cross_mark: The world creation formspec goes off-screen using the default settings.
  * If not collapsing, how about tabs?
* :negative_squared_cross_mark: The default value of (at least) `mgv6_spflags` is not read properly.
  * Example: `Trees and jungle grass` is not checked but after creating the world both is generated

Screens are usually wider than tall, thus it might also fit to re-arrange the layout as follows:
grafik

builtin/mainmenu/dlg_create_world.lua
SmallJoker: Also: Missing indent. These are continued lines.

Wuzzy2:
> The default value of (at least) mgv6_spflags is not read properly.

Oh crap! :-(

You're right for all `*_spflags`, actually.

So yeah, current code simply doesn't read the defaults, that's the reason. If the setting is not in `minetest.conf`, all flags default to false. Not good.
Apparently it is impossible to read the default value for settings with the current API. So I would have to implement that as well. Oh boy. (I could also hardcode the defaults but that's terrible.)

What should I do?

> Please can the disallowed_settings feature be left to another PR?

Yes, but I don't like it. This seems pointless additional workload for me for no benefit. Either `disallowed_settings` is accepted, then a separate PR is not needed. Or it is rejected, in which case a separate PR is completely pointless. My point is: It's much easier for me if `disallowed_settings` is decided in this issue. If you want to reject it, please reject it *here* so I can remove it *here*.

Wuzzy2:
Hmm, I just figured out something weird:

With a clean and empty `minetest.conf`, when I do `core.settings:get("mg_flags")`, I get `"caves,dungeons,light,decorations,biomes"`.
But when I do `core.settings:get("mgv6_spflags")` (or any other `*_spflags`), I get `nil`. That explains why the defaults work for the main flags but not for the `*_spflags`.

SmallJoker:
Apparently the default values are missing. Adding them here should solve the problem:
https://github.com/minetest/minetest/blob/44458896d9097c15c722c5bffb3d72e14962b7ce/src/defaultsettings.cpp#L424-L433

Wuzzy2:
https://github.com/minetest/minetest/pull/9279

paramat:
> My point is: It's much easier for me if disallowed_settings is decided in this issue

Yes, please can other core devs consider that feature?
Keep in mind though, the feature is very separate from what this PR does, so it should really have been a separate PR or at least a separate commit.

Currently i am -1 for it because it has not been mentioned or requested anywhere, and games are much less sensitive to mapgen settings than mapgens.
The justification given above for it was:
> Well, same reason as disabled_mapgens, basically. Except that it's not for the mapgen itself, but for mapgen settings. If a game wants to force its mapgen settings, those should be hidden from the UI as irrelevant. Note this still does not affect All Settings.

I cannot see any explanation of need or usefulness in that.
This says, "same reason as disallowed_mapgens", but the features are different so the same reason does not apply.
This says "If a game wants to force its mapgen settings", with an "if". It does not state that games need to, or how this will be useful.
"If a game wants to force its mapgen settings, those should be hidden from the UI as irrelevant." suggests that part of the reason you added the feature was due to formatting in this PR, not because it is needed or useful.
I could be convinced but there has been no valid argument made for it yet.

paramat:
grafik

The mapgen flags should be separated into global and mapgen-specific, not combined into one list. The distinction is important to make, combining will mislead and lead to confusion.
Sorry, this is unacceptable, please could you separate?

paramat:
The important size distinction between caves and caverns is not clear:
Currently:
Caves:
"Carve a large, interconnected network of tunnels and caves in the terrain"
Caverns:
"Large cavities will be carved in the ground"

'Large' is not sufficient to describe caverns and make them distinct from other caves.
Using the word 'large' for a worldwide system is redundant, and 'interconnected' is redundant because 'network' is present.
It is also obvious that caves are in the ground or terrain.
'Carve' or 'carving' is unnecessary,.
The fact that caverns are deep underground could be included, as this is another distinction.

For obvious reasons (formatting, space, translation), text in menus and tooltips must be as short and concise as possible, with no unnecessary or redundant information.
My request/ suggestion:
"A network of tunnels and caves"
Caverns:
"Very large caverns deep underground"

paramat:
For obvious reasons (formatting, space, translation), text in menus and tooltips must be as short and concise as possible, with no unnecessary or redundant information.
My requests/suggestions for the others:

V7
{ "ridges", fgettext("Rivers"), "ridges",
fgettext("Sea level rivers") },
{ "floatlands", fgettext("Floatlands (experimental)"), "floatlands",
fgettext("Floating landmasses in the sky") },
Floatlands are not always large, and 'high' is redundant. The experimental warning is already present in the label.

Valleys
{ "vary-river-depth", fgettext("Vary river depth"), "vary_river_depth",
fgettext("Low humidity and high heat causes shallow or dry rivers") }
'If enabled' is redundant.

Fractal
{ "terrain", fgettext("Terrain"), "terrain",
fgettext("Generate non-fractal terrain: Ocean and underground") },
Any islands are small and rare, not worth mentioning.

V6
{ "mudflow", fgettext("Mud flow"), "mudflow",
fgettext("Terrain surface erosion") },
'If enabled' is redundant.

Decorations
fgettext("Structures appearing on the terrain, such as trees and plants")
'Decorations are', 'created by the game' and 'may appear' are unnecessary. These are usually trees and plants.

paramat:
> With a clean and empty minetest.conf, when I do core.settings:get("mg_flags"), I get "caves,dungeons,light,decorations,biomes".
But when I do core.settings:get("mgv6_spflags") (or any other *_spflags), I get nil. That explains why the defaults work for the main flags but not for the *_spflags.

Due to possible issues with #9279 you can workaround this problem by hardcoding the defaults, because the defaults for V6 are highly unlikely to change now.

paramat:
The Carpathian river flag has no tooltip, rivers are sea level as in V7, so should be consistent with that flag.
However, i doubt a tooltip is needed for V7 and Carpathian rivers, the fact the river is at sea level is not essential information.

Wuzzy2:
> This says "If a game wants to force its mapgen settings", with an "if". It does not state that games need to, or how this will be useful.

Sigh. This feature is clearly useful to games, I don't understand why you want to shut it down. Of course some games want to force some mapgen flags, it's absurd to assume otherwise. If this is not added, we are just forcing all games to have all mapgen flags displayed, even if they are meaningless for that game. And that's just poor UI design.
Case in point: Haded Revisited, which wants to overwrite various v6 flags.

> The mapgen flags should be separated into 'Mapgen Flags' and 'Mapgen Specific Flags' (should be the same terminology as in All Settings), not combined into one list. The distinction is important to make, combining will mislead and lead to confusion because flags settings in All Settings and the usage in .conf will then be different.

I'm not happy with using the word “flags” here as it's technical programmer lingo. Outside of the programmer world, “flags” WILL be misunderstood. It might be OK in All Settings (as it's the more “technical” menu), but remember the Create World dialog is a high-level representation of the settings, so stuff should be simplified here.
Also note the point here is not exactly to represent all possible flags, but to represent all relevant mapgen-related settings in a simple manner. It's just a coincidence that currently, that's only flags.

I think that using 2 headings might be a bit overkill. The distinction does not really matter to the player in this menu, the result is the same either way. Whether `caves` is available for all mapgens or only `v6` doesn't really make a difference from a high-level point of view. The fact that `caves` appears for all mapgens (except `singlenode`) is already evidendence that it's generic.

You seem to want a perfect 1:1 mapping to All Settings (in terms of UI), but this goes directly against a major design goal of this PR: Abstraction. For the same reason, I call the `ridges` flag “Rivers” instead of “Ridges”.

> (a long list of comments on tooltips and stuff)

A lot of good suggestions here, I approve. Yeah, wordiness can be reduced.

> Due to possible issues with #9279 you could possibly workaround this problem by hardcoding the defaults if core.settings:get() returns nil.

This is terrifying.

paramat:
I can see your intention is good in using a single list of mapgen options, but this will actually cause harm to both inexperienced and more experienced players. This has been dumbed-down a little too much, to the point where it actually causes problems.

Inexperienced players will get used to a single list of mapgen options called 'mapgen settings'. As they progress and start using All Settings and maybe adding lines to minetest.conf, they will then be more confused and have more difficulty because these settings are actually called 'mapgen flags' and are split into general flags and mapgen-specific flags.
Everyone who is familiar with mapgen flags calls them mapgen flags, so any discussion in the forum or on IRC will use these terms, this again will cause difficulty and confusion for inexperienced players in discussions when they seek help.

When they start looking at minetest.conf they will not make the connection with 'mg_flags' and 'mgXXX_spflags'.
The 'readable, friendly name' for these settings, as used in All Settings, is 'mapgen flags'.
Currently it looks like each mapgen has a list of options, where every mapgen has 2 options called 'caves' and 'dungeons', this is very different from how it actually is so will be confusing. 'Caves' and 'dungeons' look like mapgen-specific flags.

I agree that 'mapgen flags' is slightly odd sounding terminology for an inexperienced player, but this is far outweighed by the advantages, explained above, of using the common terminology from the start.
From the setting names and the tooltips it will quickly be obvious that these are mapgen options even if they are called 'flags'.
It is better to give inexperienced players a proper start that brings benefits for them later.

For slightly more experienced players, how it is currently is also confusing, because they expect mapgen flags to be called 'flags', and split into general and mapgen-specific.
Currently it looks like 'caves' and 'dungeons' are mapgen-specific flags for each mapgen.

So :-1: for the PR until this is changed.

Thanks for the tooltip changes.

///////////

> Of course some games want to force some mapgen flags, it's absurd to assume otherwise.

This is an assumption. If it is so desired, why has it not been mentioned or requested at all until this PR?
This was clearly added primarily due to formatting issues of this PR, not due to an expressed need or usefulness.
I do not 'want to shut it down', i was just waiting for valid arguments or game makers expressing a need or a significant usefulness. I am happy to change my mind.
The 'disallowed mapgens' feature is certainly needed and was added mainly due to V6 being so different from the other mapgens.

>  If this is not added, we are just forcing all games to have all mapgen flags displayed, even if they are meaningless for that game.

That is an invalid argument. This argument could be applied to many other settings and used to justify many other similar features which are actually unnecessary.
You are starting with the assumption the feature is needed to justify that the feature is needed.

To be clear, i wait for the other core devs to decide and i will not disapprove if 2 other core devs support 'disallowed settings'. But i do insist this is called 'disallowed mapgen flags' because this is a feature used by games so should use the correct, common and technical terminology.

Also, sorry this PR has been quite hard going =)

Wuzzy2:
> (about mapgen flags)

Note that “mapgen settings” is a superset of “mapgen flags”, so the term “mapgen settings” is still true. And the goal is not only to represent mapgen flags, but *all* relevant mapgen settings. It's just a coincidence that currently from all the possible mapgen settings, only the flags were chosen.

I am not convinced that it will cause serious confusion. I am waiting for a 2nd opinion.

Also, what will happen when in future a mapgen setting appears that is so useful that it gets promoted to world creation dialog, but is not a flag?

Another option would be to avoid the word “flags” altogether and rename the setting itself. But this is out of scope for this PR.

> This is an assumption. If it is so desired, why has it not been mentioned or requested at all until this PR?

Very simple: These settings were invisible in the world creation dialog before. So these checkboxes simply did not exist before. Obviously there was no need to request the hiding of checkboxes that do not exist. But with the checkboxes now existing, things change. Just because nobody requested it so far does not make it invalid.

Let's make it simple:

Why were mapgens hidden? Because some games would not work with some mapgens, or they force the mapgen to a particular value.
Why do I want to hide mapgen settings? Because some games would not work with some mapgen settings, or they for them to a particular value.

See? It's the exact same rationale.

Finally, you are wrong that this is only an assumption. You have ignored Hades Revisited. :-1:

> This was clearly added primarily due to formatting issues of this PR, not due to an expressed need or usefulness.

I have no idea what you mean with “formatting issues of this PR”. [confused face]

> > If this is not added, we are just forcing all games to have all mapgen flags displayed, even if they are meaningless for that game.

> That is an invalid argument. This argument could be applied to many other settings and used to justify many other similar features which are actually unnecessary.

No, that's not true. This argument can *only* applied to settings that are actually highly visible (=in a place other than All Settings). Mapgen flags will be made highly visible, therefore, the argument applies for them. The argument does NOT apply to other settings, like mapgen noises, because they are not highly visible.

Therefore, the argument is NOT invalid.

> But i do insist this is called 'disallowed mapgen flags' because this is a feature used by games so should use the correct, common and technical terminology.

The name `disallowed_settings` is intentionally generic. With `disallowed_settings`, the feature can later be trivially extended to include other settings that are not mapgen related, like `enable_damage` and `creative_mode`.

If it's called `disallowed_mapgen_flags`, we restrict ourselves to mapgen flags only, and would have to add a new `game.conf` variable for every single new highly visible setting that comes along.

But I do not have strong feelings about this. I am waiting for a 2nd opinion on this.

> Also, sorry this PR has been quite hard going =)

This is because progress is delayed by many poor, fallacious arguments that need to be debunked first in order to proceed.

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -51,46 +159,173 @@ local function create_world_formspec(dialogdata)
 				end
 			end
 		end
+
+		local ds = (gameconfig:get("disallowed_settings") or ""):split()
```
paramat: The game setting is better called 'disallowed_mapgen_flags', because that is what it is. 'settings' can mean almost anything, and in code means all settings.

SmallJoker: It's literally settings. All settings.

doc/lua_api.txt
```diff
@@ -67,6 +67,10 @@ The game directory can contain the following files:
     * `disallowed_mapgens = <comma-separated mapgens>`
       e.g. `disallowed_mapgens = v5,v6,flat`
       These mapgens are removed from the list of mapgens for the game.
+    * `disallowed_settings = <comma-separated mapgens>`
```
paramat: "comma-separated mapgen flags"

builtin/mainmenu/dlg_create_world.lua
paramat: Yes that is what i mean, settings can mean any setting.
The usage of the feature is to disallow certain mapgen flags, so that is how it should be named, otherwise someone may think it can be used to disallow something that is not a mapgen flag.
@rubenwardy what is your objection to my suggestion?

paramat:
> Note that “mapgen settings” is a superset of “mapgen flags”, so the term “mapgen settings” is still true.

In MT we try to be as descriptive as possible.
> And the goal is not only to represent mapgen flags, but all relevant mapgen settings. It's just a coincidence that currently from all the possible mapgen settings, only the flags were chosen.

If this is just for flags currently then it should refer to flags.
> Also, what will happen when in future a mapgen setting appears that is so useful that it gets promoted to world creation dialog, but is not a flag?

If non-flag settings are added later those should be under additional and descriptive list titles. Putting everything into one list is what i object to.
> Another option would be to avoid the word “flags” altogether and rename the setting itself.

That is not an option, we are not going to dumb-down MT code itself.

I wrote:
> This is an assumption. If it is so desired, why has it not been mentioned or requested at all until this PR?

You replied:
> Very simple: These settings were invisible in the world creation dialog before. So these checkboxes simply did not exist before. Obviously there was no need to request the hiding of checkboxes that do not exist. But with the checkboxes now existing, things change.

That is not an answer. Until this PR, no one had requested a need for the disallowing of mapgen flags. I am referring to the ability itself, not the hiding of disallowed flag checkboxes.
You are repeating your circular argument. If no one requires the disallowing of mapgen flags then they do not need mapgen flag checkboxes hiding in the menu.

I did not ignore Hades Revisited, that is why i wrote 'until this PR'. If you needed the ability you would probably have requested it before this PR.
> I have no idea what you mean with “formatting issues of this PR”.

Sorry, i mean the hiding of mapgen flag checkboxes.
> Therefore, the argument is NOT invalid.

It is still invalid because it is a circular argument, i also wrote "You are starting with the assumption the feature is needed to justify that the feature is needed." which you ignored.
> The name disallowed_settings is intentionally generic. With disallowed_settings, the feature can later be trivially extended to include other settings

That is a good point. But it does not seem good practice to pile many different types of settings into one configuration string. This needs to be considered now.
This is why 'disallowed settings' is better left to another PR and not rushed in this one.
However, best i remain neutral for now.
> poor, fallacious arguments that need to be debunked first

I am reviewing just as i would any PR, the arguments arise due to your occasionally ridiculous logic and arguments. If i see nonsense i will point it out.

builtin/mainmenu/dlg_create_world.lua
paramat: The documentation of this does not make it clear this is for mapgen flags only.

Anyway, from discussion:
> The name disallowed_settings is intentionally generic. With disallowed_settings, the feature can later be trivially extended to include other settings that are not mapgen related

I am not sure it is acceptable to pile many different types of setting into one string. It may be better to keep this for mapgen flags only and add more configs if required.

Wuzzy2:
As I already said, I have no strong feelings about the name “mapgen settings” vs “mapgen flags” anymore but I am waiting for a 2nd opinion.

> That is not an answer. Until this PR, no one had requested a need for the disallowing of mapgen flags. I am referring to the ability itself, not the hiding of disallowed flag checkboxes.

But the setting is not about *disallowing* of mapgen flags. It's about hiding them in the highly visible part of the GUI like world creation dialog. **This is only a GUI thing, not more, not less.** The setting still exists, it is only hidden in the world creation dialog.

How the game treats the actual mapgen flags or whether it overwrites anything is entirely up to the Lua code in the game.

> I did not ignore Hades Revisited, that is why i wrote 'until this PR'. If you needed the ability you would probably have requested it before this PR.

I did not need the feature *before* this PR. Nobody did. It was impossible to request this feature before. Let me explain why …

Let's pretend I will remove `disallowed_settings`/`disallowed_mgflags` ...

* This PR will expose almost all mapgen flags to the world creation dialog
* But Hades Revisited overrides the mapgen flags of v6. **IMPORTANT: This is already the case, independent of this PR**
* When a player creates a new world in said game, the player will see the mapgen flags for v6
* But these checkboxes are pointless, the game will always overwrite the v6 flags
* Therefore, the player will see useless checkboxes that don't do anything
* It's bad UI design to display widgets that literally don't do anything. Ask @rubenwardy

And this would be bad UI design *regardless of how many people request this feature*. Just because nobody requested this feature before does not mean it is not required.
In fact, the problem I just described did not even exist before this PR. Because none of the mapgen flags appeared in the world creation dialog. So that's the reason why nobody requested it, not even me. The problem that this feature solves **did not exist before this PR**.

I hate it when widgets are useless and don't do anything. Anticipating this problem, I decided to immediately solve it. Therefore, this feature.

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -17,13 +17,120 @@
...
 
+	local flags = {}
+	flags.main = flags_to_table(core.settings:get("mg_flags"))
+	flags.v5 = flags_to_table(core.settings:get("mgv5_spflags"))
+	flags.v6 = flags_to_table(core.settings:get("mgv6_spflags"))
+	flags.v7 = flags_to_table(core.settings:get("mgv7_spflags"))
+	flags.fractal = flags_to_table(core.settings:get("mgfractal_spflags"))
+	flags.carpathian = flags_to_table(core.settings:get("mgcarpathian_spflags"))
+	flags.valleys = flags_to_table(core.settings:get("mgvalleys_spflags"))
+	flags.flat = flags_to_table(core.settings:get("mgflat_spflags"))
```
paramat: Maybe these should use:
https://github.com/minetest/minetest/blob/7c0e0c34ecb66e044f59cbab03a4c12fc7838e97/doc/lua_api.txt#L4532-L4542
?

paramat:
See https://github.com/minetest/minetest/pull/9284#issuecomment-570960119
Seems you can already do what you want by using whatever method 'All Settings' uses.

builtin/mainmenu/dlg_create_world.lua
paramat: Apparently this does not work in main menu.

Wuzzy2:
Ugh, I have lost track …
Is there a TODO list?

Wuzzy2:
@paramat: Setting renamed, captions added.

paramat:
Thanks, will retest. Removing my -1.

Wuzzy2:
Screenshot:
https://satoshiupload.com/images/OBCKN4ireq.png

paramat:
> But the setting is not about disallowing of mapgen flags. It's about hiding them in the highly visible part of the GUI like world creation dialog. This is only a GUI thing, not more, not less.

I misunderstood, i am so sorry ...
I think the 'disallowed' word also helped to confuse me.
It seems that 'disallowed mapgens' works the same way too: it only hides mapgens from the dropdown, it does not stop setting certain mapgens in All Settings or in minetest.conf (and maybe it should?). I think i have misunderstood that feature too.
So your addition seems consistent.
Maybe these settings should have been called 'hidden mapgens' and 'hidden flags' instead? But, too late.

paramat:
Retested, all looks and acts really good now.

Wuzzy2:
Just a reminder: This PR is currently blocked by #9284.

(This little fact can get easily drowned in this gigantic discussion, so I better mention it explicitly.)

Wuzzy2:
#9284 was merged! This PR is now un-blocked!

I have added the required `get_flags` function calls. Have fun! :-)

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -150,11 +375,53 @@ local function create_world_buttonhandler(this, fields)
...
+		local split = string.split(k, "_", nil, 3)
+		if split and split[1] == "flag" then
+			local setting
+			if split[2] == "mg" then
+				setting = "mg_flags"
+			else
+				setting = split[2].."_spflags"
+			end
+			-- We replaced the underscore of flag names with a dash.
+			local flag = string.gsub(split[3], "-", "_")
```
SmallJoker: If splits are already limited to 3 splits, in which case is this workaround needed?

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -51,46 +142,180 @@ local function create_world_formspec(dialogdata)
...
+		y = y + 0.5
+
+		form = form .. "tooltip[flag_mg_caves;" ..
+		fgettext("Network of tunnels and caves")
+		.. "]"
+		return form, y
+	end
+
+	local mg_specific_flags = function(mapgen, y)
+		if flag_checkboxes[mapgen] then
```
SmallJoker: Same functionality, but easier to read due to reduced indents:
```Lua
if not flag_checkboxes[mapgen] then
	return "", y
end
```

Wuzzy2:
> If splits are already limited to 3 splits, in which case is this workaround needed?

The 3rd split does not return the remainder of the string, but only until the next underscore. Therefore, `split[3]` of `"flag_mgvalleys_altitude_chill"` would yield `"altitude"` and not `"altitude_chill"`.

So, this cannot be simplified as implied. I tried.

> Same functionality, but easier to read due to reduced indents:

Refactored as requested.

SmallJoker:
> So, this cannot be simplified as implied. I tried.

Too bad. So it splits anyway rather than returning the leftover text. Latter would be move intuitive to me but well :man_shrugging: I guess the workaround is the only option here,

Wuzzy2:
Bump after 2 months.

rubenwardy:
This breaks the game warning: 
image


I'm also sad to see you didn't benefit from containers when making the two halves of this formspec

Wuzzy2:
* Containers added
* Fixed broken formspec when no game or only `minimal` is installed
* Rebased
* Squashed

Wuzzy2:
I changed the tooltip for the `decorations` checkbox when in v6. It was:

> Structures appearing on the terrain, typically trees and plants

This was misleading (because it doesn't affect v6 trees and junglegrass). It is now:

> Structures appearing on the terrain

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -35,78 +136,211 @@ local function create_world_formspec(dialogdata)
...
 
 	local game_by_gameidx = core.get_game(gameidx)
+	local disallowed_mapgen_flags = {}
 	if game_by_gameidx ~= nil then
 		local gamepath = game_by_gameidx.path
 		local gameconfig = Settings(gamepath.."/game.conf")
 
-		local allowed_mapgens = (gameconfig:get("allowed_mapgens") or ""):split()
-		for key, value in pairs(allowed_mapgens) do
-			allowed_mapgens[key] = value:trim()
```
sfan5, line 44: so is this removed without a replacement or?

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -17,13 +17,114 @@
...
+	for flag, is_set in pairs(ftable) do
+		str[#str + 1] = is_set and flag or ("no" .. flag)
+	end
+	return table.concat(str, ",")
+end
+
+-- Returns true if flag in a given table of flags is true
+local function check_flag(flags, flag)
+	return flags[flag] == true
+end
```
sfan5, line 32: surely this should just be inlined

Wuzzy2:
Oh crap, yes, the allowed mapgens were accidentally removed.
In my defense: This PR is very old and was created long before `allowed_mapgens` was a thing. And I must have screwed up the merge then … :-(

Anyway, ready for re-review.

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -17,13 +17,109 @@
...
 
+	local flags = {}
+	flags.main = core.settings:get_flags("mg_flags")
+	flags.v5 = core.settings:get_flags("mgv5_spflags")
+	flags.v6 = core.settings:get_flags("mgv6_spflags")
+	flags.v7 = core.settings:get_flags("mgv7_spflags")
+	flags.fractal = core.settings:get_flags("mgfractal_spflags")
+	flags.carpathian = core.settings:get_flags("mgcarpathian_spflags")
+	flags.valleys = core.settings:get_flags("mgvalleys_spflags")
+	flags.flat = core.settings:get_flags("mgflat_spflags")
```
sfan5, line 121: This can be declared inline:
```
local flags = {
    main = core.settings:get(...),
    ...
}
```

builtin/mainmenu/dlg_create_world.lua
```diff
@@ -64,49 +161,194 @@ local function create_world_formspec(dialogdata)
...
+	y = y + 0.3
+	str_flags, y = mg_main_flags(current_mg, y)
+	if str_flags ~= "" then
+		label_flags = "label[0,"..y_start..";" .. fgettext("Mapgen flags") .. "]"
+		y_start = y + 0.4
+	else
+		y_start = 0.0
+	end
+	y = y_start + 0.3
+	str_spflags = mg_specific_flags(current_mg, y)
```
sfan5, line 302:
```suggestion
	str_spflags, y = mg_specific_flags(current_mg, y)
```

doc/lua_api.txt
```diff
@@ -75,6 +75,10 @@ The game directory can contain the following files:
       When both `allowed_mapgens` and `disallowed_mapgens` are
       specified, `allowed_mapgens` is applied before
       `disallowed_mapgens`.
+    * `disallowed_mapgen_flags = <comma-separated mapgen flags>`
```
sfan5, line 78: this is extremely misnamed: `mgv5_spflags` is not a mapgen flag, `biomes` would be one
I suggest `disallowed_mapgen_settings`

Wuzzy2:
> this is extremely misnamed: mgv5_spflags is not a mapgen flag, biomes would be one
I suggest disallowed_mapgen_settings

LOL. This was called `disallowed_mapgen_settings` before, but IIRC I changed it because of coredev request … Oh man. How about `disallowed_mapgen_spflags`?

sfan5:
> This was called `disallowed_mapgen_settings` before, but IIRC I changed it because of coredev request

It was called `disallowed_settings` before, paramat [suggested](https://github.com/minetest/minetest/pull/9254#issuecomment-570732244) `disallowed_mapgen_flags` (which is the wrong technical term...) but I think his main reason was so it doesn't give the impression that it applies to <i>more</i> than just mapgen.
So I don't think he'd disagree with `disallowed_mapgen_settings`.

> How about `disallowed_mapgen_spflags`?

No, this is just as incorrect as `disallowed_mapgen_flags`.

Wuzzy2:
2 new commits!

> str_spflags = mg_specific_flags(current_mg, y)

This is intentional, I don't want to overwrite `y` at this point.

paramat:
> this is extremely misnamed: mgv5_spflags is not a mapgen flag, biomes would be one

Not really, 'mapgen specific flags' are still mapgen flags, so in a non-technical way this is correct.
I cannot remember now, what values can the `disallowed_mapgen_settings` configuration actually take? I seem to remember it was only for 'mapgen specific flags', in which case `disallowed_mapgen_specific_flags' would be the most correct. However that is long and technical.

So i am ok with `disallowed_mapgen_settings`.

sfan5:
> Not really, 'mapgen specific flags' are still mapgen flags

But `mgv5_spflags` is not a mapgen flag, it is a setting that specifies mapgen flags.

paramat:
> it is a setting that specifies mapgen flags.

Uh of course, good point, it is a collection of flags =)

FreeLikeGNU:
Just want to say thank you for this!

PR #9255, opened 2019-12-27
New object property to hide selection box. (#7998) rebased.

Unarelith:
See #7998. I made this PR because the other one couldn't be reopened.

~What's left to do:~
- [x] ~Show the selection box in debug mode~

~Can I get any help on how to do that? I've no idea on where I can find this.~

### DIFF
--- a/src/client/content_cao.cpp
+++ b/src/client/content_cao.cpp
@@ -24,6 +24,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include <IMeshManipulator.h>
 #include <IAnimatedMeshSceneNode.h>
 #include "client/client.h"
+#include "client/gameui.h"
 #include "client/renderingengine.h"
 #include "client/sound.h"
 #include "client/tile.h"
@@ -1655,6 +1656,12 @@ std::string GenericCAO::debugInfoText()
 	}
 	os<<"}";
 	return os.str();
+}
+
+bool GenericCAO::doShowSelectionBox()
+{
+	return m_prop.show_selection_box
+		|| (m_client && m_client->getGameUI()->getFlags().show_debug);
 }
 
 // Prototype
--- a/src/client/content_cao.h
+++ b/src/client/content_cao.h
@@ -263,4 +263,8 @@ public:
 	{
 		return m_prop.infotext;
 	}
+
+	bool doShowSelectionBox() override {
+		return m_prop.show_selection_box;
+	}
 };
--- a/src/client/content_cao.h
+++ b/src/client/content_cao.h
@@ -264,7 +264,8 @@ public:
 		return m_prop.infotext;
 	}
 
-	bool doShowSelectionBox() override {
+	bool doShowSelectionBox() override
+	{
 		return m_prop.show_selection_box;
 	}
 };
--- a/src/client/content_cao.h
+++ b/src/client/content_cao.h
@@ -21,14 +21,15 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 
 #include <map>
 #include "irrlichttypes_extrabloated.h"
+#include "client.h"
 #include "clientobject.h"
+#include "gameui.h"
 #include "object_properties.h"
 #include "itemgroup.h"
 #include "constants.h"
 #include <cassert>
 
 class Camera;
-class Client;
 struct Nametag;
 
 /*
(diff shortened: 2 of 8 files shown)


DS-Minetest:
> Can I get any help on how to do that? I've no idea on where I can find this.

You could add a public getter for [`m_game_ui`](https://github.com/minetest/minetest/blob/0b2f0914cc800bb3ae4c1f21e0c28cfdacc65f3f/src/client/client.h#L583) in `client.h` and then do in `content_cao.h`:
```cpp
	bool doShowSelectionBox() override {
		return m_client->getGameUI().getFlags().show_debug || m_prop.show_selection_box;
	}
```
(See also [here](https://github.com/minetest/minetest/blob/e462a9a5ef46776d974203bb44e0b89e8c980b29/src/client/gameui.h#L60). Note: You might need to check for nullptrs. And it is possible that this suggestion is bad.)

src/client/content_cao.h
```diff
@@ -263,4 +263,8 @@ class GenericCAO : public ClientActiveObject
 	{
 		return m_prop.infotext;
 	}
+
+	bool doShowSelectionBox() override {
```
SmallJoker:
```suggestion
	bool doShowSelectionBox() override
	{
```

src/object_properties.cpp
```diff
@@ -167,4 +169,5 @@ void ObjectProperties::deSerialize(std::istream &is)
 	eye_height = readF32(is);
 	zoom_fov = readF32(is);
 	use_texture_alpha = readU8(is);
+	show_selection_box = readU8(is);
```
SmallJoker: This will read `0` (or `false`) on older servers. Check the remaining data bytes or maybe a try/catch is needed here for compatibility.

src/script/common/c_content.cpp
```diff
@@ -410,6 +411,8 @@ void push_object_properties(lua_State *L, ObjectProperties *prop)
 	lua_setfield(L, -2, "zoom_fov");
 	lua_pushboolean(L, prop->use_texture_alpha);
 	lua_setfield(L, -2, "use_texture_alpha");
+	lua_pushboolean(L, prop->show_selection_box);
+	lua_setfield(L, -2, "show_selection_box");
```
SmallJoker: Documentation -> `lua_api.txt`

src/object_properties.cpp
Unarelith: Which server versions can have this problem?
I tried to fix it, tell me if it's ok.

SmallJoker: Basically any that were build prior this PR. The client expects a byte there, but that isn't sent by the server. Thus, it will either default to 0 or throw some exception.
To test this use an older server build and join with a new client. The client should see the selection boxes.

Unarelith:
Ok ready for review, seems like everything works as expected.

src/client/content_cao.h
```diff
@@ -263,4 +264,12 @@ class GenericCAO : public ClientActiveObject
 	{
 		return m_prop.infotext;
 	}
+
+	bool doShowSelectionBox() override
+	{
+		if (m_client && m_client->getGameUI())
```
SmallJoker:
```C++
return m_prop.show_selection_box || (m_client && m_client->getGameUI()->getFlags().show_debug);
```
1) Slightly faster due to short-circuit `OR` if the first condition is `true`
2) Removes superfluous `m_client->getGameUI()` check because this cannot be `nullptr` (see `game.cpp`)

src/client/content_cao.h
```diff
@@ -263,4 +264,12 @@ class GenericCAO : public ClientActiveObject
 	{
 		return m_prop.infotext;
 	}
+
+	bool doShowSelectionBox() override
```
SmallJoker: `content_cao.h` is included 6x in a source file, no header files.
For incremental builds this means that 6 files will need to be recompiled whenever this or one of the included header files change. In the worst case, there's a chain of header inclusions (`log.h` for example) - resulting in many, many source files that need to be recompiled even though there are no changes for them.

Hence, please move the definition to `content_cao.cpp` and keep the header light as possible.

Unarelith:
Ok I made the requested changes.

src/client/content_cao.cpp
```diff
@@ -1657,5 +1658,11 @@ std::string GenericCAO::debugInfoText()
 	return os.str();
 }
 
+bool GenericCAO::doShowSelectionBox()
+{
+	return m_prop.show_selection_box
+		|| (m_client && m_client->getGameUI()->getFlags().show_debug);
```
sfan5: this kinda breaks the abstraction
I suggest moving the show_debug check to `Game::updatePointedThing`

Unarelith:
Sorry, I'm dropping this PR, I don't have time for this because of [OpenMiner](https://github.com/Unarelith/OpenMiner).

PR #9258, opened 2019-12-28
GUIFormspecMenu: Add animated_image[] element 

ghost:
This is a rebased/updated version of #9032, intended to address the issues that were brought up during review. Since @kilbith wasn't in a position to rebase and complete the original PR, I am adopting in hopes of getting this past the home stretch!

A word of caution. The work on this was split across about a month (with a ~3-week gap) due to some irl issues, so while the main functionality works I can't guarantee that everything was properly handled. *Please test/review carefully!*

I should also point out that I haven't squashed the original PR's changes yet, I wanted the separation to be clear so that folks could more easily see what I've changed.

## How to Test
To test this PR, create a formspec using the new `animated_image` element and ensure that it renders as expected.

For your convenience, I've attached a minimal test mod: \
[test_animimage.zip](https://github.com/minetest/minetest/files/4145293/test_animimage.zip) This mod adds the `/anim` chat command, which displays a fomspec with an animated image inside. I recommend making further changes to the formspec for testing, but this should simplify the process of setting up your test environment.


src/client/texture.cpp
```diff
@@ -0,0 +1,64 @@
...
+{
+	m_name = _name;
+	m_frame_count = _frame_count;
+	m_frame_idx = 0;
+	m_frame_duration = _frame_duration;
+	m_frame_time = 0;
+	m_frame_names.resize(m_frame_count);
+
+	for (s32 i = 0; i < m_frame_count; ++i)
+		m_frame_names[i] = base_name + "_" + (char)(i + 1 + '0') + ".png";
```
DS-Minetest: This cast won't work correctly for `m_frame_count` greater than 9. I suggest using something like [`std::to_string` ](https://en.cppreference.com/w/cpp/string/basic_string/to_string) instead:
```suggestion
		m_frame_names[i] = base_name + "_" + std::to_string(i + 1) + ".png";
```

src/client/texture.cpp
```diff
@@ -0,0 +1,64 @@
...
+	else
+		return tsrc->getTexture(m_name);
+}
+
+void Texture::step(u64 step_duration)
+{
+	m_frame_time += step_duration;
+	m_frame_idx += u32(m_frame_time / m_frame_duration);
+	m_frame_idx %= m_frame_count;
+	m_frame_time %= m_frame_duration;
```
DS-Minetest: Shouldn't this be `m_frame_time %= m_frame_duration * m_frame_count;`?

src/client/texture_pool.h
```diff
@@ -0,0 +1,56 @@
...
+with this program; if not, write to the Free Software Foundation, Inc.,
+51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+#pragma once
+
+#include "texture.h"
+#include <string>
+#include <vector>
+#include <iostream>
```
DS-Minetest: What do you need `<iostream>` for?
```suggestion
```

src/gui/guiAnimatedImage.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+		const video::SColor color(255, 255, 255, 255);
+		const video::SColor colors[] = {color, color, color, color};
+
+		draw2DImageFilterScaled( driver, texture, AbsoluteRect,
+				core::rect<s32>(core::position2d<s32>(0, 0), texture->getOriginalSize()),
+				NULL, colors, true);
+	} else {
+		errorstream << "GUIAnimatedImage::draw() unable to load texture:" << std::endl;
+		errorstream << "\t" << m_name << std::endl;
```
DS-Minetest: Error message spam in draw functions can be problematic.

src/client/texture.cpp
Df458: No, it's correct. `m_frame_time` only refers to the current frame here, it's used to advance `m_frame_idx` at an interval of `m_frame_duration` ms.

If we change the line to `m_frame_time %= m_frame_duration * m_frame_count;`, what happens is that the value gets high enough to advance `m_frame_idx` up to `m_frame_count` times per-frame which messes up the playback.

src/gui/guiAnimatedImage.cpp
Df458: I just ported that log over from the original code. Would you prefer I remove it altogether, or should it be moved somewhere specific?

src/client/texture.cpp
Df458: Good catch! I've applied the fix and updated the test mod to use 10+ frames to confirm that it works.

src/gui/guiAnimatedImage.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+	if (texture) {
+		video::IVideoDriver *driver = Environment->getVideoDriver();
+
+		const video::SColor color(255, 255, 255, 255);
+		const video::SColor colors[] = {color, color, color, color};
+
+		draw2DImageFilterScaled( driver, texture, AbsoluteRect,
+				core::rect<s32>(core::position2d<s32>(0, 0), texture->getOriginalSize()),
+				NULL, colors, true);
```
DS-Minetest: Why shouldn't the new element be clippable?
```suggestion
				NoClip ? nullptr : &AbsoluteClippingRect, colors, true);
```

src/gui/guiAnimatedImage.cpp
DS-Minetest: I don't know. Finding this error elsewhere (eg, in the parser) will probably be difficult.

Df458: Actually... when the file isn't there it will normally generate a replacement and log an error anyway. I've decided to remove it, but I'll leave this unresolved in case others disagree.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -781,6 +782,54 @@ void GUIFormSpecMenu::parseImage(parserData* data, const std::string &element)
 	errorstream<< "Invalid image element(" << parts.size() << "): '" << element << "'"  << std::endl;
 }
 
+void GUIFormSpecMenu::parseAnimatedImage(parserData *data, const std::string &element)
+{
+	std::vector<std::string> parts = split(element, ';');
+
+	if ((parts.size() == 3) ||
+		((parts.size() > 3) && (m_formspec_version > FORMSPEC_API_VERSION)))
+	{
```
DS-Minetest: (`{` should be on the same line as the `if` parenthesis (`if (blabla\nbla) {`.)

```cpp
	if (parts.size() != 3 &&
			(parts.size() < 3 || m_formspec_version <= FORMSPEC_API_VERSION)) {
		errorstream << "Invalid animated image element(" << parts.size() << "): '" << element << "'"  << std::endl;
		return;
	}
...
```

src/gui/guiFormSpecMenu.cpp
```diff
@@ -781,6 +782,54 @@ void GUIFormSpecMenu::parseImage(parserData* data, const std::string &element)
...
+		v2s32 geom;
+
+		if (data->real_coordinates) {
+			pos = getRealCoordinateBasePos(v_pos);
+			geom = getRealCoordinateGeometry(v_geom);
+		} else {
+			pos = getElementBasePos(&v_pos);
+			geom.X = stof(v_geom[0]) * (float)imgsize.X;
+			geom.Y = stof(v_geom[1]) * (float)imgsize.Y;
+		}
```
DS-Minetest: I'd not support the old image pos and geom format or what that is.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -781,6 +782,54 @@ void GUIFormSpecMenu::parseImage(parserData* data, const std::string &element)
...
+
+		if (!data->explicit_size)
+			warningstream << "invalid use of image without a size[] element" << std::endl;
```
```suggestion
			warningstream << "invalid use of animated image without a size[] element" << std::endl;
```

src/gui/guiFormSpecMenu.cpp
```diff
@@ -781,6 +782,54 @@ void GUIFormSpecMenu::parseImage(parserData* data, const std::string &element)
...
+			L"",
+			L"",
+			258 + m_fields.size()
+		);
+
+		core::rect<s32> rect = core::rect<s32>(pos, pos + geom);
+
+		gui::IGUIElement *e = new GUIAnimatedImage(Environment, this, spec.fid,
+				rect, name, m_tsrc, &m_texture_pool);
+		e->setNotClipped(true);
```
DS-Minetest: Clipping should be allowed imo.
```suggestion
```

src/gui/guiFormSpecMenu.cpp
```diff
@@ -781,6 +782,54 @@ void GUIFormSpecMenu::parseImage(parserData* data, const std::string &element)
...
+			name,
+			L"",
+			L"",
+			258 + m_fields.size()
+		);
+
+		core::rect<s32> rect = core::rect<s32>(pos, pos + geom);
+
+		gui::IGUIElement *e = new GUIAnimatedImage(Environment, this, spec.fid,
+				rect, name, m_tsrc, &m_texture_pool);
```
DS-Minetest: `e` has to be dropped somewhere (eg. at the end of this function).

src/gui/guiFormSpecMenu.cpp
```diff
@@ -3296,6 +3350,11 @@ void GUIFormSpecMenu::drawMenu()
 		e->setVisible(false);
 	}
 
+	/*
+		Update animated images
+	*/
+	m_texture_pool.step();
+
```
DS-Minetest: This should probably be done a bit earlier, before anything is drawn, eg. in line 3316.

src/gui/guiFormSpecMenu.h
```diff
@@ -32,6 +32,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include "util/string.h"
 #include "util/enriched_string.h"
 #include "StyleSpec.h"
+#include "client/texture_pool.h"
```
DS-Minetest: Move the include into the .cpp file and add a `class TexturePool;` to the header (below this comment).

src/gui/guiAnimatedImage.cpp
Df458: Interestingly, `image[]` doesn't appear to be clippable either. I'll make a PR for that while I'm at it.

DS-Minetest:
The new files could use some code comments.

src/gui/guiFormSpecMenu.cpp
Df458: I like that, but it doesn't follow the existing convention used in every other `parseX()` function, as far as I can tell. I'll make the change if another reviewer agrees with this.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -103,6 +105,7 @@ GUIFormSpecMenu::GUIFormSpecMenu(JoystickController *joystick,
 	m_tsrc(tsrc),
 	m_client(client),
 	m_formspec_prepend(formspecPrepend),
+	m_texture_pool(new TexturePool()),
```
DS-Minetest: You have to delete this in the destructor (line 148).

Df458: Whoops, you're right. I need to be more careful, been making some really dumb errors on this PR...

src/gui/guiFormSpecMenu.cpp
Df458: So just to confirm, you're suggesting that we just always use real coordinates since this is a new control?

DS-Minetest: Yes, I think so.

Df458: The result looks weird, but the old non-real coordinates result also looks weird and real coordinates are preferable anyway so I've pushed the change.

src/client/texture_pool.cpp
```diff
@@ -0,0 +1,105 @@
...
+		else
+			return addTexture(base_name);
+	}
+
+	return addTexture(name);
+}
+
+//! Get a handle to the named texture, creating it and setting texture_index if necessary
+video::ITexture *TexturePool::getTexture(const std::string &name,
+	ISimpleTextureSource *tsrc, s32 &texture_idx)
```
DS-Minetest: Could you maybe instead use a pointer for `texture_idx` instead of a reference? That would make it clearer at the function call that it will be overwritten by the function.
The code style guidelines also say:
> Don't pass non-const references to functions. 

https://dev.minetest.net/Code_style_guidelines#Do_not_be_too_C.2B.2By

src/gui/guiAnimatedImage.cpp
Df458: Since an error is still logged with the logs removed altogether and no-one is piping up about this, I'm marking this as resolved.

src/gui/guiFormSpecMenu.cpp
Df458: Actually, I'm just going to be the change I want to see in the world and do it. If anyone disagrees in later reviews, I can always revert the commit.

Df458:
There hasn't been any activity in about a week, but aside from squashing this PR has been all set for review/approval/merge for a while. Any maintainers have the time to look into it this week?

v-rob:
Could support for other file formats like `.jpg` be added by splitting the filename at the last period and then concatenating the file number between the name and the extension?

Df458:
> Could support for other file formats like .jpg be added by splitting the filename at the last period and then concatenating the file number between the name and the extension?

Sounds reasonable. I'll look into that tomorrow. Another possibility would be to have some sort of token representing the number in the filename, but I'm not sure it's worth the effort to do. It could potentially make the API clearer though... any opinions on this?

v-rob:
I don't think having symbols is necessary; it'll just complicate everything.  If we leave the numbers as-is, it will be a simple naming convention rule that everyone will consistently follow.  But having support for multiple filenames is a little more of important functionality, even if most people use PNG anyway.

Df458:
I've added support for supplying an extension and rebased. For simplicity, I'm currently assuming `.png` if the user doesn't supply an extension since I'm fairly certain that anything without an extension will fail to load since the extension seems to matter to the loader (indeed, `.jpeg` appears to fail loading despite support for `.jpg`, I believe the bug is unrelated to this feature though).

ClobberXD:
Tested using the test mod provided in the first post, and also tried different combinations of frame duration and frame count. PR works as expected :+1: 

I thought the duration of jpg images were inconsistent, until I realised that test_1.jpg and test_6.jpg were the same image, lol :)

kilbith:
@SmallJoker [Rebase needed] can be removed.

Other plans I had in the original PR:
- Add `animated_image_button[]`
- Add a `on_hover` spec (bool - default to false), so the image animates when hovered.

SmallJoker:
@kilbith What was the idea of using `texture_name_%i.png`? Sounds are named `sound_name.%i.ogg` and animated frames in the node definition is a single texture, but with all frames stacked along the Y axis.
I'm asking because it could be helpful to adapt one practice from there rather than introducing a new mechanism. The latter (stacked frames) would allow re-using node definition textures and improve the media loading & saving process due to less files, but more data.

kilbith:
@SmallJoker IMO stacked frames fit to a tile because the display's width is guaranteed to be the same. For the GUI or anything else, it's a different story - you may have various textures of very different sizes, and it's not easy to stack them when the gap is too important. Moreover, one could extract a video frame by frame with `ffmpeg`, and play this "video" in a formspec.

EDIT: This PR should adopt the `filename.%i.extension` convention.

Df458:
I've updated the code to use the convention mentioned above, and updated the test mod to do the same. As for the other comments:

- I can appreciate the benefits of a button with an animated image, though at this point it might make sense to add support for buttons with custom content instead (I expect this would be fairly straightforward thanks to #9090). Such a feature would fill the need, and open the door to further possibilities at the same time.
- Having inherited this PR rather than starting it myself, I have no strong opinion re: stacked frames. I'm a little worried about the potential for further prolonging a PR that's been mostly done for quite some time, but I can see the argument for consistency there as well. I'll leave the final decision to the core devs, and make changes if necessary.

Df458:
Since no-one has come in with further opinions and git makes backing out of changes trivial, I've taken the liberty of changing this PR to use stacked frames instead of separate files. Hopefully this will let the conversation continue.

@SmallJoker feel free to take another look and give your take on this.

src/gui/guiAnimatedImage.cpp
```diff
@@ -0,0 +1,85 @@
...
+		m_texture = m_tsrc->getTexture(name);
+	}
+
+	if (m_texture != nullptr) {
+		core::dimension2d<u32> size = m_texture->getOriginalSize();
+		if (size.Height < (u64)m_frame_count) {
+			m_frame_count = size.Height;
+		}
+	} else {
+        // No need to step an animation if we have nothing to draw
```
SmallJoker: Please use tabs.

src/gui/guiAnimatedImage.cpp
```diff
@@ -0,0 +1,85 @@
...
+	if (comma_position > colon_position && comma_position < name.size()) {
+		m_texture = m_tsrc->getTexture(name.substr(0, colon_position));
+
+		m_frame_count = std::max(stoi(name.substr(
+					colon_position + 1, comma_position - colon_position - 1)), 1);
+
+		m_frame_duration = std::max(stoi(name.substr(comma_position + 1)), 1);
+	} else {
+		// Leave the count/duration and display a static image
+		m_texture = m_tsrc->getTexture(name);
```
SmallJoker: I'd output an error here. It's clearly a wrong format for this formspec element.

src/gui/guiAnimatedImage.cpp
```diff
@@ -0,0 +1,85 @@
...
+		size.Height /= m_frame_count;
+
+		draw2DImageFilterScaled( driver, m_texture, AbsoluteRect,
+				core::rect<s32>(core::position2d<s32>(0, size.Height * m_frame_idx), size),
+				NoClip ? nullptr : &AbsoluteClippingRect, colors, true);
+	}
+
+	// Step the animation
+	if (m_frame_count > 1) {
+        // Determine the delta time to step
```
SmallJoker: Tabs again

src/gui/guiAnimatedImage.cpp
```diff
@@ -0,0 +1,85 @@
...
+		u64 step_duration;
```
SmallJoker: Getting rid of `step_duration` 
```C++
u64 new_global_time = porting::getTimeMs();
// Determine the delta time to step
if (m_global_time > 0)
	m_frame_time += new_global_time - m_global_time;
m_global_time = new_global_time;
```

src/gui/guiFormSpecMenu.cpp
```diff
@@ -779,6 +780,48 @@ void GUIFormSpecMenu::parseImage(parserData* data, const std::string &element)
 	errorstream<< "Invalid image element(" << parts.size() << "): '" << element << "'"  << std::endl;
 }
 
+void GUIFormSpecMenu::parseAnimatedImage(parserData *data, const std::string &element)
+{
+	std::vector<std::string> parts = split(element, ';');
+
+	if (parts.size() != 3 &&
+			(parts.size() < 3 || m_formspec_version <= FORMSPEC_API_VERSION)) {
+		errorstream << "Invalid animated image element(" << parts.size() << "): '" << element << "'"  << std::endl;
```
SmallJoker: Line is longer than 80 characters. Please break it.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -779,6 +780,48 @@ void GUIFormSpecMenu::parseImage(parserData* data, const std::string &element)
...
+
+	if (!data->explicit_size)
+		warningstream << "invalid use of animated_image without a size[] element" << std::endl;
+
+	FieldSpec spec(
+			"",
+			L"",
+			L"",
+			258 + m_fields.size()
+			);
```
SmallJoker: This would look better IMO:
```C++
FieldSpec spec(
	"",
	L"",
	L"",
	258 + m_fields.size()
);
```

src/gui/guiFormSpecMenu.h
```diff
@@ -388,6 +389,7 @@ class GUIFormSpecMenu : public GUIModalMenu
 	void parseListRing(parserData* data, const std::string &element);
 	void parseCheckbox(parserData* data, const std::string &element);
 	void parseImage(parserData* data, const std::string &element);
+	void parseAnimatedImage(parserData* data, const std::string &element);
```
SmallJoker: Old code does not always follow the coding style guidelines. Please move the pointer placement towards the variable.

SmallJoker (commented):
The code is slim and works as expected. It needs a few stylistic changes, though.

SmallJoker:
A second core dev opinion would indeed be better than to rely on my comments only. Now that you've changed the PR there are no concept objections I could possibly have.

src/gui/guiFormSpecMenu.cpp
Df458: You're right that the line is too long, and I've gone in and broken it. I think the result is harder to read though, I'd like your opinion on how I did the break and if you see a better spot to do it.

src/gui/guiAnimatedImage.cpp
Df458: Done, I think the message is good but feel free to re-check now!

SmallJoker: Good so far, but the line needs breaking (>80 characters). It also wouldn't hurt to include a hint about the error source, example: `animated_image[]: Invalid texture ...`. I know the element name does not fit into this class but it'll be helpful to modders.

SmallJoker (approved):

src/gui/guiAnimatedImage.cpp
Df458: Looks like you went and fixed it for me, thank you!
Does that mean this conversation can be resolved?

kilbith:
I think it's a very bad idea to force using real coordinates implicitly. I do have some complex formspecs which are hard to transit to real coordinates, so they keep using the old system, but I still want to pick up animated images for them.

And it was the same mistake for `hypertext`.

src/gui/guiAnimatedImage.cpp
SmallJoker: Yes. If requested changes were addressed feel free to mark my comments as resolved.

src/gui/guiFormSpecMenu.cpp
SmallJoker: For two lines it's fine to break it like this. If it were longer, passing the template to `snprintf` would be a good solution.
Optimally binary operators would be placed to the end of the previous line. I think both is fine, so I'm not going to enforce that.

Df458:
There was support previously, but it @DS-Minetest suggested that I remove it. I could potentially reinstate support, but I'd rather stick to the current policy...

@SmallJoker, I believe you were involved in the approval process for the hypertext PR. What should be done with non-`real_coordinates` mode on new elements? Same as there, or was that an oversight?

SmallJoker:
> or was that an oversight?

That, and here the same thing happened. Copying the scaling code from the `image[]` element should be possible and provide "good enough" results. Prior `real_coordinates` the scaling and positioning is broken anyway, so I wouldn't spend too much effort into making new elements backwards compatible.

kilbith:
So I have to spend hours to convert my formspec to real coordinates in order to use `animated_image` or `hypertext`? How is that hurting to add a one-liner to support the old coordinates system? And how is that clear to someone that these elements works only correctly with real coordinates, since it's implicit?

Are you stupid or what?

Df458:
> Copying the scaling code from the image[] element should be possible and provide "good enough" results.

That's more or less what was there originally. Since we've confirmed that the original change wasn't deliberate, interest has been shown, and the cost is negligible, I've re-instated the feature. @SmallJoker  let me know if that's a deal-breaker and I can back out, but since it's so cheap to support I don't really mind keeping it.

Slightly off-topic: @kilbith, I understand that getting things right is important to you and I'm legitimately glad that you're advocating for it... but I feel that your comments sometime cross the line from spirited debate into creating unnecessary drama. It doesn't bother me personally, but I get the feeling that this sort of thing will be a source of issues sooner or later.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -779,6 +780,58 @@ void GUIFormSpecMenu::parseImage(parserData* data, const std::string &element)
...
+    v2s32 geom;
+
+    if (data->real_coordinates) {
+        pos = getRealCoordinateBasePos(v_pos);
+        geom = getRealCoordinateGeometry(v_geom);
+    } else {
+        pos = getElementBasePos(&v_pos);
+        geom.X = stof(v_geom[0]) * (float)imgsize.X;
+        geom.Y = stof(v_geom[1]) * (float)imgsize.Y;
+    }
```
SmallJoker: Please use tabulators.

Df458: Sorry for screwing that up a second time...I've taken measures to ensure that it won't happen again.

SmallJoker:
:+1: PR looks good again.

src/gui/guiAnimatedImage.cpp
```diff
@@ -0,0 +1,83 @@
...
+{
+	// Expected format: "texture_name:frame_count,frame_duration"
+	// If this format is not met, the string will be loaded as a normal texture
+
+	m_frame_idx = 0;
+
+	std::string::size_type colon_position = name.find(':', 0);
+	std::string::size_type comma_position = name.find(',', 0);
+
+	if (comma_position > colon_position && comma_position < name.size()) {
```
rubenwardy: Why isn't colon_position checked?

```suggestion
	if (comma_position != std::string::npos && colon_position != std::string::npos &&
			comma_position > colon_position) {
```

src/gui/guiAnimatedImage.cpp
```diff
@@ -0,0 +1,83 @@
...
+		s32 id, const core::rect<s32> &rectangle, const std::string &name,
+		ISimpleTextureSource *tsrc) :
+		gui::IGUIElement(gui::EGUIET_ELEMENT, env, parent, id, rectangle),
+		m_name(name), m_tsrc(tsrc), m_texture(nullptr), m_global_time(0),
+		m_frame_idx(0), m_frame_count(1), m_frame_duration(1), m_frame_time(0)
+{
+	// Expected format: "texture_name:frame_count,frame_duration"
+	// If this format is not met, the string will be loaded as a normal texture
+
+	m_frame_idx = 0;
```
rubenwardy: this can be initialised in the definition

src/gui/guiAnimatedImage.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include "log.h"
+#include "porting.h"
+#include <string>
+
+GUIAnimatedImage::GUIAnimatedImage(gui::IGUIEnvironment *env, gui::IGUIElement *parent,
+		s32 id, const core::rect<s32> &rectangle, const std::string &name,
+		ISimpleTextureSource *tsrc) :
+		gui::IGUIElement(gui::EGUIET_ELEMENT, env, parent, id, rectangle),
+		m_name(name), m_tsrc(tsrc), m_texture(nullptr), m_global_time(0),
+		m_frame_idx(0), m_frame_count(1), m_frame_duration(1), m_frame_time(0)
```
rubenwardy: duplicated initialisation

src/gui/guiFormSpecMenu.cpp
```diff
@@ -779,6 +780,58 @@ void GUIFormSpecMenu::parseImage(parserData* data, const std::string &element)
 	errorstream<< "Invalid image element(" << parts.size() << "): '" << element << "'"  << std::endl;
 }
 
+void GUIFormSpecMenu::parseAnimatedImage(parserData *data, const std::string &element)
+{
+	std::vector<std::string> parts = split(element, ';');
+
+	if (parts.size() != 3 &&
+			(parts.size() < 3 || m_formspec_version <= FORMSPEC_API_VERSION)) {
```
rubenwardy: Please don't De Morgan this - I would prefer the more obvious condition and the one that is throughout this fileused 

```suggestion
			!(parts.size() > 3 && m_formspec_version > FORMSPEC_API_VERSION)) {
```

SmallJoker: The current logic also shows that `parts.size() != 3 &&` is redundant.

Df458: Fine by me, that's more or less what I originally had when I submitted the PR.

src/gui/guiAnimatedImage.cpp
Df458: It is...just in a dangerously indirect fashion. Good catch.

src/gui/guiFormSpecMenu.cpp
rubenwardy: Were you asked to do this by someone?

Df458: Sort of? It's in one of the reviews by @DS-Minetest. The main aspect of the review was to reduce nesting though, so this particular mistake is still on me.

rubenwardy (approved):

PR #6839, opened 2017-12-26
[WIP] Dynamic Shadows

kilbith:
Dynamic Shadows as seen in this [video](https://www.youtube.com/watch?v=PBMOG_dTbWo), using native OpenGL.

Code is a mess, but the fundamentals of shadow mapping are there.

I still need to send an additional matrix (`mShadowMatrix`) to the shaders. Everything is OK on the shaders side but on the C++ side it just lacks the ability to update this uniform variable.

A function `SetShadowMatrix()` needs to be added to the shaders callback (which already calculate `mWorldViewProj`) or something similar but the most important is where to call this function, I don't know how yet. Maybe @numberZero can help out.

PR #5687, opened 2017-05-01
Artificial bank

juhdanad:
**WARNING: if you open your world with this version, it gets converted and you won't be able to open it with other versions anymore!**

The goal is this PR to change the light banks' meaning to sunlight + artificial light.

The request is still WIP, because the conversion sometimes causes dark blocks, and I don't know yet why.

What will change:
- the map format increases
- old server - new server: when first loaded, the world is automatically converted to the new format, erasing the old light information. This might slow down the map loading, but once the conversion is done, no more processing is required.
- old client - new server: the blocks get converted back when they are sent to the clients. Old clients should see no change in the lighting. Currently all `param1` values are converted, even if they contain no light information. I hope this won't cause any side effects.
- new client - old server: the client has to guess the amount of sunlight. The current implementation works well with artificial sources if they are either in direct sunlight or in darkness. But there is an ugly jump in brightness if the artificial light and sunlight meet each other, while the artificial source is in darkness. Again, currently all `param1` values are converted, even if they contain no light information. I hope this won't cause any side effects.
light_edge
- `sunlight_source` for nodes was introduced to make light calculations uniform. This property is not exposed to Lua yet.
- graphics: now artificial light sources are brighter if they are in sunlight. Since the light sources produce maximal brightness without sunlight, in sunlight their brightness is clamped.
light_clamp

I hope I did not forget anything. Please let me know what you think, how should this pull change!

PR #9272, opened 2020-01-01
Time-of-day-dependent face shading

ghost:
This PR is a continuation of #7987 which seeks to correct the remaining major issues in the feature so that it can be merged. In addition to rebasing on the current HEAD, this PR has a couple of additional changes:

1. The invisible entity bug has been completely fixed (although there are still one or two issues in the way mesh entities are lit, they're fairly minor by comparison)
2. Various changes to the lighting calculations and variables used, with the goal of fixing the overly-dark nights and improving the overall look of the lighting. (Whether or not the latter is true is up for debate)

Like with the previous PR, this feature is still considered experimental. I think it's acceptable to label it as such initially and merge with some minor roughness remaining, but the goal is for the major functionality to be correct.

Since it affects the rendering (mainly when enabled, but it also introduces a shader used by entities), this PR will need quite a bit of testing. I'd really like to see people try it out and give feedback on what feels good and what doesn't, so that we can continue iterating on the shader code to produce something good. In particular, I'd like for folks to try testing with various drawtypes, entities, mods, games, mapgens etc. so that we can get holistic view of what effect this feature has on Minetest's visuals. From my testing, I get the impression that the feature has a lot of promise but hasn't yet reached its full potential.

## How to Test
In the main settings tab, check 'Shaders' and 'Directional Shading'. After that, the feature should be active ingame.

## Known Issues
- Mesh entities are lit with seemingly-incorrect normals
- Nights, while brighter than the last PR, are still too dark
- `override_day_night_ratio` is not respected

These are the last major issues that I know of. This issue will remain a draft until these issues are resolved.


benrob0329:
Haven't spotted any issues in testing other than the wierd entity lighting, which does not seem to apply to every model and is already a known issue.

I'd like to see this respect `override_day_night_ratio` but that is firmly in the feature-request category, and not the bug category.

NathanSalapat:
I did some testing, seems to work as expected.
screenshot_20191231_183958
Some faces that probably shouldn't have highlights get them, but due to the way the lighting system works, there is probably nothing that can be done about that.  The lower sand nodes would be in the shade of the dirt and stone, but I still think it works as expected.

LoneWolfHT:
Tested and it's much better than the original PR. :+1:

kilbith:
https://github.com/minetest/minetest/pull/6839 has more potential. I hardly call that **Directional** Shading...

Df458:
#6839 certainly has more visual potential. *However*, I think most of the proponents of that method are missing the gameplay implications of such a change. Minetest allows games to use light as part of their mechanics, and true dynamic shadows masks the 'in-game' lighting from the player.

Honestly, I support eventually merging both this PR (or I wouldn't have made it) and that feature, I believe this is a good compromise that adds some visual niceness while being usable in all gameplay contexts. For contexts that don't require that information to be known, your feature would be great as well.

I think it's also worth mentioning that some of the groundwork being laid here would be useful for something like #6839 as well, by exposing entities to shaders and fixing their normals.

hecktest:
Respecting override_day_night_ratio is not a feature request, it's a must have. But it should just be an intensity override, not affecting the angle. (DNR's most legitimate use is for dimming the sky in overcast weather, and for finetuning the night's darkness)

Df458:
> Respecting override_day_night_ratio is not a feature request, it's a must have.

I'm inclined to agree. Thanks for mentioning that by the way, adding that to the issues list slipped my mind and I might've missed fixing that otherwise.

benrob0329:
> But it should just be an intensity override, not affecting the angle

DNR changes the sun/moon angle, so changing the angle seems necessary.

dcbrwn:
@Df458 I would suggest extracting code for CAO shading into a separate PR. It will get merged sooner. Even without DNR magic, simple shader with one dot product makes **HUGE** difference to minetest. 

scr

paramat:
Here is something that i think needs addressing https://github.com/minetest/minetest/pull/7987#issuecomment-447710228
I think that the effect should rapidly fall off as light level reduces from 15 (open to sky), such that only a few of the highest light levels have this effect.

Df458:
@dcbrwn I've been planning. to do this. I have a few other PRs that I'm trying to focus on right now, but when I have the time to return to this that's exactly what's going to happen.

With that said, it will still probably take a hot minute since there are a few remaining issues with the CAO lighting.

benrob0329:
@paramat Small tweaks like that can be experimented with later. Please don't force this PR through another batch of WIP-ism because of a slight change you'd like to be made that might end up looking worse.

dcbrwn:
@paramat while I completely agree on that for terrain lighting, practice shows that for meshes it's not that critical. Especially without DNR, because with it players will be able to tell the time just by looking at mobs, being deeply underground XD

Room from screenshot is at ~5 level, completely closed, no holes, no windows nor doors, with 10 block thick walls.

minecraft

paramat:
benrob0329 that is not a small tweak or slight change for a later PR, it is fundamental to the effect and an issue that results in very wrong looking lighting in many situations.
It is ok for me or anyone to make suggestions for improvement, that does not deserve negative emoticons. I did not insist on a change or state i would officially disapprove over this.
I suggest you now criticise everyone who pointed out issues and made suggestions in the original PR, therefore delaying it =)
Attending to this could possibly be done after a merge, but i cannot see why, as it is a very simple change that could be experimented with now without delaying the PR.

LoneWolfHT:
> It is ok for me or anyone to make suggestions for improvement

It's more than a suggestion if you add the WIP label because of it
image

dcbrwn:
Guys, chill! My point was that this PR mixes up two features and therefore takes longer to review, discuss, agree upon and merge. No need to blame each other! As to outdoor ratio thing @paramat is completely right. Releasing DNR without taking this into account is more harmful, that not releasing it at all. @Df458 still did a great job by adding shaders for models. I occasionally return to MT for about 5 years and I think that this PR is a big milestone!

Df458:
To follow that up: @LoneWolfHT, check the top of the page. This PR is a draft, it is *by definition* still WIP.

As for @paramat's feedback, my current PC is recording/stream-capable so I'll probably be testing several tweaks and configurations with a few 'preset' scenes and share the results for comparison. But that will come after the major known bugs are fixed.

lhofhansl:
For the record I'm a big fan of getting this in.

As for CAO shading... I had experimented with this in the past and found that while the code is easy, many (most in fact) model did not have correct normals, and so the shading would be incorrect. Enough for me to give up on it - since I obviously cannot fix every model on the planet.
Glad to do see that someone picket it up again. I wonder whether the lack of normals is still an issue?

Generally I would also agree splitting this into two. Not needed obviously, just to keep each change's complexity lower.

Lastly, let's get this in. We can spend all day discussion nuances and losing the bug picture over this. :)

What's missing? How can I help?

Df458:
> What's missing? How can I help?

For the time being, dev time is missing (because I'm busy on some other PRs), which is why this hasn't budged in a while. @dcbrwn offered to help as well the other day, but the more the merrier! Step one is simply splitting the shading changes to put them in a separate PR, and getting that in. I'm not too concerned with models having broken normals, because many models already have blatantly broken UVs as-is and few players seem to comment on it. 

On the side of this graphical feature, it's mostly a matter of getting DNR support, clean-up, and a general pass of testing/tweaking to get better visual results. Once that's done, we can un-WIP this and start the review process proper.

dcbrwn:
@lhofhansl outdoor ratio calculation is missing. It is taken into account in shaders, but calculated incorrectly. My guess is that we need to update mesh generation to include this ratio as a vertex attribute, if IrrLicht allows that

lhofhansl:
The sunlight ratio is already available in the shader, it is indeed passed in the alpha channel. I used for my perlin-noise fog PR: #8889

dcbrwn:
Yes, but its derived from day time and light intesity, not from geometry as it should. Or I'm missing something.

lhofhansl:
It's the actual amount sunlight as calculated by the light spread when the map is generated or changed. Should be fairly accurate.

Df458:
Should probably have closed this sooner.

Rationale:
- With the current state that light encoding is in, some of the existing bugs *cannot* be fixed. There's someone working on a PR that will address this issue, but for the time being we're not going to get a complete solution.
- A fair amount of the affected code has changed in the interim, rather than continuing to rebase over time I'd rather recreate the feature if/when the necessary fixes are in place.
- Since this feature's going dormant for a while, it's better to close the PR and open a fresh one later to help keep the list of open PRs relevant.

lhofhansl:
Just came back to this.

:( :(

This is the 3rd (or 4th?) attempt at doing this. And every time we give up.
Is MT fundamentally broken? Are we resistant to change? Are we looking for perfection?
Or why is that? (I mean these as actual question, not to start a flame war)

IMHO MT is showing its age. The scenes looks stale compared to other, similar games.

Df458:
Re-appearing for a brief moment to answer:

> Is MT fundamentally broken?

Fundamentally is a strong word. However, Minetest's lighting _is_ broken in the sense that the calculations discard important information that is **necessary** in order to implement this feature. The fix is not trivial, and there's a non-zero chance that it could break existing worlds.

Since I'm no longer a part of the project I don't really feel like digging too deep to explain the findings that lead me to close this PR, so instead I'll just refer you to #5155 (And the relevant failed PR in #5687). If that issue is resolved, this feature will be feasible. Someone on the unofficial Discord (can't remember who) was tackling this issue in the process of implementing colored lighting. I'm not planning to comment on this further, but maybe they'll chime in here?

GreenXenith:
I was the one attempting to fix the lighting system and add colored lighting.

The Minetest lighting system uses "light banks" to store data about light for each node. These banks store 8 bits of data about "daytime light" and "nighttime light" (4 bits each), and then calculates artificial lighting somehow based on those two values. It is both named and stored horribly.

I reworked the lighting system to use 4 bits for "natural" or "sky" light, and 12 bits for artificial colored light (4 bit channels). That was as far as I got before I iced the endeavor to work on more fruitful projects. The lighting worked, I just never got around to implementing color display, color propagation, or map handling.

In theory, colored lighting can be implemented without any breaking previous maps or clients. This can be achieved by converting map values during mapblock send based on the mapblock version, as well as converting mapblocks as they are loaded. It would also be pretty simple to implement light filtering (e.g. colored glass) if done properly.

It is not a trivial task to implement colored lighting, but fixing the light storage itself may be a good first step and make a good PR.  If anyone wishes to try their hand at implementing proper light storage with map compatibility, you can take a look at [what I had done](https://github.com/GreenXenith/minetest/commit/cfa5af94307dbabd0a7152aaf1c0a7fab6523911) to implement reasonable light storage.

lhofhansl:
@Df458 @GreenXenith at the risk of going out on a tangent... Why did you leave MT? Was frustration of not getting in, or other reasons?

GreenXenith:
Oh, I haven't left at all. I just dropped that specific Minetest project to work on others.

Pevernow:
@Df458 Please don't give up.

Pevernow:
Many of minetest's innovative proposals have been rejected on the grounds of poor or imperfect compatibility. As a result, even if it is written in C + +, its efficiency is not different from that of minecraft.



Mojang of minecraft and its community workers are actively pushing for new changes. With one idea after another, it gradually becomes more and more perfect. However, minetest has gradually lagged behind the times in vetoing again and again.



We need to change, the whole core developer and the reviewer of pull request need to change.



For example, the pull request has undergone enough revisions. I think it should be merged, even if there are some imperfect small defects. As long as it doesn't affect the game, most of the time it presents a correct and beautiful picture.

rubenwardy:
Lol, I personally prefer software to have some stability rather than be completely broken.

Bug fixes should always have priority over new features. Sorry if you don't like to hear this, but that's how software should be done. Features create bugs, and if you don't fix them they will accumulate into unstable software

We have a core developer actively working on a new GUI system

Pevernow:
This pull requests gives me an idea. Considering that no core developer is good at writing advanced shaders. Maybe some APIs can be provided to make it possible for players to customize shaders.

Minecraft's optifine built-in shader is not gorgeous, but it provides an interface to custom shaders. Seus, Kuda and other shader packages made by the community use these interfaces to provide gorgeous pictures for minecraft.

@rubenwardy  What do you think of the idea?

Pevernow:
> Lol, I personally prefer software to have some stability rather than be completely broken.
> 
> Bug fixes should always have priority over new features. Sorry if you don't like to hear this, but that's how software should be done. Features create bugs, and if you don't fix them they will accumulate into unstable software
> 
> We have a core developer actively working on a new GUI system. We don't have any core developers with the graphics programming ability to work on gorgeous shaders, but we do have attempts to replace the dinosaur rendering library Irrlicht - by switching input handling to SDL.
> 
> The shadow mapping patches are much more promising than this PR, imo

Do you have a concept map for GUI? We need a concept map so that people other than core developers can easily participate in GUI improvements. Reduce the conflict of GUI style selection between developers.

Maybe we can invite some people who write minecraft advanced shaders to participate in minetest's shader writing?

dcbrwn:
@Pevernow the problem with graphics currently is not the shaders, but the rendering engine itself. To refactor it, it will not only take expertise in CG, but also good knowledge of MT internals and C++.

PR #9284, opened 2020-01-04
Settings: Add get_flags API for mapgen flags (mg_flags, mgv6_spflags, ...)

SmallJoker:
Replaces #9279 by implementing a proper API to setting flags.
In order to get the flags to work, their default value has to be specified using an additional `FlagDesc` struct array. The mapgen default flags are now defined in `mapgen.cpp`.

**Implementation goals:**

1) Unified flags handling in C++ and Lua Settings API
    * Reading only, for now. Writing can be implemented later, if needed.
2) API function to read the currently active flags
    * Currently impossible from Lua

## To do

This PR is Ready for Review.

## How to test

1) Run unittests. The Settings tests may not fail.
2) Code for mainmenu or a server-side mod:
```Lua
core.settings:set("mg_flags", "nocaves,nodecorations")
local t

t = core.settings:get_flags("mg_flags")
print(dump(t))
assert(t.caves == false);
assert(t.dungeons == true);
assert(t.light == true);
assert(t.decorations == false);
assert(t.biomes == true);

core.settings:set("mg_flags", "nodecorations,nobiomes")
t = core.settings:get_flags("mg_flags")
print(dump(t))
assert(t.caves == true);
assert(t.dungeons == true);
assert(t.light == true);
assert(t.decorations == false);
assert(t.biomes == false);

core.settings:remove("mg_flags")
```

@Wuzzy2 and @pauloue might be interested.

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -6040,6 +6040,9 @@ It can be created via `Settings(filename)`.
     * `default` is the value returned if `key` is not found.
     * Returns `nil` if `key` is not found and `default` not specified.
 * `get_np_group(key)`: returns a NoiseParams table
+* `get_flags(key)`:
+    * Returns `{flag = true/false, ...}` according to the set flags.
+    * Is currently limited to mapgen settings like `mgv5_spflags`.
 * `set(key, value)`
     * Setting names can't contain whitespace or any of `="{}#`.
     * Setting values can't contain the sequence `\n"""`.
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -6042,7 +6042,8 @@ It can be created via `Settings(filename)`.
 * `get_np_group(key)`: returns a NoiseParams table
 * `get_flags(key)`:
     * Returns `{flag = true/false, ...}` according to the set flags.
-    * Is currently limited to mapgen settings like `mgv5_spflags`.
+    * Is currently limited to mapgen flags `mg_flags` and mapgen-specific
+      flags like `mgv5_spflags`.
 * `set(key, value)`
     * Setting names can't contain whitespace or any of `="{}#`.
     * Setting values can't contain the sequence `\n"""`.
--- a/src/mapgen/mapgen.cpp
+++ b/src/mapgen/mapgen.cpp
@@ -215,6 +215,25 @@ void Mapgen::getMapgenNames(std::vector<const char *> *mgnames, bool include_hid
 	}
 }
 
+void Mapgen::setDefaultSettings(Settings *settings)
+{
+	settings->setDefault("mg_flags", flagdesc_mapgen,
+		 "caves,dungeons,light,decorations,biomes");
+	settings->setDefault("mgv5_spflags", flagdesc_mapgen_v5,
+		"caverns");
+	settings->setDefault("mgv6_spflags", flagdesc_mapgen_v6,
+		"jungles,biomeblend,mudflow,snowbiomes,noflat,trees");
+	settings->setDefault("mgv7_spflags", flagdesc_mapgen_v7,
+		"mountains,ridges,nofloatlands,caverns");
+	settings->setDefault("mgcarpathian_spflags", flagdesc_mapgen_carpathian,
+		"caverns,norivers");
+	settings->setDefault("mgflat_spflags", flagdesc_mapgen_flat,
+		"nolakes,nohills");
+	settings->setDefault("mgfractal_spflags", flagdesc_mapgen_fractal,
+		"terrain");
+	settings->setDefault("mgvalleys_spflags", flagdesc_mapgen_valleys,
+		"altitude_chill,humid_rivers,vary_river_depth,altitude_dry");
+}
 
 u32 Mapgen::getBlockSeed(v3s16 p, s32 seed)
 {
--- a/src/mapgen/mapgen.cpp
+++ b/src/mapgen/mapgen.cpp
@@ -217,10 +217,6 @@ void Mapgen::getMapgenNames(std::vector<const char *> *mgnames, bool include_hid
 
 void Mapgen::setDefaultSettings(Settings *settings)
 {
-	// Option 1: List all defaults here as an overview
-	// Option 2: Each mapgen contains its own default setter function
-	// This is option 2.
-
 	settings->setDefault("mg_flags", flagdesc_mapgen,
 		 MG_CAVES | MG_DUNGEONS | MG_LIGHT | MG_DECORATIONS | MG_BIOMES);
 
--- a/src/mapgen/mapgen.cpp
+++ b/src/mapgen/mapgen.cpp
@@ -217,22 +217,18 @@ void Mapgen::getMapgenNames(std::vector<const char *> *mgnames, bool include_hid
 
 void Mapgen::setDefaultSettings(Settings *settings)
 {
+	// Option 1: List all defaults here as an overview
+	// Option 2: Each mapgen contains its own default setter function
+	// This is option 2.
+
 	settings->setDefault("mg_flags", flagdesc_mapgen,
-		 "caves,dungeons,light,decorations,biomes");
-	settings->setDefault("mgv5_spflags", flagdesc_mapgen_v5,
-		"caverns");
-	settings->setDefault("mgv6_spflags", flagdesc_mapgen_v6,
-		"jungles,biomeblend,mudflow,snowbiomes,noflat,trees");
-	settings->setDefault("mgv7_spflags", flagdesc_mapgen_v7,
-		"mountains,ridges,nofloatlands,caverns");
-	settings->setDefault("mgcarpathian_spflags", flagdesc_mapgen_carpathian,
-		"caverns,norivers");
-	settings->setDefault("mgflat_spflags", flagdesc_mapgen_flat,
-		"nolakes,nohills");
-	settings->setDefault("mgfractal_spflags", flagdesc_mapgen_fractal,
-		"terrain");
-	settings->setDefault("mgvalleys_spflags", flagdesc_mapgen_valleys,
-		"altitude_chill,humid_rivers,vary_river_depth,altitude_dry");
+		 MG_CAVES | MG_DUNGEONS | MG_LIGHT | MG_DECORATIONS | MG_BIOMES);
+
+	for (int i = 0; i < (int)MAPGEN_INVALID; ++i) {
+		MapgenParams *params = createMapgenParams((MapgenType)i);
+		params->setDefaultSettings(settings);
+		delete params;
+	}
 }
 
 u32 Mapgen::getBlockSeed(v3s16 p, s32 seed)
@@ -1087,7 +1083,7 @@ void MapgenParams::writeParams(Settings *settings) const
 	settings->setS16("water_level", water_level);
 	settings->setS16("mapgen_limit", mapgen_limit);
 	settings->setS16("chunksize", chunksize);
-	settings->setFlagStr("mg_flags", flags, flagdesc_mapgen, U32_MAX);
+	settings->setFlagStr("mg_flags", flags, flagdesc_mapgen);
 
 	if (bparams)
 		bparams->writeParams(settings);
--- a/src/mapgen/mapgen_carpathian.cpp
+++ b/src/mapgen/mapgen_carpathian.cpp
@@ -187,7 +187,7 @@ void MapgenCarpathianParams::readParams(const Settings *settings)
 
 void MapgenCarpathianParams::writeParams(Settings *settings) const
 {
-	settings->setFlagStr("mgcarpathian_spflags", spflags, flagdesc_mapgen_carpathian, U32_MAX);
+	settings->setFlagStr("mgcarpathian_spflags", spflags, flagdesc_mapgen_carpathian);
 
 	settings->setFloat("mgcarpathian_base_level",   base_level);
 	settings->setFloat("mgcarpathian_river_width",  river_width);
@@ -226,6 +226,12 @@ void MapgenCarpathianParams::writeParams(Settings *settings) const
 	settings->setNoiseParams("mgcarpathian_np_dungeons",      np_dungeons);
 }
 
+
+void MapgenCarpathianParams::setDefaultSettings(Settings *settings)
+{
+	settings->setDefault("mgcarpathian_spflags", flagdesc_mapgen_carpathian,
+		MGCARPATHIAN_CAVERNS);
+}
 
 ////////////////////////////////////////////////////////////////////////////////
 
--- a/src/defaultsettings.cpp
+++ b/src/defaultsettings.cpp
@@ -427,6 +427,13 @@ void set_default_settings(Settings *settings)
 	settings->setDefault("mapgen_limit", "31000");
 	settings->setDefault("chunksize", "5");
 	settings->setDefault("mg_flags", "caves,dungeons,light,decorations,biomes");
+	settings->setDefault("mgv5_spflags", "caverns");
+	settings->setDefault("mgv6_spflags", "jungles,biomeblend,mudflow,snowbiomes,noflat,trees");
+	settings->setDefault("mgv7_spflags", "mountains,ridges,nofloatlands,caverns");
+	settings->setDefault("mgcarpathian_spflags", "caverns,norivers");
+	settings->setDefault("mgflat_spflags", "nolakes,nohills");
+	settings->setDefault("mgfractal_spflags", "terrain");
+	settings->setDefault("mgvalleys_spflags", "altitude_chill,humid_rivers,vary_river_depth,altitude_dry");
 	settings->setDefault("fixed_map_seed", "");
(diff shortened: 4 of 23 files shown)


Wuzzy2:
>     * Is currently limited to mapgen settings like `mgv5_spflags`.

This text is a little unclear to me.

Otherwise this looks OK.

SmallJoker:
@Wuzzy2 There are many kinds of flags, but I couldn't find better words to explain what exact kind of flags that are meant. I'd be glad to hear another suggestion.

paramat:
Please can you add explanation to the first post of why this is needed? (i know it probably is, i am not disputing that).
Currently all this says is that it replaces #9279 , but that PR also does not state why it is needed. To find out it is necessary to read through quite technical stuff buried in #9254 , which #9279 does not link to.
I have read through all the PRs but even i am still a little unclear on why this is needed and what it does.
Links to relevant comments in #9254 would be useful.

paramat:
Also, please see https://github.com/minetest/minetest/pull/9279#issuecomment-570828275

SmallJoker:
@paramat Added the goals of this PR to the topic. I hope these are clear enough.

> Actually, only stating one flag is known to cause problems, flags settings added to conf must always specify all flags, either with prefix 'no' or not, such as:

Not according to the code (neither in master nor this PR). `nodecorations` will be applied "on top" of the mapgen default `spflags` which has the effect of turning off decorations - and only decorations. All other flags still have their default value.
Feel free to do some local tests using partially set flags (new worlds, for example).

paramat:
Odd, i cannot see any explanation =) Forgot to 'submit'?

> Not according to the code

Weird, will test. I might be remembering incorrectly.
I seem to remember that any flag not specified when adding a line to .conf is automatically set to disabled.

///////////

>  The mapgen default flags are now defined in mapgen.cpp.

```
	// TODO: Remove the default "spflags" values from the mapgens
	// They are now overwritten by Settings::getFlagStr
	//u32 spflags = 0;
```
I support this feature, and it is part of the request in #6711 . However, it seems obvious that moving the definition of the mapgen-specific flag defaults out of the mapgen files is a horrible and unacceptable change. Also bad, they are defined in user-facing form instead of something like `MG_CAVES | MG_LIGHT | MG_DECORATIONS | MG_BIOMES`.
This will also make mapgen development more difficult, anything that is specific to a mapgen that is not in the specific mapgen files is a pain (from experience).
All mapgen default parameters are in the specific mapgen files so that is where the mapgen-specific flags should be too.
There must be a way to do this that keeps the spflags defined in the mapgen files.

Now see #6711 and specifically https://github.com/minetest/minetest/issues/6711#issuecomment-348046665
The problem occurs for other parameters too, not just mapgen flags, but we cannot do similar to all mapgen parameters. I think this PR is best changed to a more general feature that addresses all the 'ungettable' parameters and all of #6711
Maybe, as the issue suggests, we can extract the default settings from settingtypes?

So, i disapprove of this PR in its current form.
We need to take time and do this properly, and not rush it just for Wuzzy's PR (a workaround can be used in the meantime, hardcoding the defaults).

SmallJoker:
> Odd, i cannot see any explanation =) Forgot to 'submit'?

First post, after `Implementation goals:`.

> All mapgen default parameters are in the specific mapgen files so that is where the mapgen-specific flags should be too.

I get your point. The reason for a central `mapgen.cpp` is the PR/code size. I can add a static function to each `Mapgen*Params` class if you really want to have the defaults in the same file, on cost of at least double the lines of what it currently is.

> I think this PR is best changed to a more general feature that addresses all the 'ungettable' parameters and all of #6711

No, you missed the point. This PR does not offer an API to get the default flags. It is the complement to `get()` (string values). Like there, the default value is only used as some sort of a fallback.

> Maybe, as the issue suggests, we can extract the default settings from settingtypes?

Could be done, but is definitely out of scope for this PR. It is about providing an API function to read mapgen flags from minetest.conf.

paramat:
> First post, after Implementation goals:.

Oh ok, sorry, i was expecting something more descriptive.
> I can add a static function to each Mapgen*Params class if you really want to have the defaults in the same file, on cost of at least double the lines of what it currently is.

Yes please. But also, mapgen flags should still be defined as they are now, in the form `MG_CAVES | MG_LIGHT | MG_DECORATIONS | MG_BIOMES`.
> This PR does not offer an API to get the default flags

Sorry, i got that wrong.
However, it does seem related to my code shown in https://github.com/minetest/minetest/issues/6711#issuecomment-348046665 That code results because i was unable to get the currently active setting for floatland level.
> Could be done, but is definitely out of scope for this PR.

I agree, nevermind that =)

paramat:
What about:
https://github.com/minetest/minetest/blob/7c0e0c34ecb66e044f59cbab03a4c12fc7838e97/doc/lua_api.txt#L4532-L4539
Maybe we have this feature already?

paramat:
Even if not, if these existing APIs can get the active settings without disrupting mapgen files i am wondering why your approach needs to disrupt them?

paramat:
In MT master:
I tried adding to .conf:
mg_flags = light
Result is a world with:
mg_flags = caves, nodungeons, light, decorations, biomes
Somehow dungeons have been disabled.

Doing similar for mapgen-specific flags seems to work properly though.

paramat:
https://github.com/minetest/minetest/blob/7c0e0c34ecb66e044f59cbab03a4c12fc7838e97/src/mapgen/mapgen.h#L118-L127

Seems to occur because dungeons are not set as the default in mapgen.h.
I remember trying to add them here but hmmmm warned this would then enable dungeons in worlds that had them deliberately disabled.

So, mg_flags must be stated on full in .conf, this is harmless and good practice anyway, the 'no'prefix is the way to disable them, not leaving them out.
This is why we removed the following lines from settingtypes.txt, which were incorrect:
```
#    Flags that are not specified in the flag string are not modified from the default.
#    Flags starting with 'no' are used to explicitly disable them.
```
The 'default' in mapgen.h is dungeons disabled, btu the 'default' in defaultsettings.cpp is dungoens enabled. Again this refers to my discussion with hmmmm, we decided to make dungeons default enabled, but this could only be done in defaultsettings.cpp not in mapgen.h.

paramat:
Just thought of something ...
Why is this feature needed at all when we already have a way to get the active flags?
'All Settings' does this already:
With no mapgen flags set, defaults are correctly detected:



With `mg_flags = nocaves,dungeons,light,nodecorations,biomes` in .conf, also correct:

Screenshot from 2020-01-05 23-18-06

So, it seems that Wuzzy's PR can use whatever method 'All Settings' uses to get the active mapgen flags, which i assume includes reading settingtypes.txt, which is exactly what settingtypes.txt exists for, to support configuration in the menu.

paramat:
I am guessing the need for this PR is to have a high performance way to get active mapgen flags, without having to parse settingtypes.txt?
```
1. Unified flags handling in C++ and Lua Settings API
Reading only, for now. Writing can be implemented later, if needed.
```
Why is this needed? Is it to solve the problem stated below? Any other reasons?:
```
2. API function to read the currently active flags
Currently impossible from Lua
```

Anyway, 5.2.0 is approaching and i feel this PR needs time and consideration, it may need to wait until after 5.2.0. It certainly should not be rushed for Wuzzy's PR.
I think Wuzzy's PR should use a workaround for now (there are 2 workaorunds: hardcoding defaults or reading settingtypes.txt) if it is merged for 5.2.0.

paramat:
Apparently `minetest.get_mapgen_setting()` does not work in the main menu, however, putting aside use in the main menu, does this existing API cover what this PR is trying to do?

pauloue:
The 'All Settings' checkboxes are not correct if the setting is in `minetest.conf`, but some enabled by default flags are not specified. I explained in https://github.com/minetest/minetest/pull/7847#issuecomment-526795242.

This PR is supposed to make it possible to accurately get the current value of mapgen flags from the main menu. Also `get_flags` returns a table instead of a comma-separated string.

SmallJoker:
Reading flags currently sucks. I want an API that works for the `minetest.conf` settings - no matter whether used in main menu or in-game (could be extended for `map_meta.txt` later if needed).

Since the addition of `minetest.settings`, `minetest.get_mapgen_setting()` became a bad design idea. I should probably start deprecating those in favour of `minetest.mapgen_settings` (Settings reference).

**What would you like me to do in this PR?**
 * Move the flag registration to each mapgen file
 * Set flag defaults using `u32` values

**Would you like to see something of this?**
 * New Settings object for in-game named `minetest.mapgen_settings`
 * Deprecate `minetset.*et_mapgen_setting*()`
 * Give up this PR because it cannot be implemented the way you'd like to have it

paramat:
pauloue,
thanks, your linked comment covers the issue i described above.
And thanks for pointing out that All Settings also does not correctly work.

SmallJoker,
My comments above were random thoughts for discussion, not confident objections, most were misunderstandings, now edited. I can see the usefulness of what you are trying to do and do not object.
> Move the flag registration to each mapgen file

Yes please.
> Set flag defaults using u32 values

Yes please, unless this causes a lot of extra complexity.
> Would you like to see something of this?
New Settings object for in-game use, named minetest.mapgen_settings
Deprecate minetset.*et_mapgen_setting*()

I cannot judge as i have always been bad with settings architecture, i will leave this to the other core devs. However, i have no objections to those 2 proposals if they are suitable.

Wuzzy2:
If I understand correctly, this PR is required to make #9254 possible.

SmallJoker:
@Wuzzy2 Yes. This PR provides the function for reading the currently active mapgen flags from `minetest.conf` - or default setting, if not specified.

SmallJoker:
@paramat After the most recent commit:

**Map compatibility**
Old `map_meta.txt` file (look at dungeons)
   * `mg_flags = caves, light, decorations, biomes`
After rejoin
   * `mg_flags = caves, nodungeons, light, decorations, biomes`
   * Flags are taken literally, no on/off switching

**minetest.conf settings**
`mg_flags = nobiomes` new world
   * `map_meta.txt` -> `mg_flags = caves, dungeons, light, decorations, nobiomes`
   * Flags are applied with on/off switching as documented in `settingtypes.txt`
No `mg_flags` defined
   * `map_meta.txt` -> `mg_flags = caves, dungeons, light, decorations, biomes`

Unittests and the testing mod still passes.

paramat:
Thanks, the move to the specific mapgen files looks fine.
> Map compatibility
Old map_meta.txt file (look at dungeons):
mg_flags = caves, light, decorations, biomes

In MT master it seems mg_flags in map_meta.txt always contains all 5 flags, so that would not appear in map_meta.txt unless edited that way?
> minetest.conf settings
mg_flags = nobiomes new world:
map_meta.txt -> mg_flags = caves, dungeons, light, decorations, nobiomes

Nice, that fixes the issue i reported.

I still think that documentation should recommend that all flags are specified when adding a line to .conf, even though this will become non-essential if this PR is merged.
Because there is 'flag' and 'noflag' those should be used instead of leaving a flag missing which is sloppy.

src/mapgen/mapgen.cpp
```diff
@@ -215,6 +215,21 @@ void Mapgen::getMapgenNames(std::vector<const char *> *mgnames, bool include_hid
...
 }
 
+void Mapgen::setDefaultSettings(Settings *settings)
+{
+	// Option 1: List all defaults here as an overview
+	// Option 2: Each mapgen contains its own default setter function
+	// This is option 2.
+
+	settings->setDefault("mg_flags", flagdesc_mapgen,
+		 MG_CAVES | MG_DUNGEONS | MG_LIGHT | MG_DECORATIONS | MG_BIOMES);
```
paramat: Does this avoid forcing dungeons to enabled for old worlds that had them disabled? Just checking as this default is different from before.

paramat:
All this looks fine to me so far.

src/mapgen/mapgen_carpathian.cpp
```diff
@@ -227,6 +227,12 @@ void MapgenCarpathianParams::writeParams(Settings *settings) const
 }
 
 
+void MapgenCarpathianParams::setDefaultSettings(Settings *settings)
```
paramat: Would these functions be better named `setDefaultFlags()` as that is what they do?

src/mapgen/mapgen.cpp
SmallJoker: This is exactly why I posted an example about the map compatibility in my last post. If it's not specified, it will be read as "not set" without any defaulting mechanisms that happen in `minetest.conf`.

Feel free to create a world, then remove `dungeons` from the flags to see what happens (becomes `nodungeons`).

src/mapgen/mapgen_carpathian.cpp
SmallJoker: It's what they currently do. This function is intended to be used for further settings, such as defaults for noise params.

src/mapgen/mapgen.cpp
paramat: Ok, no problem.

src/mapgen/mapgen_carpathian.cpp
paramat: Ok, i suspected these might be extended to other settings. This is fine.

doc/lua_api.txt
```diff
@@ -6040,6 +6040,9 @@ It can be created via `Settings(filename)`.
     * `default` is the value returned if `key` is not found.
     * Returns `nil` if `key` is not found and `default` not specified.
 * `get_np_group(key)`: returns a NoiseParams table
+* `get_flags(key)`:
+    * Returns `{flag = true/false, ...}` according to the set flags.
+    * Is currently limited to mapgen settings like `mgv5_spflags`.
```
paramat: "Is currently limited to mapgen flags like `mg_flags` and `mgv5_spflags`."

paramat:
My mod testing code:
```
t = core.settings:get_flags("mg_flags")
print(dump(t))
t = core.settings:get_flags("mgv5_spflags")
print(dump(t))
t = core.settings:get_flags("mgv6_spflags")
print(dump(t))
t = core.settings:get_flags("mgv7_spflags")
print(dump(t))
t = core.settings:get_flags("mgvalleys_spflags")
print(dump(t))
t = core.settings:get_flags("mgcarpathian_spflags")
print(dump(t))
t = core.settings:get_flags("mgflat_spflags")
print(dump(t))
t = core.settings:get_flags("mgfractal_spflags")
print(dump(t))
```
1st test, with no mapgen flags present in .conf, output:
```
{
	decorations = true,
	caves = true,
	biomes = true,
	dungeons = true,
	light = true
}
{
	caverns = true
}
{
	flat = false,
	mudflow = true,
	trees = true,
	snowbiomes = true,
	biomeblend = true,
	jungles = true
}
{
	caverns = true,
	mountains = true,
	ridges = true,
	floatlands = false
}
{
	humid_rivers = true,
	altitude_chill = true,
	altitude_dry = true,
	vary_river_depth = true
}
{
	rivers = false,
	caverns = true
}
{
	hills = false,
	lakes = false
}
{
	terrain = true
}
```
Default mapgen flags correctly fetched.

doc/lua_api.txt
SmallJoker: Well so: `Is currently limited to mapgen settings like 'mgv5_spflags' or ' mg_flags'.`?

paramat:
2nd test, i disabled/enabled various general and mapgen-specific mapgen flags, all changes were correctly detected by this PR.
So, PR seems to work fine.

src/mapgen/mapgen.cpp
```diff
@@ -215,6 +215,21 @@ void Mapgen::getMapgenNames(std::vector<const char *> *mgnames, bool include_hid
 	}
 }
 
+void Mapgen::setDefaultSettings(Settings *settings)
+{
+	// Option 1: List all defaults here as an overview
+	// Option 2: Each mapgen contains its own default setter function
+	// This is option 2.
```
paramat: Remove comment?

doc/lua_api.txt
paramat: Hmm, what i meant was, why call these 'settings' when it is more descriptive to call them 'mapgen flags'.
But i can see the problem with my suggestion now.
How about:
"Is currently limited to the mapgen flags 'mg_flags' and 'mgv5_spflags'."
?

paramat: "Is currently limited to mapgen flags 'mg_flags' and mapgen-specific flags like 'mgv5_spflags'"
?

paramat:
> In MT master:
I tried adding to .conf:
mg_flags = light
Result is a V7 world with:
mg_flags = caves, nodungeons, light, decorations, biomes
Somehow dungeons have been disabled.

This problem is now fixed by the PR.

paramat:
Actually, i understand almost all of this, so trusting you with the rest.
+1 once my line comments are attended to.

doc/lua_api.txt
paramat: Resolved.

src/mapgen/mapgen.cpp
paramat: Resolved.

paramat:
:+1:

PR #9287, opened 2020-01-05
Formspec: create a new class for inventorylists

### LINKED ISSUE #4844: It's not possible anymore to write text with label[ ... ] over item images or inventory slots in formspecs

DS-Minetest:
- Goal of the PR: Complete #8740.
- There's a new class `GUIInventoryList` which is the element class for inventorylists.
- Fixes #4844 (in formspec version 3).
(Btw. there was a little clipping-bug with slotborders (probably introduced by #8740) which is fixed now.)
- This PR also makes it easier to implement things like customizable slot-spacing and slot-size and per-inventorylist colors and borders.
- It might be possible to move more code to the new class, eg. the event handler, and to improve the code-structure. But I didn't want to make this PR even larger (I'm lazy).

## To do

(Note that there's still a new `todo`, see line comment.)

## How to test

I've used this for testing:
```lua
local elements = [[
	list[current_player;main;0,0;4,4;]
	button[5,0;2,1;btn;Button]
	button[3,1.5;3,1;btn;Button2]
]]

local fs = [[
	size[11,10]
]] .. elements .. [[
	tooltip[0,0;10,5;<rect_mode tooltip defined after other elements>]

	listcolors[#000;#fff;#f00;#0f0;#00f]

	container[0,6]
	tooltip[0,0;10,5;<rect_mode tooltip defined before other elements>]
]] .. elements .. [[
	container_end[]
]]

	minetest.show_formspec(player:get_player_name(), "fs_test", fs)
end)

minetest.register_chatcommand("fs", {
	params = "",
	description = "Show the test formspec.",
	privs = {},
	func = function(name, param)
		minetest.show_formspec(name, "fs", fs)
		return true
	end,
})
```
It's suggest testing by clicking around.

### DIFF
--- a/src/gui/guiFormSpecMenu.cpp
+++ b/src/gui/guiFormSpecMenu.cpp
@@ -442,10 +442,7 @@ void GUIFormSpecMenu::parseList(parserData *data, const std::string &element)
 
 		GUIInventoryList *e = new GUIInventoryList(Environment, this, spec.fid,
 				rect, m_invmgr, loc, listname, geom, start_i, imgsize, slot_spacing,
-				this, data->inventorylist_options.slotbg_n,
-				data->inventorylist_options.slotbg_h,
-				data->inventorylist_options.slotborder,
-				data->inventorylist_options.slotbordercolor, m_font);
+				this, data->inventorylist_options, m_font);
 
 		m_inventorylists.push_back(e);
 		m_fields.push_back(spec);
--- a/src/gui/guiFormSpecMenu.cpp
+++ b/src/gui/guiFormSpecMenu.cpp
@@ -61,6 +61,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include "guiButtonImage.h"
 #include "guiButtonItemImage.h"
 #include "guiEditBoxWithScrollbar.h"
+#include "guiInventoryList.h"
 #include "guiItemImage.h"
 #include "guiScrollBar.h"
 #include "guiTable.h"
@@ -130,7 +131,7 @@ GUIFormSpecMenu::~GUIFormSpecMenu()
 	for (auto &table_it : m_tables)
 		table_it.second->drop();
 	for (auto &inventorylist_it : m_inventorylists)
-		inventorylist_it.e->drop();
+		inventorylist_it->drop();
 	for (auto &checkbox_it : m_checkboxes)
 		checkbox_it.second->drop();
 	for (auto &scrollbar_it : m_scrollbars)
@@ -347,7 +348,7 @@ void GUIFormSpecMenu::parseContainerEnd(parserData* data)
 	}
 }
 
-void GUIFormSpecMenu::parseList(parserData* data, const std::string &element)
+void GUIFormSpecMenu::parseList(parserData *data, const std::string &element)
 {
 	if (m_client == 0) {
 		warningstream<<"invalid use of 'list' with m_client==0"<<std::endl;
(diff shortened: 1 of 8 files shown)


src/gui/guiFormSpecMenu.cpp
```diff
@@ -2681,7 +2686,7 @@ void GUIFormSpecMenu::regenerateGui(v2u32 screensize)
 	// Base position of contents of form
 	mydata.basepos = getBasePos();
 
-	/* Convert m_init_draw_spec to m_inventorylists */
+	/* Convert m_init_draw_spec to m_inventorylists (todo: what does that mean?) */
```
DS-Minetest: Can I remove this comment? What is `m_init_draw_spec`?

src/gui/guiFormSpecMenu.cpp
```diff
@@ -429,36 +430,31 @@ void GUIFormSpecMenu::parseList(parserData* data, const std::string &element)
...
+				pos.Y + (geom.Y - 1) * slot_spacing.Y + imgsize.Y);
 
-		// the element the list is bound to should not block mouse-clicks
-		e->setVisible(false);
+		GUIInventoryList *e = new GUIInventoryList(Environment, this, spec.fid,
+				rect, m_invmgr, loc, listname, geom, start_i, imgsize, slot_spacing,
+				this, data->inventorylist_options.slotbg_n,
+				data->inventorylist_options.slotbg_h,
+				data->inventorylist_options.slotborder,
+				data->inventorylist_options.slotbordercolor, m_font);
```
SmallJoker: Link target placeholder

src/gui/guiInventoryList.h
```diff
@@ -0,0 +1,129 @@
...
+		const std::string &listname,
+		const v2s32 &geom,
+		const s32 start_item_i,
+		const v2s32 &slot_size,
+		const v2f32 &slot_spacing,
+		GUIFormSpecMenu *fs_menu,
+		const video::SColor &slotbg_n,
+		const video::SColor &slotbg_h,
+		bool slotborder,
+		const video::SColor &slotbordercolor,
```
SmallJoker: Link target placeholder

src/gui/guiFormSpecMenu.h
```diff
@@ -403,6 +350,13 @@ class GUIFormSpecMenu : public GUIModalMenu
 		GUITable::TableOptions table_options;
 		GUITable::TableColumns table_columns;
 
+		struct {
+			bool slotborder = false;
+			video::SColor slotbg_n = video::SColor(255, 128, 128, 128);
+			video::SColor slotbg_h = video::SColor(255, 192, 192, 192);
+			video::SColor slotbordercolor = video::SColor(200, 0, 0, 0);
+		} inventorylist_options;
```
SmallJoker: Consider moving this to `guiInventoryList.h` so that you can pass/use the entire struct in those two places;

https://github.com/minetest/minetest/pull/9287/files#r363431743
https://github.com/minetest/minetest/pull/9287/files#r363431900

src/gui/guiFormSpecMenu.cpp
SmallJoker: Lost comment. Just remove it.

src/gui/guiFormSpecMenu.h
DS-Minetest: Done.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: Thanks! Done.

src/gui/guiFormSpecMenu.cpp
DS-Minetest: ?
Is this the same as the next comment?

DS-Minetest:
The visual change for buttons when hovering currently doesn't work if there's a list in front of the button.
I'm not sure how this can be fixed.

Issue #4087, opened 2016-05-04
Node metadata is lost on falling nodes


jhcole:
Node metadata set on a `falling_node` is lost if the node falls. For example, a falling chest will lose its inventory if it falls.

sofar:
This is due to falling.lua only serializing and deserializing node data, not metadata. Should be fixable in falling.lua.

sofar:
It should also serialize and deserialize any active node timers.

PilzAdam:
@sofar maybe we should just call a callback in nodedef if a node starts falling? This way we don't have to consider every corner case and chests could just serialize the stuff they actually need.

sofar:
@pilzadam yeah, that's a very good idea. I'll probably look into this after 0.4.14...

PR #9294, opened 2020-01-08
Add callbacks for falling and attached nodes

### LINKED ISSUE #6943: Add callback for attached node detaching
##### Issue type
 - Feature request
 
##### Minetest version
```
0.4.16
```

##### Summary
`attached_node` is a special group to make a node attached to other nodes. When the block to which the node is attached to is destroyed, the attached node drops.

As far I know, there's no callback called when this happens. It's impossible to detect. So I request to add a callback for when an attached node is detached and dropped. Note this is distinct from just digging or destroying the node directly. Other callbacks should not be affected.

Proposed functions for the node definition:

* `on_detach_node(pos, node)`: Called before node detaches.
* `after_detach_node(pos, oldnode)`: Called after node has been detached. `oldnode` is the node table of the node before it detached.

These callbacks *only* apply to attached nodes. Other nodes are not affected by this.

Example: You attach a picture frame (the attached node) to stone. You dig the stone. Now the picture frame is dropped as well. This needs to trigger the callbacks above.

#### Relevance
Mesecons, redstone and similar mods need to propagate wire updates. The wire shape is updated when e.g. a neighbor wire is dug or placed. Things get tricky if you make the wire an `attached_node` however. If you dig the node below the wire, the wire drops but this can't be detected by the mod. I see no possible way to propagate any wire updates, thereby seriously messing up the wire state. E.g. a wire might stay active even if the connection to the power source is broken.
With the proposed callbacks, this bug can be fixed.

Since the callback is pretty basic, I'm sure there are other use cases as well.

Wuzzy2:
This adds 4 callbacks for nodes:

- `on_fall`: Before node falls
- `after_fall`: After node *started* to fall
- `on_detach`: Before an `attached_node` detaches
- `after_detach`: After node has been detached

See the changes in `lua_api.txt` for details.

It's also possible to overwrite the builtin handling of node falling/detaching.

Fixes #6943.

## Justification

See #6943 for justification of `on_detach` and `after_detach`. This issue is supported by core dev.

`on_fall` and `after_fall` are added because they are functionally similar.

## To do

This PR is ready.

## How to test

Here are two test nodes to play around with (basically just outputting the params into chat):

```
-- Is supposed to fall when it doesn't rest on solid ground
minetest.register_node("testnodes:falling", {
	description = "Falling Node",
	tiles = {
		"testnodes_node.png",
		"testnodes_node.png",
		"testnodes_node_falling.png",
	},
	groups = { falling_node = 1, dig_immediate = 2 },

	on_construct = function(pos)
		local m = minetest.get_meta(pos)
		m:set_string("hello", "world")
	end,
	on_fall = function(pos, node)
		minetest.log("error", "on_fall")
		minetest.log("error", "   pos = "..minetest.pos_to_string(pos))
		minetest.log("error", "   node = "..tostring(node.name))
		--Uncomment to test overwrite behavior
		--minetest.remove_node(pos)
		--return true
	end,
	after_fall = function(pos, oldnode, oldmeta, obj)
		minetest.log("error", "after_fall")
		minetest.log("error", "   pos = "..minetest.pos_to_string(pos))
		minetest.log("error", "   oldnode = "..tostring(oldnode.name))
		minetest.log("error", "   oldmeta = "..tostring(dump(oldmeta)))
		minetest.log("error", "   obj = "..tostring(obj))
	end,
})

-- This node attaches to the floor and drops as item
-- when the floor is gone.
minetest.register_node("testnodes:attached", {
	description = "Floor-Attached Node",
	tiles = {
		"testnodes_attached_top.png",
		"testnodes_attached_bottom.png",
		"testnodes_attached_side.png",
	},
	groups = { attached_node = 1, dig_immediate = 2 },

	on_construct = function(pos)
		local m = minetest.get_meta(pos)
		m:set_string("hello", "world")
	end,
	on_detach = function(pos, node)
		minetest.log("error", "on_detach")
		minetest.log("error", "   pos = "..minetest.pos_to_string(pos))
		minetest.log("error", "   node = "..tostring(node.name))
		--Uncomment to test overwrite behavior
		--minetest.remove_node(pos)
		--return true
	end,
	after_detach = function(pos, oldnode, oldmeta)
		minetest.log("error", "after_detach")
		minetest.log("error", "   pos = "..minetest.pos_to_string(pos))
		minetest.log("error", "   oldnode = "..tostring(oldnode.name))
		minetest.log("error", "   oldmeta = "..tostring(dump(oldmeta)))
	end,
})
```

### DIFF
--- a/builtin/game/falling.lua
+++ b/builtin/game/falling.lua
@@ -402,22 +402,39 @@ core.register_entity(":__builtin:falling_node", {
 })
 
 local function convert_to_falling_node(pos, node)
-	local obj = core.add_entity(pos, "__builtin:falling_node")
-	if not obj then
-		return false
-	end
 	-- remember node level, the entities' set_node() uses this
 	node.level = core.get_node_level(pos)
 	local meta = core.get_meta(pos)
 	local metatable = meta and meta:to_table() or {}
+	local oldmeta = metatable.fields
 
 	local def = core.registered_nodes[node.name]
+
+	if def and def.on_fall then
+		local ok, obj = def.on_fall(vector.new(pos),
+			{name=node.name, param1=node.param1, param2=node.param2})
+		if type(ok) == "boolean" then
+			return ok, obj
+		end
+	end
+
+	local obj = core.add_entity(pos, "__builtin:falling_node")
+	if not obj then
+		return false
+	end
 	if def and def.sounds and def.sounds.fall then
 		core.sound_play(def.sounds.fall, {pos = pos}, true)
 	end
 
 	obj:get_luaentity():set_node(node, metatable)
 	core.remove_node(pos)
+
+	if def and def.after_fall then
+		def.after_fall(vector.new(pos),
+				{name=node.name, param1=node.param1, param2=node.param2},
+				oldmeta, obj)
+	end
+
 	return true, obj
 end
 
@@ -498,23 +515,37 @@ function core.check_single_for_falling(p)
 			if same and d_bottom.paramtype2 == "leveled" and
 					core.get_node_level(p_bottom) <
 					core.get_node_max_level(p_bottom) then
-				convert_to_falling_node(p, n)
-				return true
+				return convert_to_falling_node(p, n)
 			end
 			-- Otherwise only if the bottom node is considered "fall through"
 			if not same and
 					(not d_bottom.walkable or d_bottom.buildable_to) and
 					(core.get_item_group(n.name, "float") == 0 or
 					d_bottom.liquidtype == "none") then
-				convert_to_falling_node(p, n)
-				return true
+				return convert_to_falling_node(p, n)
 			end
 		end
 	end
 
 	if core.get_item_group(n.name, "attached_node") ~= 0 then
+		local def = core.registered_nodes[n.name]
 		if not builtin_shared.check_attached_node(p, n) then
+			if def and def.on_detach then
+				local ok = def.on_detach(vector.new(p),
+						{name=n.name, param1=n.param1, param2=n.param2})
+				if type(ok) == "boolean" then
+					return ok
+				end
+			end
+			local oldmeta = core.get_meta(p):to_table().fields
+
 			drop_attached_node(p)
+
+			if def and def.after_detach then
+				def.after_detach(vector.new(p),
+						{name=n.name, param1=n.param1, param2=n.param2},
+						oldmeta)
+			end
 			return true
 		end
 	end
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -7885,6 +7885,30 @@ Used by `minetest.register_node`.
         -- Default: nil
         -- Warning: making a liquid node 'floodable' will cause problems.
 
+        on_fall = function(pos, node)
+        -- Called before node starts to fall.
+        -- If a boolean is returned, the builtin falling node
+        -- handling is skipped. `true` denotes successful falling.
+        -- By default, the node is removed and a falling node
+        -- entity is spawned at `pos`.
+        -- default: nil
+
+        after_fall = function(pos, oldnode, oldmeta, obj),
+        -- Called when node started to fall and a falling node entity
+        -- has been spawned. `obj` is the falling node entity.
+        -- default: nil
+
+        on_detach = function(pos, oldnode)
+        -- Called when an `attached_node` is about to be detached.
+        -- If a boolean is returned, the builtin node detachment
+        -- handling is skipped. `true` denotes a successful detachment.
+        -- By default, the node is removed and is turned into drops.
+        -- default: nil
+
+        after_detach = function(pos, oldnode, oldmeta)
+        -- Called when an `attached_node` was detached.
+        -- default: nil
+
         preserve_metadata = function(pos, oldnode, oldmeta, drops),
         -- Called when oldnode is about be converted to an item, but before the
         -- node is deleted from the world or the drops are added. This is
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -7887,26 +7887,38 @@ Used by `minetest.register_node`.
 
         on_fall = function(pos, node)
         -- Called before node starts to fall.
-        -- If a boolean is returned, the builtin falling node
-        -- handling is skipped. `true` denotes successful falling.
-        -- By default, the node is removed and a falling node
-        -- entity is spawned at `pos`.
+        -- Optionally should return `<boolean>, <obj>`.
+        -- If `boolean` is returned as a boolean, the builtin falling node
+        -- handling is skipped.
+        -- `boolean=true` means an object was created by the function which
+        -- must be returned in `obj`.
+        -- `boolean=false` means no object was created and `obj` must be `nil`;
+        -- `boolean=nil` means the builtin falling handling is executed.
+        -- The builtin falling node falling means that the node at `pos` is
+        -- removed and a `__builtin:falling_node` entity spawns in its place.
+        -- `after_fall` is only called if `boolean` `nil`.
         -- default: nil
 
         after_fall = function(pos, oldnode, oldmeta, obj),
         -- Called when node started to fall and a falling node entity
-        -- has been spawned. `obj` is the falling node entity.
+        -- has been spawned due to the builtin falling node handler.
+        -- `obj` is the falling node entity.
         -- default: nil
 
-        on_detach = function(pos, oldnode)
+        on_detach = function(pos, node)
         -- Called when an `attached_node` is about to be detached.
         -- If a boolean is returned, the builtin node detachment
-        -- handling is skipped. `true` denotes a successful detachment.
-        -- By default, the node is removed and is turned into drops.
+        -- handling is skipped.
+        -- `true` means that one or more objects were created,
+        -- `false` means this was not the case.
+        -- The builtin node attachment handling means that the node
+        -- is removed, drops as item(s).
+        -- `after_detach` is only called if `nil` is returned.
         -- default: nil
 
         after_detach = function(pos, oldnode, oldmeta)
-        -- Called when an `attached_node` was detached.
+        -- Called when an `attached_node` was detached using the builtin
+        -- node detachment handler.
         -- default: nil
 
         preserve_metadata = function(pos, oldnode, oldmeta, drops),


builtin/game/falling.lua
```diff
@@ -243,14 +257,27 @@ function core.check_single_for_falling(p)
 				core.get_node_max_level(p_bottom))) and
 
 				(not d_bottom.walkable or d_bottom.buildable_to) then
-			convert_to_falling_node(p, n)
-			return true
+			local ok = convert_to_falling_node(p, n)
+			return ok
```
MoNTE48: -local ok = convert_to_falling_node(p, n)
-return ok

+return  convert_to_falling_node(p, n)

builtin/game/falling.lua
```diff
@@ -144,21 +144,35 @@ core.register_entity(":__builtin:falling_node", {
...
 	node.level = core.get_node_level(pos)
 	local meta = core.get_meta(pos)
 	local metatable = meta and meta:to_table() or {}
+	local oldmeta = metatable.fields
 
 	local def = core.registered_nodes[node.name]
+
+	local ok
+	if def and def.on_fall then
+		ok = def.on_fall(table.copy(pos), table.copy(node))
```
SmallJoker: `ok` is only used within this scope.

builtin/game/falling.lua
```diff
@@ -144,21 +144,35 @@ core.register_entity(":__builtin:falling_node", {
...
+		if type(ok) == "boolean" then
+			return ok
+		end
+	end
+
 	if def and def.sounds and def.sounds.fall then
 		core.sound_play(def.sounds.fall, {pos = pos})
 	end
-
+	local obj = core.add_entity(pos, "__builtin:falling_node")
```
SmallJoker: The entity should be added before playing the sound to ensure there's no sound when it fails.

builtin/game/falling.lua
```diff
@@ -243,14 +257,26 @@ function core.check_single_for_falling(p)
...
+				local ok = def.on_detach(table.copy(p), table.copy(n))
+				if type(ok) == "boolean" then
+					return ok
+				end
+			end
+			local oldmeta = core.get_meta(p):to_table().fields
+
 			drop_attached_node(p)
+
+			if def and def.after_detach then
```
SmallJoker: L267 and here: `def` cannot be `nil`. The `core.get_item_group ... ~= 0` always requires a valid node def.

Wuzzy2:
@SmallJoker: Done.

Wuzzy2:
Bump after 2 months.

builtin/game/falling.lua
```diff
@@ -144,21 +144,34 @@ core.register_entity(":__builtin:falling_node", {
...
 	node.level = core.get_node_level(pos)
 	local meta = core.get_meta(pos)
 	local metatable = meta and meta:to_table() or {}
+	local oldmeta = metatable.fields
 
 	local def = core.registered_nodes[node.name]
+
+	if def and def.on_fall then
+		local ok = def.on_fall(table.copy(pos), table.copy(node))
+		if type(ok) == "boolean" then
```
appgurueu, line 156: Should probably be a bit stricter, as in `assert(type(ok) == "boolean")` or throwing a warning

builtin/game/falling.lua
```diff
@@ -243,14 +256,26 @@ function core.check_single_for_falling(p)
...
-			convert_to_falling_node(p, n)
-			return true
+			return convert_to_falling_node(p, n)
 		end
 	end
 
 	if core.get_item_group(n.name, "attached_node") ~= 0 then
+		local def = core.registered_nodes[n.name]
 		if not builtin_shared.check_attached_node(p, n) then
+			if def.on_detach then
```
sfan5, line 266: `def and ...` missing

builtin/game/falling.lua
```diff
@@ -243,14 +256,26 @@ function core.check_single_for_falling(p)
 				core.get_node_max_level(p_bottom))) and
 
 				(not d_bottom.walkable or d_bottom.buildable_to) then
-			convert_to_falling_node(p, n)
-			return true
+			return convert_to_falling_node(p, n)
```
sfan5, line 259: The current state of the code has a second call to `convert_to_falling_node(p, n)`, on merge git will only apply one change.
So please rebase.

Wuzzy2:
@appgurueu: This is intentional, the return value is allowed to be a non-boolean. Re-read `lua_api.txt`.

Wuzzy2:
Ready for re-review!
As I needed to touch a lot of code, I'm afraid the review process is back to square one. :-(

sfan5:
> If you don't want to make this stricter, maybe check for truthiness?

^ I would consider this useful though.

sfan5:
```diff
-		if type(ok) == "boolean" then
-			return ok
+		if ok then
+			return true
 		end
```
The reason is that a boolean check is weirdly specific, but no there is no real difference.

Wuzzy2:
But it's intentional:

> If a boolean is returned, the builtin falling node
> handling is skipped. `true` denotes successful falling.

Implying you can also return `false` to skip builtin falling, and “unsuccessful” falling.

So, `true`, `false` and `<anything else>` are valid return values.

Wuzzy2:
Any chance for the second approval? I have given a justification of why I cannot change said code.

Wuzzy2:
Bump after 3 months.

appgurueu:
Bump.

builtin/game/falling.lua
```diff
@@ -369,22 +369,36 @@ core.register_entity(":__builtin:falling_node", {
...
 	-- remember node level, the entities' set_node() uses this
 	node.level = core.get_node_level(pos)
 	local meta = core.get_meta(pos)
 	local metatable = meta and meta:to_table() or {}
+	local oldmeta = metatable.fields
 
 	local def = core.registered_nodes[node.name]
+
+	if def and def.on_fall then
+		local ok = def.on_fall(table.copy(pos), table.copy(node))
```
Desour, line 381:
```suggestion
		local ok = def.on_fall(vector.new(pos), {name = node.name, param1 = node.param1, param2 = node.param2})
```

builtin/game/falling.lua
```diff
@@ -369,22 +369,36 @@ core.register_entity(":__builtin:falling_node", {
...
+	end
 	if def and def.sounds and def.sounds.fall then
 		core.sound_play(def.sounds.fall, {pos = pos}, true)
 	end
 
 	obj:get_luaentity():set_node(node, metatable)
 	core.remove_node(pos)
+
+	if def and def.after_fall then
+		def.after_fall(pos, node, oldmeta, obj)
```
Desour, line 399:
```suggestion
		def.after_fall(vector.new(pos), node, oldmeta, obj)
```
The `pos` table can be the same as the one given to `core.spawn_falling_node` or `core.check_single_for_falling`. (It would be problematic if user provided arguments are modified by callbacks and the argument is used for something else afterwards.)

builtin/game/falling.lua
```diff
@@ -465,23 +479,34 @@ function core.check_single_for_falling(p)
...
+				return convert_to_falling_node(p, n)
 			end
 		end
 	end
 
 	if core.get_item_group(n.name, "attached_node") ~= 0 then
+		local def = core.registered_nodes[n.name]
 		if not builtin_shared.check_attached_node(p, n) then
+			if def and def.on_detach then
+				local ok = def.on_detach(table.copy(p), table.copy(n))
```
Desour, line 498:
```suggestion
				local ok = def.on_detach(vector.new(p), {name = n.name, param1 = n.param1, param2 = n.param2})
```
(`table.copy` is slow as it uses `pairs`, and can copy garbage.)

builtin/game/falling.lua
```diff
@@ -465,23 +479,34 @@ function core.check_single_for_falling(p)
...
+				if (type(ok) == "boolean") then
```
Desour, line 499:
```suggestion
				if type(ok) == "boolean" then
```

builtin/game/falling.lua
```diff
@@ -465,23 +479,34 @@ function core.check_single_for_falling(p)
...
+					return ok
+				end
+			end
+			local oldmeta = core.get_meta(p):to_table().fields
+
 			drop_attached_node(p)
+
+			if def and def.after_detach then
+				def.after_detach(p, n, oldmeta)
```
Desour, line 508:
```suggestion
				def.after_detach(vector.new(p), n, oldmeta)
```

(Also: Giving the oldmeta here seems a little bit overkill. A mod could still get it by saving it in a variable in `on_detach`, (which might be hacky).)

Desour (commented):
Sorry for late review.
Api improvements are important, :+1: for adding these callbacks.

Wuzzy2:
Updated!

I agree on the `vector.new` but not on the `table.copy`. Is it really that big of a deal? It feels bad that such an essential function is de facto banned. I'm waiting for a 2nd opinion on this.

appgurueu:
I'm not a fan of pass-by-value for reference types at all. Yes, it might be consistent to do this, as the API implemented on the C++ side does it (as it has to convert C++ types to Lua tables/userdata), but usually it is a waste of performance: vector functions return copies anyways, the node passed in a callback is rarely modified. If a modder chooses to modify the passed tables in the callback for little performance gain, relying on a greater performance loss in the engine, that's on them.

Desour:
It seems to be common practice in builtin to deepcopy vectors, nodes and co., without `table.copy`, eg.: 
https://github.com/minetest/minetest/blob/531e7ef8ebbb9c7e08bf9043d96a44558c7a6d7f/builtin/game/item.lua#L393-L398


> I agree [..] not on the `table.copy`. Is it really that big of a deal? It feels bad that such an essential function is de facto banned. I'm waiting for a 2nd opinion on this.

Not a 2nd opinion. Performance-wise it is a big deal. `table.copy` can make no assumptions about the table (hence uses NYI `pairs`, looks at every type for potential deepcopy, ...), but we know that we have a node (the most expensive part in the non-`table.copy` is creating the new table object (which also has to be garbage collected later) afaik). 

>I'm not a fan of pass-by-value for reference types at all. ...

Yes, me neither. It could be changed by mentioning in the documentation that one should never modify given objects. But old, and also new, mods would not comply with this and cause many bugs that are hard to find. Also, pass-by-value is easier for modders.

Wuzzy2:
After having consulted IRC, I now have confirmation that the explicit table format is indeed preferred. I have now switched to the `{name=x, param1=y, param2=z}` syntax.

Wuzzy2:
Bump.

Wuzzy2:
Since the code got old again, I rebased it and pushed the rebase.

doc/lua_api.txt
```diff
@@ -7885,6 +7885,30 @@ Used by `minetest.register_node`.
         -- Default: nil
         -- Warning: making a liquid node 'floodable' will cause problems.
 
+        on_fall = function(pos, node)
+        -- Called before node starts to fall.
+        -- If a boolean is returned, the builtin falling node
+        -- handling is skipped. `true` denotes successful falling.
+        -- By default, the node is removed and a falling node
+        -- entity is spawned at `pos`.
+        -- default: nil
```
Desour, line 7894: - There should be more information on what successful falling means (update spread to neighbours).
- The "builtin" falling node handling should probably also be called "default". Otherwise one could understand the "By default" in the last sentence as "If this callback is not defined (`on_fall = nil`)".
- The default / builtin falling node handling includes calling the `after_fall` callback. Imo that's an important detail to be mentioned here.

doc/lua_api.txt
```diff
@@ -7885,6 +7885,30 @@ Used by `minetest.register_node`.
...
+        -- Called when node started to fall and a falling node entity
+        -- has been spawned. `obj` is the falling node entity.
+        -- default: nil
+
+        on_detach = function(pos, oldnode)
+        -- Called when an `attached_node` is about to be detached.
+        -- If a boolean is returned, the builtin node detachment
+        -- handling is skipped. `true` denotes a successful detachment.
+        -- By default, the node is removed and is turned into drops.
+        -- default: nil
```
Desour, line 7906: Same as above.

builtin/game/falling.lua
```diff
@@ -498,23 +515,37 @@ function core.check_single_for_falling(p)
...
+				end
+			end
+			local oldmeta = core.get_meta(p):to_table().fields
+
 			drop_attached_node(p)
+
+			if def and def.after_detach then
+				def.after_detach(vector.new(p),
+						{name=n.name, param1=n.param1, param2=n.param2},
+						oldmeta)
```
Desour, line 547: (You could pass a table containing the added items by `drop_attached_node()`. Not important though.)

Desour (commented):
- `minetest.spawn_falling_node(pos)` was changed: It can now return `true` but no objref on success.
- (The api is of course still not perfect. Ie. there's no global `register_after_node_fall()` or similar. And you can not let a non-leveled falling node fall into itself. But overwriting `core.check_single_for_falling()` can probably do this well enough, so this is fine.)
- PR is mostly fine and a good enhancement to the api. :+1:

Wuzzy2:
I've did the neccessary changes. Thanks for catching the missing `obj`.

doc/lua_api.txt
```diff
@@ -7885,6 +7885,42 @@ Used by `minetest.register_node`.
...
+        -- If `boolean` is returned as a boolean, the builtin falling node
+        -- handling is skipped.
+        -- `boolean=true` means an object was created by the function which
+        -- must be returned in `obj`.
+        -- `boolean=false` means no object was created and `obj` must be `nil`;
+        -- `boolean=nil` means the builtin falling handling is executed.
+        -- The builtin falling node falling means that the node at `pos` is
+        -- removed and a `__builtin:falling_node` entity spawns in its place.
+        -- `after_fall` is only called if `boolean` `nil`.
+        -- default: nil
```
Desour, line 7900: Now the api states that the `on_fall` callback must always add a falling node entity to allow to propagate the falling check. Ie. it's impossible to spawn a totally different object.
This makes returning something else than `nil` pretty useless.

Also, if the returned bool is `false`, the node must not have been removed, else we have floating nodes, the doc is missing this information. (I'm referring to [this](https://github.com/minetest/minetest/blob/c6e03b5ba91c6578f16f7ee2b6818cea4aa70992/builtin/game/falling.lua#L596).)

I think anything that doesn't involve forcing to return a `__builtin:falling_node` entity will probably come with an api break of `minetest.spawn_falling_node()`.
Possible options include:
- 1. Allow `minetest.spawn_falling_node()` to return `nil` as 2nd return value.
- 2. Make `minetest.spawn_falling_node()` return `false` if `on_fall` has returned a boolean. Maybe return `false, bool`.
- 3. Make `minetest.spawn_falling_node()` return `false` and do nothing if `def.on_fall` exists. And add a 2nd function that also calls `on_fall`, naming suggestions follow:
  * 1. Add a function version number: `minetest.spawn_falling_node2()`
  * 2. Use a super descriptive and long name: `minetest.spawn_falling_node_and_also_call_on_fall_in_node_definition()`
  * 3. Reorder the words, à la `.node_dig()` and `.dig_node()`: `minetest.spawn_node_falling()`
  * 4. Add spelling errors: `minetest.spawn_faling_node()`

appgurueu:
This PR must be merged if if https://github.com/minetest/minetest_game/issues/2247 is to be fixed without hackily hooking the falling entity, as https://github.com/minetest/minetest/issues/4087 has been closed in favor of this approach (callbacks).

Wuzzy2:
@Desour: I have read your complaint multiple times and I still can't make a lot of sense of it. Where does it say that `on_fall` must always spawn a falling node?

Can you please show real code to me that will break things?

Df458 (changes requested):
Started testing, immediately found an issue: If `on_fall` or `on_detach` returns true but _doesn't remove the node_, the lua thread will be trapped in an infinite loop. We can discuss whether or not that should be acceptable behavior, but the fact of the matter is that _this requirement isn't documented either_, just assumed. It's also not necessarily obvious to modders, since a return value of false doesn't trigger this issue.

Wuzzy2:
I give up on this. This code is now stone old (>2 years), it seems to be a trash design so probably needs to be nuked and started from scratch anyway. Sorry. :-(

Closing.

Df458:
Unfortunate but understandable. I think the feature is genuinely useful though, so I hope we'll see a follow-up without these issues in the future, be it by you or someone else!

PR #9298, opened 2020-01-09
Fix z-fighting of large firelike, torchlike, signlike, plantlike nodes

### LINKED ISSUE #9292: Large firelike nodes z-fight with each other
##### Minetest version
```
5.2.0-dev @ 44458896d9097c15c722c5bffb3d72e14962b7ce
```

##### Summary
Nodes of `drawtype="firelike"` and with a large value for `visual_scale` will z-fight with each other.

##### Steps to reproduce
1. Place a node with `drawtype="firelike"` and `visual_scale=2.0` (or another large-ish value)
2. Place more of this node around it, in all directions (left, right, diagonal)
3. Walk around and move camera and see if you spot z-fighting

If you have installed `devtest` <https://forum.minetest.net/viewtopic.php?f=50&t=24030>, use the node `testnodes:firelike_double` for testing.

KqDeLavjNV

##### Additional context
Compare with `drawtype="plantlike"` and large `visual_scale` (also included in `devtest`). Those nodes don't z-fight, even if large.
I think the trick that `plantlike` nodes use is that there's a tiny offset and/or difference in angle for each node, thus avoiding the z-fighting.

So I suppose the fix for `firelike` nodes is to implement a similar strategy.

Wuzzy2:
Fixes #9292.

Nodes of drawtype `firelike`, `torchlike` and `signlike` will no longer z-fight if they have a large `visual_scale`.

This is done by applying a tiny X/Z offset based on their position in the world. For `torchlike` and `firelike`, the diagonal quads are also rotated by 1° (like for `plantlike`).

This also fixes slightly different z-fighting with `plantlike` when those are stacked vertically. Here, the 1° difference is changed based on the Y level, so that it's +1°, -1°, +1°, -1°. That way, even vertically stacked `plantlike` nodes will no longer z-fight.

## To do

Testing by someone who is not me.
Code review by someone who is not me.

## How to test

Place many large (`visual_scale > 1.25`) `firelike`, `torchlike`, `signlike`, `plantlike` nodes next to each other or stack them and try everything to make them z-fight.

In `devtest`, you will find example nodes for everything. https://git.minetest.land/Wuzzy/devtest

### DIFF
--- a/src/client/content_mapblock.cpp
+++ b/src/client/content_mapblock.cpp
@@ -892,14 +892,25 @@ void MapblockMeshGenerator::drawSignlikeNode()
 {
 	u8 wall = n.getWallMounted(nodedef);
 	useTile(0, MATERIAL_FLAG_CRACK_OVERLAY, MATERIAL_FLAG_BACKFACE_CULLING);
-	static const float offset = BS / 16;
+	static const float wall_offset = BS / 16;
 	float size = BS / 2 * f->visual_scale;
+
+	// Anti-z-fighting offset
+	int o;
+	if (p.Z % 2 == 0) {
+		o = p.X % 4;
+	} else {
+		o = p.X % 4 + 2;
+	}
+	o = (o + (p.Y % 4)) % 4;
+	float azf_offset = (o * 0.01 - 0.04);
+
 	// Wall at X+ of node
 	v3f vertices[4] = {
-		v3f(BS / 2 - offset,  size,  size),
-		v3f(BS / 2 - offset,  size, -size),
-		v3f(BS / 2 - offset, -size, -size),
-		v3f(BS / 2 - offset, -size,  size),
+		v3f(BS / 2 - wall_offset + azf_offset,  size,  size),
+		v3f(BS / 2 - wall_offset + azf_offset,  size, -size),
+		v3f(BS / 2 - wall_offset + azf_offset, -size, -size),
+		v3f(BS / 2 - wall_offset + azf_offset, -size,  size),
 	};
 
 	for (v3f &vertex : vertices) {
--- a/src/client/content_mapblock.cpp
+++ b/src/client/content_mapblock.cpp
@@ -857,12 +857,23 @@ void MapblockMeshGenerator::drawTorchlikeNode()
 		v3f(-size, -size, 0),
 	};
 
+	// Tiny offset to fix z-fighting
+	offset = v3f(0, 0, 0);
+	if (wall == DWM_YP) {
+		offset.X = BS * (p.Y % 4) * 0.001 - 0.004;
+	} else if (wall == DWM_YN) {
+		offset.Z = BS * (p.Y % 4) * 0.001 - 0.004;
+	} else {
+		offset.X = BS * ((p.Z % 4 + p.Y % 4) * 0.001 - 0.004);
+		offset.Z = BS * ((p.X % 4 + p.Y % 4) * 0.001 - 0.004);
+	}
+
 	for (v3f &vertex : vertices) {
 		switch (wall) {
 			case DWM_YP:
-				vertex.rotateXZBy(-45); break;
+				vertex.rotateXZBy(-44); break;
 			case DWM_YN:
-				vertex.rotateXZBy( 45); break;
+				vertex.rotateXZBy( 46); break;
 			case DWM_XP:
 				vertex.rotateXZBy(  0); break;
 			case DWM_XN:
@@ -872,6 +883,7 @@ void MapblockMeshGenerator::drawTorchlikeNode()
 			case DWM_ZN:
 				vertex.rotateXZBy(-90); break;
 		}
+		vertex += offset;
 	}
 	drawQuad(vertices);
 }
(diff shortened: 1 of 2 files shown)


paramat:
The chance of very large firelike, signlike or torchlike nodes being used close to each other is quite small. Those drawtypes are quite specialised and not used much.
For those drawtypes this is clever but unnecessary, overkill, adds complexity and reduces meshgen performance a little too.

:-1: for PR as it is. We should only add stuff like this when very much needed.
So far no one apart from yourself has complained about those drawtypes z-fighting, and your issue was due to a test node, not a node you actually needed for a mod.
I think you are overdoing things, please remember simplicity.

However, the new approach to plantlike has potential, the fixing of vertical z-fighting is useful. If you make this PR only for plantlike i might approve this.
Plantlike is where z fighting is noticed a lot, so fixing that is justified.

src/client/content_mapblock.cpp
```diff
@@ -969,35 +992,42 @@ void MapblockMeshGenerator::drawPlantlike()
 		break;
 	}
 
+	int o;
+	if (p.Y % 2 == 0) {
+		o = 1;
+	} else {
+		o = -1;
+	}
```
paramat: Surely variation according to X and Z is required too to avoid side-by-side z-fighting, because the angles have returned to 45 degrees?

Wuzzy2:
:-(
Minetest is now at a point at which you have to justify bugfixes …

I wonder how much the performance is *actually* affected. Any ideas on how to optimize performance?

paramat:
Nuuu ... that is an unreasonable way to portray this. It is not really a bug, more an uncommon visual glitch. If a particular visual glitch very rarely occurs then it may not be worth the extra code complexity, maintenance, review workload and slightly lower performance.
In the issue you did agree this was low priority for firelike ;) So i am a little bemused.

Here is what i am currently thinking ...

Torchlike and signlike:

These are usually placed manually by players, so they would notice z-fighting and adjust the placement.
It seems quite rare that players will want to place scaled up nodes of these drawtypes close together such that they overlap. Torchlike clips into its supporting node if scaled up. Signlike i can understand being used scaled up, but how many players would then want these overlapping?
Both are rarely used for mapgen decorations.

Firelike:

This is automatically placed, not manually.
Rarely used for anything other than fire because the node alters its appearence depending on surrounding nodes in a way only really useful for fire.
Rarely scaled up because it looks so weird that way.
Used for fire, even if scaled up, the z fighting just adds flicker which is not really an issue for fire, and maybe even looks good.

Plantlike:

This is very different from the others.
Scaling up plantlike is fairly common.
Very often it is automatically and randomly placed as mapgen decorations, and so placed close to each other, so scaled up nodes will cause a vast amount of z-fighting nodes.
This is why avoiding z-fighting has been done for plantlike and nothing else. It was done from actual need instead of imagining a problem that has not arisen, which is how we should approach issues like this.
I support the plantlike part of this.

The fact that no-one (as far as i know, and i spend a lot of time reviewing and managing issues) has complained about torchlike, signlike and firelike z-fighting (for an actual mod usage, not a test node) seems to support my reasoning.

NathanSalapat:
Everything works as expected.

Wuzzy2:
I'm waiting for a 2nd core dev opinion before I change anything.

SmallJoker:
Rotation calculations alone will perform the same as before. It does not matter how many degrees you rotate the face - the function calls are always the same.

However, the modulo (`%`) operator is very slow and mesh generating is important to be fast as possible. In place of module, I'd prefer to use bitwise AND wherever possible (powers of 2) to not rely on compiler optimizations here.
`x % 4` -> `x & 3`, `x % 2` -> `x & 1`

hecktest:
I use diagonal plantlikes with sqrt(2) scale so that they fill their bounding square. They do not z-fight, and any random offsets will create gaps in the geometry where there were previously none.
objection

I'll repeat here the same thing I repeat every time this comes up; opengl allows you to bias the depth without moving the actual object physically.

SmallJoker:
> I'll repeat here the same thing I repeat every time this comes up; opengl allows you to bias the depth without moving the actual object physically.

Do you have a code snippet by any chance? That would be a welcome alternative to solve the Z-fighting problem. PRs are welcome.

Wuzzy2:
Lol, so this entire PR is garbage.

Wuzzy2:
Well, except for the plantlike rotation thingie, as @paramat suggested. I will destroy everything in this PR except the plantlike tweak.

Yeah, to be honest, I was not really happy with the world position offset as well, it felt pretty hacky to me, too. I just wanted to see what the reactions will be. @hecktest showed me the light! :D

Frankly, I'm not really experienced in 3D so far, I need to learn/practice more.

Wuzzy2:
This PR has been replaced by #9302.

paramat:
> I'll repeat here the same thing I repeat every time this comes up; opengl allows you to bias the depth without moving the actual object physically.

I would much prefer that too, so will not approve #9302 yet, even though it looks ok.
Then the 'degrotate' feature will also be able to place plantlike at exact circle divisions.

PR #9332, opened 2020-01-21
Formspec: Don't start a button click when the pointer isn't on top

### LINKED ISSUE #9330: Formspec elements can be pressed without clicking on them.
##### Minetest version
```
Minetest 5.2.0-0877587-0877587-dirty
```

##### OS / Hardware
Operating system:  Windows 10
CPU: Intel Core i5 x64

##### Summary
In any formspec element that has focus, clicking outside the formspec form will activate the thing with focus, and releasing the mouse button will let go of the element without activating it (as though you pressed on a button and then released outside of it)

##### Steps to reproduce
Open Minetest, make an element have focus (like a button by clicking and releasing outside the button) and then click the mouse again outside of the gray form border not on another element.  The button will be pressed when the mouse if pressed

I knew that so many formspec PRs would introduce some weird bugs.

Df458:
Partially resolves #9330 by preventing buttons from entering pressed state from a click event if the mouse is somewhere else. The GUIButton code already checks if the mouse is over it on release, so I've simply added a second check on press.

## Limitations
As mentioned above, this fix is only for buttons. I expect that any formspec element receiving mouse clicks can be affected by this (I can confirm that this affects `dropdown` elements as well, which appear to be a builtin Irrlicht gui element), and honestly the ideal would be to fix the issue at the source... However, I get the impression that this might be an Irrlicht bug so workarounds might be the best we can do.

## Testing
To test, just open a formspec and click any button once to make it active. After that, click outside of the formspec window and remark that it doesn't get pressed. I've taken the liberty of testing how this interacts with buttons clipping with the window, this works on my end but you should probably test it as well.

### DIFF
--- a/src/gui/guiButton.cpp
+++ b/src/gui/guiButton.cpp
@@ -203,8 +203,12 @@ bool GUIButton::OnEvent(const SEvent& event)
 	case EET_MOUSE_INPUT_EVENT:
 		if (event.MouseInput.Event == EMIE_LMOUSE_PRESSED_DOWN)
 		{
-			if (!IsPushButton)
+			// Sometimes formspec elements can receive mouse events when the
+			// mouse is outside of the formspec. Thus, we test the position here.
+			if ( !IsPushButton && AbsoluteClippingRect.isPointInside(
+						core::position2d<s32>(event.MouseInput.X, event.MouseInput.Y ))) {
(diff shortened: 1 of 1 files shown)


nerzhul:
i understand that point, i don't want to clic on button if i release outside of button :) approved

PR #9350, opened 2020-02-01
Disallow the unpack crash

HybridDog:
Without this fix here or in the mesecons mod, players on servers with mesecons can crash the server, which might lead to data loss.
luac

### DIFF
--- a/builtin/common/strict.lua
+++ b/builtin/common/strict.lua
@@ -55,3 +55,10 @@ end
 
 setmetatable(_G, meta)
 
+--------------------------------------------------------------------------------
+-- Workaround for bug https://www.lua.org/bugs.html#5.2.3-1
+local actual_unpack = unpack
+function unpack(t, a, b)
+	assert(not b or b < 2^30)
+	return actual_unpack(t, a, b)
+end


sfan5:
This works in LuaJIT too, so patching our built-in Lua with the fix would not suffice.

However, isn't it the sandbox's job to guard against implementation bugs?

HybridDog:
> patching our built-in Lua with the fix would not suffice.

This workaround overrides the unpack function, so it should work with all lua versions.

> However, isn't it the sandbox's job to guard against implementation bugs?

This bug causes a crash without an error message, so it is difficult to find if it happens in a mod.
The luacontroller does not have a proper sandbox; I think it is also possible to cause a crash (with traceback) with very long strings.

sfan5:
> This workaround overrides the unpack function, so it should work with all lua versions.

Sure. That might have been a little unclear, I was just thinking what if we applied the patch on lua.org to the Lua we ship in `lib/lua/`.

thomasrudin:
> Without this fix here or in the mesecons mod, players on creative servers with mesecons can crash the server, which might lead to data loss.

Why creative? That can be executed on survival-servers with mesecons too, right?

HybridDog:
> Why creative? That can be executed on survival-servers with mesecons too, right?

Of course it can. I'm not sure why I wrote that and fixed it now, thank you.

thomasrudin:
Can't this be fixed with an entry/override in the LuaC's environment: https://github.com/minetest-mods/mesecons/blob/master/mesecons_luacontroller/init.lua#L446 ?

HybridDog:
This only fixes it for the luacontroller: https://github.com/HybridDog/mesecons-pack/commit/130b5508f5271639f3e6041dc1dfedb8674172b4

There exist other mods which allow players to execute lua code, and the bug causes a memory access error, so I think it should be fixed in minetest.

nerzhul:
we may fix our lua embedded instead, if possible no ?

SmallJoker:
> we may fix our lua embedded instead, if possible no ?

This won't help if the system Lua installation is used.
Meanwhile Minetest has a mod security, which already provides basic protection. It's not great but if the root of the problem won't be fixed, it should be done in the next layer (Minetest API). So to say -  to follow the Linux kernel's idea with their Spectre/Meltdown workarounds.

HybridDog:
Applying the bugfix to LuaJIT should be straightforward.
The patch and the bug is described there: https://www.lua.org/bugs.html#5.2.3-1
I assume that LuaJIT's unpack is defined at [lib_base.c#L221-L240](https://github.com/LuaJIT/LuaJIT/blob/9143e86498436892cb4316550be4d45b68a61224/src/lib_base.c#L221-L240) and [minilua.c#L6365-L6378](https://github.com/LuaJIT/LuaJIT/blob/9143e86498436892cb4316550be4d45b68a61224/src/host/minilua.c#L6365-L6378).
The bug is explained at an official list of Lua bugs, which the LuaJIT developers are probably aware of, so maybe there is a reason that the patch has not been applied to LuaJIT yet?

Server owners may use an outdated LuaJIT, so a bugfix in the official LuaJIT may not affect those server owners.

sfan5:
> so maybe there is a reason that the patch has not been applied to LuaJIT yet?

Yes, the reason is that LuaJIT upstream development is pretty dead.

HybridDog:
https://github.com/LuaJIT/LuaJIT/pull/574

SmallJoker:
Fixed upstream. Can this be closed now?

Desour:
This was never fixed in our bundled PUC lua, right? And system-wide PUC lua (maybe also old luajit versions in distributions that have blood-less-edge packages) can possibly also segfault.
As we don't want mods to cause segfaults, shouldn't we fix this somewhere, ie. in builtin as this PR would've done?

PR #9374, opened 2020-02-08
Basic model shading

dcbrwn:
Adds bare bones shading for meshes. Meshes are lit with constant light vector.

screenshot_20200208_230326
screenshot_20200208_230343
screenshot_20200208_230407
screenshot_20200208_230433
screenshot_20200208_230455
screenshot_20200208_230723

## To do

Ready for Review.

## How to test

You'll need some mod with models like mobs or trains, for conveniences sake.


dcbrwn:
I tried to make it as less invasive as possible. Code is based on https://github.com/minetest/minetest/pull/9272, with only exception that instead of ambient, I use emissive color to pass light level. Looks like there is a possibility to set ambient color via scripts somehow, so I decided to save it.

benrob0329:
Does this happen to fix #8780 ?

dcbrwn:
@benrob0329 sadly, no. Let's try to not pile up features and fixes into one beast PR. One step at a time. Given that models weren't shaded at al, this is a huge leap already.

client/shaders/model_shader/opengl_fragment.glsl
```diff
@@ -0,0 +1,133 @@
...
+#else
+	color = base.rgb;
+#endif
+
+	vec4 col = vec4(color.rgb, base.a);
+	float diffuseIntensity = dot(vNormal, vec3(0.0, 1.0, 0.2));
+	vec3 diffuseColor = vec3(1.0) * clamp(diffuseIntensity, -0.3, 1.0);
+	vec3 lightColor = emissiveColor.rgb;
+
+	col.rgb *= lightColor * 0.8 + (lightColor + 0.1) * diffuseColor * 0.3;
```
HybridDog: Directional ambient lighting could give better results: https://github.com/HybridDog/minetest/blob/19d7a6c8d7c5a287c6fedbefe4f7d9828aaa497c/client/shaders/nodes_shader/opengl_fragment.glsl#L85-L107

Diffuse lighting with a single light direction has the disadvantage that there are always at least three faces on the cube with the same light strength, unless you do it incorrectly (here you allow negative diffuseIntensity values).
Directional ambient light is a bit more realistic than constant ambient light combined with diffuse light with constant light direction because directional ambient light approximately incorporates indirect lighting, which is similar to Physically Based Rendering I think.

src/client/clientenvironment.cpp
```diff
@@ -47,6 +102,8 @@ ClientEnvironment::ClientEnvironment(ClientMap *map,
 	m_texturesource(texturesource),
 	m_client(client)
 {
+	auto *shdrsrc = (IWritableShaderSource *)m_client->getShaderSource();
```
sfan5: perhaps the client method should be changed to return a writable one then

src/client/content_cao.cpp
```diff
@@ -563,6 +566,23 @@ void GenericCAO::removeFromScene(bool permanent)
...
+void GenericCAO::setSceneNodeMaterial(video::E_MATERIAL_TYPE material_type)
+{
+	scene::ISceneNode *node = getSceneNode();
+
+	node->setMaterialFlag(video::EMF_LIGHTING, false);
+	node->setMaterialFlag(video::EMF_BILINEAR_FILTER, false);
+	node->setMaterialFlag(video::EMF_FOG_ENABLE, true);
+
+	if (m_enable_shaders) {
+		node->setMaterialType(m_material_type);
```
sfan5: what is `m_material_type` and why is it different (is it?) from `material_type`?

src/client/shader.h
```diff
@@ -67,6 +67,8 @@ class IShaderConstantSetter {
 	virtual ~IShaderConstantSetter() = default;
 	virtual void onSetConstants(video::IMaterialRendererServices *services,
 			bool is_highlevel) = 0;
+	virtual void onSetMaterial(const video::SMaterial& material)
+	{ }
```
sfan5: `= 0`?

dcbrwn: This will force GameGlobalShaderConstantSetter and MainShaderConstantSetter to implement empty onSetMaterial callbacks. If that's ok I'll make the change

src/client/content_cao.cpp
dcbrwn: I renamed parameter names, so now intent should be clear even w/o comments

src/client/clientenvironment.cpp
dcbrwn: done

client/shaders/model_shader/opengl_fragment.glsl
dcbrwn: I'll try it now. But it looks like a lot of operations only to make other faces to look brighter. Maybe with a few more dot products (i.e. light sources) we can achieve the same effect?

dcbrwn: изображение

Here. With just one additional "light" source it doesn't look so flat on the sides. And it's almost as cheap as dirt:

```GLSL

	// Top "light"
	float diffuseIntensity = dot(vNormal, normalize(vec3(0.0, 1.0, 0.0)));
	// Front "light"
	diffuseIntensity += dot(vNormal, normalize(vec3(0.0, 0.2, 0.7))) * 0.5;

	// No more negative "light"
	diffuseIntensity = clamp(diffuseIntensity, 0.0, 1.0);

	col.rgb *= lightColor * 0.7 + (lightColor + 0.1) * diffuseIntensity * 0.3;
```

src/client/shader.h
sfan5: Nah, that's okay as is.

src/client/content_cao.cpp
```diff
@@ -580,6 +602,13 @@ void GenericCAO::addToScene(ITextureSource *tsrc)
 	video::E_MATERIAL_TYPE material_type = (m_prop.use_texture_alpha) ?
 		video::EMT_TRANSPARENT_ALPHA_CHANNEL : video::EMT_TRANSPARENT_ALPHA_CHANNEL_REF;
 
+	IShaderSource *shdrsrc = m_client->getShaderSource();
+
+	if (m_enable_shaders) {
+		u32 shader_id = shdrsrc->getShader("model_shader", TILE_MATERIAL_BASIC, NDT_NORMAL);
+		m_material_type = shdrsrc->getShaderInfo(shader_id).material;
```
sfan5: is it correct that `m_material_type` is only for shaders? it should be renamed too then

sfan5:
Your code currently doesn't [build](https://travis-ci.org/minetest/minetest/jobs/647988725):
<pre>
In file included from /home/travis/build/minetest/minetest/src/client/clientenvironment.cpp:36:
/home/travis/build/minetest/minetest/src/client/content_cao.h:230:19: error: 
      extra qualification on member 'setSceneNodeMaterial'
        void GenericCAO::setSceneNodeMaterial(
             ~~~~~~~~~~~~^
</pre>

src/client/content_cao.cpp
dcbrwn: also in the original PR, model_shader was applied not only to meshes, but to all types of props in addToScene. Not sure if that's correct. Maybe you have any advice on this?

client/shaders/model_shader/opengl_fragment.glsl
HybridDog: > But it looks like a lot of operations only to make other faces to look brighter.

It's because that code supports bumpmapping, where the normal vectors from the normal map need to be rotated to the surface face rotation. (normalmaps are in tangent space)
Without bumpmapping support, the code should be fairly simple, and can be put in the vertex shader:
```glsl
// untested
float directional_ambient(vec3 normal)
{
	vec3 v = normal * normal;
	if (normal.y < 0)
		return dot(v, vec3(0.670820f, 0.447213f, 0.836660f));
	return dot(v, vec3(0.670820f, 1.000000f, 0.836660f));
}
```
Original code, which is used for the nodes: https://github.com/minetest/minetest/blob/b14aa1c84714a4800d214768ff7868a7bb76f7ae/src/client/mesh.cpp#L45-L63

dcbrwn: Ok, that model better follows overall shading model. Done.

src/client/content_cao.cpp
sfan5: Apply model_shader to all kinds of active objects sounds correct to me.

benrob0329:
Testing with my game:
Screenshot from 2020-02-10 18-24-33

The "Staff" are lit about how I'd expect, but the player sprite is completely black.

client/shaders/model_shader/opengl_fragment.glsl
paramat: "Use shading that matches light distribution on world nodes"
Yes, this seems a good idea.

src/client/content_cao.cpp
dcbrwn: Probably I should rename it to "object_shader" then.

dcbrwn:
@benrob0329 should be ok now

src/client/content_cao.cpp
```diff
@@ -563,6 +566,21 @@ void GenericCAO::removeFromScene(bool permanent)
...
+	node->setMaterialFlag(video::EMF_LIGHTING, false);
+	node->setMaterialFlag(video::EMF_BILINEAR_FILTER, false);
+	node->setMaterialFlag(video::EMF_FOG_ENABLE, true);
+	node->setMaterialType(m_material_type);
+
+	if (m_enable_shaders) {
+		node->setMaterialFlag(video::EMF_GOURAUD_SHADING, false);
+		node->setMaterialFlag(video::EMF_NORMALIZE_NORMALS, true);
+	}
+}
```
sfan5: For code quality, it'd be nice to turn this into a lambda inside `addToScene`.
Like this:
```cpp
auto setSceneNodeMaterial = [=] (scene::ISceneNode *node) {
	node->setMaterialFlag(video::EMF_LIGHTING, false);
	node->setMaterialFlag(video::EMF_BILINEAR_FILTER, false);
	node->setMaterialFlag(video::EMF_FOG_ENABLE, true);
	node->setMaterialType(m_material_type);

	if (m_enable_shaders) {
		node->setMaterialFlag(video::EMF_GOURAUD_SHADING, false);
		node->setMaterialFlag(video::EMF_NORMALIZE_NORMALS, true);
	}
};
```

dcbrwn: Done

Df458:
Confirmed working on my end as well.

One quick note to other testers: On my end, I had to temporarily "uninstall" minetest (by moving the directory elsewhere) to get the right shaders loading. If you don't see a change, this may be why. Another possibility is broken normals on the model, which should hopefully be fixed by modders once this feature is out.

On my end, I tested Repixture's mobs and they seemed to be lit correctly:
shading

The game already tried to emulate shading, but mobs had a light underbelly because the top/bottom of the body uses the same part of the texture. With this patch, the model is clearly shaded correctly.

SmallJoker:
For me the shaders applied after changing the settings and restarting Minetest: (PR / master)

grafik

Also starting with no `minetest.conf` did not change this. The mobs are not lit up correctly in OpenGL 3.3.

dcbrwn:
@SmallJoker make sure models have correct normals. It's a common issue with mods.

Df458:
It would be good to name/link the mod in question as well so that folks with other hardware configurations can test.

Df458:
I had the same issue testing with the turret from [Simple Shooter](https://content.minetest.net/packages/stu/shooter/). I loaded the model into Blender, recalculated the normals, then re-exported to the same format and this fixed the issue. If you have a copy of Blender, this is something you can test on your end as well.

SmallJoker:
I used [mob-horse](https://notabug.org/TenPlus1/mob_horse) which depends on [mobs_redo](https://notabug.org/TenPlus1/mobs_redo/). In order to use this mod based on current master, it needs the following fixes:

```diff
diff --git a/api.lua b/api.lua
index 2ba4395..adf1673 100644
--- a/api.lua
+++ b/api.lua
@@ -214,7 +214,9 @@ end
 
 -- move mob in facing direction
 function mob_class:set_velocity(v)
-
+	if not self.object:get_velocity() then
+		return
+	end
 	local c_x, c_y = 0, 0
 
 	-- can mob be pushed, if so calculate direction
@@ -248,14 +250,18 @@ end
 function mob_class:get_velocity()
 
 	local v = self.object:get_velocity()
-
+	if not v then
+		return 0
+	end
 	return (v.x * v.x + v.z * v.z) ^ 0.5
 end
 
 
 -- set and return valid yaw
 function mob_class:set_yaw(yaw, delay)
-
+	if not self.object:get_yaw() then
+		return 0
+	end
 	if not yaw or yaw ~= yaw then
 		yaw = 0
 	end

```

I will test the Repixture game as soon I resolved the mapgen flags issues that arise upon world creation.

Df458:
I can repro this issue with the horses on my end... unfortunately, the model format is .x so I can't test re-exporting from Blender.

dcbrwn:
It's not a hardware bug. But the issue with horses mod. I'll look if I can do something about it in shaders. There might be an issue with normals length.
screenshot_20200216_175041

dcbrwn:
@SmallJoker should be ok now.

dcbrwn:
Build is failing because of network issues. Is there a way to retrigger it?

lhofhansl:
As I suggested (somewhere else, forget where) when I tried this, I found that many mods have models without normals and hence will be completely black (or white). I'm seeing this now.

Sigh. As with other improvements we did before we're just finding so many faulty mods.
As is, though, for many older mods (where nobody will ever make new models) the experience is gone.

dcbrwn:
@lhofhansl give an example mod. Best we can do is to check if normals have length. In that case it's easy to fallback to old shading. Otherwise It's up to modders to fix models (or _you_, yes _you_. It's opensource! :D). And it shouldn't be hard at all. Every 3d editing software I've seen had automatic normals calculation.

NathanSalapat:
Flipping normals isn't the hard part, it's finding a way to import the .x or .b3d files that keeps the animation and uv mapping.

HybridDog:
> Best we can do is to check if normals have length. In that case it's easy to fallback to old shading.

The model can only look broken if the normal is zero or has special float values such as nan or inf. Instead of falling back to old shading, maybe we can test if the normal is invalid during model loading, and if so set the normal to point up (0,1,0).

dcbrwn:
@HybridDog I wanted to check normals length in vertex shader, and if it's 0, instead of invoking directional_ambient, just set vIDiff to 1.0.

HybridDog:
You can do that too, but setting it when the model is loaded is probably more efficient.

SmallJoker:
New bugs: #9482 and #9481

dcbrwn:
@SmallJoker Is there any map or mod that allows to test all the rendering capabilities of minetest engine? That would be very useful

SmallJoker:
@dcbrwn I was able to reproduce this issue using the instructions from #9482.
Devtest: https://git.minetest.land/Wuzzy/devtest

hecktest:
#9985
Real sad that this got merged in the state it is. What happened to _no breaking changes_?

A lack of normals is not "faulty mods", it can be a choice. All art until now was made with the lack of shading in mind.

Forced vertex shading makes things look like printed and glued papercraft models; that is, rubbish. This must be optional.

dcbrwn:
@hecktest #9969 has fallback code for models w/o normals. smooth shading while looking like crap on its own, still magnitudes better than large polygon blobs that look brighter than everything on the screen. Try to concentrate on fixing problems instead of complaining about them.

hecktest:
@dcbrwn 
I've seen PRs rejected for much less, this is why I'm stumped that this got approved without much thinking or testing behind it. It somehow even found its way into a stable release in this state. The review for this PR has been very sloppy. The lesson to learn here is to **test harder**. Try every supported format, because some of them might behave differently.

>still magnitudes better than large polygon blobs

No, a lack of shading is a legitimate style choice. Knowing that the engine has no shading, an artist will shade the texture instead. Vertex shading is pretty much forbidden for certain styles because it will never look good. It looks like painted paper, look at how it ruins the highlights. Without customizable shaders and materials to finely control the look, it's usually better to have no shading at all.

unlit lit

A change this significant should always be optional.

dcbrwn:
In your example with a relatively high level of detail this is true. For low-poly models like animals mod, trans, furniture, mobs, etc. shading is pretty much required even when it's kind of backed into textures.

As for this PR, it was an attempt to push feature of shading to stable version, because it was long overdue. And the situation with mods only confirms that. Several people made an attempt of delivering this or something similar to this, but PRs got stuck because of different reasons. There will be transition period, and there will be people who are for and against this feature, but in essence minetest as an engine got better.

And btw, if you don't like the shading you have several options to disable it:
- set normals in your models to 0
- edit shader
That way you'll get tone-mapping at least.

As for "sloppiness" of testing, well at the point of writing, there wasn't tool as great as Wuzzy's devtest, and all mods I checked worked fine. And also the changes were once part of other PR, which also was tested by several people.

benrob0329:
~~Good thing Minetest is PEGI 7 or we'd have to care about the oversexualized anime girls~~

There was lots of time to come and mention your art style, and quite a bit of testing was done for this across multiple games. Yes it could have been better, but now `devtest` has tests specifically because of bugs that were missed here.

But to put some standard convention into play, most game engines shade models by default, and anything not shaded is in "fullbright" mode. Minetest has always shaded entities by default, just not in a way that was consistent with nodes (it's still not consistent with nodes, but its closer). **Shading entities like nodes is sane behavior and should be the default.** Anything other than that is a Feature Request.

PR #9390, opened 2020-02-11
Script API: Check that SAOs are still usable before attempting to use them

### LINKED ISSUE #8587: Segfault caused by chat sending scrollbar change
##### Issue type
 - Bug report
 
##### Minetest version
```
5.0.1
```

##### OS / Hardware
Operating system: Windows 10, Linux Mint
CPU: Intel something-or-other

##### Summary
If you have a formspec with a scrollbar, and you send the value using `minetest.chat_send_all` _without_ using `minetest.explode_scrollbar_event`, after closing the form, Minetest will give the following error and segfault: `WARNING[Server]: ScriptApiBase::objectrefGetOrCreate(): Pushing ObjectRef to removed/deactivated object, this is probably a bug.`

##### Steps to reproduce
Paste the following code in any file, enter Minetest, press space to open the form, move the scrollbar and notice the chat, and then close the form.

```lua
        if player then
                if controls.jump then
                        minetest.show_formspec("singleplayer", "formspec:test", 
					"size[10,1]"..
					"scrollbar[0,0;10,1;;name;0]"..
                        "")
                end
                player:set_look_vertical(math.rad(0))
        end
end)

minetest.register_on_player_receive_fields(function(player, formname, fields)
        if not player or not formname then
                return
        end
	minetest.chat_send_all(fields.name)
end)
```

### LINKED ISSUE #9387: Segfault on Server::SendMovePlayer
##### Minetest version
```
stable-5
c02f13d33f87e030e89a63767f5f96e03fedc6fe
```

##### OS / Hardware
Operating system: Linux, ubuntu (container)
CPU: x86_64

##### Summary
Medium-sized server with about 15 players average online.
Crashes every few days on `Server::SendMovePlayer` with a segfault.

Output from `gdb`:
```
# gdb minetestserver dump.Server.7 

[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `minetestserver --config /data/minetest.conf --world /data/world/ --quiet'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000561bcd155116 in Server::SendMovePlayer (this=0x7ffd25957260, peer_id=0) at /git/minetest/src/server.cpp:1796

(gdb) backtrace
#0  0x0000561bcd155116 in Server::SendMovePlayer (this=0x7ffd25957260, peer_id=0) at /git/minetest/src/server.cpp:1796
#1  0x0000561bcd09d30b in PlayerSAO::setPlayerYawAndSend (this=this@entry=0x7f057d0ab220, yaw=<optimized out>) at /git/minetest/src/serverenvironment.h:217
#2  0x0000561bccfdd29b in ObjectRef::l_set_look_horizontal (L=0x4088c378, L@entry=<error reading variable: value has been optimized out>) at /git/minetest/src/script/lua_api/l_object.cpp:1215
#3  0x0000561bccfa0fca in script_exception_wrapper (L=0x4088c378, f=<optimized out>) at /git/minetest/src/script/common/c_internal.cpp:38
#4  0x00007f0637818e97 in ?? ()
#5  0x000000024089b368 in ?? ()
#6  0x0000561bced3d360 in ?? ()
#7  0x0000001000000002 in ?? ()
#8  0x40c4c3104088c378 in ?? ()
#9  0x0000000000000000 in ?? ()


[Current thread is 1 (Thread 0x7f0626ffd700 (LWP 17))]
(gdb) display player
1: player = (RemotePlayer *) 0x0
(gdb) display sao
2: sao = <optimized out>
(gdb) 

```

Source-code:

https://github.com/minetest/minetest/blob/c02f13d33f87e030e89a63767f5f96e03fedc6fe/src/server.cpp#L1796

My guess: `m_env->getPlayer(peer_id)` returns an invalid pointer (somehow) and the `assert` statement does not fail because it is non-NULL.

##### Steps to reproduce
I couldn't reliably reproduce this, happens from time to time...

sfan5:
fixes #8587, fixes #9387

## To do

~~Q: Should this print a warning so the mod this happened in can be investigated?~~ (not necessary)

### DIFF
--- a/src/content_sao.cpp
+++ b/src/content_sao.cpp
@@ -864,7 +864,7 @@ PlayerSAO::PlayerSAO(ServerEnvironment *env_, RemotePlayer *player_, session_t p
 	m_peer_id(peer_id_),
 	m_is_singleplayer(is_singleplayer)
 {
-	assert(m_peer_id != 0);	// pre-condition
+	SANITY_CHECK(m_peer_id != PEER_ID_INEXISTENT);
 
 	m_prop.hp_max = PLAYER_MAX_HP_DEFAULT;
 	m_prop.breath_max = PLAYER_MAX_BREATH_DEFAULT;
@@ -1393,7 +1393,7 @@ bool PlayerSAO::setWieldedItem(const ItemStack &item)
 
 void PlayerSAO::disconnected()
 {
-	m_peer_id = 0;
+	m_peer_id = PEER_ID_INEXISTENT;
 	m_pending_removal = true;
 }
 
--- a/src/script/lua_api/l_object.cpp
+++ b/src/script/lua_api/l_object.cpp
@@ -60,6 +60,8 @@ LuaEntitySAO* ObjectRef::getluaobject(ObjectRef *ref)
 		return NULL;
 	if (obj->getType() != ACTIVEOBJECT_TYPE_LUAENTITY)
 		return NULL;
+	if (obj->isGone())
+		return NULL;
 	return (LuaEntitySAO*)obj;
 }
 
@@ -69,6 +71,8 @@ PlayerSAO* ObjectRef::getplayersao(ObjectRef *ref)
 	if (obj == NULL)
 		return NULL;
 	if (obj->getType() != ACTIVEOBJECT_TYPE_PLAYER)
+		return NULL;
+	if (obj->isGone())
 		return NULL;
 	return (PlayerSAO*)obj;
 }
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -1768,10 +1768,7 @@ void Server::SendTimeOfDay(session_t peer_id, u16 time, f32 time_speed)
 void Server::SendPlayerHP(session_t peer_id)
 {
 	PlayerSAO *playersao = getPlayerSAO(peer_id);
-	// In some rare case if the player is disconnected
-	// while Lua call l_punch, for example, this can be NULL
-	if (!playersao)
-		return;
+	assert(playersao);
 
 	SendHP(peer_id, playersao->getHP());
 	m_script->player_event(playersao,"health_changed");
@@ -2701,10 +2698,7 @@ void Server::sendDetachedInventories(session_t peer_id, bool incremental)
 void Server::DiePlayer(session_t peer_id, const PlayerHPChangeReason &reason)
 {
 	PlayerSAO *playersao = getPlayerSAO(peer_id);
-	// In some rare cases this can be NULL -- if the player is disconnected
-	// when a Lua function modifies l_punch, for example
-	if (!playersao)
-		return;
+	assert(playersao);
 
 	infostream << "Server::DiePlayer(): Player "
 			<< playersao->getPlayer()->getName()


thomasrudin:
I'm now getting crashes from nil-values returned from `self.object:get_velocity()` not sure if they are related.

I'm opening an issue if it is cause by this..

sfan5:
Hm, well this fix has the side effect that the ObjectRefs that were buggy before are now entirely non-functionality with all relevant methods returning `nil`.

It would be possible to work around this in the affected mods, but that would be incorrect.
Do the `Pushing ObjectRef to removed/deactivated object, this is probably a bug.` warnings still appear?

thomasrudin:
>  Do the Pushing ObjectRef to removed/deactivated object, this is probably a bug. warnings still appear?

No, not in that time-frame, the patch was only active for about 15 minutes..

sorcerykid:
> Hm, well this fix has the side effect that the ObjectRefs that were buggy before are now entirely non-functionality with all relevant methods returning `nil`.

I'm definitely glad to hear that there is now fully consistent behavior for all methods of a freed ObjectRef. Perhaps it might be worth mentioning that point in the documentation.

lhofhansl:
Hmm... I'm seeing this too with various mods that use mobs_redo. Perhaps we should undo this one (I agree it's correct behavior, yet, there are mods that are broken by this).

sfan5:
I believe the ultimately better outcome is to get mods fixed quickly right now instead of turning it into a soft warning and getting mods fixed _sometime somewhen_ (or never).

Also, has anyone noticed crashes with a mod other than `mobs_redo`?

thomasrudin:
> Also, has anyone noticed errors with a mod other than mobs_redo?

Not yet, no, but i'm guessing this will be a mine-field... :confounded: 

I fixed the `3d_armor` issues in my own fork (https://github.com/mt-mods/3d_armor/pull/2) because the original mod by stujones11 is not maintained anymore...

lhofhansl:
I'm also not 100% sure the change has the behavior we want.
The engine calls on_step, as it should. If "remove" is called inside that, why wouldn't we allow still accessing the object for this step. It is weird that a reference that was valid before suddenly turns into nil.

I haven't looked closely at this part of the code, is there a risk of accessing freed memory?

sorcerykid:
I have to wonder how this change broke the 3d_armor mod. I looked at the PR linked above and don't see any changes dealing with ObjectRefs, just a lot of renaming of variables. Perhaps I missed something.

sfan5:
> why wouldn't we allow still accessing the object for this step?

I get that it's more convenient here but why would we? The mod has just asked to delete the object, it should cease to exist.

> is there a risk of accessing freed memory?

There isn't and unlike the reported bug (#9387) for players I don't think using entities after their removal could crash anything either.

sorcerykid:
I can sort of see @lhofhansl's point. It might be more intuitive that calls to obj:remove() only add the object to a queue to be removed. But the object is not actually removed from the environment until control is returned back to the environment at the conclusion of the server step. (Personally, I thought that's how it already worked on the CPP side, but I suppose I'm recalling wrong).

Nevertheless, I must agree with @sfan5, that if an object is removed, then rightfully it should cease to exist at that point going forward. Mod developers should properly document their own API so that there is no risk of objects still being referenced after removal. Or at least they should provide some kind of callback mechanism so that other mods can be notified of a pending removal. These are mod-specific implementation details, and technically shouldn't be the responsibility of the engine.

lhofhansl:
> I get that it's more convenient here but why would we? The mod has just asked to delete the object, it should cease to exist.

Upon further reflection I disagree.
The mod did not set the reference to nil, it simply called a method on the object - indicating that the object should be removed from the set of active objects, so that no future callbacks are called. Which is precisely the reason why the engine has an isgone method.

lhofhansl:
> I get that it's more convenient here but why would we? The mod has just asked to delete the object, it should cease to exist.

Upon further reflection I disagree.
The mod did not set the reference to nil, it simply called a method on the object - indicating that the object should be removed from the set of active objects, so that no future callbacks are called. Which is precisely the reason why the engine has an isgone method.

sorcerykid:
If an object is removed from the set of active objects, then what type of object is it at that point? Its not a LuaEntitySAO nor is it a PlayerSAO. Perhaps it is a DeletedSAO? Shouldn't there be documentation on the API for this new class of DeletedSAOs?

sfan5:
> The mod did not set the reference to nil, it simply called a method on the object - indicating that the object should be removed from the set of active objects, so that no future callbacks are called.

Well not exactly. The mod has marked the object for removal, the earliest point at which that can happen is the current server step (after all step callbacks have run).
It's true that this PR artificially makes the object unusable despite it still existing.

> I vote for reverting... 

Do you wish to revert this for players and luaentities or just luaentities?
Opening up interaction with "deleted" players will open up at least one segfaulting code path in the server code up again, so that would have to dealt with in an other way.

> If an object is removed from the set of active objects, then what type of object is it at that point?

lhofhansl:
> Would you want this reverted for players and Luaentities or just Luaentities?

Just Luaentities, I think - assuming others agree obviously.

lhofhansl:
So should we revert the Luaentities part of this?

sfan5:
My vote is still no, I don't know the opinions of other coredevs.

lhofhansl:
My vote is still yes :)
(and document that soa.remove just removes the object from the set of active objects)

Perhaps some other core devs could weigh in.

SmallJoker:
There were already warnings when invalid/deleted objects were accessed. This PR "only" enforced the object lifespan which now became an issue in mods that relied on it.
The handling after this PR is exactly as intended, and mod authors should fix their mods. BUT to ease this process an API function, say `obj:is_valid()` must be provided to perform a proper check whether the entity is already gone (LuaEntity and Players).

lhofhansl:
Honestly I'd rather get rid of the warnings then.

The object is not deleted, it's still there, just not in the set of currently active object anymore. That's easy to explain, easy to understand, and easy to use in mods.

We do the same in core. The object is still there, references are not set to null (not possible in C++ anyway) and there's an is_gone method.

Can't do obj:is_valid() if you pull the rug from underneath by returning a null-reference, right?

PR #9392, opened 2020-02-11
guiHyperText: Fix blinky cursor on link hover

### LINKED ISSUE #9385: Formspecs / Hypertext : Action tags hover cursor changes back to normal
##### Minetest version
```
Minetest 5.2.0-dev-b14aa1c8 (Linux)
Using Irrlicht 1.8.4
```
##### Summary
When hovering an action tag, the cursor is changed to gui::ECI_HAND (normal behavior) but changes back to gui::ECI_NORMAL after a short while.

##### Steps to reproduce
Create a simple formspec like this one :
```
minetest.show_formspec(name, "test", "size[4.5,2]hypertext[1,1;1.5,1;xxx;<action name=foo>test</action>]")
```
Move cursor over "test" text and stop any move. Cursor should keep being display as a hand but changes back to normal.

##### Fixing
This is due to hover cursor management by hypertext conflicting with cursor management in 4c8af5492ba9c3b9671f5213e81d28a7b856e134

I'm wondering about best way to solve it : 
- Create an exception in GUIFormSpecMenu::drawMenu() for Hypertext elements;
- Have a special value in FieldSpec::fcursor_icon which says "leave element manage cursor";
- Other solution ?

Another question : should Hypertext use FieldSpec::fcursor_icon as default cursor instread of gui::ECI_NORMAL (and still use gui::ECI_HAND when hovering) ?

SmallJoker:
Fixes #9385

## To do

This PR is Ready for Review.

## How to test

See #9385, alternatively https://paste.ubuntu.com/p/Nf6XfXnSZX/ (from #8680)

### DIFF
--- a/src/gui/guiFormSpecMenu.cpp
+++ b/src/gui/guiFormSpecMenu.cpp
@@ -4181,9 +4181,9 @@ bool GUIFormSpecMenu::OnEvent(const SEvent& event)
 				(event.GUIEvent.EventType == gui::EGET_CHECKBOX_CHANGED) ||
 				(event.GUIEvent.EventType == gui::EGET_COMBO_BOX_CHANGED) ||
 				(event.GUIEvent.EventType == gui::EGET_SCROLL_BAR_CHANGED)) {
-			unsigned int btn_id = event.GUIEvent.Caller->getID();
+			s32 caller_id = event.GUIEvent.Caller->getID();
 
-			if (btn_id == 257) {
+			if (caller_id == 257) {
 				if (m_allowclose) {
 					acceptInput(quit_mode_accept);
 					quitMenu();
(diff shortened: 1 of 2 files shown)


kilbith:
Btw that’s a good opportunity to rehabilitate the old coordinates system for this element.

SmallJoker:
@kilbith Which element's code should I take for the HyperText coordinate calculations? Alternatively, please provide a code snippet that results in the behaviour you'd like to have.

v-rob:
I would guess textarea because that used to be the text display element.  Either that, or textlists.

kilbith:
Tested and works :+1:

sfan5 (approved):
Tested and works.

SmallJoker:
766d160

Issue #9383, opened 2020-02-10
Hungarian: Some special characters don't show up


An0n3m0us:
##### Minetest version
```
Minetest 5.2.0-dev-b14aa1c-dirty
```
Works fine before: `Minetest 5.1.1-dev-5e4739b-dirty`

##### OS / Hardware
Operating system: Linux - Ubuntu 16.04

##### Summary
I've been meaning to mention this issue for a while;

With Hungarian language selected, some special characters don't show up in Minetest 5.1.1+;
for example: Ú and ö

Affects mod locales too.

##### Steps to reproduce
1. In Minetest 5.1.1, change the language to Hungarian (`hu`)
2. The "New" button for creating a new world will read "j" when it is supposed to read "Új"

An0n3m0us:
Seems to originate from here:
https://github.com/minetest/minetest/commit/b83ca324f95641d41dcff5380a7b608ec820d610#diff-fd4749a1026ea4f7b35c2c2ce943c841L876

I changed it back from 'yes' to the Hungarian translation and it fixes the issue.

SmallJoker:
The bundled fonts do not contain all language characters. Consider using droid sans (fallback). For this particular situation there's the "needs_fallback_font" translation string to chose it automatically.

An0n3m0us:
It was using `DroidSansFallbackFull.ttf`, yet the special characters are still missing.

sfan5:
Having worked with "DroidSansFallbackFull" in the context of another project I can say that it's actually missing several quite basic glyphs that will be present in any other normal font (e.g. Arimo).

What needs to be done here is to turn off the fallback font for Hungarian (and possibly other languages which don't need it).

An0n3m0us:
Can one simply set `needs_fallback_font` ([line 877](https://github.com/minetest/minetest/blob/master/po/hu/minetest.po#L877)) to `no` to solve the issue?

sfan5:
Yes, that should work.

SmallJoker:
To be kept open until the translations are updated - or someone confirms that it works with the current Hungarian translation from Weblate.

An0n3m0us:
Can confirm that it works; I downloaded the `.mo` file and placed it in the `hu/LC_MESSAGES` folder, then checked it in Minetest.

PR #9400, opened 2020-02-13
Hungarian: Disable fallback font

An0n3m0us:
Ref: #9383 

Disables fallback font for Hungarian as special characters are not showing up.

## How to test
1. Change the language to Hungarian (`hu`)
2. The "New" button for creating a new world will now correctly read "Új" instead of missing the Ú character.


sfan5:
This should also be changed in Weblate so it isn't undone on the next translation import.

An0n3m0us:
Done: https://hosted.weblate.org/translate/minetest/minetest/hu/?checksum=a52375068599ce23

SmallJoker:
Issue resolved.

PR #9405, opened 2020-02-16
Add reconnect button on error: Too many players

Lejo1:
Allows the player to easily retry connecting instead of typing the password in all the time.


SmallJoker:
So rather than delaying the join by a few seconds, you're offering them to continuously spam "Join the server" where the chance of getting in at the end stays the same. Both ways are awful.

PR #9411, opened 2020-02-18
Add starting frame to `animated_image`

v-rob:
This PR does the following to `animated_image`:
- Allows the specification of a starting frame (optional parameter).
- Setting the frame duration to 0 stops the animation.
- On any event, e.g. a button press, the current frame index is returned using the `name` parameter.
- The syntax has been changed: `animated_image[<X>,<Y>;<W>,<H>;<name>;<texture name>,<frame count>,<frame duration>;<frame start>]`.  Notice the two new parameters and the change from a colon to a comma.

This PR is useful because images can be stopped and restarted at any specific point.  For instance, a slide show that automatically goes through the images slowly, but buttons can be used to go back to previous images.  Or a primitive soundless video with pause, play, and a scrollbar for video position.  This can also be used to place specific images from spritesheets in multiple places on the formspec.  So, many different applications.

## To do

This PR is Ready for Review.

## How to test

Use the updated Minimal formspec.  The animated images have been changed to numbers to show which frame is currently shown better.  The `Current Number` buttons will send the index of the animated image to the right of them to the chat.


Df458:
@kilbith Thoughts re: the syntax?

Personally, I'm indifferent to the syntax changes and the feature seems reasonable enough... It doesn't seem to have many use-cases at the moment, but I assume this will be much more useful when your other proposals start to roll in.

sorcerykid:
Dumb question, but wouldn't the colon make more sense as a semi-colon to be consistent with other elements where the texture name is always a distinct parameter?

v-rob:
Not a dumb question.  The answer is probably, but `GUIAnimatedImage` parses that itself, so it acts as one parameter to `GUIFormSpecMenu`.  I don't think that it's a particularly good design to have parsing in the GUIElement, but that's the way it was, and I won't change it unless someone requests it.

sorcerykid:
Ah, okay that sounds like a solid rationale for the comma. I hadn't thought of the internal workings. Thanks for clarifying.

kilbith:
I am in favor of commas but `<frame start>` should be put after a semicolon to be consistent with the other elements like `dropdown` or `table` which have the `<selected idx>` after a semicolon.

v-rob:
Ok, I changed the syntax, but don't go testing it yet.  I'm going to add an event where, when any event fires, the animated image will tell you what frame is currently shown.  This will make it much easier to pause at a specific point.

v-rob:
OK, this PR is now Ready for Review.

kilbith:
IIRC, if `<name>` is specified, fields are sent every `<frame_duration>` time? :-1: This can be abused.

v-rob:
No, not every `frame duration` time.  I agree, that would be bad.  No; it's only when _other_ events happen.  For instance, if you press a button, _then_ the current frame for each image with a name is returned.

rubenwardy:
Would it make sense for the format to be `animated_image[<X>,<Y>;<W>,<H>;<name>;<texture name>;<frame count>;<frame duration>;<frame start>]`, which could make it easier to specify texture modifiers? Is there anywhere where this format is used/similar to?

rubenwardy:
Closest appears to be these:

```
* `[cracko:<tiles>:<frames>:<start>`
* `[verticalframe:<frames>:<start>`
```

rubenwardy:
I guess the benefit of commas is that you can rationalise it at the specification of an entity which is an animated texture, and reuse it in other places

v-rob:
I personally have no preference on the syntax.  Now that I've moved the parsing code to `GUIFormSpecMenu`, either one is as easy as the other.  I think the idea for commas is that each parameter is necessary to define the image, e.g. a video is meaningless without the number of frames it has.

v-rob:
Two days; I assume there are no objections.  I've changed the commas to semicolons

games/minimal/mods/test/formspec.lua
```diff
@@ -181,11 +181,19 @@ local pages = {
...
-			animated_image[3,0.5;5,2;test_animation.png:4,100]
-			animated_image[3,2.75;5,2;test_animation.jpg:4,100]
+			animated_image[0.5,0.5;1,1;;test_animation.png;4;100]
+			animated_image[0.5,1.75;1,1;;test_animation.jpg;4;100]
+			animated_image[1.75,0.5;1,1;;test_animation.png;100;100]
+			animated_image[3,0.5;1,1;ani_image;test_animation.png;4;1000]
+			button[4.25,0.5;1,1;;Current
+Number]
+			animated_image[3,1.75;1,1;;test_animation.png;4;1000;2]
+			button[4.25,1.75;1,1;;Current
```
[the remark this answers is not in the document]
v-rob: Not sure I understand.  That button sends the current frame of the image next to it.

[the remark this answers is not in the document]
rubenwardy: There's no button name. How does it know which image to send?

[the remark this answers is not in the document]
v-rob: Oh, I see.  Probably me forgetting to implement that.

doc/lua_api.txt
```diff
@@ -2133,14 +2133,17 @@ Elements
...
+### `animated_image[<X>,<Y>;<W>,<H>;<name>;<texture name>;<frame count>;<frame duration>;<frame start>]`
 
 * Show an animated image. The image is drawn like a "vertical_frames" tile
-    animation (See Tile animation definition), but uses a frame count/duration
+    animation (See [Tile animation definition]), but uses a frame count/duration
     for simplicity
-* `<texture name>` is the image to use
-* `<frame count>` is the number of frames animating the image
-* `<frame duration>` is in milliseconds
+* `<name>`: The event name. The event is the index of the current frame, and it is fired
```
rubenwardy: for consistency, don't include the `<>` when mentioning a formspec argument

games/minimal/mods/test/formspec.lua
```diff
@@ -26,7 +26,7 @@ local clip_fs = [[
 	scrollbar[0,9;3,0.8;horizontal;x9;3]
 	tablecolumns[text;text]
 	table[0,10;3,1;x10;one,two,three,four;1]
-	animated_image[0,11;3,1;test_animation.png:4,100]
+	animated_image[0,11;3,1;;test_animation.png;4;100]
```
rubenwardy:
```suggestion
	animated_image[-0.5,11;4.5,1;;test_animation.png;4;100]
```

so that it actually appears on the page

src/gui/guiAnimatedImage.cpp
```diff
@@ -81,3 +63,11 @@ void GUIAnimatedImage::draw()
...
 	}
 }
+
+
+void GUIAnimatedImage::setFrameIndex(s32 frame)
+{
+	s32 idx = std::max(frame, 0);
+	if (idx > 0 && idx < m_frame_count)
+		m_frame_idx = idx;
+}
```
rubenwardy:
```suggestion
}

```

src/gui/guiFormSpecMenu.cpp
```diff
@@ -3512,14 +3521,13 @@ void GUIFormSpecMenu::acceptInput(FormspecQuitMode quitmode=quit_mode_no)
 					if (table) {
 						fields[name] = table->checkEvent();
 					}
-				}
-				else if(s.ftype == f_DropDown) {
-					// no dynamic cast possible due to some distributions shipped
-					// without rtti support in irrlicht
+				// No dynamic cast possible for any element due to some
```
rubenwardy: this was better in its previous location, even if duplicated, because it was next to the relevant code

src/gui/guiFormSpecMenu.cpp
```diff
@@ -784,16 +784,20 @@ void GUIFormSpecMenu::parseAnimatedImage(parserData *data, const std::string &el
...
 	std::vector<std::string> parts = split(element, ';');
 
-	if (parts.size() != 3 &&
-			!(parts.size() > 3 && m_formspec_version > FORMSPEC_API_VERSION)) {
-		errorstream << "Invalid animated image element(" << parts.size()
-				<< "): '" << element << "'"  << std::endl;
+	if (parts.size() != 6 && parts.size() != 7 &&
+		!(parts.size() > 4 && m_formspec_version > FORMSPEC_API_VERSION))
+	{
+		warningstream << "Invalid animated_image element(" << parts.size()
```
rubenwardy: this should be errorstream

rubenwardy:
looks good apart from this, I hope to merge this tomorrow

src/gui/guiFormSpecMenu.cpp
v-rob: Duplicating it so many times seems really messy, though.  Is there a better way that we could do it?

src/gui/guiAnimatedImage.cpp
v-rob: We need an extra newline at the end of a file?  Why?

rubenwardy: Having an ending new line comes from unix standards, but computers are usually clever enough to work it out these days: https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

image

rubenwardy (approved):

doc/lua_api.txt
SmallJoker: Sorry, did I miss anything? Nothing's changed for this line.

doc/lua_api.txt
```diff
@@ -2120,27 +2120,30 @@ Elements
...
 
 * Show an animated image. The image is drawn like a "vertical_frames" tile
-    animation (See Tile animation definition), but uses a frame count/duration
+    animation (See [Tile animation definition]), but uses a frame count/duration
     for simplicity
-* `<texture name>` is the image to use
-* `<frame count>` is the number of frames animating the image
-* `<frame duration>` is in milliseconds
+* `name`: The event name. The event is the index of the current frame, and it is fired
+  when another event is fired, e.g. a button is pressed.
```
SmallJoker: `fired`? For formspec this might fit better:
```
* `name`: Element name to send when an event occurs.
  The value is the index of the current frame.
```

SmallJoker (approved):
Works

doc/lua_api.txt
v-rob: I'll fix ASAP (probably in about an hour).

v-rob:
Ready for merge at any time.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -784,16 +784,20 @@ void GUIFormSpecMenu::parseAnimatedImage(parserData *data, const std::string &el
 {
 	std::vector<std::string> parts = split(element, ';');
 
-	if (parts.size() != 3 &&
-			!(parts.size() > 3 && m_formspec_version > FORMSPEC_API_VERSION)) {
-		errorstream << "Invalid animated image element(" << parts.size()
-				<< "): '" << element << "'"  << std::endl;
+	if (parts.size() != 6 && parts.size() != 7 &&
+		!(parts.size() > 4 && m_formspec_version > FORMSPEC_API_VERSION))
```
Desour:
```suggestion
		!(parts.size() > 7 && m_formspec_version > FORMSPEC_API_VERSION))
```

src/gui/guiFormSpecMenu.cpp
```diff
@@ -811,21 +815,26 @@ void GUIFormSpecMenu::parseAnimatedImage(parserData *data, const std::string &el
...
 
 	core::rect<s32> rect = core::rect<s32>(pos, pos + geom);
 
-	gui::IGUIElement *e = new GUIAnimatedImage(Environment, this, spec.fid,
-			rect, name, m_tsrc);
+	GUIAnimatedImage *e = new GUIAnimatedImage(Environment, this, spec.fid,
+		rect, texture_name, frame_count, frame_duration, m_tsrc);
 
-	auto style = getStyleForElement("animated_image", spec.fname);
+	if (parts.size() == 7)
```
SmallJoker: What if there will be more params?

`>= 7`

src/gui/guiFormSpecMenu.cpp
v-rob: I find this code sooooo ugly.  Putting the brace on the next line look so much cleaner and easier to understand, and it's what `GUIFormSpecMenu.cpp` usually does.  Oh well.

Desour: Sorry, I couldn't resist doing the style "fixes" in the suggestion. The important part is the `7`.

SmallJoker:
:+1: Changes look good. Thanks.

PR #9421, opened 2020-02-22
Add waypoint precision

appgurueu:
# Overview

- Goal of the PR: Add waypoint precision ("digits-after-dot"). If set to 0, distance is not shown.
- How does the PR work? It uses the `item` field to store precision.
- If not a bug fix, why is this PR needed? What usecases does it solve? Closes [this feature request](https://github.com/minetest/minetest/issues/8141).

## To do

This PR is Ready for Review.

## How to test

Create a mod with the following `init.lua`:
```lua
minetest.register_chatcommand(
	"test_waypoints",
	{
		description = "tests waypoints",
		func = function(sendername, params)
			local player = minetest.get_player_by_name(sendername)
			minetest.chat_send_all(
				"regular waypoint: " ..
					player:hud_add(
						{
							hud_elem_type = "waypoint",
							name = "origin",
							text = "",
							number = 0xFF0000,
							world_pos = vector.add(player:get_pos(), {x = 0, y = 2, z = 0})
						}
					)
			)
			minetest.chat_send_all(
				"better waypoint: " ..
					player:hud_add(
						{
							hud_elem_type = "waypoint",
							name = "origin",
							text = "",
							item = 5,
							number = 0xFF0000,
							world_pos = player:get_pos()
						}
					)
			)
			minetest.chat_send_all(
				"better waypoint without distance: " ..
					player:hud_add(
						{
							hud_elem_type = "waypoint",
							name = "origin",
							text = "",
							item = 0,
							number = 0xFF0000,
							world_pos = player:get_pos()
						}
					)
			)
		end
	}
)
```

## Screenshot

Screenshot

### DIFF
--- a/src/client/hud.cpp
+++ b/src/client/hud.cpp
@@ -370,7 +370,7 @@ void Hud::drawLuaElements(const v3s16 &camera_offset)
 				core::rect<s32> size(0, 0, 200, 2 * text_height);
 				std::wstring text = unescape_translate(utf8_to_wide(e->name));
 				font->draw(text.c_str(), size + pos, color);
-				// waypoints reusing item for precision, item = precision + 1
+				// waypoints reuse item to store the precision, item = precision + 1
 				u32 item = e->item;
 				float precision = item == 0 ? 10.0f : (item - 1);
 
--- a/src/client/hud.cpp
+++ b/src/client/hud.cpp
@@ -373,11 +373,10 @@ void Hud::drawLuaElements(const v3s16 &camera_offset)
 				// item = precision + 1
 				float precision;
 				u32 item = e->item;
-				if (item == 0) {
+				if (item == 0)
 					precision = 10.0f;
-				} else {
+				else
 					precision = item - 1;
-				}
 
 				if (precision > 0) {
 					std::ostringstream os;
--- a/src/client/hud.cpp
+++ b/src/client/hud.cpp
@@ -370,13 +370,9 @@ void Hud::drawLuaElements(const v3s16 &camera_offset)
 				core::rect<s32> size(0, 0, 200, 2 * text_height);
 				std::wstring text = unescape_translate(utf8_to_wide(e->name));
 				font->draw(text.c_str(), size + pos, color);
-				// item = precision + 1
-				float precision;
+				// waypoints reusing item for precision, item = precision + 1
 				u32 item = e->item;
-				if (item == 0)
-					precision = 10.0f;
-				else
-					precision = item - 1;
+				float precision = item == 0 ? 10.0f : (item - 1);
 
 				if (precision > 0) {
 					std::ostringstream os;
--- a/src/client/hud.cpp
+++ b/src/client/hud.cpp
@@ -351,8 +351,6 @@ void Hud::drawLuaElements(const v3s16 &camera_offset)
 			case HUD_ELEM_WAYPOINT: {
 				v3f p_pos = player->getPosition() / BS;
 				v3f w_pos = e->world_pos * BS;
-				float distance = std::floor(10 * p_pos.getDistanceFrom(e->world_pos)) /
-					10.0f;
 				scene::ICameraSceneNode* camera =
 					RenderingEngine::get_scene_manager()->getActiveCamera();
 				w_pos -= intToFloat(camera_offset, BS);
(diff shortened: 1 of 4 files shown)


src/script/common/c_content.cpp
```diff
@@ -1852,6 +1852,9 @@ void read_hud_element(lua_State *L, HudElement *elem)
 	elem->text    = getstringfield_default(L, 2, "text", "");
 	elem->number  = getintfield_default(L, 2, "number", 0);
 	elem->item    = getintfield_default(L, 2, "item", 0);
+	if (elem->type == HUD_ELEM_WAYPOINT) {
```
appgurueu: Probably this should be changed to an if-else?

src/script/common/c_content.cpp
```diff
@@ -1852,6 +1852,9 @@ void read_hud_element(lua_State *L, HudElement *elem)
...
+		elem->item = getintfield_default(L, 2, "item", -1) + 1;
+	}
```
appgurueu: Probably this should be changed to an if-else?

MoNTE48: it seems that yes

appgurueu: Done

src/script/common/c_content.cpp
```diff
@@ -1851,9 +1851,10 @@ void read_hud_element(lua_State *L, HudElement *elem)
 	elem->name    = getstringfield_default(L, 2, "name", "");
 	elem->text    = getstringfield_default(L, 2, "text", "");
 	elem->number  = getintfield_default(L, 2, "number", 0);
-	elem->item    = getintfield_default(L, 2, "item", 0);
 	if (elem->type == HUD_ELEM_WAYPOINT) {
 		elem->item = getintfield_default(L, 2, "item", -1) + 1;
```
appgurueu: Should I rename this to `precision`?

Wuzzy2:
> If set to 0, distance is not shown.

But what if I want to show a distance with whole numbers only? Looks like then I'm screwed.

Wuzzy2:
Suggestion: Use `item=-1` for hiding the distance number. Assuming the `item` field is signed …

appgurueu:
@Wuzzy2 unfortunately, it's not: `u32 item;` in HUD.h

src/script/common/c_content.cpp
MoNTE48: and btw, { } are not needed in this case. wWthout them, the code looks neater

appgurueu: I know, but I prefer to use them.

appgurueu: Update: Will remove the curly braces according to the preferred code style

SmallJoker:
Why `item`? A field name like `scale` would be clearer for mods, and is versatile. If needed, extend the HUD packet to send an additional `f32`. Old clients won't handle the precision anyway.

appgurueu:
@SmallJoker re-using the currently unused `item` field seems fine to me, as it stores an `u32` which suffices for the precision purpose. `scale` is a vector. And I would prefer not to send an uneeded `f32`.

Wuzzy2:
Nice trick. But I would have been completely fine if the `item` field would be reused. Anyway, it's OK either way.

src/script/common/c_content.cpp
```diff
@@ -1851,7 +1851,10 @@ void read_hud_element(lua_State *L, HudElement *elem)
 	elem->name    = getstringfield_default(L, 2, "name", "");
 	elem->text    = getstringfield_default(L, 2, "text", "");
 	elem->number  = getintfield_default(L, 2, "number", 0);
-	elem->item    = getintfield_default(L, 2, "item", 0);
+	if (elem->type == HUD_ELEM_WAYPOINT)
```
sfan5: should add a comment here noting that waypoints reuse `item`

src/client/hud.cpp
```diff
@@ -372,11 +370,22 @@ void Hud::drawLuaElements(const v3s16 &camera_offset)
...
-				text = unescape_translate(utf8_to_wide(os.str()));
-				pos.Y += text_height;
-				font->draw(text.c_str(), size + pos, color);
+				// item = precision + 1
+				float precision;
+				u32 item = e->item;
+				if (item == 0)
+					precision = 10.0f;
+				else
+					precision = item - 1;
```
sfan5: `float precision = e->item == 0 ? 10.0f : (item - 1);`

Extex101:
Would you consider #9447

src/script/common/c_content.cpp
appgurueu: will do

src/client/hud.cpp
appgurueu: will change

src/client/hud.cpp
```diff
@@ -370,13 +370,9 @@ void Hud::drawLuaElements(const v3s16 &camera_offset)
 				core::rect<s32> size(0, 0, 200, 2 * text_height);
 				std::wstring text = unescape_translate(utf8_to_wide(e->name));
 				font->draw(text.c_str(), size + pos, color);
-				// item = precision + 1
-				float precision;
+				// waypoints reusing item for precision, item = precision + 1
```
sfan5:
```suggestion
				// waypoints reuse item to store the precision, item = precision + 1
```

src/script/common/c_content.cpp
sfan5: This doesn't really explain it, I was thinking more of:
`// Waypoints reuse the item property to store the precision (+1) of the displayed distance`

appgurueu:
Closing this to open a larger PR incorporating it.

PR #9424, opened 2020-02-22
Implement the 'userdata' commandline option

vilhelmgray:
This patch adds support for a 'userdata' commandline option to allow
users to configure the path for storing user data. This should give
users more control over where data is stored on their systems.

See: https://github.com/minetest/minetest/issues/6869

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -158,11 +158,15 @@ int main(int argc, char *argv[])
 
 	porting::signal_handler_init();
 
+	std::string userdata;
+	if (cmd_args.exists("userdata"))
+		userdata = cmd_args.get("userdata");
+
 #ifdef __ANDROID__
 	porting::initAndroid();
-	porting::initializePathsAndroid();
+	porting::initializePathsAndroid(userdata);
 #else
-	porting::initializePaths();
+	porting::initializePaths(userdata);
 #endif
 
 	if (!create_userdata_path()) {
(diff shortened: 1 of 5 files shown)


SmallJoker:
I'm late, but wouldn't this be a job for environment variables?

vilhelmgray:
We encountered this issue in Gentoo. Using the HOME env variable can work if the user wants to use that location, but some users prefer to store their game data in other locations outside their HOME -- this is particularly useful for cases where Minetest is run by a system user that is shared by other users. Providing a `userdata` commandline option enables users to directly specify their preferred Minetest user data location without changing the env setup of the system its running in.

SmallJoker:
How about `HOME=/path/to/data minetestserver ...etc` ? Does that not work, or not have the expected results?

vilhelmgray:
> How about `HOME=/path/to/data minetestserver ...etc` ? Does that not work, or not have the expected results?

This assumes the user is running a shell such as bash that can overload the environment variables and scope them for the process.  Unfortunately, a user could be running a different shell (csh, tcsh, etc.) that doesn't support such syntax, and they would have to clobber their HOME variable in order to launch the application. However, with the `userdata` command line option approach, it doesn't matter what shell or system setup the user has, they will be able to directly configure where the Minetest data is located.

SmallJoker:
https://irc.minetest.net/minetest-dev/2022-07-03#i_5990796

The idea is good, but generally an environment variable seems to be more welcome.

vilhelmgray:
As long as the environment variable is named something sufficiently unique (e.g. `MINETEST_USERDATA` or similar) then I think that approach should be okay because we avoid the clobbering `HOME` issue.

Zughy:
@vilhelmgray any news?

vilhelmgray:
@Zughy No one's contacted me any updates, so I don't know what the state of this is.

I think  the only a change needed is a single line in the `getHomeOrFail()`: https://github.com/minetest/minetest/blob/644f145ff260c6f9cf0e351d158c045019aa895d/src/porting.cpp#L393
Change `"HOME"` to `"MINETEST_USERDATA"` and that should resolve this issue.

Zughy:
It's your PR, can you do that and test it? 👀

vilhelmgray:
I apologize, I was under the impression someone else had already started working on this. It's no problem, I will code something up later today for this and submit it. :+1:

vilhelmgray:
I'm closing this PR in favor of https://github.com/minetest/minetest/pull/12639.

PR #9430, opened 2020-02-24
Add FFmpeg support

kilbith:
This PR offered by [EvidenceBKidscode](https://www.evidenceb.com/) adds the **`FFmpeg`** support to Minetest. **`FFmpeg`** can virtually decode [anything](https://en.wikipedia.org/wiki/FFmpeg#Supported_codecs_and_formats) (including GIF files).

Demo: https://www.youtube.com/watch?v=d9Rm1vM53tE

Required libs: `libavcodec` `libavformat` `libavutil` `libswscale` (I haven't dealt *properly* with `CMakeLists` yet, but it works at least).

A video can be rendered onto a `IMeshSceneNode` in C++ like this: https://paste.ubuntu.com/p/9jKjMXfvcy/

Remaining goals of this PR:

- [ ] Make the `ffmpeg` dependency optional, fallback consistently in-game.
- [ ] Add a `video` field to the meshnode and entity register definitions.
- [ ] Fix the segfault when drawing `video[]` in a formspec.
- [ ] Add controls to stop/~pause~/resume/loop a video.
- [x] Silence the `FFmpeg` logging.
- [ ] Add the audio support.

All help would be appreciated.
@Df458 @v-rob @DS-Minetest

Addendum: from my tests, rendering a video in-game does not affect the performance.


sorcerykid:
I will be very excited to see how this will work in multiplayer, in particular whether it will be possible to serve the video on demand (streaming live from a media server or via HTTP download).

benrob0329:
Does this just use FFmpeg for decoding videos, or does it use it as the sole media (textures aside) decoding library (replacing libogg and libvorbis directly, though I think ffmpeg makes use of them)?

v-rob:
Very, very nice!  If this does support audio (or will in the future), with API stuff for playing, pausing, looping, setting video position, getting video length, maybe even tracks (just throwing a bunch of stuff out there), etc., we might have better way of doing in-game music instead of the less-than-adequate sound API.

Question: Is the video supposed to be mirrored in the example video?

Also kind of ironic that the YouTube video nearly crashed my computer somehow...

benrob0329:
> Very, very nice! If this does support audio (or will in the future), with API stuff for playing, pausing, looping, setting video position, getting video length, maybe even tracks (just throwing a bunch of stuff out there), etc., we might have better way of doing in-game music instead of the less-than-adequate sound API.

I really hope you mean using a generic FFmpeg API for sound files, and not actually using video files to store and transport background music just because MT's sound API is a little basic...

My biggest concern is a sudden increase in media download sized, going from maybe a hundred Megabytes under any sane circumstance to hundreds of Megabytes for even just one video of a decent resolution and framerate. Minetest really needs an on-demand media fetcher for stuff like this, which would also help games with large amounts of background music but is not something we have at the moment.

My question is, aside from the added format support, what are the practical use-cases of full-motion video in Minetest? Since MT is catered towards pretty low-tech graphics overall, I don't see this fitting in all that well under most circumstances. You could say tutorials or walkthroughs, but that seems like a lot of wasted bandwidth and library dependencies just to play a glorified YouTube video for something that is generally only needed because of bad game design.

TumeniNodes:
Will video and audio capturing be added, or possible?
It would be nice to exploit as many ffmpeg features as possible

kilbith:
> Question: Is the video supposed to be mirrored in the example video?

In this demo, the video has been applied on the 6 faces of a cube, one of its face had the mirrored view.

> My biggest concern is a sudden increase in media download sized

That's a valid concern, but isn't the Public Remote Media Server by @sofar already helps in this regard?

> My question is, aside from the added format support, what are the practical use-cases of full-motion video in Minetest?

One of the best games of all time, Half-Life 2, had memorable videos all along the game.

sofar:
>> My biggest concern is a sudden increase in media download sized
>
>That's a valid concern, but isn't the Public Remote Media Server by
>@sofar already helps in this regard?

I'm not worried there. If videos are part of the static mod content just like textures, servers will make the choice for their users. If the users complain too much then servers will take these down, or shrink their size.

I am worried that this will force a dependency on ffmpeg. At a minimum this should remain optional, because ffmpeg ships with patented codec support by default, which makes it impossible for some distros to include it. In Clearlinux we ship a stripped ffmpeg that can't decode x264, so we should at least attempt to promote open codecs like vp8/9 and av1, and avoid aac/mp3 and x264/5, so that people can choose the unencumbered codecs when they want.

It's possible that distros remove minetest if ffmpeg becomes a required dependency.

Rendering in game should remain consistent if ffmpeg isn't available or rendering fails due to missing codec support.

sfan5:
This is a cool feature but I think it <i>might</i> be out-of-scope of the engine (just like that chromium-embedded thing someone demoed once).
Cutscenes are a good point but I wonder about other usecases for this (that aren't "look, you can watch this tutorial video inside MT instead of opening your browser")

I'm also wary of adding a 20MB dependency to what might be a feature of minor importance in day-to-day gameplay. If added, this should certainly be optional.

In the code, the given filename is passed directly to `avformat_open_input`. There will need to be some validation, FFmpeg has lots of protocols which may allow doing undesirable things (especially if you manage to invoke the [filtering infrastructure](http://ffmpeg.org/ffmpeg-filters.html), though I don't know if that's possible). Security concerns will need to be looked into to ensure ffmpeg can't be convinced to do anything except normal video decoding.

> with API stuff for playing, pausing, looping, setting video position, getting video length, maybe even tracks (just throwing a bunch of stuff out there), etc., we might have better way of doing in-game music instead of the less-than-adequate sound API.

If the sound API is lacking features that modders need, it should be improved instead of repurposing an entirely different feature.

> Will video and audio capturing be added, or possible?
It would be nice to exploit as many ffmpeg features as possible

Oh god, please don't. What would you even want to use this for?

kilbith:
Telling a student to open the web browser to watch a tutorial is a straight no-go in the educational world, and I know what I'm talking about. And the BAW team understood that pretty well too.

I was not referring to cutscenes, but videos which do not break up the gameplay, which are part of an environment and that deepen your immersion (cf. Dr Breen's speech in train station of Half-Life 2). Portal 2 has in-game videos too, IIRC. The concern there *might* be a lack of imagination.

sofar:
>Oh god, please don't. What would you even want to use this for?

Abuse.

I'm neutral on videos in general (not saying anything about this PR just yet) but allowing a server to capture even just audio is, to me, a full-stop no.

sfan5:
> Telling a student to open the web browser to watch a tutorial is a straight no-go in the educational world, and I know what I'm talking about.

Sure that makes sense. A feature that is an absolute necessity in the education context is not necessarily very useful for MT in general.

> I was not referring to cutscenes, but videos which do not break up the gameplay, which are part of an environment and that deepen your immersion

Oh I misunderstood that then, though in it's current state this PR does not yet allow that.

TumeniNodes:
> Oh god, please don't. What would you even want to use this for?
> Abuse. I'm neutral on videos in general (not saying anything about this PR just yet) but allowing a server to capture even just audio is, to me, a full-stop no.

I was thinking more along the likes of recording episodes, mod reviews, or flybys, etc...
But yeh... I can see where this ability would be abused. I hadn't thought of that, I tend to forget about "because, society"

benrob0329:
If I remember correctly, Source engine games use render-to-texture and an off-screen models, not full motion video for screens.

kilbith:
Because the teacher don’t know much if not less than the students oftentimes. Yeah, true stories.

sofar:
> I was thinking more along the likes of recording episodes, mod reviews, or flybys, etc...

You would just use OBS studio to do those, and them embed that video.

Allowing ffmpeg to open the microphone and then record the data, and send that data back to the server would allow servers to listen in on what happens in player's rooms.

Obviously that's a lot more complex than this patch, but, it might be a good idea to have this discussion now.

This patch looks pretty nice, though, from a quick code glance.... I'm surprised it's this simple.

paramat:
The playing of videos in-game is an impressive ability, though i agree it is fairly low priority for MT. Obviously video file sizes will have to be restrained, but then, low-res low-framerate videos will look right in MT.

For those of us, like me, who are not knowledgeable about ffmpeg: Apart from in-game videos, what other uses will this PR make possible? The first post does not suggest anything other than videos but i suspect there are more uses?
Unless i misunderstood, the below suggests more uses:
> FFmpeg can virtually decode anything (including GIF files).

sofar:
> Obviously video file sizes will have to be restrained,

that is of no concern to the *engine*. The *engine* here (mt core) should not restrict anything. We don't restrict sizes of textures - which can be gigabytes. Neither does the core limit the size of `ogg` files for audio, which could have 256 channel audio at 2000kbit rates and be 250 million years long (the ogg standard doesn't actually seem to have a maximum length(!)). So it makes no sense to put any limits in - server operators will do those checks, as they should.

> For those of us, like me, who are not knowledgeable about ffmpeg: Apart from in-game videos, what other uses will this PR make possible? The first post does not suggest anything other than videos but i suspect there are more uses?

ffmpeg is a system library that can potentially do some of the following:
- record video from a webcam
- record audio
- grab or record your entire screen, not just something inside the game
- encode this
- send the data out over a streaming protocol

This patch doesn't do anything like that.

TumeniNodes:
> Allowing ffmpeg to open the microphone and then record the data, and send that data back to the server would allow servers to listen in on what happens in player's rooms.

I had not even thought of that.

> Obviously that's a lot more complex than this patch, but, it might be a good idea to have this discussion now.

Agreed. Especially related to ensuring ffmpeg features cannot be hacked from within MT to do such things.
Though patches like this seem pretty isolated/restricted, unless I am mistaken. Which would not be surprising.

runsy:
Hytale will have this feature:
imagen
- The Size of videos is a matter of servers/games, as for audio or textures. But it would be could an option to load multimedia on demand.

sorcerykid:
I think that the ability to show videos in-world and to embed videos in formspecs is essential, as it opens up so many more possibilities for fresh and innovative games in Minetest.

However, given the current state of affairs, and the very poignant comments that kilbith made on IRC today, I think it's time to rethink the direction of Minetest as a game development platform, otherwise it is almost certainly doomed to obsolescence like Irrlicht.

The whole "purist" stance that I've often witnessed when it comes to PR review and approval is very clearly working against this project in every respect. Nothing beneficial is ever achieved if the goal is always devout adherence to 100% perfection. I say screw "perfection" and let's get on with it! 

I get that Minetest shouldn't try to be or do everything. Obviously there's no point in adding every cool little widget or fancy little gadget. But it still has to remain competitive in the marketplace. And let's be honest: Both the  engine and Minetest Game are so far behind the curve compared even to its sibling Minecraft, that we just can't afford to keep on nitpicking about which feature "fits" exactly into the vision of what Minetest "really is".

I can tell you right now what Minetest _isn't_ to most people: Minetest is not cool, nor hip, nor worthy of mention in any popular gaming circles. And I can  guarantee it has nothing to do with the esoteric name. Other than being FOSS, there's literally nothing else appealing or exciting at all about Minetest that even makes it worthy of comparison to other PC games.

**I really hope that will change. I want to see Minetest succeed, obviously!** And I say this not just as a server operator, but as a devoted member of the community.

I wish to see more players, more mod developers, more game designers, more contributors. I want to see a growing, thriving, active online community once again. But if we keep going down this same path, then quite frankly the future looks very bleak for Minetest.

/steps off soapbox :)

celeron55:
@kilbith if we were to by default disable patented formats like h.264 (which frankly is the format that everyone will try to use at first) as @sofar mentions necessary for some linux distributions and probably some organizations that actually try to follow the law, does it affect your intended educational use in practice? I am guessing not, because your particular educational use will have its own builds and configurations anyway.

So that only leaves those people who don't care about patents crying, which might be 99%. But then again there's a valid reason for it.

We don't often hear people complaining not being able to use mp3 for audio though, so it could be similarly painless to require using vp8 or some such for video. I think the main difference is that converting video takes some know-how and a lot of processing compared to audio.

I don't see it becoming a huge issue that some builds will not include ffmpeg support, but maybe the formspec video element should support a placeholder image for those cases? If a video was used in eg. a tutorial, the same information could easily be shortened into an image, just with less details. It could also allow people to switch off videos without losing important content if they have a slow connection.

appgurueu:
@kilbith @DS-Minetest is now called @Desour

kilbith:
🖕🏽

appgurueu:
This is sad.

PR #9440, opened 2020-02-28
Chatcommand: Show help message if func returns false without message

HybridDog:
This makes it easier to show how to use a chatcommand to the player if he/she uses it incorrectly.
Usually, `/help <command>` should show how to invoke `<command>`, so there's often no need for individual parsing error messages.

### DIFF
--- a/builtin/game/chat.lua
+++ b/builtin/game/chat.lua
@@ -70,9 +70,11 @@ core.register_on_chat_message(function(name, message)
 		if success == false and result == nil then
 			core.chat_send_player(name, "-!- Invalid command usage")
 			local help_def = core.registered_chatcommands["help"]
-			local _, helpmsg = help_def.func(name, cmd)
-			if helpmsg then
-				core.chat_send_player(name, helpmsg)
+			if help_def then
+				local _, helpmsg = help_def.func(name, cmd)
+				if helpmsg then
+					core.chat_send_player(name, helpmsg)
+				end
 			end
 		elseif result then
 			core.chat_send_player(name, result)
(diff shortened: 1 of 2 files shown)


SmallJoker:
The idea is good, but the implementation looks hacky. I'd prefer to call the help function if `result == nil and success == false`, although this might cause side-effects in a few mods.

sorcerykid:
I concur with SmallJoker, as that would be a more logical implementation.

appgurueu:
A suggestion: Instead of modifiying on_chat_message, just modify register_chatcommand like this:
```lua
...
func = function(sendername, params)
       local success, response = def.func(sendername, params)
       if response == nil and success == false then
           success, response = core.registered_chatcommands["help"].func(name, cmd)
           return success, response
       end
end
...
```

This would also be compatible with cmdlib.

builtin/game/chat.lua
```diff
@@ -66,8 +66,15 @@ core.register_on_chat_message(function(name, message)
 	local has_privs, missing_privs = core.check_player_privs(name, cmd_def.privs)
 	if has_privs then
 		core.set_last_run_mod(cmd_def.mod_origin)
-		local _, result = cmd_def.func(name, param)
-		if result then
+		local success, result = cmd_def.func(name, param)
+		if success == false and result == nil then
+			core.chat_send_player(name, "-!- Invalid command usage")
+			local help_def = core.registered_chatcommands["help"]
```
SmallJoker, line 72: Servers might remove the help command and implement something on their own. I think a `nil` check for `help_def` would be appropriate here.

HybridDog: I've added a check.

rubenwardy (approved):

PR #9443, opened 2020-02-29
Improve minetest.spawn_item and include in API documentation

sorcerykid:
This PR from the _Minetest S3_ fork allows the builtin `spawn_item()` function (currently aliased from `add_item()` within lua_api/l_env.cpp) to accept a horizontal and vertical velocity parameter for a randomly oriented "tossing" effect.

Unlike `item_drop()`, this function is more suitable for incidental drops that are independent of the player, such as the drops from TNT explosions.

explosion

The underlying logic is the same as the `minetest.drop_item()` function [that I published a few years ago](https://forum.minetest.net/viewtopic.php?p=295360#p295360) on the forums, but it has since been integrated into `spawn_item()` instead (which makes more sense, although I didn't know about this builtin function at the time, since it's undocumented :)

## To do

This PR is Ready for Review.

## How to test

Create a mod with the following init.lua. Then, place a Nyan Cat node and punch it for goodies :)

```
minetest.override_item("nyancat:nyancat", {
	on_punch = function (pos)
		minetest.spawn_item(pos, "nyancat:nyancat", 2.5, 5.0)
	end,
})
```

### DIFF
--- a/builtin/game/item_entity.lua
+++ b/builtin/game/item_entity.lua
@@ -1,12 +1,19 @@
 -- Minetest: builtin/item_entity.lua
 
-function core.spawn_item(pos, item)
+function core.spawn_item(pos, item, horz, vert)
 	-- Take item in any format
 	local stack = ItemStack(item)
 	local obj = core.add_entity(pos, "__builtin:item")
+
 	-- Don't use obj if it couldn't be added to the map.
 	if obj then
 		obj:get_luaentity():set_item(stack:to_string())
+
+		if horz then
+			obj:set_velocity(vector.new(
+				math.random(-horz, horz), vert or 0, math.random(-horz, horz)
+			))
+		end
 	end
 	return obj
 end
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -4461,6 +4461,11 @@ Environment access
     * Returns `ObjectRef`, or `nil` if failed
 * `minetest.add_item(pos, item)`: Spawn item
     * Returns `ObjectRef`, or `nil` if failed
+* `minetest.spawn_item(pos, item, horz, vert)`
+    * Same as `add_item, but with an optional spontaneous velocity applied
+    * `horz` is the maximum velocity along both horizontal axes.
+    * `vert` is the vertical velocity, which may be negative for downward or
+      positive for upward
 * `minetest.get_player_by_name(name)`: Get an `ObjectRef` to a player
 * `minetest.get_objects_inside_radius(pos, radius)`: returns a list of
   ObjectRefs.


SmallJoker:
I can't think of many situations where exactly this behaviour is wanted. Even if the behaviour fits, it only saves one additional line to call `set_velocity()`. I'd prefer to keep the builtin functions simple and generic as possible,

Wuzzy2:
I agree. I also don't see why these additional parameters are needed. Just call `set_velocity` afterwards.
:-1:

sorcerykid:
> I can't think of many situations where exactly this behaviour is wanted. 

The behavior is wanted in my TNT Redux mod, my Bones Redux mod, my Macro Crafting Manager mod, and my fork of Mobs Redo. 

> I agree. I also don't see why these additional parameters are needed. Just call set_velocity afterwards.

The parameters aren't "needed" they are optional. And it's not as simple as "just call set_velocity afterwards". In fact, let's take a look at a concrete example from my TNT Redux Mod.

minetest.spawn_item(drop_pos, item_name, 5, 10)

Using your suggestion would require multiple temporary variables, as well as a conditional branch, an in-line vector object, and two duplicate calls to math.random. Talk about excessive, just for a random spawn-propulsion effect.

local temp_obj = minetest.add_item(drop_pos, item_name)
if temp_obj then
        local horz = 5
        local vert = 10
        temp_obj:set_velocity({
              x = math.random(-horz, horz),
              y = vert,
              z = math.random(-horz, horz)
        })
end

> Even if the behaviour fits, it only saves one additional line to call set_velocity()

It doesn't look like "one additional line" to me. Having to manually and consistently re-implement the same functionality across several mods undermines the principle of code-reuse and abstraction.

> I'd prefer to keep the builtin functions simple and generic as possible,

There are a lot of builtin functions that are neither simple nor generic. A perfect example is minetest.get_player_radius_area(). What is that even doing in the core? It can also be implemented with just "one additional line/;. I can't think of a single situation where I've needed that function.

I did a search of the "minetest-mods" repo, and it looks like minetest.get_player_radius_area() is only used by the Enhanced Chat Commands mod. Even the engine itself only uses this function once in chat.lua. And I can think of several other builtin functions that are rarely ever needed except under very specialized circumstances, yet they are part of the Minetest API anyway.

Wuzzy2:
Oh, I didn't know the velocity is random. It wasn't clear from the first post.

But still! I'm still not entirely convinced.
I think this would be more suited to put into a a separate utility function to set random velocity instead of overloading `spawn_item`.

But since you already went for the nuclear option, that doesn't matter anyway now. XD

PR #9450, opened 2020-03-02
 Replacement for Minimal Development Test (PR)

Wuzzy2:
Here I represent to you a completely reworked Minimal Development Test. This is the PR form of #9260 (see this issue for details).

Note I have changed the name back to “Minimal development test” [`minimal`]. I plan to change the name to “Development Test” [`devtest`] in a later PR.

## Review remarks
Note this is a *complete* replacement of Minimal, with major structural changes and tons of new features. You will probably be better off reviewing the entire game instead of just the diff.
See #9260 for a changelog.

Originally, I started this as a fork of the `minimal` codebase and worked on it on a separate repo. To see the full commit log of my “working repo”, go here: https://git.minetest.land/Wuzzy/devtest/commits/branch/master
This PR is based on commit `7eee66d199283d3de3937b6abbece5c676a3dfa4` (this might change if I am required to make changes).

The single commit in this PR is basically just a squashed version of my changes.

## How I will apply changes
If I am required to make any changes, I will first apply them to the external repo, then (eventually) apply them back to this PR via force-push from time to time. This is only done for practical reasons, to keep me out of merge hell.
Don't worry, I will inform you when updates happen to this PR.


Wuzzy2:
Updates from the latest force-push:

- Some code style cleanup
- Improved command descriptions
- Move 2 entities from `experimental` to `testentities`
- Add 2 new attached soundtest nodes

To make reviewing worry-free, I promise to no longer touch the PR until one of 3 things happen:

- I find an IMPORTANT bug
- I was asked to touch it
- MT 5.2.0 is released

My strongest argument for this PR is that it already helped uncovering numerous bugs in the engine, especially drawtypes. The latest blockers #9482, #9481 were also discovered with this.

appgurueu:
I would love to see `devtest` replace `minimal`. I don't see a single reason to stick with `minimal`; it has many tests running by default (which I don't approve) and is not much lighter. I'm wondering whether you (sh/c)ould add [my waypoint capability test](https://gist.github.com/appgurueu/79fee6fa43f802907f7b34b3dba05163) (but probably you shouldn't touch the PR until it is merged) ?

It should become common to not only add documentation when adding a feature, but to also include relevant tests in `devtest` :)

Wuzzy2:
Yeah. The tests of old `minimal` are still there, but don't run by default (there are now game settings and commands for this stuff).

And yes, I totally agree that in the future, new Minetest features should automatically come with appropriate test cases in the test game. :) We need to work out a precise policy on this later. The reason why `minimal` is in such a bad shape is because it has been neglected for far too long.

---

For the review, I like you to make you aware that I defined design principles in the README. Take a look of it and tell me if you agree.

appgurueu:
I had already read it and definitely agree! Until now, I've been using `minimal` to test my mods, which had quite some noise, which devtest improves upon (less noise). As you have shown, it's also far more usefull for engine testing.

appgurueu:
Crashing on MT 5.1.1
```
2020-03-13 12:25:20: ACTION[Main]: [modchannels] Code loaded!
2020-03-13 12:25:20: ACTION[Main]: [experimental] on_mods_loaded()
2020-03-13 12:25:20: ACTION[Main]: World at [/home/lars/.minetest/worlds/cellestial]
2020-03-13 12:25:20: ACTION[Main]: Server for gameid="devtest" listening on 0.0.0.0:62240.
```

CBugDCoder:
@appgurueu It is likely using a function that is only available in 5.2.0-dev because devtest master is designed to test new engine features so won't work with the stable release.

dcbrwn:
Not sure if it's related, but latest minetest from master branch crashes with

```
Assertion failed: liquid_type == LIQUID_FLOWING, file C:\lab\minetest\src\nodedef.cpp, line 741
```

This is having old "minimal test" game working properly. After I commented out asserts from that file, game started seemingly fine.

Win10, x64

dcbrwn:
After using this for a while, I must say it's brilliant! Definitely a huge improvement over "minimal". Minimal game serves as a good start for custom games, so stripping out testing tools from there will make it even better suited for that purpose.

Wuzzy2:
OK. I just looked at the assert. So basically this assert requires that nodes with `drawtype = "flowingliquid"` also have `liquid_type = "flowing"`.

In `devtest`, I have intentionally included some test nodes with liquid/flowingliquid drawtypes that are not liquids themselves (`liquid_type = "none"`) in order to test the liquid/flowingliquid drawtypes without side effects. The nodes in question are `testnodes:liquid` and `testnodes:liquid_flowing`.

Is it OK to use a liquid/flowingliquid drawtype for non-liquids? If yes, then the assert is unjustified and must be removed. If not, then my test code is wrong and needs to be changed?

Oddly enough, the assert never affected me.

Wuzzy2:
Warning: I will make a force-push soon. There are new features coming, and tests relating to the 5.2.0 release.

Wuzzy2:
Force-push done.
New stuff compared to the previous commit:
* A couple of new sound test nodes
* Added hypertext[] test
* Updated other formspec tests

games/minimal/README.md
```diff
@@ -0,0 +1,49 @@
...
+* Experimental/strange/unstructured tests can be found in the `experimental` mod
+* Textures of drawtype test nodes have a red dot at the top left corner. This is to see whether the textures are oriented properly
+
+## Design philosophy
+
+This should loosely follow the following principles:
+
+* Engine testing: The main focus of this is to aid testing of *engine* features, such as mapgen or node drawtypes
+* Mod testing: The secondary focus is to help modders as well, either as a minimal testbed for mods or even as a code example
+* Minimal interference: Under default settings, it shall not interfere with APIs except on explicit user wish. Non-trivial tests and features need to be enabled by a setting first
```
rubenwardy, line 44: I'm not sure about the `devtest_unittests_autostart` settings, most of the tests are non-intrusive. The player hp change reasons one is intrusive because it changes HP, but it will never hill the player

games/minimal/game.conf
```diff
@@ -1,2 +1,2 @@
 name = Minimal development test
-
+description = Testing environment to help with testing the core features of Minetest. It can also be helpful in mod development.
```
rubenwardy, line 2: Engine is more understandable than core, and core usually refers to the c++ of the engine rather than its entirety

```suggestion
description = Testing environment to help with testing the engine features of Minetest. It can also be helpful in mod development.
```

games/minimal/mods/testentities/armor.lua
```diff
@@ -0,0 +1,37 @@
...
+-- Rightclick to change armor group
+
+local phasearmor = {
+	[0]={icy=100},
+	[1]={firy=100},
+	[2]={fleshy=100}
+}
+
+minetest.register_entity("testentities:armorball", {
+	hp_max = 20,
```
rubenwardy, line 11: defining initial properties has been deprecated since the early days of 0.4, but never warned about. A warning will be added soon (#9650)

```lua
initial_properties = {
    hp_max = 20
	physical = false,
	collisionbox = {-0.4,-0.4,-0.4, 0.4,0.4,0.4},
	visual = "sprite",
	visual_size = {x=1, y=1},
	textures = {"testentities_armorball.png"},
	spritediv = {x=1, y=3},
	initial_sprite_basepos = {x=0, y=0},
},
```

Also, the docs say that hp_max is a player only object 

```lua
hp_max = 1,
-- For players only. Defaults to `minetest.PLAYER_MAX_HP_DEFAULT`.
```

games/minimal/mods/testentities/callbacks.lua
```diff
@@ -0,0 +1,72 @@
...
+	return name
+end
+
+local spos = function(self)
+	return minetest.pos_to_string(vector.round(self.object:get_pos()))
+end
+
+-- Callback test entity (all callbacks except on_step)
+minetest.register_entity("testentities:callback", {
+	visual = "upright_sprite",
```
rubenwardy, line 26: same here

games/minimal/mods/testentities/visuals.lua
```diff
@@ -0,0 +1,60 @@
+-- Minimal test entities to test visuals
+
+minetest.register_entity("testentities:sprite", {
+	visual = "sprite",
```
rubenwardy, line 4: and here

games/minimal/mods/testtools/README.md
```diff
@@ -0,0 +1,99 @@
...
+## Object Property Editor
+Edits properties of objects.
+
+Usage:
+
+* Punch object to open a formspec that allows you to view and edit properties
+* Punch air to edit properties of your own player object
+
+To edit a property, select it in the list, enter a new value and select a data type (boolean, string, number).
+Note that tables are not supported yet.
```
rubenwardy, line 64: I suggest that you use loadstring() to support arbitrary data types here, and not need to specify the type

games/minimal/mods/testtools/edit_object_property_table.patch
```diff
@@ -0,0 +1,176 @@
+diff --git a/mods/testtools/helpers.lua b/mods/testtools/helpers.lua
+new file mode 100644
+index 0000000..06f2bda
+--- /dev/null
++++ b/mods/testtools/helpers.lua
+@@ -0,0 +1,82 @@
++-- FIXME: Quotation marks in strings are not escaped
++-- TODO: Support nested tables
++
++function testtools.parse_table(input)
```
rubenwardy, line 10: noooooo, use loadstring

rubenwardy:
How to use loadstring:

```lua
local func, error = loadstring("return (" .. param .. ")")
if not func then
	return false, "E " .. error
end

-- optional: apply sandbox here using setfenv (probably not needed)
setfenv(func, {})

local good, errOrResult = pcall(func)
if not good then
	return false, "E " .. errOrResult
end

-- errOrResult will be the value
```

rubenwardy:
This is super nice, thanks for doing this

Wuzzy2:
About `loadstring`: I would normally be concerned about mod security but as this thing is just for testing I think it can be neglected.

I suggest to put all other PRs that want to touch `minimal` on hold for now, until this one is merged. I do not want to duplicate the work.

rubenwardy:
> About loadstring: I would normally be concerned about mod security but as this thing is just for testing I think it can be neglected.

The two vulnerabilities you'd need to be worried about with a sandboxed `loadstring`:

* performance issues due to infinite loops
* malicious-crafted return values - care should be taken

both of these are fine in this context. I'm not sure how mesecons gets around the first in their lua controllers

Wuzzy2:
I have updated the original DevTest repository (not the PR):
* Added support for all data types in the entity property editor using `loadstring` (Thanks for the hints! This missing feature always bothered me. I have already tried to figure out how to “parse” Lua values, but didn't know how.)
* As a security precaution, you need the `server` priv to change any object property, because it involves `loadstring`. Just in case …
* Updated `game.conf` description as requested
* Fix deprecated entity syntax as requested
* More armorball modes added: `immortal` and `punch_operable`

https://git.minetest.land/Wuzzy/devtest/commits/branch/rename_minimal

I will update this PR when requested. Question: Should I force-push to this PR after every change in future? Note I also like to wait until more review drop in.

rubenwardy:
I'd prefer that you do no squash the commits whilst reviewing, as squashing means I cannot see what you've changed since the last look

Wuzzy2:
OK, here's how I proceed: I will copy over changes from the DevTest repo to this PR as new commit. I have just copied over a bunch of changes.

I will only squash, rebase, [do whatever] on explicit request. Without request, I will do nothing.

games/minimal/mods/craft/init.lua
```diff
@@ -0,0 +1 @@
+
```
rubenwardy, line 1: empty mod

games/minimal/mods/initial_message/init.lua
```diff
@@ -0,0 +1,6 @@
+minetest.register_on_joinplayer(function(player)
+	local cb = function(player)
+		minetest.chat_send_player(player:get_player_name(), "This is the \"Minimal development test\" [minimal], meant only for testing and development. Use Minetest Game for the real thing.")
```
rubenwardy, line 3: This will crash if the player leaves within two seconds. Although, minimal does tend to only be used in singleplayer :shrug:

games/minimal/mods/testpathfinder/init.lua
```diff
@@ -0,0 +1,136 @@
...
+-- End of config parameters
+
+local timer = 0
+local algorithms = {
+	"A*_noprefetch",
+	"A*",
+	"Dijkstra",
+}
+
+local function call_find_path(pos1, pos2, searchdistance, max_jump, max_drop, algo)
```
rubenwardy, line 25: What's the point of this? For debugging?

```lua
local find_path = minetest.find_path
```

games/minimal/mods/testtools/init.lua
```diff
@@ -0,0 +1,749 @@
+local S = minetest.get_translator("testtools")
+local F = minetest.formspec_escape
+
+-- TODO: Add a Node Metadata tool
+
+local function check_priv(player, priv)
```
rubenwardy, line 6: I don't get the point of this at all. It's also potentially dangerous, if someone tries to check a non-whitelisted priv. You can use:

```lua
minetest.check_player_privs(player, "server")
```

to check a single privilege

Wuzzy2:
* `craft` removed
* Priv check simplified
* Remove pointless pathfinder wrapper function
* Fix `minetest.after` crash

Changesets on DevTest repo:

https://git.minetest.land/Wuzzy/devtest/commit/4ea7d47a01c9059c667315c1445e24931318ebcd
https://git.minetest.land/Wuzzy/devtest/commit/c719f55a696ccf3ac81f1c6b593aaf0f908682fc
https://git.minetest.land/Wuzzy/devtest/commit/942e0d55dd84367c84be22cfafc3fa18360a4bc0
https://git.minetest.land/Wuzzy/devtest/commit/e448104a089aebf2f6c149c4829bc112396231f0

Since you have not requested an update to this PR, I have not touched this PR.

Wuzzy2:
How to proceed?

rubenwardy:
Please may you push the changes here and fix the conflicts?

Wuzzy2:
OK, done. If requested, I will squash everything (if you're unhappy with the history).

Sorry for being super annoying to work with right now, but I do this “I do nothing without request” strategy to be sure I do not interfere with the review process (which could be even more annoying).

games/minimal/README.md
appgurueu: I'd keep those settings. Not everyone wants auto unit tests, modders using this as lightweight testbed usually don't.

games/minimal/game.conf
```diff
@@ -1,2 +1,2 @@
 name = Minimal development test
-
+description = Testing environment to help with testing the engine features of Minetest. It can also be helpful in mod development.
```
appgurueu, line 2: Can probably be shortened a bit to something along the lines of `Minimal testbed for engine or mod development`

games/minimal/mods/basetools/init.lua
```diff
@@ -0,0 +1,295 @@
...
+--
+minetest.register_tool("basetools:dagger_steel", {
+	description = "Steel Dagger",
+	inventory_image = "basetools_steeldagger.png",
+	tool_capabilities = {
+		full_punch_interval = 0.5,
+		max_drop_level=0,
+		damage_groups = {fleshy=2},
+	}
+})
```
appgurueu, line 295: While no huge issue, you could get rid of lots of redundant code by having a table with tool tiers and their stats and then looping over that for registration.

games/minimal/mods/experimental/commands.lua
```diff
@@ -0,0 +1,215 @@
...
+	minetest.chat_send_player(name, "Placing nodes …")
+	local nodes_placed = 0
+	for n=1, #nodes do
+		local itemstring = nodes[n]
+		local def = minetest.registered_nodes[itemstring]
+		local p2_max = 0
+		if param.param ~= "no_param2" then
+			-- Also test the param2 values of the nodes
+			-- ... but we only use permissible param2 values
+			if def.paramtype2 == "wallmounted" then
```
appgurueu, line 123: If-chain can be replaced by using lookup table (small code quality note).

games/minimal/mods/experimental/commands.lua
```diff
@@ -0,0 +1,215 @@
...
+			end
+		end
+		-- Emerge area to make sure that all nodes are being placed.
+		-- Note we will emerge much more than we need to (overestimation),
+		-- the estimation code could be improved performance-wise …
+		local length = 16 + math.ceil(emerge_estimate / 24) * 2
+		minetest.emerge_area(start_pos,
+			{ x = start_pos.x + 46, y = start_pos.y, z = start_pos.z + length },
+			after_emerge, { nodes = nodes, name = name, pos = pos, start_pos = start_pos, param = param })
+		return true, "Emerging area …"
```
appgurueu, line 212: Might want to replace the Unicode `…` with ASCII `...` (three dot characters).

games/minimal/mods/experimental/mod.conf
```diff
@@ -1,3 +1,2 @@
 name = experimental
-description = Minimal mod to test features
-depends = default, stairs
+description = Chaotic mod containing unstructured tests for testing out engine features. The features in this mod should be moved to other mods.
```
appgurueu, line 2: I see three tests in here: node callbacks, node meta and particles. Node callbacks & meta could be merged to a node functionality testing mod?

games/minimal/mods/soundstuff/mod.conf
```diff
@@ -0,0 +1,2 @@
+name = soundstuff
```
appgurueu, line 1: Probably rename to `audiovisuals` or something similar.

games/minimal/mods/testentities/armor.lua
```diff
@@ -0,0 +1,41 @@
+-- Armorball: Test entity for testing armor groups
+-- Rightclick to change armor group
+
+local phasearmor = {
+	[0]={icy=100},
+	[1]={firy=100},
```
appgurueu, line 6: `[1]` .. `[4]` (everything except `[0]`) can be omitted. Might be more readable this way, though.

games/minimal/mods/testentities/mod.conf
```diff
@@ -0,0 +1,2 @@
+name = testentities
```
appgurueu, line 1: You could leave the `test` in the name away, just call it `entities` - this entire game is a testbed ;)

games/minimal/mods/testfood/mod.conf
```diff
@@ -0,0 +1,2 @@
+name = testfood
```
appgurueu, line 1: Again, the `test` could probably be omitted.

games/minimal/mods/testformspec/callbacks.lua
```diff
@@ -0,0 +1,51 @@
...
+	description = "Test: Change formspec callbacks testing mode",
+	func = function(name, param)
+		local mode = tonumber(param)
+		if not mode then
+			callback_test = (callback_test + 1 % 3)
+		else
+			callback_test = mode
+		end
+		if callback_test == 1 then
+			minetest.chat_send_player(name, "Formspec callback test mode 1 enabled: Logging only")
```
appgurueu, line 43: Probably just return instead of `minetest.chat_send_player`? Would also save you the `else`s.

games/minimal/mods/testformspec/mod.conf
```diff
@@ -0,0 +1,2 @@
+name = testformspec
```
appgurueu, line 1: ~~test~~ - probably put this in a modpack named test, but having test in all names is irritating and redundant...

games/minimal/mods/testformspec/formspec.lua
```diff
@@ -0,0 +1,381 @@
...
+		"dropdown[0,6;2;hmdrpdwn;apple,bulb;1]"..
+		"image_button[0,4;2,2;testformspec_button_image.png;imagebutton;bbbbtt;false;true;testformspec_pressed.png]"..
+		"box[1,22.5;4,1;#a00a]"..
+	"scroll_container_end[]"..
+	"scrollbaroptions[max=170]".. -- lowest seen pos is: 0.1*170+6=23 (factor*max+height)
+	"scrollbar[7.5,0;0.3,4;vertical;scrbar;0]"..
+	"scrollbar[8,0;0.3,4;vertical;scrbarhmmm;0]"..
+	"dropdown[0,6;2;hmdrpdwnnn;Outside,of,container;1]"
+
+--style_type[label;textcolor=green]
```
appgurueu, line 231: Outcommented code should either be removed or made optional using configuration variables...

games/minimal/mods/testnodes/mod.conf
```diff
@@ -0,0 +1,2 @@
+name = testnodes
```
appgurueu, line 1: ~~test~~ again, and this mod could probably incorporate some of the tests currently in `experimental`?

games/minimal/mods/testpathfinder/mod.conf
```diff
@@ -0,0 +1,2 @@
+name = testpathfinder
```
appgurueu, line 1: ...

games/minimal/mods/testtools/mod.conf
```diff
@@ -0,0 +1,2 @@
+name = testtools
```
appgurueu, line 1: ...

games/minimal/mods/unittests/crafting.lua
```diff
@@ -0,0 +1,120 @@
...
+	input = {
+		method = "cooking",
+		width = 1,
+		items = {"unittests:iron_lump"}
+	}
+	minetest.log("info", "[unittests] iron lump cooking input: "..dump(output))
+	output, decremented_input = minetest.get_craft_result(input)
+	minetest.log("info", "[unittests] iron lump cooking output: "..dump(output))
+	minetest.log("info", "[unittests] iron lump cooking decremented input: "..dump(decremented_input))
+	assert(output.time)
```
appgurueu, line 68: Probably `type(output.time) == "number"` ?

games/minimal/mods/unittests/random.lua
```diff
@@ -0,0 +1,10 @@
+function unittests.test_random()
+	-- Try out PseudoRandom
+	minetest.log("action", "[unittests] Testing PseudoRandom ...")
+	local pseudo = PseudoRandom(13)
```
appgurueu, line 4: Should work differently; PseudoRandom is not guaranteed to always return those numbers, it should only return the same numbers each run, depending on the seed. So I'd suggest going with something like:
```lua
for i = 1, 13 do
    local pseudo = PseudoRandom(i)
    local pseudo2 = PseudoRandom(i)
    for j = 1, 10 do
         assert(pseudo:next() == pseudo2:next())
    end
end
```
But we can probably drop this test altogether.

games/minimal/mods/experimental/commands.lua
SmallJoker: `drawtype` checks are needed too, which defeats lookup tables. The code works as-is and is acceptable in size. No change needed.

games/minimal/mods/basetools/init.lua
SmallJoker: This is subjective code style. Serves its purpose and tables add unnecessary effort for little benefit here.

Wuzzy2:
I'm not really convinced by any of appgurueu's complaints. Any core dev opinions on this?

games/minimal/mods/testnodes/mod.conf
SmallJoker: This decision was likely made to avoid mod name conflicts, but also to clarify it's a mod for testing. I don't see an issue there.

Wuzzy2:
And yes, the reason why I made the automated tests disabled by default is because of the Pinciple of Minimal Interference. If you want to have this test, just enable the corresponding setting.
I don't agree that having this setting is a bad thing.

The goal is to make Minimal useful for engine testing, but ALSO for mod testing. Therefore, screwing as little as possible with default engine behavior *by default* is the most sensible thing to do.

rubenwardy:
You made a merge mistake - `test/formspec.lua` exists with no `init.lua`. This file was moved to `testformspec/init.lua`

Wuzzy2:
`test` mod is removed now. I should probably squash everything when we're done here, do you agree?

rubenwardy:
I haven't tested everything, but I'm happy enough that this is an improvement

games/minimal/mods/unittests/random.lua
sfan5: This is actually guaranteed.

sfan5:
* infinite placing not enabled by default if `creative_mode = true`
* keeping the original dirt/grass walk sounds would've been nice
* liquid test node is not `buildable_to` 
* I can't figure out what the formspec tester node and item do
* Sunlight filter doesn't seem to do anything
* Pathfinder tool always prints "Destination too far away."
* Break sound tool is useless since it doesn't wear
* Zoom is disabled
* Somewhat disappointingly, lava didn't kill me
* Lava does not generate underground (isn't this a default mapgen feature?)
* Glasslike Liquid level node is useless without a way to change the level

Wuzzy2:
> infinite placing not enabled by default if creative_mode = true

There's a setting to enable infinite placing. Creative Mode by default does nothing, except for whatever the engine does.

> keeping the original dirt/grass walk sound would've been nice

It annoyed me. If footstep sounds are added back, they should be added to many more nodes.

> liquid test node is not buildable_to

Valid.

> I can't figure out what the formspec tester node and item do

These are dummy items for the image formspec elements. I should probably add code comments.

> Sunlight filter doesn't seem to do anything

It's a node that lets light through but not sunlight (`sunlight_propagates=false`)

> Pathfinder tool always prints "Destination too far away."

Read the README.

> Break sound tool is useless since it doesn't wear

Turn off Creative Mode.

> Zoom is disabled

OK, this is bad. Maybe I add a command or item.

> Somewhat disappointingly, lava didn't kill me

Damage intentionally disabled so you can fly in lava to check out mapgen.

> Lava does not generate underground (isn't this a default mapgen feature?)

It does definitely generate in v6. Lava does not glow, however.
I'm not sure about other mapgens, however.

> Glasslike Liquid level node is useless without a way to change the level

Use the Param2 Tool. This also applies to all other drawtype test nodes.

Wuzzy2:
I updated the `devtest` repository (not the PR).

* Liquid nodes are now `buildable_to`
* Lava now glows
* Added code comments for some of the more esoteric nodes
* Added `/zoom` command
* Improved error message when pathfinder destination is too far

About pathfinder distance: This is a safeguard I have added to avoid freezing Minetest for extremely far distances. There are a couple of in-code variables to configure and/or disable the safeguard if you like to play around with this. Maybe the distance limit is too low (please comment)?

Wuzzy2:
This PR has been updated.
I'm not sure how I should proceed now. Do I have to squash?

Wuzzy2:
I have just changed the behavior of infinite node placement:

By default, infinite node placement will be enabled in Creative Mode.

This behavior can be controlled by the setting `devtest_infplace`, which is now an enum (`true`, `false` or `auto`).

sfan5:
> It's a node that lets light through but not sunlight (sunlight_propagates=false)

I know. Turns out it does work, but not on a small scale:
screenshot_20200428_160938

> Turn off Creative Mode.

Ah of course. It could warn me about this though.

>  I'm not sure about other mapgens, however.

Couldn't find any with v7, v6 generates lava though so that's fine.

> Do I have to squash?

Not necessary, can be done on merge.

Wuzzy2:
> > Turn off Creative Mode.
> 
> Ah of course. It could warn me about this though.

Disabling tool wear is (as far I know) the only thing that the engine does when Creative Mode is enabled. And this is hardcoded behavior. Everthing else about Creative comes from mods/games. It's basically just the engine being weird. I believe it has been proposed in the past to change this weirdness, but this is unrelated to DevTest.

SmallJoker:
> Damage intentionally disabled so you can fly in lava to check out mapgen.

In this case the player can disable damage in the main menu, or you can add a command to set the `immortal` group on a player. If creative respects the settings, why not damage too?

 * Creative inventory would be very helpful. Previously `unified_inventory` was very handy, but it depends on `default`.
   * Also no search function within the "Chest of Everything". Finding things takes way too long
 * Aliases to `default: ` nodes would be very helpful to use old testing world in the new game
 * Ice could made slippery for gameplay comparisons (mainly `LocalPlayer` changes)
 * `testtools:remover` is currently superfluous. The initial mese pickaxe digs everything but liquids
 * No thin slabs or L-shaped nodeboxes for collision and sneak testing -> moreblocks does not work, `default` mod missing

Wuzzy2:
OMG. If you are constantly adding feature request after feature request, this PR will never get finished. :-( This amount of ridiculous perfectionism is very demotivating. Sorry, at one point the feature requests have to stop.

> Creative inventory would be very helpful.

I intentionally did not do this so the original player invnetory does not have to be overridden. The chest of everything allows me to have a formspec without replacing the player inventory formspec.

But this is a huge feature request anyway, way too big for this PR.

> Previously unified_inventory was very handy, but it depends on default.

That's totally `unified_inventory`'s fault, not mine. :P

> Also no search function within the "Chest of Everything". Finding things takes way too long.

Good idea, but it's a lot of work. This feature request is too big for me, at least for this PR. At one point the feature requests have to stop. This PR is already starting to become very unmanagable at this huge size. Let's move this to future PRs that will not only be smaller but also much less painful to review.

> Trash slot is not obvious, missing labels

Will be added, as it's trivial to do.

> Aliases to default: nodes would be very helpful to use old testing world in the new game

I consider it as it's cheap to do.

> Ice could made slippery for gameplay comparisons (mainly LocalPlayer changes)

There are already several nodes that test slipperiness. Note that the `basenodes` are there to test mapgen, not node properties. Nodes to test node stuff are in `testodes`.

> testtools:remover is currently superfluous. The initial mese pickaxe digs everything but liquids

Incorrect. It's special in that it calls `minetest.remove_node`, the hardcore method to remove nodes (bypassing all checks). `dignodes:none` is a node without groups, therefore it cannot be dug by mining tools, that's where this tool is useful.
This tool also removes entities with a single punch, which is very useful.

Reading the README files would have helped you btw. :P

> No thin slabs or L-shaped nodeboxes for collision and sneak testing -> moreblocks does not work, default mod missing

Again, this is another big feature request, too big for this PR. Sorry, it's not my fault that moreblocks depends on default. :P

Part of the reason why I removed `default` is to force mods to move away from the ugly `default` dependency. :-) Since this rework is also meant to be usable by mods, this can be a great way to test if you mod actually works in an `default`-less environment. Besides, `default` is mostly a MTG thing. The MTG `default` and the Minimal `default` are already different anyway.

Wuzzy2:
OK, new commit added.

But I'm not sure if I should really add aliases for `default`. I'm not happy in adding a bias towards MT. It conflicts with the goal of minimal interference IMO. I'm waiting for other core devs to weigh in.

And frankly, I don't care at all about compability (neither does paramat). It's a test game. Worlds are expected to break all the time.

Wuzzy2:
Oops. I meant to say “I'm not happy in adding a bias towards MTG.”.

LoneWolfHT:
Couldn't just pack sfinv and the creative mod?

Wuzzy2:
Maybe, but as I said, at one point the feature requests just get too much. This PR just grows and grows and grows … I have been developing `devtest` for 5 MONTHS now, and it's still not good enough for the quality standards of Minetest. For something that a normal player will never see.

Take it, or leave it. I will gladly fix bugs or annoying weirdnesses but that's it for now. I'm getting exhausted. Feel free to post your feature request in future issues.

This PR is already a huge improvement over current Minimal. Don't you agree? It seems everyone expects me now to turn Minimal from kinda shit to absolutely perfect in a single PR. No, just no.

Wuzzy2:
Besides, I consider `sfinv` and `creative` mods to be quite “heavyweight”. They also overwrite the player inventory, which I want to avoid. Principle of Minimal Interference (look it up in the README).

Even if I still had motivation left, I woudn't do it.

Wuzzy2:
Anyway. Is there anything left that is blocking this PR?

paramat:
Is there a reason for the delayed merge?

I agree with Wuzzy2, time to merge, more can always be done later.
> Aliases to default

No thanks, it is pointless to support old testing worlds, they have zero value.
>  I consider sfinv and creative mods to be quite “heavyweight”.

They are, best left out of this game.

Wuzzy2:
> Is there a reason for the delayed merge?

Yes, it's because nobody that is allowed to merge actually bothers to merge. XD There's no other reason I can think of.

Well, except this PR has collected only two core dev approvals, and has been reviewed over the course of only 6 months, so …

I totally understand that this might be way too early. Let's wait 6 months for no reason so this PR collects more conflicts and must be rebased for the 100th time. Two core dev approvals are also clearly not enough. We must first wait for literally everyone on the Minetest community to give their approval. If anyone disagrees, delete this PR without trace, it will be better to completely start from scratch. It's less efficient that way, which is a good thing because efficiency gives us a false sense of actually accomplishing something. A good programmer is one who works more hours, not less. I'm obviously writing PRs only for PR's sake, not to actually get them into Minetest, because that would actually make sense. The endless development circle has to continue. The Minetest development process is still too efficient, this has to change. ;-)

> After merge, a follow-up PR to change thename to 'devtest'? I think the game name is best changed to actually be descriptive.

Yes. Easy-peasy.

Wuzzy2:
Working on renaming the game to devtest. Stay tuned for an upcoming PR!

paramat:
I will make a PR to update the registered biomes to the latest form, including defined cave liquids and dungeon nodes. Then non-V6 mapgens will have lava (and water) caves, and the dungeons will have stairs.

paramat:
> Lava does not generate underground (isn't this a default mapgen feature?)

Tested, it does in non-V6 mapgens, below y = -256.
However this is due to an engine fallback, the nodes should be defined by the biomes, which i will do in my PR.

PR #9459, opened 2020-03-04
Fixes around ServerActiveObject on_punch handling

sfan5:
* Fixes <i>2.</i> from this comment https://github.com/minetest/minetest/pull/8959#issuecomment-594632035
* Fixes inaccurate hpchange reason (currently irrelevant, since unused)
* Document return value of LuaEntity `on_punch`
* Fix ObjectRef being unusable at the time `on_death` is called

## To do

### DIFF
--- a/src/network/serverpackethandler.cpp
+++ b/src/network/serverpackethandler.cpp
@@ -1166,6 +1166,8 @@ void Server::handleCommand_Interact(NetworkPacket *pkt)
 			u16 wear = pointed_object->punch(dir, &toolcap, playersao,
 					time_from_last_punch);
 
+			// Callback may have changed item, so get it again
+			playersao->getWieldedItem(&selected_item);
 			bool changed = selected_item.addWear(wear, m_itemdef);
 			if (changed)
 				playersao->setWieldedItem(selected_item);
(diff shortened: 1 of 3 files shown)


MoNTE48:
...continue the discussion...
@sfan5 although this made the `mobs_redo` code work again, why does the builtin wear processing still not work?
Pls, add Blocker / High priority labels.

sfan5:
Works for me:
gif

MoNTE48:
@sfan5 You're right. But only after I set `punch_attack_uses`.
Also does not work with mobs_redo when i remove `hitter:set_wielded_item(weapon)`.
But that is a different issue.

src/network/serverpackethandler.cpp
```diff
@@ -1150,8 +1150,8 @@ void Server::handleCommand_Interact(NetworkPacket *pkt)
 			if (pointed_object->isGone())
 				return;
 
-			ItemStack selected_item, hand_item;
-			ItemStack tool_item = playersao->getWieldedItem(&selected_item, &hand_item);
+			ItemStack selected_item;
+			ItemStack tool_item = playersao->getWieldedItem(&selected_item, nullptr);
```
SmallJoker: This will ignore custom `hand` tools (toolcaps). Intended?

sfan5: Oh right, providing `&hand_item` changes the behavior of `getWieldedItem` even if you leave the return unused doesn't it?

sfan5: Indeed, fixed.

PR #9462, opened 2020-03-05
Add support for statbar “off state” icons

### LINKED ISSUE #6030: Add statbar background icons
I suggest to add background icons for the statbars by default. This includes the heart and bubble icons, but also the statbar API.

**Rationale:**
Your brain can not instantly tell how much health you have *without counting each heart one-by-one* here:

Some hearts. Maybe 9? Or 10? Or just 8?

But with background icons, you can:

Obviously 9 hearts

Oh, how often have I already pointlessly wasted food because of eating at full health! :grinning: 

**Technical notes:**
First, in the HUD element definition, another field for a background icon must be added. For health and air, the default image files could be `heart_bg.png` and `bubble_bg.png`, respectively.
Also, each statbar must know its maximum count, obviously.

**Suggested rendering algorithm:**
* On a “classic” health bar (20 half hearts), it could work like this:
    * The number of total half hearts (both “active” and “inactive”) rendered equals the maximum statbar size. In this case: 20
    * An “active” (red) heart is always drawn by drawing `heart.png`.
    * An “inactive” (gray) heart is always drawn by drawing `heart_bg.png`.
    * A “half” heart is drawn by drawing the left heart of `heart.png`, directly followed by the right half of `heart_bg.png`.
    * Exception: No background image is rendered if the required variable set to `nil`.

Note this rendering algorithm intentionally does *not* involve any texture overlays. This is done to give artists more artistic freedom. (OK, pedantically, that would mean there would also be a separate icon for a “half” symbol (half active, half inactive)) Of course, the same algorithm applies to bubbles and custom statbars.

**Justification for engine feature (rather than mod feature):**
This will be an obvious and easy improvement for all subgames. Nobody will seriously miss the current hard-to-count statbars. But there is still backwards-compability in mind.

**Data files for you:**
Here you can find free background icons for free, in the HUD Bars mod: https://forum.minetest.net/viewtopic.php?f=11&t=11153

HUD Bars icons

Wuzzy2:
Closes #6030.

This adds support for optional “off state” icons for statbars. “off state icons” can be used to denote the lack of something, like missing hearts or bubbles.

## Implementation

This implementation is 100% network-compatible.

~~The “off state” image is set by using the `text` field of the HUD definition. To add the “off state” image, append a slash to the original texture name, then the additional texture name to the `text` field of the HUD definition. Example: `heart.png/heart_gone.png`. Here, `heart.png` is the normal icon, and `heart_gone.png` is the “off state”. If no slash is found, the background image is disabled.~~

The “maximum value” is set by using the `item` field.

For the builtin health bar and breath bar, `health_gone.png` and `breath_gone.png` are used, respectively. Note the default images are empty, so games need to add these images first.

### Implementation rationale and details
~~The slash was chosen as this is character is forbidden in most OSes.~~

I would have loved to provide non-empty default `_gone` images, but decided aginst that because adding visible images would break compability of existing games. Existing games don't have those images yet and add default images would not fit well together with the game-provided `heart.png`/`bubble.png`.

Technically, the “off state” does not act as a background of the normal image. Instead, the “off state” icons *replace* the normal icons. I decided against using overlays because it would have made drawing semitransparent icons (like bubbles) quite annoying.

## To do

This PR is ready for review.

## How to test

Use this test mod and use the `/statbar` command to generate a random statbar (random image count and direction):

```
local bar = nil

minetest.register_chatcommand("statbar", {
	privs = {},
	description = "Render test statbar",
	params = "[<textures>] | clear",
	func = function(name, param)
		if not player then
			return false, "No player"
		end
		if bar then
			player:hud_remove(bar)
			bar = nil
		end
		if param == "clear" then
			return true
		end
		if param == "" then
			param = "heart.png/heart_gone.png"
		end
		local dir = math.random(0,3)
		local max = math.random(1,15)
		local num = math.random(0,max)
		minetest.chat_send_player(name, "direction="..dir.."; value="..num.."/"..max)
		bar = player:hud_add({
			hud_elem_type = "statbar",
			number = num,
			item = max,
			size = { x = 24, y = 24 },
			position = { x = 0.5, y = 0.5 },
			offset = { x = - 12, y = -12 },
			text = param,
			direction = dir,
		})
		return true
	end,
})
```

Additionally, you want to test the behavior or the builtin health and breath statbars. You need to replace `heart_gone.png` and `bubble_gone.png` for testing purposes because the default ones are empty/transparent.


MoNTE48:
Server 5.2, client 5.0. The client will see only "on state", the behavior will not change?

builtin/game/statbars.lua
```diff
@@ -4,8 +4,9 @@ local enable_damage = core.settings:get_bool("enable_damage")
 local health_bar_definition = {
 	hud_elem_type = "statbar",
 	position = { x=0.5, y=1 },
-	text = "heart.png",
+	text = "heart.png/heart_gone.png",
```
sfan5: I know this is extremely convenient, but no.
The two texture names need to be separate fields.

Wuzzy2:
@sfan5: OK, I introduced a new field `text2` for the 2nd texture name.

@MoNTE48: The alt. texture will not appear by default. Read the whole 1st post again.

src/network/clientpackethandler.cpp
```diff
@@ -1104,6 +1105,11 @@ void Client::handleCommand_HudAdd(NetworkPacket* pkt)
 	}
 	catch(PacketError &e) {}
 
+	try {
+		*pkt >> text2;
+	}
+	catch(PacketError &e) {}
```
sfan5: All the previous try-catch blocks should be merged together actually

src/network/networkprotocol.h
```diff
@@ -568,6 +568,7 @@ enum ToClientCommand
 		v3f1000 world_pos
 		v2s32 size
 		s16 z_index
+		u8[len2] text2
```
sfan5: `u32 len3` (actually not sure if its u16)
`u8[len3] text2`

Wuzzy2:
@sfan5: Done.

src/network/clientpackethandler.cpp
sfan5: By <i>all</i> I really meant all of them.
I don't know why the other catches `PacketError`, just make it catch `SerializationError`.

src/client/hud.cpp
```diff
@@ -380,16 +381,28 @@ void Hud::drawLuaElements(const v3s16 &camera_offset)
 }
 
 
-void Hud::drawStatbar(v2s32 pos, u16 corner, u16 drawdir, const std::string &texture,
-		s32 count, v2s32 offset, v2s32 size)
+void Hud::drawStatbar(v2s32 pos, u16 corner, u16 drawdir,
+		const std::string &texture, const std::string &bgtexture,
+		s32 count, s32 maxcount, v2s32 offset, v2s32 size)
 {
+	const char separator = '/';
```
sfan5, line 388: unused

src/client/hud.cpp
```diff
@@ -380,16 +381,28 @@ void Hud::drawLuaElements(const v3s16 &camera_offset)
...
 		return;
 
+	video::ITexture *stat_texture_bg;
+	bool uses_bg_texture = false;
+	if (!bgtexture.empty()) {
+		uses_bg_texture = true;
+		stat_texture_bg = tsrc->getTexture(bgtexture);
+	}
+	if (!stat_texture_bg)
+		uses_bg_texture = false;
```
sfan5, line 404: the bool is not necessary here, just use `stat_texture_bg == nullptr` to indicate absence

doc/texture_packs.txt
```diff
@@ -64,6 +64,8 @@ by texture packs. All existing fallback textures can be found in the directory
 
 * `bubble.png`: the bubble texture when the player is drowning
                 (default size: 12×12)
+* `bubble_gone.png`: like `bubble.png`, but denotes lack of breath
+                     (empty by default)
```
sfan5, line 68:
```suggestion
                     (transparent by default, same size as `bubble.png`)
```

Wuzzy2:
Done, rebased and squashed!

doc/texture_packs.txt
```diff
@@ -64,6 +64,8 @@ by texture packs. All existing fallback textures can be found in the directory
 
 * `bubble.png`: the bubble texture when the player is drowning
                 (default size: 12×12)
+* `bubble_gone.png`: like `bubble.png`, but denotes lack of breath
+                     (transparent by default)
```
sfan5, line 68: still doesn't say that the size should be the same

sfan5:
<pre>
textures/base/pack/heart.png:      PNG image data, 12 x 12, 4-bit colormap, non-interlaced
textures/base/pack/heart_gone.png: PNG image data, 12 x 12, 4-bit colormap, non-interlaced
</pre>
grafik

uh?

src/client/hud.cpp
```diff
@@ -411,6 +413,11 @@ void Hud::drawStatbar(v2s32 pos, u16 corner, u16 drawdir, const std::string &tex
 	if (!stat_texture)
 		return;
 
+	video::ITexture *stat_texture_bg;
```
sfan5, line 416:
```suggestion
	video::ITexture *stat_texture_bg = nullptr;
```

Wuzzy2:
Weird about the heart thing. Also tested in Minimal with game-provided `heart.png` removed. Still not reproducible. Note that `heart.png` and `bubble.png` are not even touched in this PR.

Wuzzy2:
Note that MTG uses the wrong size for both `heart.png` and `bubble.png` (16×16 instead of 12×12).

SmallJoker:
@Wuzzy2 Would you please be so nice and rebase this PR?

Wuzzy2:
This is ready now!

Wuzzy2:
Looks good.

ClobberXD:
The new `item` field is somewhat non-descriptive, IMHO. If it denotes the maximum value, why not call it `max`?

SmallJoker:
@sfan5 Outdated review. Does the PR need any chance, or does it look good?

src/network/networkprotocol.h
```diff
@@ -573,6 +573,8 @@ enum ToClientCommand
 		v3f1000 world_pos
 		v2s32 size
 		s16 z_index
+		u32 len3
+		u8[len3] text2
```
sfan5, line 577: This reminds me some of these are inaccurate, `len`, `len2` and `len3` are u16 not u32

sfan5:
Still looks good, you can fix the comment before merge if you want.

src/network/clientpackethandler.cpp
```diff
@@ -1102,22 +1102,17 @@ void Client::handleCommand_HudAdd(NetworkPacket* pkt)
...
-	try {
 		*pkt >> size;
-	} catch(SerializationError &e) {};
-
-	try {
 		*pkt >> z_index;
+		*pkt >> text2;
 	}
-	catch(PacketError &e) {}
+	catch(SerializationError &e) {};
```
SmallJoker, line 1115: Needs to catch `PacketError`. Why was this changed?
Discovered while testing another PR.

PR #9465, opened 2020-03-05
Fix linking failures when compiled with Clang 9

sfan5:
Clang has broken `-ffast-math` on glibc in their 9.0 release and they don't plan on fixing it until 10.x comes out.
Sooner or later someone is going to hit this bug and complain to us, so it's better to add a workaround right now.

https://bugs.llvm.org/show_bug.cgi?id=45034


SmallJoker:
:+1:

PR #9476, opened 2020-03-07
Add LevelDB auth database.

luk3yx:
- Goal of the PR

To allow LevelDB to be used as the auth database (`auth.txt`/`auth.sqlite`).

No.

This provides an alternative authentication database format for people who do not want to use sqlite but don't want the performance hit of using the legacy text-based format.

## To do

This PR is ready for review.

## How to test

Create a new world and add `auth_backend = leveldb` to world.mt.


src/database/database-leveldb.cpp
```diff
@@ -97,5 +98,101 @@ void Database_LevelDB::listAllLoadableBlocks(std::vector<v3s16> &dst)
...
+}
 
+bool AuthDatabaseLevelDB::saveAuth(const AuthEntry &authEntry)
+{
+	std::ostringstream os;
+	writeU8(os, 0);
+	os << serializeString(authEntry.password);
+
+	size_t privilege_count = authEntry.privileges.size();
+	FATAL_ERROR_IF(privilege_count > 0xFFFF,
```
weqqr:
```suggestion
	FATAL_ERROR_IF(privilege_count > U16_MAX,
```

src/database/database-leveldb.cpp
```diff
@@ -97,5 +98,101 @@ void Database_LevelDB::listAllLoadableBlocks(std::vector<v3s16> &dst)
...
 
-#endif // USE_LEVELDB
+AuthDatabaseLevelDB::AuthDatabaseLevelDB(const std::string &savedir)
+{
+	leveldb::Options options;
+	options.create_if_missing = true;
+	leveldb::Status status = leveldb::DB::Open(options,
+		savedir + DIR_DELIM + "auth.db", &m_database);
+	ENSURE_STATUS_OK(status);
+	m_savedir = savedir;
```
weqqr: Looks like `m_savedir` isn't used anywhere.

luk3yx: Thank you for pointing this out.

src/database/database-leveldb.cpp
```diff
@@ -97,5 +98,100 @@ void Database_LevelDB::listAllLoadableBlocks(std::vector<v3s16> &dst)
...
+	u8 version = 0
+	std::string password
+	u16 number of privileges
+	for each privilege {
+		std::string privilege
+	}
+	s64 last_login
+	*/
+
+	if (readU8(is) != 0)
```
SmallJoker: For future changes, following code would help to reduce missing adjustments:
`if (readU8(is) > AUTH_VERSION)` where `AUTH_VERSION` is a macro defined above (and undefined later).

src/database/database-leveldb.cpp
```diff
@@ -97,5 +98,100 @@ void Database_LevelDB::listAllLoadableBlocks(std::vector<v3s16> &dst)
...
+	leveldb::Status s = m_database->Delete(leveldb::WriteOptions(), name);
+	return s.ok();
+}
+
+void AuthDatabaseLevelDB::listNames(std::vector<std::string> &res)
+{
+	leveldb::Iterator* it = m_database->NewIterator(leveldb::ReadOptions());
+	res.clear();
+	for (it->SeekToFirst(); it->Valid(); it->Next()) {
+		res.push_back(it->key().ToString());
```
sfan5: should use `emplace_back`

src/database/database-leveldb.cpp
```diff
@@ -97,5 +98,100 @@ void Database_LevelDB::listAllLoadableBlocks(std::vector<v3s16> &dst)
...
+	for each privilege {
+		std::string privilege
+	}
+	s64 last_login
+	*/
+
+	if (readU8(is) != 0)
+		return false;
+
+	res.id = 1;
```
sfan5: is it safe to use a constant ID?
does the auth code (or anything else) use this?

src/database/database-leveldb.cpp
luk3yx: Fixed, however the sqlite3 database uses `push_back` as well: https://github.com/minetest/minetest/blob/625b1005939e76d6a5b2d241635be3b8a89eb128/src/database/database-sqlite3.cpp#L750-L758

src/database/database-leveldb.cpp
luk3yx: The plaintext database uses a constant ID.
https://github.com/minetest/minetest/blob/625b1005939e76d6a5b2d241635be3b8a89eb128/src/database/database-files.cpp#L263-L269

src/database/database-leveldb.cpp
nerzhul, line 133: first version must be 1.
After that i will add my approval and we can merge

luk3yx: Done.

nerzhul:
thanks

PR #9489, opened 2020-03-09
Fix memory leak in guihypertext

Desour:
- The dynamically allocated GUIHyperText in parseHyperText wasn't deallocated before. And dropping it would have resulted in double-freeing of an attribute of ParsedText (`m_root_tag`).
`m_vscrollbar` wasn't dropped either.
- Note: I've taken this from #9487 so that this can be merged in feature freeze.

## To do

## How to test

Look for segfaults, use valgrind, idk.
Or just look at the code, it should be obvious.
Formspec I've used for testing:
```lua
local my_formspec =
	"formspec_version[3]"..
	"size[10,8]"..
	"hypertext[1,1;5,5;htxt;Bla]"..
	""
```

### DIFF
--- a/src/gui/guiFormSpecMenu.cpp
+++ b/src/gui/guiFormSpecMenu.cpp
@@ -1662,8 +1662,9 @@ void GUIFormSpecMenu::parseHyperText(parserData *data, const std::string &elemen
 	);
 
 	spec.ftype = f_Unknown;
-	(new GUIHyperText(spec.flabel.c_str(), Environment, this, spec.fid, rect,
-			m_client, m_tsrc))->drop();
+	GUIHyperText *e = new GUIHyperText(spec.flabel.c_str(), Environment, this,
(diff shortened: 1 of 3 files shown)


src/gui/guiFormSpecMenu.cpp
```diff
@@ -1662,8 +1662,8 @@ void GUIFormSpecMenu::parseHyperText(parserData *data, const std::string &elemen
 	);
 
 	spec.ftype = f_Unknown;
-	new GUIHyperText(
-		spec.flabel.c_str(), Environment, this, spec.fid, rect, m_client, m_tsrc);
+	(new GUIHyperText(spec.flabel.c_str(), Environment, this, spec.fid, rect,
```
sfan5: please use a variable here

Desour: Done.

PR #8964, opened 2019-09-19
set_sky improvements, set_sun, set_moon and set_stars

Jordach:
Implements Player only controls for the sun, moon, stars and skybox.

- To allow full mod or game control over the default sky colours, sun, moon and stars.
- By adding Player specific functions that allow for fine control over the sun, moon, stars and sky.
- Removes Irrlicht's texture filtering on the sun, moon, to allow for low resolution textures.
- Has full backwards compatibility with the original set_sky call, and hides the sun, moon and stars when used for textured skyboxes.
- Supports changing the colours of the default skybox, day, dawn, night and indoors, as well as sun / moon tinting at dawn and sunrise.

## To do:

Ready for Review.

## How to test

@sofar 's skybox mod: https://content.minetest.net/packages/sofar/skybox/
An experimental version of atmos containing all four calls: [atmos.zip](https://github.com/minetest/minetest/files/3633436/atmos.zip)

## Fancy Screenshots:

(EDIT by paramat: Note these are all of a skybox using 6 supplied textures.)

screenshot_20190822_014327
screenshot
screenshot
screenshot
screenshot

PR #9502, opened 2020-03-10
Color gradient for default and 'regular' type sky

TheTermos:
Adds sky color gradient which extends from the horizon up to roughly 25 degrees above it.
Should also increase client performance somewhat, as one of the 'far cloudy fog things' has been removed as redundant.
It works with default and type='regular' sky. Tested with #8964 in effect.

Change of default colors is not essential for this PR, however I put it up for serious consideration.
- sky color, previously it extended from the horizon line all the way to the zenith. Here, flat sky color occupies only the top part of the sky 'dome', so in my opinion, it's no longer justified for it to be this dull.
- background color, it's dirty gray, not a good choice because fog inherits this color effectively overlaying everything with it, I think it has to go.

The picture below uses my proposed colors. I'm not claiming they are the best possible, but I think it's the right direction. Please don't judge the colors based on the image, check em out in game.

grad2

## To do

This PR is Ready for Review.

Please consider not doing coding style reviews. This PR doesn't add a single line of code, it's all modifications and deletions, all coding style is received

### DIFF
--- a/src/client/sky.cpp
+++ b/src/client/sky.cpp
@@ -252,7 +252,6 @@ void Sky::render()
 		if (m_visible) {
 			driver->setMaterial(m_materials[1]);
 			for (u32 j = 0; j < 4; j++) {
-				video::SColor c = cloudyfogcolor.getInterpolated(m_skycolor, 0.45);
 				vertices[0] = video::S3DVertex(-1, -0.02, -1, 0, 0, 1, m_bgcolor, t, t);
 				vertices[1] = video::S3DVertex( 1, -0.02, -1, 0, 0, 1, m_bgcolor, o, t);
 				vertices[2] = video::S3DVertex( 1, 0.45, -1, 0, 0, 1, m_skycolor, o, o);
--- a/src/client/sky.cpp
+++ b/src/client/sky.cpp
@@ -253,34 +253,10 @@ void Sky::render()
 			driver->setMaterial(m_materials[1]);
 			for (u32 j = 0; j < 4; j++) {
 				video::SColor c = cloudyfogcolor.getInterpolated(m_skycolor, 0.45);
-				vertices[0] = video::S3DVertex(-1, 0.08, -1, 0, 0, 1, c, t, t);
-				vertices[1] = video::S3DVertex( 1, 0.08, -1, 0, 0, 1, c, o, t);
-				vertices[2] = video::S3DVertex( 1, 0.12, -1, 0, 0, 1, c, o, o);
-				vertices[3] = video::S3DVertex(-1, 0.12, -1, 0, 0, 1, c, t, o);
-				for (video::S3DVertex &vertex : vertices) {
-					if (j == 0)
-						// Don't switch
-						{}
-					else if (j == 1)
-						// Switch from -Z (south) to +X (east)
-						vertex.Pos.rotateXZBy(90);
-					else if (j == 2)
-						// Switch from -Z (south) to -X (west)
-						vertex.Pos.rotateXZBy(-90);
-					else
-						// Switch from -Z (south) to +Z (north)
-						vertex.Pos.rotateXZBy(-180);
-				}
-				driver->drawIndexedTriangleFan(&vertices[0], 4, indices, 2);
-			}
-
-			// Draw far cloudy fog thing at and below all horizons
-			for (u32 j = 0; j < 4; j++) {
-				video::SColor c = cloudyfogcolor;
-				vertices[0] = video::S3DVertex(-1, -1.0, -1, 0, 0, 1, c, t, t);
-				vertices[1] = video::S3DVertex( 1, -1.0, -1, 0, 0, 1, c, o, t);
-				vertices[2] = video::S3DVertex( 1, 0.08, -1, 0, 0, 1, c, o, o);
-				vertices[3] = video::S3DVertex(-1, 0.08, -1, 0, 0, 1, c, t, o);
+				vertices[0] = video::S3DVertex(-1, -0.02, -1, 0, 0, 1, m_bgcolor, t, t);
+				vertices[1] = video::S3DVertex( 1, -0.02, -1, 0, 0, 1, m_bgcolor, o, t);
+				vertices[2] = video::S3DVertex( 1, 0.45, -1, 0, 0, 1, m_skycolor, o, o);
+				vertices[3] = video::S3DVertex(-1, 0.45, -1, 0, 0, 1, m_skycolor, t, o);
 				for (video::S3DVertex &vertex : vertices) {
 					if (j == 0)
 						// Don't switch
--- a/src/skyparams.h
+++ b/src/skyparams.h
@@ -76,12 +76,12 @@ public:
 	{
 		SkyColor sky;
 		// Horizon colors
-		sky.day_horizon = video::SColor(255, 149, 219, 255);
+		sky.day_horizon = video::SColor(255, 144, 211, 246);
 		sky.indoors = video::SColor(255, 100, 100, 100);
 		sky.dawn_horizon = video::SColor(255, 186, 193, 240);
 		sky.night_horizon = video::SColor(255, 64, 144, 255);
 		// Sky colors
-		sky.day_sky = video::SColor(255, 101, 188, 253);
+		sky.day_sky = video::SColor(255, 97, 181, 245);
 		sky.dawn_sky = video::SColor(255, 180, 186, 250);
 		sky.night_sky = video::SColor(255, 0, 107, 255);
 		return sky;
--- a/src/skyparams.h
+++ b/src/skyparams.h
@@ -76,12 +76,12 @@ public:
 	{
 		SkyColor sky;
 		// Horizon colors
-		sky.day_horizon = video::SColor(255, 155, 193, 240);
+		sky.day_horizon = video::SColor(255, 149, 219, 255);
 		sky.indoors = video::SColor(255, 100, 100, 100);
 		sky.dawn_horizon = video::SColor(255, 186, 193, 240);
 		sky.night_horizon = video::SColor(255, 64, 144, 255);
 		// Sky colors
-		sky.day_sky = video::SColor(255, 140, 186, 250);
+		sky.day_sky = video::SColor(255, 101, 188, 253);
 		sky.dawn_sky = video::SColor(255, 180, 186, 250);
 		sky.night_sky = video::SColor(255, 0, 107, 255);
 		return sky;


src/client/sky.cpp
```diff
@@ -253,34 +253,10 @@ void Sky::render()
 			driver->setMaterial(m_materials[1]);
 			for (u32 j = 0; j < 4; j++) {
 				video::SColor c = cloudyfogcolor.getInterpolated(m_skycolor, 0.45);
```
sfan5: unused

sfan5:
This isn't a new thing, but I think the gradient should not be visible during night time. The (top) skybox should be an uniform dark color.
screenshot

The day_sky color is a bit too bright for my taste. Making it less dull is good but I'd prefer a "healthy" mix e.g. `(255, 122, 187, 252)`

paramat:
Although this may surprise some, i am fine with changing the stepped horizon colour change to a gradient, but there is a certain way to do this i prefer and that was discussed with Jordach during one of his previous sky PRs which implemented a gradient in this way.
It looks like you are already some way towards what i am about to suggest.

To confirm what this PR does, it seems that ('horizon') colour below horizon is unchanged and flat, the ('sky') colour above 25 degrees above horizon is unchanged and flat, and from horizon to 25 degrees up is a gradient between the two.

The discussion i had with Jordach was that i feel it is important that a colour gradient near the horizon continues to give a clear 'sense of horizon position', for when the player is high up and there is no rendered terrain creating a horizon, which is a very common situation. Without this the world will appear disorientating and lacking a sense of 'what is level', as well as of course, lacking a visual suggestion of a horizon.

So the colour gradient needs to create a noticeable horizon position, and therefore it needs to be fairly narrow, although i am not suggesting it needs to be as narrow as the previous stepped transition. However, 25 degrees is large, i will test your PR to see how it is but i suspect the gradient will need to be narrower.

> It works with default and type='regular' sky

Those 2 are the same thing, 'regular' is what you use to change the sky type back to the standard design.
> sky color, previously it extended from the horizon line all the way to the zenith. Here, flat sky color occupies only the top part of the sky 'dome', so in my opinion, it's no longer justified for it to be this dull.

Actually this logic works the other way around. Real skies are brighter near the horizon, as is MT's sky.
With a larger transition zone (as in this PR), the larger the brighter zone is, which would allow the flat colour sky zone above to be darker.

Real sky gets darker towards zenith, so if the flat colour sky zone is moved towards zenith it could actually be made darker. Making it brighter makes no sense at all, and your logic here does not either.
Anyway, i am not suggesting the flat sky colour is made darker, but there is now less reason to make it brighter, as the gradient is now providing more of the desired brightness.

I do not mind the sky becoming slightly brighter, but what your screenshot shows is too bright for me, although i need to test in-game.

> background color, it's dirty gray, not a good choice because fog inherits this color effectively overlaying everything with it, I think it has to go.

It is obviously not a 'dirty gray' =) It is a hazy blue-white, the hue is blue. If you look at it in a graphics program it is a pleasant colour.

Colours expressed as HSV:

Old horizon: H 213 (blue) S 35 V 94
PR horizon: H 200 (greenish-blue) S 42 (richer colour) V 100 (maximum brightness)

Old sky: H 215 (blue) S 44 V 98
New sky: H 206 (greenish-blue) S 60 (richer colour) V 99 (slightly brighter)

I will test the PR soon.

paramat:
> This isn't a new thing, but I think the gradient should not be visible during night time. The (top) skybox should be an uniform dark color.

'top' is confusing, do you mean the entire sky being one colour?
I think we still need a suggested horizon at night, for the same reasons i gave above. Space mods or games can alter sky to make it uniform if desired.
Moonlit night terrain still has a noticeable horizon, otherwise it will look like deep space and be very disorientating.
So, i strongly disagree.

src/client/sky.cpp
```diff
@@ -253,34 +253,10 @@ void Sky::render()
...
-						// Switch from -Z (south) to -X (west)
-						vertex.Pos.rotateXZBy(-90);
-					else
-						// Switch from -Z (south) to +Z (north)
-						vertex.Pos.rotateXZBy(-180);
-				}
-				driver->drawIndexedTriangleFan(&vertices[0], 4, indices, 2);
-			}
-
-			// Draw far cloudy fog thing at and below all horizons
```
paramat: Careful ... there may be a reason why this farcloudyfogthing is drawn down to -1.0, even though it is later overlaid by another layer at 'Draw far cloudy fog thing below all horizons in front of sun, moon and stars.'.
There is often a reason for the way things are.
It is possible that you cannot remove a drawing stage like this.

src/client/sky.cpp
```diff
@@ -253,34 +253,10 @@ void Sky::render()
...
-			for (u32 j = 0; j < 4; j++) {
-				video::SColor c = cloudyfogcolor;
-				vertices[0] = video::S3DVertex(-1, -1.0, -1, 0, 0, 1, c, t, t);
-				vertices[1] = video::S3DVertex( 1, -1.0, -1, 0, 0, 1, c, o, t);
-				vertices[2] = video::S3DVertex( 1, 0.08, -1, 0, 0, 1, c, o, o);
-				vertices[3] = video::S3DVertex(-1, 0.08, -1, 0, 0, 1, c, t, o);
+				vertices[0] = video::S3DVertex(-1, -0.02, -1, 0, 0, 1, m_bgcolor, t, t);
+				vertices[1] = video::S3DVertex( 1, -0.02, -1, 0, 0, 1, m_bgcolor, o, t);
+				vertices[2] = video::S3DVertex( 1, 0.45, -1, 0, 0, 1, m_skycolor, o, o);
+				vertices[3] = video::S3DVertex(-1, 0.45, -1, 0, 0, 1, m_skycolor, t, o);
```
paramat: The previous transition was from horizon up to 0.12.
To maintain a clear horizon position, and for some degree of continuity of design, i think this 0.45, being roughly 4 times larger, is far too large.
It seems a rough equivalent to before would be twice of 0.12, so i suggest 0.24, however i need to test because this may not work well for some reason with a gradient, it may need to be larger than 0.24.

TheTermos:
> The day_sky color is a bit too bright for my taste. Making it less dull is good but I'd prefer a "healthy" mix e.g. `(255, 122, 187, 252)`

This one is cool too, I like mine a bit better but that may be because of different displays we're using.

grad_sfan



> This isn't a new thing, but I think the gradient should not be visible during night time. The (top) skybox should be an uniform dark color.

This can be achieved the simplest by setting both sky and background the same color, if it gets accepted.

sfan5:
> 'top' is confusing, do you mean the entire sky being one colour?

By "top" I meant everything above the horizon should be one color.

> This can be achieved the simplest by setting both sky and background the same color

Sure, but I'd prefer if the sky code did this on its own rather than mods having to change sky colors as soon as night starts/ends.

TheTermos:
> Sure, but I'd prefer if the sky code did this on its own rather than mods having to change sky colors as soon as night starts/ends.

Yes, I meant setting default night sky colors, in the sky code.

As to mods, since #8964 they don't have to detect time of day, night/day sky colors can be set separately, so they can do whatever, night sky and background can be equal or not.

src/client/sky.cpp
TheTermos: There often is, but it doesn't seem to be the case here, both 'things' are drawn under the same condition.

I don't know how well gpus handle fill rate these days, but drawing two large, opaque surfaces on top of one another probably isn't a great idea.

src/client/sky.cpp
TheTermos: > however i need to test because this may not work well for some reason with a gradient, it may need to be larger than 0.24.

Please do,
In fact, the horizon line doesn't seem to do anything for me, would be great if if it could be done away with because it messes up the fog.

TheTermos:
> Actually this logic works the other way around. Real skies are brighter near the horizon, as is MT's sky.

Current MT background color isn't brighter than the sky color at all, it's just desaturated, then there's less than 7 degrees of transition, so it's basically flat color all the way from just above the horizon, so no, it isn't anywhere near real sky.

> Real sky gets darker towards zenith, so if the flat colour sky zone is moved towards zenith it could actually be made darker. Making it brighter makes no sense at all, and your logic here does not either.

Brightness is relative, so the right question is, darker than what? 
Not the random old color obviously, it should be darker than the sky color just above the horizon.
And it is, just look at it.

But it still should be brighter than average, because during the day the atmosphere is a light source afterall, otherwise it looks like it's going to rain.
How much brighter exactly, I hope we come to some sort of agreement about that here.

src/client/sky.cpp
appgurueu: I agree with @TheTermos - if it can't be justified now, it shouldn't be left there. If it's overlaid, there's room for optimization, and should be optimized.

src/client/sky.cpp
```diff
@@ -253,34 +253,10 @@ void Sky::render()
 			driver->setMaterial(m_materials[1]);
 			for (u32 j = 0; j < 4; j++) {
 				video::SColor c = cloudyfogcolor.getInterpolated(m_skycolor, 0.45);
-				vertices[0] = video::S3DVertex(-1, 0.08, -1, 0, 0, 1, c, t, t);
-				vertices[1] = video::S3DVertex( 1, 0.08, -1, 0, 0, 1, c, o, t);
-				vertices[2] = video::S3DVertex( 1, 0.12, -1, 0, 0, 1, c, o, o);
-				vertices[3] = video::S3DVertex(-1, 0.12, -1, 0, 0, 1, c, t, o);
-				for (video::S3DVertex &vertex : vertices) {
```
appgurueu: Are you sure that you can remove those transformations?

paramat:
Comparison screenshots of master against PR at ground level and at alititude:

comp_master_PR45

comp_master_PR45_s

Colours expressed as HSV:

Old horizon: H 213 S 35 V 94
PR horizon: H 200 S 42 V 100

Old sky: H 215 S 44 V 98
PR sky: H 206 S 60 V 99

Colours: Hue
////////////

The hues have changed, becoming slightly more cyan-blue. I did not see a need for a change of hues but the new hues look good, so i am neutral about these changes.

Colours: Saturation
///////////////////

The saturations (richness of colour) are increased. This improves the upper sky area, it was quite hazy-whitish before. Supported.

Colours: Brightness
///////////////////

Oddly enough the sky colour is not actually any brighter (V 99 versus 98, and visually no brighter). The extra brightness comes from the horizon colour and the blended area.
I find the horizon colour perhaps a little too bright. However, this is a difficult balance, as a lower brightness would reduce the contrast against the sky colour, and this contrast is needed to make the horizon position clear.
So, maybe the horizon colour should be a little darker, unsure about this.

Gradient
////////

Supported.

Gradient size
/////////////

PR at altitude:

PR45s

Due to the large gradient size this is now disorientating, with a very unclear horizon position. There is not much sense of the world below, or of horizon, or of atmospheric haze.

I tried 0.24 as i suggested earlier, but that was not enough.
In master the full strength horizon colour extends up to 0.08, so it seemed that twice that, 0.16, would be a rough equivalent. I tried this and here is the comparison with master:

PR16s

comp_master_PR16_s

Now the horizon position is clear at the base of the gradient, creating visual orientation.
There is a sense of being at high altitude with a world below, a sense of a horizon, and of atmospheric haze above the horizon.
This also helps at ground level to create a sense of distant hazy world.
This is also consistent with the previous transition area size.

This PR is a radical change of sky design. I feel there needs to be at least some kind of continuity of design from before, for obvious reasons.
The change to a gradient is good and i support this.
The changes to colours seem generally good, with perhaps slight changes needed.
However, the change to the size of the transition area is not necessary, and there are very good reasons to keep it as before.
So i think the size of the transition area should be what provides some continuity of design.

I disapprove the PR with the current 0.45 :-1:
I will probably approve the PR if it uses 0.16 or something very close to that.

src/client/sky.cpp
paramat: These removed lines are part of what i am saying below to be careful about removing.

TheTermos:
> Due to the large gradient size this is now disorientating, with a very unclear horizon position. There is not much sense of the world below, or of horizon, or of atmospheric haze.

The problem with this opinion is, it's untestable.
As I said earlier, the horizon line does nothing for me, I could argue it's existence is actually more disorienting, because there's terrain, and then it vanishes completely in the fog, and then there's the line which shouldn't really be there because of the fog, that looks just like a disconnected flat vertical surface (which it is).

In my opinion, fake horizon makes sense only if fog distance is greater than terrain draw distance, so it never gets completely covered by fog.

Both opinions are subjective and untestable and there are no no good ways of arguing them, but there's one objective consideration, the fog: It is single, flat color, so too much color variation in the horizon area makes it unclear what color terrain should blend into.

The size of gradient area and fake horizon are unrelated things, shrinking the gradient area would defeat the purpose, so here's a better bad idea:

horizon1

Top edge of the bottomest fogthing has been colored with background blended with 15% sky, this creates faint but well defined horizon line, and gradient area can be as wide as it needs to be.

Still, my opinion is drop the fake horizon altogether, it's useless.

src/client/sky.cpp
TheTermos: As sure as I can be based on code analysis and some testing, these two things are drawn either both together or none.

appgurueu:
What about a vote regarding this (as it's definetely a question of taste, and "de gustibus non est disputandum") ? Probably on Minetest Forum?

I would rather prefer a smooth gradient instead of a horizon. But as said, this is mostly personal preference, and we need a vote to justify decisions.

TheTermos:
> > Moonlit night terrain still has a noticeable horizon, otherwise it will look like deep space and be very disorientating.
> 
> Hmm, does it? I thought it didn't.

In fact, it never does, horizon is an **imaginary** line.
All you ever see is the edge of visible terrain, which just sometimes happens to be close to the horizon. 

And this is what occurs in MT naturally, only often closer than irl which is consistent with way shorter view ranges, just look:

horizon2

Terrain edge forms perfect line exactly where it should be based on view distance and camera elevation. Does the fake horizon look like it belongs there?

Now look below how it has been in MT. Can anyone honestly say that grayish stuff between water and sky on the right looks like anything? Do we really need that?

sky_before

sfan5:
> Does the fake horizon look like it belongs there?

Have to agree, it looks out of place there. I am fine with the large gradient introduced with this PR.

TheTermos:
**New evidence regarding fake horizon**

A recent forum post: https://forum.minetest.net/viewtopic.php?f=3&t=24431
> Why do the sun and moon appear to rise and set a couple of degrees _above_ the horizon?

The person evidently doesn't perceive the line as the horizon, instead their reference seems to be the edge of visible tarrain.

@paramat, you're the only one who claimed fake horizon was useful because its absence was disorienting, but actually it seems to be the opposite.
Do you still stand by that opinion?

Wuzzy2:
And what if my game doesn't want a gradient and it liked the old, simple style more?

rubenwardy:
> And what if my game doesn't want a gradient and it liked the old, simple style more?

set_sky can customise it. The default sky behaviour is not a supported part of the API

paramat:
TheTermos
> there's terrain, and then it vanishes completely in the fog, and then there's the line which shouldn't really be there because of the fog

It should be there, and this was by design by celeron55. The horizon line is there to suggest 'distant hazy world, distant clouds, atmospheric haze and true horizon' because the edge of the loaded terrain is almost always a large angle below the true horizon. This is why it is called 'farcloudyfogthing'.
> The size of gradient area and fake horizon are unrelated things

The smaller gradient makes the position of the horizon more defined, as my screenshots show. A large gradient area makes the horizon les defined. They are not unrelated. This is visually obvious and not subjective or untestable.
> so here's a better bad idea:
Top edge of the bottomest fogthing has been colored with background blended with 15% sky, this creates faint but well defined horizon line, and gradient area can be as wide as it needs to be.

That looks far worse to me because the horizon is now too obvious, it also clashes with loaded terrain edge if that just happens to be close to the true horizon, as your later screenshots show.
A small gradient area works well because it very subtly suggests where the horizon is.
> Terrain edge forms perfect line exactly where it should be based on view distance and camera elevation

The terrain edge is almost always a large angle below the true horizon, because view distance is limited and the player is often high up. On a mountan the terrain edge is often below 45 degrees down. So it is not "exactly where it should be" if you mean that is an accurate true horizon.
Without the true horizon, the terrain edge gets mistakes for the horizon, which looks terrible and is disorientating.
> Does the fake horizon look like it belongs there?

No, it looks bad in those screenshots, but only because you are using a sharp-edged horizon line coded for the purpose of this discussion. It would look fine in MT master.
> Now look below how it has been in MT. Can anyone honestly say that grayish stuff between water and sky on the right looks like anything? Do we really need that?

Yes it looks like distant hazy world, distant clouds, atmospheric haze and horizon. This is by design and has been this way for 8 years with very little complaint. Something like this is needed because it looks far worse without.

@sfan5 
> Have to agree, it looks out of place there.

Are you aware that the screenshot you refer to is not MT master? It contains a sharp-edged horizon line that the author coded for the purpose of this discussion. And i agree it looks bad. You cannot judge based on that screenshot.
> New evidence regarding fake horizon
The person evidently doesn't perceive the line as the horizon, instead their reference seems to be the edge of visible tarrain.

This is only 1 person who misperceives, and seeing the terrain edge as the horizon is ridiculous, it should be obvious it is not the true horizon. This is hardly representative of all MT users.
1 person is hardly 'evidence', there may be 10000+ users who are not seeing a problem here and are quietly satisfied.

Besides, the forum thread continues:
Thread author:
> Why do the sun and moon appear to rise and set a couple of degrees _above_ the horizon?

Answer:
> It is because the loaded area ends at some point and unloaded areas look the same as air. If you loaded infinite area, the Sun and Moon would rise and set exactly at the horizon.

Thread author:
> Ok, I'll buy that one. It makes enough sense for going on with.

So they are fairly satisfied now.

> paramat, you're the only one who claimed fake horizon was useful because its absence was disorienting, [...] Do you still stand by that opinion?

Yes certainly, because it is visually disorientating. It does not matter if no-one agrees with me, that will still be the case. If there is no clear horizon, you lose orientation, this is simply a fact.

Now about gradient size:
I tried 0.24, it did not seem ideal but it was an improvement, so i am ok with some value around that.
How the PR is now at 0.45 is far preferable to this:
> Top edge of the bottomest fogthing has been colored with background blended with 15% sky, this creates faint but well defined horizon line,

Also i should clarify, 1 disapproval cannot not stop a PR, if the PR then gets 2 approvals it will be merged.

Please can you reduce the brightness of the lower half of the sky a little? It seems just a little too bright, to sfan5 too.

src/client/sky.cpp
paramat: Yes, it is probably ok.
appgurueu, i am only saying 'be careful', i am not opposing this change.

src/client/sky.cpp
paramat: I tested 0.24 and it was still too large for me, 0.16 is a good match for classic sky and does a good job of visually suggesting distant hazy world, atmospheric haze and horizon, see my screenshots in the discussion.
I prefer 0.16 but 0.24 is acceptable.

paramat:
> Both night sky colors get multiplied by quite low brightness value, the effect is they end up so similar so there's not enough variation to fill in the gradient area smoothly and flat stripes become visible. All the more reason not to bother and set them equal.

Actually, good point, so i would be ok with this. Where the stars get cutoff defines the horizon anyway.

sfan5:
> Please can you reduce the brightness of the 'horizon' colour slightly? (hue and saturation are fine) It seems just a little too bright, to sfan5 too.

benrob0329:
Wait, that gray blob is supposed to be unloaded terrain? I always thought it was a gradiant compromise due to engine limitations. Isn't the unloaded terrain thing the fog's job anyways?

TheTermos:
> So they are fairly satisfied now.

Believe it or not, I've foreseen you'd totally miss the point there.
It's irrelevant if they're satisfied with the explanation, what matters is the fact that it required explaining what the hell that line thing was.
What could be better evidence that the line doesn't serve its intended purpose. 

>  seeing the terrain edge as the horizon is ridiculous, it should be obvious it is not the true horizon.

You don't say. True horizon is an imaginary line, you never get to see it.

> Are you aware that the screenshot you refer to is not MT master?

Are you aware there are two screenshots in that post, one form MT master? Yup, both look bad, because flat fake horizon is a bad, redundant, ugly looking concept.
Distant terrain probably could be adequately faked with a texture, or a mesh - a better option, never flat color.

> This is only 1 person who misperceives

Don't know about that, based on this thread so far it looks like it's you who's the one misperceiving person.

> it is visually disorientating. It does not matter if no-one agrees with me, that will still be the case. If there is no clear horizon, you lose orientation, this is simply a fact.

Nope, this is made up nonsense, here are some facts.
- You hardly ever see anything close to the horizon line IRL unless you're a sailor or a farmer on the plains of Arizona or some such. In urban setting and hilly areas people can go on for weeks without seeing it once
- You only sometimes see the line in MT.

If our orientation relied on seeing a line close to the horizon we'd be screwed, but luckily this is not the case.

> Please can you reduce the brightness of the 'horizon' colour slightly? (hue and saturation are fine) It seems just a little too bright, to sfan5 too.

I could, if you stopped making it look like a waste of time for silly reasons.

paramat:
Actually ... the gradient top being located at 0.24 is fine for me, i will approve the PR with that.
0.16 is quite small. 0.24 makes the gradient more visually subtle but still gives a subtle sense of where the horizon is:

PR24s

TheTermos:
> Actually ... the gradient top being located at 0.24 is fine for me, i will approve the PR with that.

Yup, you said that before.
This PR is about introducing gradient and getting rid of the line. Gradient is already narrower than I'd like it to be, only because it's painted onto vertical walls, we'd need a proper sky dome for it to be wider.
You seem to be trying to hijack this PR to do exactly the opposite, squash the gradient even more to form a blurry line nobody else seems to need.

You don't have to approve, I'm only asking you to not get in the way for a change.

appgurueu:
General consensus seems to be that this PR is mostly supported (only saw :+1: reactions).

sfan5:
I'd still like to see the day_sky color changed to the one I suggested in https://github.com/minetest/minetest/pull/9502#issuecomment-597332587

TheTermos:
> I'd still like to see the day_sky color changed to the one I suggested in [#9502 (comment)](https://github.com/minetest/minetest/pull/9502#issuecomment-597332587)

Sure, the proposed colors aren't set in stone, I just want to double check the collision code and then come back to this.
Maybe I'll propose another compromise set of colors I stumbled upon in the meantime.

TheTermos:
My final suggested colors:
sky: 97, 181, 245
horizon: 144, 211, 246

Horizon has been slightlty desaturated, it looks less vibrant but it goes much better with distant fog
screenshot_20200501_203118

@sfan5 please see how you like em in game, my sky is slightly darker but more saturated compared to your suggested color.
I like mine better obviously, but now that sky colors can be changed easily, I'll go with yours if you insist.

sfan5:
your horizon + your sky: [screenshot_20200502_121832](https://user-images.githubusercontent.com/1042418/80861762-cfdfd500-8c70-11ea-8938-909054fb7a77.jpg)
your horizon + my sky: [screenshot_20200502_121937](https://user-images.githubusercontent.com/1042418/80861772-da9a6a00-8c70-11ea-8260-a564f817df5f.jpg)

The darkened sky indeed looks much better.

src/client/sky.cpp
TheTermos: done

paramat:
Tested new colours, very nice, not too bright. Thanks.
The slightly darker 'horizon' colour makes the height of the gradient a little less of a problem for me, it also preserves a good contrast against the clouds.

Apart from the height of the gradient i really like what this PR does, and therefore would like to approve it, but due to the gradient height unfortunately i cannot.
This might sound weird but, despite the gradient height, i think this should be merged.
So balancing all this up consider my disapproval removed.

Note that the large gradient height makes the corners of the skybox more noticeable, where the top of the gradient dips down to form a 'V' shape, noticeable when panning view.

I still prefer a smaller gradient height like 0.24 (double the height of the classic transition), for more definition, orientation and a less radical change from classic MT sky. I think we should have at least a little design continuity from classic sky.

PR #9503, opened 2020-03-10
Fix some chatcommands not returning a value

Wuzzy2:
Some chat commands in `builtin` did not have a boolean return value. This PR fixes this.

Affected commands: `me`, `shutdown`, `clearobjects` and `help`.

## To do

This PR is ready for review.

### DIFF
--- a/builtin/common/information_formspecs.lua
+++ b/builtin/common/information_formspecs.lua
@@ -136,14 +136,14 @@ help_command.func = function(name, param)
 		core.show_formspec(name, "__builtin:help_privs",
 			build_privs_formspec(name))
 		if name ~= admin then
-			return
+			return true
 		end
 	end
 	if param == "" or param == "all" then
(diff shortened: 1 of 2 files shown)


SmallJoker:
What's the difference?

Wuzzy2:
Well. The function `register_chatcommand.action` is supposed to return either true or false, according to `lua_api.txt`, to indicate success. The 4 offending function however, always return `nil`. The difference is that I added the missing return value. See “Files changed” tab …

This is basically done for completion's sake and code quality. Builtin should set a good example of what good code looks like, and be compliant with `lua_api.txt` recommendations.

Wuzzy2:
For a purely practical reason: Also, there are some mods that like to *call* the chatcommands themselves (e.g. `mesecons_commandblock`). It'd b useful to them to always know the success of chatcommands for at least all builtin commands.

appgurueu:
To give another example, it's usually useful for relays to know what happened.

PR #9505, opened 2020-03-11
Mainmenu Improvements: "Join Game" tab + code cleanup

appgurueu:
- Goal of the PR? Improves the mainmenu, in particular the "Join Game" tab.
- How does the PR work? Cleaned up & tweaked builtin mainmenu stuff.
- Does it resolve any reported issue? Works towards #6733
- If not a bug fix, why is this PR needed? What usecases does it solve? Takes less space, makes the "Join Game" tab more intuitive, cleans up code. See [To do].

## Before

Screenshot

## After

Screenshot

## To do

This PR is Ready For Review.

- [X] Remove simple main menu
- [X] Clean up code, migrate to real coordinates
- [X] Organize form, improve style
- [X] Merge gamemode indicators
- [X] Add dividers, sort servers as favorites - discover - incompatible
- [x] Network indicators: Replaced with lag + ping lagometers
- [x] Gamemode indicators: Creative + Damage combination

## How to test

Fire up Minetest.


builtin/mainmenu/common.lua
```diff
@@ -128,44 +100,35 @@ function render_serverlist_row(spec, is_favorite)
...
 
-	return details .. (grey_out and '#aaaaaa,' or ',') .. text
+	return table.concat(details, ",")
 end
 
---------------------------------------------------------------------------------
 os.tempfolder = function()
 	if core.settings:get("TMPFolder") then
-		return core.settings:get("TMPFolder") .. DIR_DELIM .. "MT_" .. math.random(0,10000)
+		return core.settings:get("TMPFolder") .. "/MT_" .. math.random(0,10000)
```
MoNTE48: Important for various OS.

builtin/mainmenu/common.lua
```diff
@@ -93,30 +66,29 @@ function render_serverlist_row(spec, is_favorite)
...
-			details = details .. "5,"
+			table.insert(details, "4")
 		end
 	else
-		details = details .. "0,"
+		table.insert(details, "0")
 	end
 
+	table.insert(details, "")
+	table.insert(details, "")
```
MoNTE48: What for?

builtin/mainmenu/common.lua
```diff
@@ -14,31 +14,18 @@
...
---------------------------------------------------------------------------------
 -- Menu helper functions
---------------------------------------------------------------------------------
-
---------------------------------------------------------------------------------
-local function render_client_count(n)
-	if     n > 99 then return '99+'
-	elseif n >= 0 then return tostring(n)
-	else return '?' end
-end
```
MoNTE48: But what if I specify the maximum client count number in a number with a hundred digits and break the table?

builtin/mainmenu/common.lua
```diff
@@ -93,30 +66,29 @@ function render_serverlist_row(spec, is_favorite)
...
-	else
-		details = "0,"
-	end
+	local details = {}
 
 	if spec.ping then
 		local ping = spec.ping * 1000
 		if ping <= 50 then
-			details = details .. "2,"
+			table.insert(details, "1")
```
MoNTE48: table.insert is about 7 times slower.

builtin/mainmenu/common.lua
appgurueu: No, `DIR_DELIM` is being deprecated for good. Lua's `stdlib` *and* Minetest are both supposed to work with `"/"`, automatically converting it to the proper delimiter, as does LuaFileSystem. I see no issue here.

builtin/mainmenu/common.lua
appgurueu: The dividers. I will probably attach a screenshot.

builtin/mainmenu/common.lua
appgurueu: No, using a rope is the recommended way if you have to do lots of string concatenations. Try LuaJIT if your performance is bad.

builtin/mainmenu/common.lua
appgurueu: Indeed, I should readd this.

builtin/mainmenu/common.lua
appgurueu: Update: done

builtin/mainmenu/common.lua
rubenwardy: There may still be bugs in Minetest that don't allow this. It's best for this to be fixed in another PR which sets DIR_DELIM to `/`

builtin/mainmenu/common.lua
rubenwardy: I believe monte means slower than `details[#details + 1] = "1"`

paramat:
I am :-1: for combining the game mode icons in this way.
See https://github.com/minetest/minetest/issues/6733#issuecomment-597900316 for why.

If done, this needs discussion in an issue first of what information needs to be shown, then how to display that in a single column, probably by using a new set of icons with superimposed elements.

appgurueu:
Well, [here goes my (disagreeing) reply](https://github.com/minetest/minetest/issues/6733#issuecomment-598116917).

rubenwardy:
The problem is that creative is a concept that is defined by minetest game. Creative, damage, and PvP is perfectly valid in certain games - the server may not have a way to heal using creative. I can imagine a PvP game where you can place infinite blocks whilst attempting to knock people off their blocks - like spleef

appgurueu:
A spleef server should be marked as just PvP. There shouldn't be three flags at all.

rubenwardy:
In which case, the order of your priorities in this pr is wrong

appgurueu:
Let me rephrase: **A server should have one gamemode.**

Current common gamemodes are creative, survival and PvP, so they are supported by the serverlist & my mainmenu improvements.

rubenwardy:
Minetest has no concept of a game mode

appgurueu:
Well, servers which fall outside of the common categories can just set none and get no image at all.

rubenwardy:
The flags are set using the setting which control the various things. It's not possible to set none and retain the functionality

MoNTE48:
In my fork I use
Mode: Creative / Survival
PVP: On / Off
The most understandable for the player!

MoNTE48:
Your X to clear the search looks so disgusting. Sorry, but it’s better to remove it.
And you can separate
Address: Port:
Name: Password:

builtin/settingtypes.txt
```diff
@@ -1413,12 +1413,6 @@ curl_file_download_timeout (cURL file download timeout) int 300000
 #    Makes DirectX work with LuaJIT. Disable if it causes troubles.
 high_precision_fpu (High-precision FPU) bool true
 
-#    Changes the main menu UI:
```
MoNTE48: You can separate it for another PR. It is deprecated and not used due to criticism of all Android players. But this is another issue.

appgurueu: I already have 2 PRs running (this and the waypoint one). This is part of mainmenu improvements and therefore included.

appgurueu:
> And you can separate [...]

Done.

appgurueu:
@kilbith Why the thumbs down?

appgurueu:
Here's my compromise: the "unknown" indicator, which is applied for creative + (damage enabled or pvp enabled).
Screenshot

MoNTE48:
> @kilbith Why the thumbs down?

@appgurueu Because it’s a complete mess, sir. The menu was so complicated. Your changes make it worse.
Other core devs also told you this

(And let the input field breathe! Or did you want to steal it?)
image

Wuzzy2:
Ugh, this is terrible.

* Minetest has no concept of “game modes”, so it is just wrong to pretend otherwise. If you want game modes, propose them in a new issue first
* Using the icon combination is bad because it introduces ambiguity. The question mark is even worse. Congratulations, you have just removed vital information from the server list! If you really insist on squashing it into one icon, make one icon for every possible combination (2^3 = 8), althrough I think that's overkill and might not even look nice (too small icon size)
* The green sidebar makes me sick just by looking at it. The contrast is really bad

The only good things that I can approve of are:
* Added a “search clear” icon
* Used icon for search button
* Added headers for favourites and non-favourites

Wuzzy2:
I thought a bit about icons. I came to the conclusion that the 3-icon system is redundant and can be reduced to 2 icons.
PvP has no effect if damage is disabled. So PvP and heart can indeed be squashed into one without loss of information. PvP icon should only be shown if also damage is active, because PvP doesn't work if damage is disabled.

This leads me to the following icon system:

```
Damage	PvP	Creative	Icons
y	y	y		Creative + Sword
y	y	n		Sword
y	n	y		Creative + Heart
y	n	n		Heart
n	y	y		Creative
n	y	n		(empty)
n	n	y		Creative
n	n	n		(empty)
```

Maybe this can be even reduced to 1 icon if you draw new icons for “Creative+Sword” and “Creative+Heart”. Maybe use the normal creative icon, but add a smaller heart/sword in the corner.

MoNTE48:
It is 16 pixels. You cannot unite many things. On devices with a high pixel density, it will not be clear what is drawn on the icon. Although generally agree with @Wuzzy2

rubenwardy:
You can reduce this into two icons - one for creative, one for damage type - and avoid the issues of combining icons

appgurueu:
If you insist on it, I'll follow the MultiCraft approach and have Creative + (PvP or Damage).

MoNTE48:
This, by chance, is not the place to say - remove fake ping and replace it with lag?
https://github.com/minetest/minetest/blob/master/src/script/lua_api/l_mainmenu.cpp#L402-L407
https://github.com/MultiCraft/MultiCraft/blob/master/src/script/lua_api/l_mainmenu.cpp#L598-L603

builtin/mainmenu/common.lua
```diff
@@ -81,8 +55,13 @@ function order_favorite_list(list)
 	return res
 end
 
---------------------------------------------------------------------------------
-function render_serverlist_row(spec, is_favorite)
+local function render_client_count(n)
+	if     n > 99 then return '99+'
```
MoNTE48: `if     n > 999 then return '999'`
_My servers easily cope with 120+ players. It was also 150+ several times!_

appgurueu:
> This, by chance, is not the place to say - remove fake ping and replace it with lag?

In fact, I have already suggested it, and will probably implement it.

There is one valid concern, though: what about servers which are not negatively affected by Lua Lag, due to the nature of their game?

builtin/mainmenu/common.lua
appgurueu: Well, in fact that is only used to render the possible client count AFAIK.

MoNTE48: And the servers look like 99+ / 99+
instead of 110 / 200.
Table cell size? Same.
Informative? Worse.
Inexperienced player thinks that on his favorite server the maximum number of players!

MoNTE48:
> what about servers which are not negatively affected by Lua Lag, due to the nature of their game?

Lag is the speed of inventory and all formspecs as well.
At Lag 300+, between pressing the inventory button and receiving the result, it will take about 0.5s, which is equal to 2 FPS.
The same problem with loading map areas, chat, chat-commands, building blocks, movements of other players...

In any case, there is more information from Lag than from `10ms fake-ping` from a Germany server to a server-list (which is located in Germany) for a Canadian or Russian players

appgurueu:
Done. Now showing lag instead of ping.

appgurueu:
Separated creative and damage.
Screenshot

appgurueu:
Have replaced the (indeed too bright) green with a far more subtle gray.

Screenshot

Other suggestions?

rubenwardy:
Please don't combine multiple things into one PR. Just change the server list list and (at most) search box in this PR

Wuzzy2:
I think you could also accept the icons as this is a lightweight change (unless you disapprove of it).

Forcing contributors to break up PRs in 1000 small sub-PRs is extremely frustrating and it doesn't really make sense anyway.

appgurueu:
So, are you happy with the lastest design?

Regarding the breakup argument: I also prefer monolithic PRs.

rubenwardy:
> I think you could also accept the icons as this is a lightweight change (unless you disapprove of it).

My issue is with the sidebar. The other changes are fine for this PR

appgurueu:
@rubenwardy No. I haven't changed the other tabs. I have changed the code so that the tabs needn't all have the same size anymore.

MoNTE48:
In general, I will try to do my little review. If anyone is interested in my vision of the interface. Many MultiCraft players have noted that it is more enjoyable than vanilla Minetest.
---

1) the clear button is useless (Ctrl + A -> Backspace does the same) In addition, the upper left corner is a very important place from the point of view of perception, and can be associated with closing, rather than clearing, the text entered in the future.
2) The input field and the search and update button have different heights. Very annoying
3) There should be free space between the buttons. Otherwise, it looks overloaded and aesthetically unpleasant. As if there’s an extra button there, which they tried their best to put in an inappropriate place for this.
4) There is too much empty space between the server icons and the name. Think of mobile players who often do not even fit the full server name.
5) Another background for the sidebar makes the interface look like 2 separate panels. It should look one. -1 for it
6) you replaced ping with a lag. Thanks for this. But you chose the value 75 to maximize the filling of the scale. Minetest uses globalstep 100 by default. <= 100 will be optimal.
7) Let the username and password entry fields breathe! Increase the distance between them by at least 0.1!
8) Resizing the tab while we use the tab interface is a bad idea. In the future, if Minetest abandons tabs, it will look better. -1 now.
9) Server description. On 4.5 "phones, even now only half of the description for most servers is placed. After your changes, nothing will fit at all. -1 for this.

---
Otherwise, I support your PR.
But there are so many things that are not related to the "Improve serverlist" ... Because of this, your PR will have a long review, constant complaints from Core Devs and will not be accepted in the near future. And you will suffer from constant rebase! Just a comment from experience.

appgurueu:
Thanks for your review. I'll address your points by their numbers:

1. No, it's not. I imagine the clear button being useful to mobile players. Furthermore, it exists to shrink the gargantuan search bar (nobody types more than 5 letters in there anyways). It's oversized currently.

2. Indeed, the height differs slightly. Could be fixed hackily. But I don't think that's the way to go. In theory, they should have the same height (as I'm using real coordinates). I'm afraid that if I tweak this, there will be differences if it's scaled. In other words: not my job, this should be fixed in the engine.

3. No, I don't think that there should necessarily be free space. Having no free space makes the buttons appear grouped to the search bar as they should. See [Bootstrap](https://getbootstrap.com/docs/4.4/components/input-group/) for examples.

4. Indeed. Will tweak.

5. Disagree. Again, grouping. Those are two separate panels.

6. Yes. And I'd like to keep it as is, "maximizing" the filling of the scale. I'm wondering whether there is another option, though: involving the ping, one could calculate a new "lag-score". Ping should mean less than lag, but shouldn't be meaningless.

7. I'd like to keep the grouping, again. No need for unused space.

8. No, it's not. The panels have enormous width, and my PR reduces the width for one panel. Not going to stretch it artificially. There are no issues with this.

9. Please elaborate. The description is displayed in the box, which should be fine.

appgurueu:
Points addressed, screenshots updated. Feedback is welcome.

src/script/lua_api/l_mainmenu.cpp
```diff
@@ -328,82 +328,42 @@ int ModApiMainMenu::l_get_favorites(lua_State *L)
...
-			lua_pushstring(L,topush.c_str());
-			lua_settable(L, top_lvl2);
-		}
-
-		if (!server["name"].asString().empty()) {
-			lua_pushstring(L,"name");
-			std::string topush = server["name"].asString();
-			lua_pushstring(L,topush.c_str());
-			lua_settable(L, top_lvl2);
+		for (const auto& string_member : {"version", "description", "name", "address", "port"})
```
sfan5: please declare these explicitly on a separate line

sfan5:
#### PR structure

You need to either split refactoring/maintenance changes into a separate PR \*or\* squash the commits in this PR so that there is a separate commit for <b>at least</b>:
* general improvements (removing `-------------------------`, changing `DIR_DELIM`, ...) to any file that isn't "tab_online.lua"
* removing the simple main menu
* improving the online tab
 
#### Design feedback

The improvements are very good overall, but I have a few points:
* the clear button is useless
* at the default window size the text fields are so close together that their borders touch
  grafik
* instead of "Discover Servers" I'd perhaps call it "Public Servers"
* and it should say "Favorited Servers" (or "Server Favorites"?) too
* the discover icon could use a better look

#### Code feedback

(see review)

builtin/mainmenu/common.lua
```diff
@@ -93,30 +72,28 @@ function render_serverlist_row(spec, is_favorite)
...
-	if spec.ping then
-		local ping = spec.ping * 1000
-		if ping <= 50 then
-			details = details .. "2,"
-		elseif ping <= 100 then
-			details = details .. "3,"
-		elseif ping <= 250 then
-			details = details .. "4,"
+	if spec.lag or spec.ping then
+		local lag = (spec.lag or 0) * 1000 + (spec.ping or 0) * 250
```
sfan5: this definitely need some kind of revision, too
I can't find a single server on the list that is below 4 bars which makes the entire indicator useless

sfan5:
> If you change the side bar, you need to then change the other tabs in the main menu.

I disagree here, it's fine for one tab to have a colored sidebar.

> I have changed the code so that the tabs needn't all have the same size anymore.

Hmm, this might be an issue though. Would it be detrimental to the design to expand the servers tab to be sized consistently with the other ones?

builtin/mainmenu/tab_online.lua
```diff
@@ -15,7 +15,7 @@
 --with this program; if not, write to the Free Software Foundation, Inc.,
 --51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 
---------------------------------------------------------------------------------
+local server_lookup
```
sfan5: almost forgot: why does this need to be global variable and which purpose does it serve?

rubenwardy:
> I disagree here, it's fine for one tab to have a colored sidebar.

It is inconsistent for one tab to add this color, and it adds nothing

sfan5:
> It is inconsistent for one tab to add this color

Is it really? None of the menu tabs have matching layout, so there is barely any consistency you can take as a measure what is allowed to be different and what isn't.

> it adds nothing

Increased visual appeal is definitely not "nothing"

rubenwardy:
I'm in favour of adding a distinction to the sidebar - this [exact same background is in my mainmenu tabs redesigns](https://user-images.githubusercontent.com/2122943/60265323-64550a80-98dd-11e9-9c4f-408cc6ddd554.png).

I'm don't like how this is done as part of this PR, or without doing other tabs.

I'm not going to oppose it in this PR, but you should at least use the correct spacings: there needs to be 0.375 padding and 0.25 spacing. Padding is the padding from an important container to its contents. In most cases, this is the formspec window, in this case the two important containers should be the server list and the sidebar. Spacing is the distance between not directly related elements, such as the search bar and the server table. Directly related elements should be touching, like the search bar and the search button. These values of padding and spacing are done semi automatically by the formspec code when you don't have real coordinates on. See the design I linked for an example

Tabs need to be the same size.

> None of the menu tabs have matching layout

The content tab does. The start game tab could be layouted to match it, but doesn't currently

rubenwardy:
Ok I've changed my mind-go ahead with the sidebar but use consistent spacing

appgurueu:
Hmm.
> There needs to be 0.375 padding and 0.25 spacing
I've currently worked with a consistent spacing of 0.25, the sidebar however has 0.125. Will increase that to 0.25.

builtin/mainmenu/tab_online.lua
appgurueu: `server_lookup` is a local variable I introduced. It basically represents the table rows. It is used while generating the table & when to identify selected entries.

sfan5:
> Regarding PR structure, I'd like to save me the hassle of creating multiple PRs/commits. Is this possible, at least if I get my design improvements accepted?

I don't understand your question but having commits as useful "units" is a hard requirement either way.
The reason for this is if e.g. the DIR_DELIM changes cause a problem, we don't want to pick apart a huge commit that touches everything related to the menu just to revert a small portion of changes.

Depending on which of your current commits touches what, it may be sufficient to squash a few to get some organization into them.

builtin/fstk/tabview.lua
```diff
@@ -56,32 +53,29 @@ local function add_tab(self,tab)
...
-		formspec = formspec ..
-				self.tablist[self.last_tab_index].get_formspec(
-					self,
-					self.tablist[self.last_tab_index].name,
-					self.tablist[self.last_tab_index].tabdata,
-					self.tablist[self.last_tab_index].tabsize
-					)
+		formspec = formspec .. tab_content
 	end
 	return formspec
```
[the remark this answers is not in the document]
appgurueu: will do when I have time

builtin/mainmenu/common.lua
```diff
@@ -93,30 +72,28 @@ function render_serverlist_row(spec, is_favorite)
...
 		else
-			details = details .. "5,"
+			table.insert(details, "4")
 		end
 	else
-		details = details .. "0,"
+		table.insert(details, "0")
 	end
 
+	table.insert(details, ",")
```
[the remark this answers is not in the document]
appgurueu: I originally had it this way, but it doesn't matter - because `table.concat({"whatever", "", "", "whatever"}, ",") == table.concat({"whatever", ",", "whatever"}, ",") == ` - and this way, it should be better for performance.

src/script/lua_api/l_mainmenu.cpp
appgurueu: this defeats the purpose of shortening the code; if the arrays are only used at one place, why explicitly declare them?

sfan5: For readability?

builtin/mainmenu/common.lua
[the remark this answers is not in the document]
sfan5: I know that, but for correctness I consider it better not to mix the separator into the real data just because it gets the right result.

Wuzzy2:
I have the impression many people misinterpret “Rebase needed” as “Please squash into one commit” … Note sure if I'm right.

appgurueu:
Opinions needed: What about having all tabs of equals size, but not fully using up the available width (leaving free space). The free space won't be as noticeable, as it will have no background there.

Wuzzy2:
I think the current tab size is fine, i.e. the tabs are as large as they need to be. This tab design is pretty standard across the board in software UIs, as far I can tell. So, no change to the tabs is needed in your PR IMO.

The real problem with Minetest tabs IMO is that they're forced to be semi-transparent, but that's completely off-topic.

sfan5:
Any progress? I think these changes are good and would like to see then in 5.4.0.

appgurueu:
Closing this. The mainmenu has to be redone entirely.

sfan5:
Who knows when that will come. I'd still prefer these improvements in the meantime.

PR #9513, opened 2020-03-14
Translate builtin

### LINKED ISSUE #7883: Translate builtin

Wuzzy2:
This PR translates `builtin`. Fixes #7883.

## To do

This PR works …

But you probably want the loading of the translation to be put somewhere else …

## How to test

* Add a translation for testing purposes: https://pastebin.com/raw/GFqK1Dpk (`builtin/locale/__builtin.de.tr`)
* Test in-game (commands, privs, help window, etc.)
* Test if all the builtin stuff still works

## In future
I will post German translation after this is merged.


MoNTE48:
Using gettext, which is used to translate the engine (C++ & builtin) for all these years it seems too complicated for you?

Wuzzy2:
What are you talking about?

MoNTE48:
https://hosted.weblate.org/projects/minetest/minetest/
https://github.com/minetest/minetest/tree/master/po
https://github.com/minetest/minetest/blob/master/builtin/mainmenu/tab_local.lua#L105-L143

Wuzzy2:
Since you continue to be cryptic, I'm assuming you want me to use Gettext for everything.

Gettext can't be used for builtin commands and privs as they are server-sent. If I use `core.gettext` on a server-sent string, this means the clients receive the *translated* string in the server's language. This would be completely wrong!
Server-sent stuff (like mods) are (currently) being translated with our mod translation system.

I know, I know. It sucks that we have this stupid `.tr` files. But this is out of scope here. Yes, it has been proposed that mod translations use the Gettext file format eventually, but this is COMPLETELY out of scope of this PR. I can't fix everything wrong in Minetest in one single PR, that's not going to happen.

SmallJoker:
Partially conflicts with #9110 (chat commands).

Wuzzy2:
@SmallJoker: So which of the PRs shall be merged first?

Wuzzy2:
Fixes #7880.

builtin/client/chatcommands.lua
```diff
@@ -7,8 +7,10 @@ core.register_on_sending_chat_message(function(message)
 	end
 
 	local first_char = message:sub(1,1)
-	if first_char == "/" or first_char == "." then
-		core.display_chat_message(core.gettext("issued command: ") .. message)
+	if first_char == "/" then
+		core.display_chat_message(core.translate("__builtin", "Issued command: @1", message))
```
sfan5, line 11: Shouldn't this be using gettext since this code runs clientside?

builtin/common/chatcommands.lua
```diff
@@ -65,9 +56,8 @@ local function do_help_cmd(name, param)
 			end
 		end
 		table.sort(cmds)
-		return true, gettext("Available commands: ") .. table.concat(cmds, " ") .. "\n"
-				.. gettext_replace("Use '$1help <cmd>' to get more information,"
-				.. " or '$1help all' to list everything.", cmd_marker)
+		return true, S("Available commands: @1", table.concat(cmds, " ")) .. "\n"
```
sfan5, line 59:
```suggestion
		return true, S("Available commands:") .. " " .. table.concat(cmds, " ") .. "\n"
```
to avoid duplicate translations

builtin/common/chatcommands.lua
```diff
@@ -76,19 +66,19 @@ local function do_help_cmd(name, param)
...
 		end
 		table.sort(privs)
-		return true, "Available privileges:\n"..table.concat(privs, "\n")
+		return true, S("Available privileges:").."\n"..table.concat(privs, "\n")
 	else
 		local cmd = param
 		local def = core.registered_chatcommands[cmd]
 		if not def then
-			return false, gettext("Command not available: ")..cmd
+			return false, S("Command not available: @1", cmd)
```
sfan5, line 81: I think this can be left as `S("Command not available:")` too

builtin/game/chat.lua
```diff
@@ -1,5 +1,8 @@
 -- Minetest: builtin/game/chat.lua
 
+local S = core.get_translator("__builtin")
+local N = function(s) return s end
```
sfan5, line 4: purpose?

builtin/game/chat.lua
```diff
@@ -52,15 +55,15 @@ core.register_on_chat_message(function(name, message)
 
 	local cmd, param = string.match(message, "^/([^ ]+) *(.*)")
 	if not cmd then
-		core.chat_send_player(name, "-!- Empty command")
+		core.chat_send_player(name, S("-!- Empty command."))
```
sfan5, line 58: Hm, perhaps the `-!-` prefix should not be part of the translated string.

builtin/game/chat.lua
```diff
@@ -109,9 +112,8 @@ end
 -- Chat commands
 --
 core.register_chatcommand("me", {
-	params = "<action>",
-	description = "Show chat action (e.g., '/me orders a pizza' displays"
-			.. " '<player name> orders a pizza')",
+	params = S("<action>"),
+	description = S("Show chat action (e.g., '/me orders a pizza' displays '<player name> orders a pizza')"),
```
sfan5, line 116:
```suggestion
	description = S("Show chat action (e.g. '/me orders a pizza' displays '* <player name> orders a pizza')"),
```

builtin/game/chat.lua
```diff
@@ -324,89 +327,88 @@ core.register_chatcommand("setpassword", {
...
 		elseif rc == 1 then
-			return true, "No such player \"" .. toname .. "\" to remove."
+			return true, S("No such player \"@1\" to remove.", toname)
 		elseif rc == 2 then
-			return true, "Player \"" .. toname .. "\" is connected, cannot remove."
+			return true, S("Player \"@1\" is connected, cannot remove.", toname)
 		end
 
-		return false, "Unhandled remove_player return code " .. rc .. ""
+		return false, S("Unhandled remove_player return code @1.", rc)
```
sfan5, line 405: Not sure if this should be translated, it's essentially debug output

builtin/game/chat.lua
```diff
@@ -500,19 +502,18 @@ core.register_chatcommand("teleport", {
...
-					.. " to " .. target_name
-					.. " at " .. core.pos_to_string(p)
+			return true, S("Teleporting @1 to @2 at @3.", teleportee_name,
+					target_name,
+					core.pos_to_string(p))
 		end
 
-		return false, 'Invalid parameters ("' .. param
-				.. '") or player not found (see /help teleport)'
+		return false, S('Invalid parameters ("@1") or player not found (see /help teleport).', param)
```
sfan5, line 510:
```suggestion
		return false, S('Invalid parameters or player not found (see /help teleport).')
```

builtin/game/chat.lua
```diff
@@ -552,26 +552,27 @@ local function emergeblocks_callback(pos, action, num_calls_remaining, ctx)
 
 	if ctx.current_blocks == ctx.total_blocks then
 		core.chat_send_player(ctx.requestor_name,
-			string.format("Finished emerging %d blocks in %.2fms.",
-			ctx.total_blocks, (os.clock() - ctx.start_time) * 1000))
+			S("Finished emerging @1 blocks in @2 ms.",
+				tostring(ctx.total_blocks),
+				string.format("%.2f", (os.clock() - ctx.start_time) * 1000)))
```
sfan5, line 557: please introduce a local variable here

builtin/game/chat.lua
```diff
@@ -552,26 +552,27 @@ local function emergeblocks_callback(pos, action, num_calls_remaining, ctx)
...
 local function emergeblocks_progress_update(ctx)
 	if ctx.current_blocks ~= ctx.total_blocks then
 		core.chat_send_player(ctx.requestor_name,
-			string.format("emergeblocks update: %d/%d blocks emerged (%.1f%%)",
-			ctx.current_blocks, ctx.total_blocks,
-			(ctx.current_blocks / ctx.total_blocks) * 100))
+			S("emergeblocks update: @1/@2 blocks emerged (@3%)",
+			tostring(ctx.current_blocks),
+			tostring(ctx.total_blocks),
+			string.format("%.1f", (ctx.current_blocks / ctx.total_blocks) * 100)))
```
sfan5, line 567: same here

builtin/game/chat.lua
```diff
@@ -648,117 +647,117 @@ local function handle_give_command(cmd, giver, receiver, stackstring)
...
-			return false, "Cannot spawn an unknown entity"
+			return false, S("Cannot spawn an unknown entity.")
 		end
 		if p == "" then
 			p = player:get_pos()
 		else
 			p = core.string_to_pos(p)
 			if p == nil then
-				return false, "Invalid parameters ('" .. param .. "')"
+				return false, S("Invalid parameters ('@1').", param)
```
sfan5, line 734:
```suggestion
				return false, S("Invalid parameters.")
```

builtin/game/chat.lua
```diff
@@ -774,157 +773,159 @@ core.register_on_punchnode(function(pos, node, puncher)
...
+	description = S("Show or set time of day"),
 	privs = {},
 	func = function(name, param)
 		if param == "" then
 			local current_time = math.floor(core.get_timeofday() * 1440)
 			local minutes = current_time % 60
 			local hour = (current_time - minutes) / 60
-			return true, ("Current time is %d:%02d"):format(hour, minutes)
+			return true, S("Current time is @1:@2",
+				tostring(hour),
```
sfan5, line 887: (unrelated, but since you are touching it anyway)
```suggestion
				string.format("%02d", hour),
```

builtin/game/chat.lua
```diff
@@ -774,157 +773,159 @@ core.register_on_punchnode(function(pos, node, puncher)
...
+		return true, S("Time of day changed.")
 	end,
 })
 
 core.register_chatcommand("days", {
-	description = "Show day count since world creation",
+	description = S("Show day count since world creation"),
 	func = function(name, param)
-		return true, "Current day is " .. core.get_day_count()
+		return true, S("Current day is @1", core.get_day_count())
```
sfan5, line 922:
```suggestion
		return true, S("Current day is @1", tostring(core.get_day_count()))
```

builtin/game/chat.lua
```diff
@@ -937,73 +938,73 @@ core.register_chatcommand("shutdown", {
 
 		if delay == 0 then
 			core.log("action", name .. " shuts down server")
-			core.chat_send_all("*** Server shutting down (operator request).")
+			core.chat_send_all(S("*** Server shutting down (operator request)."))
```
sfan5, line 941: again, prefix should probably be excluded

builtin/game/chat.lua
```diff
@@ -937,73 +938,73 @@ core.register_chatcommand("shutdown", {
...
 	privs = {ban=true},
 	func = function(name, param)
 		if param == "" then
 			local ban_list = core.get_ban_list()
 			if ban_list == "" then
-				return true, "The ban list is empty."
+				return true, S("The ban list is empty.")
 			else
-				return true, "Ban list: " .. ban_list
+				return true, S("Ban list: @1", ban_list)
```
sfan5, line 958: `S("Ban list:") .. " " .. ban_list`

builtin/game/chat.lua
```diff
@@ -1012,71 +1013,70 @@ core.register_chatcommand("clearobjects", {
...
-		core.chat_send_all("Clearing all objects. This may take a long time."
-				.. " You may experience a timeout. (by "
-				.. name .. ")")
+		core.chat_send_all(
+		S("Clearing all objects. This may take long. You may experience a timeout. (Started by @1)",
+		name))
 		core.clear_objects(options)
 		core.log("action", "Object clearing done.")
-		core.chat_send_all("*** Cleared all objects.")
+		core.chat_send_all(S("*** Cleared all objects."))
```
sfan5, line 1026: consistency
```suggestion
		core.chat_send_all(S("Cleared all objects."))
```

src/client/client.cpp
```diff
@@ -115,6 +115,29 @@ Client::Client(
...
+		} else {
+			warningstream << "No builtin translation loaded - "
+				<< "couldn't read file [" << lang << "]" << std::endl;
+		}
+	} else {
+		actionstream << "No builtin translation loaded - "
+			<< "missing LANG_CODE in locale file ["
+			<< lang << "]" << std::endl;
+	}
+
```
sfan5, line 140: please move this code to a function

src/client/client.cpp
```diff
@@ -115,6 +115,29 @@ Client::Client(
...
+		if(is.good() && is.read(&translation[0], size)) {
+			g_client_translations->loadTranslation(translation);
+			actionstream << "Builtin translation loaded ["
+				<< lang << "]" << std::endl;
+		} else {
+			warningstream << "No builtin translation loaded - "
+				<< "couldn't read file [" << lang << "]" << std::endl;
+		}
+	} else {
+		actionstream << "No builtin translation loaded - "
```
sfan5, line 136: this and the other log message are not important enough for `actionstream`

sfan5:
There are some places where a line needs wrapping because it is too long.
Also some messages e.g. "Invalid parameters." should not have a period because they aren't full sentenced, I didn't bother to mark all those.

Wuzzy2:
> Shouldn't this be using gettext since this code runs clientside?

I tested this code, only `translate` works.

> to avoid duplicate translations

No, please no string contatenation. Format strings are always better than string concatenation from the translator's view.

> purpose? [N()]

The purpose of `N()` is that this string will be collecteed by scripts while not actually translating it in this part of the code. This is very useful for tables or large string lists, etc.

> Hm, perhaps the -!- prefix should not be part of the translated string.

OK.

> There are some places where a line needs wrapping because it is too long.

Sadly not possible without breaking up the string. Adding a line break would force me to add a linebreak in the source string, which is not desirable.

> this and the other log message are not important enough for actionstream

OK.

Wuzzy2:
@sfan5: Done.

Is there anything else to do?

sfan5:
> I tested this code, only translate works.

That doesn't make it better.

> Sadly not possible without breaking up the string.

So what you are actually trying to say is that the translation collection script can't deal with string concatenations. That should rather be fixed.
The line length limit is also what is making the CI Luacheck fail.

<hr />

I only realized just now that there's also a big flaw in this PR:
The client loads translations for the version of builtin it was shipped with, not for the version of builtin the server actually has.
Even though this probably <i>mostly</i> works in practice, it is not an acceptable solution.

Wuzzy2:
Oh crap. That's indeed a big problem.
I haven't considered the possibility that the server uses a modified builtin. I just made this PR on the assumption that builtin is never touched.

I am at the end of my skill level then. Unfortunately, I'm afraid someone else has to pick up this PR. :-(

I'm so sorry.

Wuzzy2:
Any idea on how to make this PR work? I need some pointers …

Would it be enough if the server sends `__builtin.<LANG>.tr` to the clients?

SmallJoker:
> Would it be enough if the server sends __builtin.<LANG>.tr to the clients?
Yes. Builtin is much like a regular server mod.

Needs to push back the builtin path which you can get...
https://github.com/minetest/minetest/blob/7ec76e530db9c3ed093e6edeca156e9311a6378a/src/server/mods.cpp#L99-L102

.. from here. Move this function to `ServerModManager`.
https://github.com/minetest/minetest/blob/7ec76e530db9c3ed093e6edeca156e9311a6378a/src/server.cpp#L3547-L3550

Wuzzy2:
This PR has been superceded by #10693.

PR #9517, opened 2020-03-14
FormspecMenu: make drawing of backgrounds less hacky

Desour:
- Goal of the PR:
Since #8740, all backgrounds are set invisible and set visible only while they are drawn. They are drawn manually and stored separately. This is hacky and weird. The goal is to change this.
- How does the PR work?
In regenerateGui, a new element is added to hold all background elements as its children. Because it's the first element that is added, it and its children (hence all backgrounds) will be the first elements to be drawn. This means all backgrounds are in the back.
Probably not.
- Why is this PR needed?
The old way is more hacky.

## To do

(I didn't really test this yet.)

## How to test

- Look at the new `/test_formspec` tabs.
- Make screenshots of the Background tab and `compare` master with this PR.
You should see this:
scrnsht_tst_fs

### DIFF
--- a/src/gui/guiFormSpecMenu.cpp
+++ b/src/gui/guiFormSpecMenu.cpp
@@ -137,8 +137,6 @@ GUIFormSpecMenu::~GUIFormSpecMenu()
 		checkbox_it.second->drop();
 	for (auto &scrollbar_it : m_scrollbars)
 		scrollbar_it.second->drop();
-	for (auto &background_it : m_backgrounds)
-		background_it->drop();
 	for (auto &tooltip_rect_it : m_tooltip_rects)
 		tooltip_rect_it.first->drop();
 	for (auto &clickthrough_it : m_clickthrough_elements)
@@ -1124,17 +1122,15 @@ void GUIFormSpecMenu::parseBackground(parserData* data, const std::string &eleme
 		rect = core::rect<s32>(-pos, pos);
 	}
 
-	GUIBackgroundImage *e = new GUIBackgroundImage(Environment, this, spec.fid,
-			rect, name, middle, m_tsrc, clip);
+	GUIBackgroundImage *e = new GUIBackgroundImage(Environment, data->background_parent.get(),
+			spec.fid, rect, name, middle, m_tsrc, clip);
 
 	FATAL_ERROR_IF(!e, "Failed to create background formspec element");
 
 	e->setNotClipped(true);
 
-	e->setVisible(false); // the element is drawn manually before all others
-
-	m_backgrounds.push_back(e);
 	m_fields.push_back(spec);
+	e->drop();
 }
 
 void GUIFormSpecMenu::parseTableOptions(parserData* data, const std::string &element)
@@ -3059,8 +3055,6 @@ void GUIFormSpecMenu::regenerateGui(v2u32 screensize)
 		checkbox_it.second->drop();
 	for (auto &scrollbar_it : m_scrollbars)
 		scrollbar_it.second->drop();
-	for (auto &background_it : m_backgrounds)
-		background_it->drop();
 	for (auto &tooltip_rect_it : m_tooltip_rects)
 		tooltip_rect_it.first->drop();
 	for (auto &clickthrough_it : m_clickthrough_elements)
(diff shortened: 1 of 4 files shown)


rubenwardy:
how will this combine with scroll containers?

Desour:
Backgrounds can't be put into scroll containers anyway. There won't be any confilcts.

src/gui/guiFormSpecMenu.h
```diff
@@ -354,6 +353,7 @@ class GUIFormSpecMenu : public GUIModalMenu
 		std::string focused_fieldname;
 		GUITable::TableOptions table_options;
 		GUITable::TableColumns table_columns;
+		IGUIElement *background_parent;
```
SmallJoker:
```suggestion
		IGUIElement *background_parent = nullptr;
```
Needs to be initialized for security.

Desour: Done.

Desour:
Rebased.

Desour:
Squashed and rebased.

rubenwardy (approved):

src/gui/guiFormSpecMenu.cpp
```diff
@@ -3172,6 +3165,14 @@ void GUIFormSpecMenu::regenerateGui(v2u32 screensize)
 	gui::IGUIFont *old_font = skin->getFont();
 	skin->setFont(m_font);
 
+	// Add a new element that will hold all the background elements as its children
+	// because it is the first added element, all backgrounds will be behind all
+	// the other elements
```
sfan5, line 3170:
```suggestion
	// Add a new element that will hold all the background elements as its children.
	// Because it is the first added element, all backgrounds will be behind all
	// the other elements.
```

sfan5 (approved):
Looked at devtest's formspec, didn't see any changes.

kilbith:
Beware because, historically, the author is very prone to cause regressions. This must be tested with `i3` prior merging.

sfan5:
grafik
breaks MTG

rubenwardy:
Damn. I tested with Devtest and a random server

src/gui/guiFormSpecMenu.cpp
```diff
@@ -3172,6 +3165,14 @@ void GUIFormSpecMenu::regenerateGui(v2u32 screensize)
...
+	core::rect<s32> background_parent_rect(0, 0, 100000, 100000);
```
appgurueu, line 3171: This kinda looks like the cause for the bug sfan described. The numbers seem arbitrary, mustn't formspec size be factored in here?

Desour:
Good that kilbith wrote this. Additionally to me being imprudent and stupid, this is also a rather old PR in a region of code that has undergone some changes. (Also the How to test just says "todo", wtf?)

I will look into fixing this.

Desour:
Fixed.
The problem was that `GUIBackgroundImage` had used the absolute rect of its parent and not the clipped absolute rect.

Another way to handle this would be to update the desired rect and then absolute pos of the background parent element at the very end when we know the size of the formspec window. But I thought clipping is easier.
Btw. this is not the first element with a fixed huge rect, see here:
https://github.com/minetest/minetest/blob/14c7fae378fc40f88d3c430dea2cb726afc005b1/src/client/clientlauncher.cpp#L222-L228

I really should've tested this better back when I created this PR.

Desour:
Rebased.

ghost (approved):
Tested:
- devtest
- Minetest Game with formspec-heavy mods `i3` and `cartographer`. No errors found with formspec appearance (after rebasing on 8fab406, ofc) or item drop. Also tested an un-auto-clipped out of bounds background, and got identical results with master
- Mineclone 2

Also tried some private formspec code I had kicking around and got similarly-consistent results.
I should note that I _did not_ go out of my way to test code with older formspec versions, I don't have any of that on hand and I don't know the old positioning rules well enough to confirm that behavior. Ideally, this should be tested by someone with experience in that part.

On the code side, the generic element with an arbitrarily-large rect gives me some pause. Since we've done similar things before I'm not opposed to having it here, but I think this is a prime refactoring candidate. Some kind of dedicated root element class maybe?

sfan5 (approved):
Seems fine this time. Only did superficial testing, I trust Ross for the rest.

PR #9523, opened 2020-03-18
Add object crosshair and disable entity selectionboxes by default

LoneWolfHT:
## Goal of the PR
 Add an indicator on the HUD to indicate when you are pointing at an entity. Similar to Minecraft's:
 image
## How does the PR work?
 Replaces the default crosshair with a X when you are pointing at an entity
 Comparison between HUD when pointing/not pointing at an entity:
image

The object crosshair is pretty much exactly like the regular crosshair in customization, it also uses the crosshair HUD flag to determine appearance
* To define a custom texture for it use `object_crosshair.png`
* To define a custom color use the `object_crosshair_color` setting
* To define a custom alpha use the `object_crosshair_alpha` setting

This also disables entity selectionboxes by default
## Does it resolve any reported issue?
Not a reported one, but @tacotexmex has been wanting entity selectionboxes removed for a while now.
Entity selectionboxes break immersion and are better suited for dev work IMO. 
This PR will disable them by default without preventing players from knowing when they're pointing at an entity.

## To do

This PR is Ready for Review.

## How to test

* Join the game and put your crosshair over an entity while within range of it (dropped items work well)
* Make sure the object_crosshair settings (color/alpha) work
* Make sure the crosshair/object_crosshair textures work and work together
* Make sure the object crosshair doesn't appear when the HUD flag for the crosshair is disabled

I have tested all of the above without finding anything wrong

### DIFF
--- a/doc/texture_packs.txt
+++ b/doc/texture_packs.txt
@@ -74,9 +74,9 @@ by texture packs. All existing fallback textures can be found in the directory
       `crosshair_color` and `crosshair_alpha` are used to create a cross
       when no texture was found
 
-* `selectionindicator.png`
+* `object_crosshair.png`
 * the dot above the crosshair seen when pointing at an entity. Will replace crosshair when shown. The settings
-`selectionindicator_color` and `selectionindicator_alpha` are used to create a cross
+`object_crosshair_color` and `object_crosshair_alpha` are used to create a cross
 when no texture was found
 
 * `halo.png`: used for the node highlighting mesh
--- a/doc/texture_packs.txt
+++ b/doc/texture_packs.txt
@@ -72,12 +72,12 @@ by texture packs. All existing fallback textures can be found in the directory
 * `crosshair.png`
     * the crosshair texture in the center of the screen. The settings
       `crosshair_color` and `crosshair_alpha` are used to create a cross
-      when no texture was found
+      when no texture is found.
 
 * `object_crosshair.png`
     * the crosshair seen when pointing at an object. The settings
     `object_crosshair_color` and `object_crosshair_alpha` are used to create a cross
-    when no texture was found
+    when no texture is found.
 
 * `halo.png`: used for the node highlighting mesh
 
--- a/doc/texture_packs.txt
+++ b/doc/texture_packs.txt
@@ -75,7 +75,7 @@ by texture packs. All existing fallback textures can be found in the directory
       when no texture was found
 
 * `selectionindicator.png`
-* the entity selection indicator texture above the crosshair. The settings
+* the dot above the crosshair seen when pointing at an entity. Will replace crosshair when shown. The settings
 `selectionindicator_color` and `selectionindicator_alpha` are used to create a cross
 when no texture was found
 
--- a/doc/texture_packs.txt
+++ b/doc/texture_packs.txt
@@ -74,6 +74,11 @@ by texture packs. All existing fallback textures can be found in the directory
       `crosshair_color` and `crosshair_alpha` are used to create a cross
       when no texture was found
 
+* `selectionindicator.png`
+* the entity selection indicator texture above the crosshair. The settings
+`selectionindicator_color` and `selectionindicator_alpha` are used to create a cross
+when no texture was found
+
 * `halo.png`: used for the node highlighting mesh
 
 * `heart.png`: used to display the health points of the player
--- a/doc/texture_packs.txt
+++ b/doc/texture_packs.txt
@@ -75,7 +75,7 @@ by texture packs. All existing fallback textures can be found in the directory
       when no texture was found
 
 * `object_crosshair.png`
-* the dot above the crosshair seen when pointing at an entity. Will replace crosshair when shown. The settings
+* the crosshair seen when pointing at an object. The settings
 `object_crosshair_color` and `object_crosshair_alpha` are used to create a cross
 when no texture was found
 
--- a/doc/texture_packs.txt
+++ b/doc/texture_packs.txt
@@ -75,9 +75,9 @@ by texture packs. All existing fallback textures can be found in the directory
       when no texture was found
 
 * `object_crosshair.png`
-* the crosshair seen when pointing at an object. The settings
-`object_crosshair_color` and `object_crosshair_alpha` are used to create a cross
-when no texture was found
+    * the crosshair seen when pointing at an object. The settings
+    `object_crosshair_color` and `object_crosshair_alpha` are used to create a cross
+    when no texture was found
 
 * `halo.png`: used for the node highlighting mesh
 
--- a/doc/texture_packs.txt
+++ b/doc/texture_packs.txt
@@ -76,7 +76,7 @@ by texture packs. All existing fallback textures can be found in the directory
 
 * `object_crosshair.png`
     * the crosshair seen when pointing at an object. The settings
-    `object_crosshair_color` and `object_crosshair_alpha` are used to create a cross
+    `crosshair_color` and `crosshair_alpha` are used to create a cross
     when no texture is found.
 
 * `halo.png`: used for the node highlighting mesh
--- a/src/client/game.cpp
+++ b/src/client/game.cpp
@@ -3198,8 +3198,7 @@ PointedThing Game::updatePointedThing(
 	PointedThing result;
 	env.continueRaycast(&s, &result);
 	if (result.type == POINTEDTHING_OBJECT) {
-		if (!hud->can_draw_selectionindicator)
-			hud->can_draw_selectionindicator = true;
+		hud->can_draw_selectionindicator = true;
 
 		runData.selected_object = client->getEnv().getActiveObject(result.object_id);
 		aabb3f selection_box;
@@ -3231,8 +3230,7 @@ PointedThing Game::updatePointedThing(
 			result.intersection_normal.Y,
 			result.intersection_normal.Z));
 
-		if (hud->can_draw_selectionindicator)
-			hud->can_draw_selectionindicator = false;
+		hud->can_draw_selectionindicator = false;
 	}
 
 	// Update selection mesh light level and vertex colors
--- a/src/client/game.cpp
+++ b/src/client/game.cpp
@@ -3199,7 +3199,8 @@ PointedThing Game::updatePointedThing(
 	PointedThing result;
 	env.continueRaycast(&s, &result);
 	if (result.type == POINTEDTHING_OBJECT) {
-		hud->can_draw_selectionindicator = true;
+		if (!hud->can_draw_selectionindicator)
+			hud->can_draw_selectionindicator = true;
 
 		runData.selected_object = client->getEnv().getActiveObject(result.object_id);
 		aabb3f selection_box;
@@ -3231,6 +3232,11 @@ PointedThing Game::updatePointedThing(
 			result.intersection_normal.Y,
 			result.intersection_normal.Z));
 
+<<<<<<< HEAD
+=======
+		if (hud->can_draw_selectionindicator)
+			hud->can_draw_selectionindicator = false;
+>>>>>>> The PR is born
 	}
 
 	// Update selection mesh light level and vertex colors
--- a/src/client/game.cpp
+++ b/src/client/game.cpp
@@ -3193,14 +3193,13 @@ PointedThing Game::updatePointedThing(
 	const NodeDefManager *nodedef = map.getNodeDefManager();
 
 	runData.selected_object = NULL;
-	hud->can_draw_selectionindicator = false;
+	hud->can_draw_object_crosshair = false;
 
 	RaycastState s(shootline, look_for_object, liquids_pointable);
 	PointedThing result;
 	env.continueRaycast(&s, &result);
 	if (result.type == POINTEDTHING_OBJECT) {
-		if (!hud->can_draw_selectionindicator)
-			hud->can_draw_selectionindicator = true;
+		hud->can_draw_object_crosshair = true;
 
 		runData.selected_object = client->getEnv().getActiveObject(result.object_id);
 		aabb3f selection_box;
--- a/src/client/game.cpp
+++ b/src/client/game.cpp
@@ -3193,6 +3193,7 @@ PointedThing Game::updatePointedThing(
 	const NodeDefManager *nodedef = map.getNodeDefManager();
 
 	runData.selected_object = NULL;
+	hud->can_draw_selectionindicator = false;
 
 	RaycastState s(shootline, look_for_object, liquids_pointable);
 	PointedThing result;
@@ -3230,7 +3231,6 @@ PointedThing Game::updatePointedThing(
 			result.intersection_normal.Y,
 			result.intersection_normal.Z));
 
-		hud->can_draw_selectionindicator = false;
 	}
 
 	// Update selection mesh light level and vertex colors
--- a/src/client/game.cpp
+++ b/src/client/game.cpp
@@ -3193,13 +3193,13 @@ PointedThing Game::updatePointedThing(
 	const NodeDefManager *nodedef = map.getNodeDefManager();
 
 	runData.selected_object = NULL;
-	hud->can_draw_object_crosshair = false;
+	hud->pointing_at_object = false;
 
 	RaycastState s(shootline, look_for_object, liquids_pointable);
 	PointedThing result;
 	env.continueRaycast(&s, &result);
 	if (result.type == POINTEDTHING_OBJECT) {
-		hud->can_draw_object_crosshair = true;
+		hud->pointing_at_object = true;
 
 		runData.selected_object = client->getEnv().getActiveObject(result.object_id);
 		aabb3f selection_box;
--- a/src/client/game.cpp
+++ b/src/client/game.cpp
@@ -3230,12 +3230,6 @@ PointedThing Game::updatePointedThing(
 			result.intersection_normal.X,
 			result.intersection_normal.Y,
 			result.intersection_normal.Z));
-
-<<<<<<< HEAD
-=======
-		if (hud->can_draw_selectionindicator)
-			hud->can_draw_selectionindicator = false;
->>>>>>> The PR is born
 	}
 
 	// Update selection mesh light level and vertex colors
--- a/src/client/game.cpp
+++ b/src/client/game.cpp
@@ -3198,6 +3198,9 @@ PointedThing Game::updatePointedThing(
 	PointedThing result;
 	env.continueRaycast(&s, &result);
 	if (result.type == POINTEDTHING_OBJECT) {
+		if (!hud->can_draw_selectionindicator)
+			hud->can_draw_selectionindicator = true;
+
 		runData.selected_object = client->getEnv().getActiveObject(result.object_id);
 		aabb3f selection_box;
 		if (show_entity_selectionbox && runData.selected_object->doShowSelectionBox() &&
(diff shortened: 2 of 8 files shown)


paramat:
Hey nice idea.
Note: Changes to minetest.conf.example are not necessary and are wasted work, as that file is regularly autogenerated from settingtypes.txt.

LoneWolfHT:
Ok thanks. I'll remove that...

tacotexmex:
Nice improvement! 🎉

Is the position of the texture off-center? I’d suggest that it be completely centered and then offsetted in the texture itself. This way makers can choose any position of the indicator by customizing the texture.

LoneWolfHT:
Changed

rubenwardy:
I like this, but I'm concerned about how obvious this would be to the user. The sword like thing in MC looks like an entity health bar?

LoneWolfHT:
The sword is actually on the HUD most of the time, it displays the charge of the wielded tool. The plus icon below it only shows up when pointing at an entity.
I think most players will be able to figure out what is going on pretty quickly. Since they'll only notice it when pointing at an entity

tacotexmex:
Come to think of it, why not define this as an alternative crosshair feature altogether? Meaning, not add this indicator on top of the existing crosshair but instead replace it?

LoneWolfHT:
> Come to think of it, why not define this as an alternative crosshair feature altogether? Meaning, not add this indicator on top of the existing crosshair but instead replace it?

Replace it with what? Most of the time you should be able to just cover it up, the selectionindicator is drawn after the crosshair is

paramat:
> Meaning, not add this indicator on top of the existing crosshair but instead replace it?

Do you mean: the crosshair changes shape when pointing at an entity?

LoneWolfHT:
A setting for replacement could be added I guess

sorcerykid:
I really like the concept :)

But I have a dumb question: How is a selection indicator any less immersion breaking? I'm inclined to think it would be just as intuitive if not moreso if the crosshair simply changed shape to indicate that it's pointing to an entity or player. Having a dot showing the location of the entity almost seems like it would be more distracting than the selection box, esp. if the entity is moving.

Either way, it's still nice change of pace, and I welcome anything that helps to improve the gameplay experience.

LoneWolfHT:
> Having a dot showing the location of the entity almost seems like it would be more distracting than the selection box, esp. if the entity is moving.

The dot doesn't show the location of anything, it's just a dot above the crosshair that appears in the same place every time when the player can interact with an entity

sorcerykid:
Ah, thanks for the clarification, I was under the impression it was centered on the entity. If it's simply being drawn above the crosshair, then I'm curious why not simply change the crosshair image itself?

LoneWolfHT:
> Ah, thanks for the clarification, I was under the impression it was centered on the entity. If it's simply being drawn above the crosshair, then I'm curious why not simply change the crosshair image itself?

What would I change it to? Note: The default isn't an image, it's 2 lines drawn on the screen to form a '+'

src/client/game.cpp
```diff
@@ -3158,6 +3158,9 @@ PointedThing Game::updatePointedThing(
 	PointedThing result;
 	env.continueRaycast(&s, &result);
 	if (result.type == POINTEDTHING_OBJECT) {
+		if (!hud->can_draw_selectionindicator)
```
nerzhul: for me the condition are useless, just do it unconditionnaly

tacotexmex:
> Do you mean: the crosshair changes shape when pointing at an entity?

Yes. Regular crosshair would use `crosshair.png` while the entity crosshair would use `crosshair_entity.png` or similar. The default look for the inicator can still be the crosshair combined with the dot, but this way it's more versatile imo.

src/client/game.cpp
LoneWolfHT: Done

LoneWolfHT:
The selectionindicator texture now replaces the crosshair.
Really need to think up a better name for the selectionindicator...

paramat:
I have a suggestion for the form of the default selection indicator: Instead of a small solid-shaded square, use a small empty box formed by the same lines used to draw the crosshair. So 4 uses of `driver->draw2DLine()`.
This will have more design consistency, be less obstructive to vision because it is not solid-shaded, and because it looks like a box it works like an icon that suggests pointing at the entity's selectionbox.

LoneWolfHT:
I'm not sure it would indicate a selectionbox very well but I'm up for that once someone else comments, the box does seem a little intrusive

sorcerykid:
I was actually going to suggest the exact same thing as paramat, but a diamond shape instead , just to make it even more distance. Although a square is fine by me as well. It's cool that I'm not the only one that was thinking along those same lines (no pun intended :)

tacotexmex:
As long as the look is fully customizable I don’t care about the default look.

LoneWolfHT:
I like the diamond idea, I think I'll do that tonight if nobody has any objections

src/client/game.cpp
```diff
@@ -3187,6 +3189,8 @@ PointedThing Game::updatePointedThing(
 			result.intersection_normal.X,
 			result.intersection_normal.Y,
 			result.intersection_normal.Z));
+
+		hud->can_draw_selectionindicator = false;
 	}
```
Desour: What if an object is pointed and afterwards nothing (ie. toward sky) is pointed?

src/client/hud.cpp
```diff
@@ -504,6 +515,24 @@ void Hud::drawCrosshair()
...
+	int y_offset = 15;
+
+	if (use_selectionindicator_image) {
+		video::ITexture *selectionindicator = tsrc->getTexture("selectionindicator.png");
+		v2u32 size  = selectionindicator->getOriginalSize();
+		v2s32 lsize = v2s32(m_displaycenter.X - (size.X / 2),
+				m_displaycenter.Y - (size.Y / 2));
+		driver->draw2DImage(selectionindicator, lsize,
+				core::rect<s32>(0, 0, size.X, size.Y),
+				0, selectionindicator_argb, true);
```
Desour:
```suggestion
				nullptr, selectionindicator_argb, true);
```

src/client/render/core.cpp
```diff
@@ -84,8 +84,12 @@ void RenderingCore::draw3D()
 void RenderingCore::drawHUD()
 {
 	if (show_hud) {
-		if (draw_crosshair)
+		if (draw_crosshair) {
 			hud->drawCrosshair();
+
+			if (hud->can_draw_selectionindicator)
+				hud->drawSelectionIndicator();
```
Desour: I suggest to move this into `Hud::drawCrosshair`. (So that the selection indicator is the crosshair.)

LoneWolfHT:
Also changed indicator to a diamond
image

src/client/render/core.cpp
LoneWolfHT: Done

src/client/hud.cpp
LoneWolfHT: That part is code I copied. Changed both the 'copied' and 'copied from' code

src/client/game.cpp
LoneWolfHT: Fixed

paramat:
Sorry, my proposal was to replace the crosshair with an empty square (or diamond) when pointing at an entity, not add it above the crosshair. I cannot see a point in having both at once as the entity indicator can act as a crosshair itself.
So, the crosshair would change shape when pointing at an entity. Less visual obstruction and fewer lines drawn.

LoneWolfHT:
The diamond or box won't have an exact center though, it would need a dot in the middle or something.
Another possibility would be changing the crosshair into an X

tacotexmex:
> it would need a dot in the middle or something.

I really don't think so. There are plenty of first person games that works just fine with crosshairs without a pixel at the absolute center.

LoneWolfHT:
> I really don't think so. There are plenty of first person games that works just fine with crosshairs without a pixel at the absolute center. What is definitely confusing is _having two crosshairs_ at once.

I don't think the transition from a + to a diamond or box makes sense

> Is the crosshair still textureable?

Yes

paramat:
> The diamond or box won't have an exact center though

Yeah, this occurred to me too, but it seems not a problem because while you are pointing at an entity there is no need for precise aiming.
Two symbols onscreen is unnecessary and visually messy, it makes a lot of sense that the crosshair just alters design when pointing at an entity, but maybe it could change in a different way, not necessarily to a box or diamond.

LoneWolfHT:
> Yeah, this occurred to me too, but it seems not a problem because while you are pointing at an entity there is no need for precise aiming.

https://content.minetest.net/packages/GreenDimond/skinmaker/ requires precision when pointing at the canvas entity. There are quite a few mods that do things like this too

> Two symbols onscreen is unnecessary and visually messy, it makes a lot of sense to me that the crosshair just alters design when pointing at an entity, but maybe it could change in a different way, not necessarily to a box or diamond.

I'm going to try just removing the y offset and experiment with different shapes over the crosshair

Desour:
I'm for an X shape.

v-rob:
An X would make sense because 1) it has a central point and 2) many entities are mobs or players which can be hurt/killed, so an X would make sense for that.

paramat:
Ok, preserving a centre seems a good idea.

LoneWolfHT:
image

Here's what it looks like, although it would probably still be better to see for yourself ingame

LoneWolfHT:
Well I've tested this on a PvP server and it worked quite well.
Possible problem is the fact that it's less obvious what entity you are pointing at in a group of entites. Although currently it's hard to point at a specific one in a group anyway because the selectionboxes overlay (in the case of dropped items)

tacotexmex:
Last iteration looks great LoneWolfHT. I'd be happy with it 😃

src/client/hud.cpp
```diff
@@ -488,14 +496,30 @@ void Hud::drawHotbar(u16 playeritem) {
...
+			v2u32 size  = selectionindicator->getOriginalSize();
+			v2s32 lsize = v2s32(m_displaycenter.X - (size.X / 2),
+					m_displaycenter.Y - (size.Y / 2));
+			driver->draw2DImage(selectionindicator, lsize,
+					core::rect<s32>(0, 0, size.X, size.Y),
+					nullptr, selectionindicator_argb, true);
+		} else {
+			driver->draw2DLine(m_displaycenter - v2s32(9, 9), m_displaycenter + v2s32(9, 9),
+					selectionindicator_argb);
+			driver->draw2DLine(m_displaycenter + v2s32(9, -9), m_displaycenter + v2s32(-9, 9),
```
Desour, line 512: I'd suggest to use `7`, not `9`, here. `sqrt(1/2)` is circa `0.7071`. Otherwise, the `X` will have longer lines than the `+`.

paramat: I agree, the 'X' is a little unnecessarily large in the screenshot.

paramat:
Looks good but i agree it should be smaller to be no larger than the '+'.

SmallJoker:
Wouldn't it be better to add a `crosshair_object.png` texture which is used as crosshair when an object is selected? Afterwards it could be enabled/disabled locally, and changed in texture packs.

tacotexmex:
I thought that was already supported? Both crosshairs absolutely have to be customizable.

LoneWolfHT:
I feel like `crosshair_object` might be a better name for this

src/client/hud.cpp
LoneWolfHT: 7 is too small, this is what 8 looks like:
image

Desour: 7 would make the lines approximately 10 long in each direction (as with `+`).
But 8 looks fine in your screenshot.
(Maybe you should give this magic number a name.)
```
sqrt(2 * 7**2) -> 9.899494936611665
sqrt(2 * 8**2) -> 11.313708498984761
```

rubenwardy:
I would prefer two textures, to allow texture packs to customise it

LoneWolfHT:
I'll be renaming everything with `selectionindicator` in it to `crosshair_object` then (Code and all)

src/client/hud.cpp
LoneWolfHT: How do you handle names for magic numbers here? A #define?

paramat:
To be clear, this is how i think this should work:
(It is probably like this already, but there seems to be confusion)
There is an engine-drawn '+' crosshair which changes to an engine-drawn 'x' crosshair when pointing at an object.
Both crosshairs can be assigned a custom texture, but no textures should be provided by default.
>  I agree that crosshair_object.png would be a better name

+1

src/client/hud.cpp
Desour: Yes.

LoneWolfHT: How's that?

LoneWolfHT: @Desour I can review/test one of your PRs in exchange for an answer here if wanted

Desour: Oh, sorry for not answering.

It should be fine, I think.

LoneWolfHT: Alrighty, thanks!

sfan5:
This is ready for review already, right?

LoneWolfHT:
Yep

KaylebJay:
Oh dear, a whole month has passed. Someone, give this some love, this needs to be in 5.3.0 :)

ClobberXD:
@KaylebJay Too late, we're in feature freeze already. :/

KaylebJay:
Yeah, I noticed. :/

doc/texture_packs.txt
```diff
@@ -72,6 +72,11 @@ by texture packs. All existing fallback textures can be found in the directory
       `crosshair_color` and `crosshair_alpha` are used to create a cross
       when no texture was found
 
+* `object_crosshair.png`
+* the crosshair seen when pointing at an object. The settings
+`object_crosshair_color` and `object_crosshair_alpha` are used to create a cross
+when no texture was found
```
ClobberXD, line 78: These lines probably require indentation. Looks like 4 spaces, going by L71.

src/client/hud.cpp
```diff
@@ -488,19 +499,43 @@ void Hud::drawHotbar(u16 playeritem) {
 
 void Hud::drawCrosshair()
 {
-	if (use_crosshair_image) {
+	if (can_draw_object_crosshair)
+	{
```
ClobberXD, line 503:
```suggestion
	if (can_draw_object_crosshair) {
```

ClobberXD:
Tested according to the instructions provided in the first post:

- Crosshair changes to object crosshair when pointing at an object, as intended.
- Both object crosshair settings work as expected.
- Both crosshair textures work (tested simultaneously as well) as expected.
- Object crosshair correctly (dis)appears when toggling the `crosshair` HUD flag.

PR works as expected. :+1: 

One thing I noticed is that the object crosshair looks "less solid" compared to the normal crosshair, if that makes sense. This is an inherent issue when rendering diagonal lines, but I was wondering if it's possible to make the crosshair a wee bit thicker. This could be done as a separate feature though; no need to complicate this feature PR any further.

doc/texture_packs.txt
LoneWolfHT: How is it now?

paramat:
I disapprove of making the diagonal crosshair thicker in future, it is less solid due to diagonal pixels but is acceptably visible. Measured perpendicular to the lines it is actually thicker, so i think it is a good match =)
Thick crosshairs are obstructive, distracting and and intrusive.

LoneWolfHT:
Can always override it with a texturepack too I guess

ClobberXD:
Yeah, they could be overridden using texture packs if required. My request stands invalidated.

doc/texture_packs.txt
```diff
@@ -73,9 +73,9 @@ by texture packs. All existing fallback textures can be found in the directory
       when no texture was found
 
 * `object_crosshair.png`
-* the crosshair seen when pointing at an object. The settings
-`object_crosshair_color` and `object_crosshair_alpha` are used to create a cross
-when no texture was found
+    * the crosshair seen when pointing at an object. The settings
+    `object_crosshair_color` and `object_crosshair_alpha` are used to create a cross
+    when no texture was found
```
ClobberXD, line 78:
```suggestion
    when no texture was found.
```

:D

ClobberXD: Please also fix this in L73 for consistency. :)

doc/texture_packs.txt
ClobberXD: It looks about right. :+1:

doc/texture_packs.txt
LoneWolfHT: Also fixed the past tense usage. How's that?

ClobberXD: I think it sounds better now. :+1:

LoneWolfHT:
This is ready for review btw

src/client/hud.cpp
```diff
@@ -488,19 +499,42 @@ void Hud::drawHotbar(u16 playeritem) {
...
+					OBJECT_CROSSHAIR_LINE_SIZE),
+					m_displaycenter + v2s32(OBJECT_CROSSHAIR_LINE_SIZE,
+					OBJECT_CROSSHAIR_LINE_SIZE),
+					object_crosshair_argb);
+			driver->draw2DLine(
+					m_displaycenter + v2s32(OBJECT_CROSSHAIR_LINE_SIZE,
+					-OBJECT_CROSSHAIR_LINE_SIZE),
+					m_displaycenter + v2s32(-OBJECT_CROSSHAIR_LINE_SIZE,
+					OBJECT_CROSSHAIR_LINE_SIZE),
+					object_crosshair_argb);
```
SmallJoker, line 523: Rather than nested `if`/`else`, consider using `return;` here as soon a crosshair type is drawn.

src/defaultsettings.cpp
```diff
@@ -223,7 +225,7 @@ void set_default_settings(Settings *settings)
 	settings->setDefault("desynchronize_mapblock_texture_animation", "true");
 	settings->setDefault("hud_hotbar_max_width", "1.0");
 	settings->setDefault("enable_local_map_saving", "false");
-	settings->setDefault("show_entity_selectionbox", "true");
+	settings->setDefault("show_entity_selectionbox", "false");
```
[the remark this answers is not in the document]
LoneWolfHT: I'd prefer for it to be in this PR, it's easy to change back so I don't think it'll be too much of a problem.
It would be really nice if there was an ingame changelog shown for new releases...

LoneWolfHT:
The selectionbox that appears around entities when you select them will be ___DISABLED BY DEFAULT___ in favour of the object crosshair

Are there any objections?

If you're not sure what I'm talking about please say so

LoneWolfHT:
Made all but one of the requested changes and also rebased this so I could run a `git diff` to make sure I didn't miss anything

SmallJoker:
@LoneWolfHT Pro tip to avoid most of the rebase hell in the future:
`git rebase -i HEAD~20` to squash most (or all) commits prior rebasing -> results in less rebase work and less errors.

paramat:
> The selectionbox that appears around entities when you select them will be DISABLED BY DEFAULT in favour of the object crosshair

This is fine for me too.

src/defaultsettings.cpp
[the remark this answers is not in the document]
paramat: This change is fine for me, and apparently for others too, seeing as the PR was merged.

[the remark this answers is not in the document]
LoneWolfHT: I'll be around to revert if there is a massive uproar over it. I'll be sneaking a Windows build with it to some people who play Minetest too

PR #9572, opened 2020-03-31
Some vector functions useful for working with rotations

NetherEran:
This PR adds three vector functions useful for rotating objects:

* `vector.rotate(v, rot)`: applies the rotation vector rot as returned by object:get_rotation() to v using a rotation matrix
* `vector.rotate_around_axis(v, axis, angle)`: rotates vector v around vector axis by angle according to the right hand rule.
* `vector.dir_to_rotation(forward, [up])`: converts a unit vector pointing forward and one that points up to a rotation vector. If `up` is omitted, assumes roll to be zero.

Rotating entities by more than just yaw is currently kind of hard to do. This PR aims to make it easier.

## To do

This PR is Ready for Review.

## How to test

Write code using these functions. 
I've also attached a dumbed down version of the airplane mod I wrote these functions for. You can spawn in an airplane with `/spawnentity eg_airplanes:airplane`, right click to enter it, Up and Down to control pitch,  Left and Right to control roll and left and right mouse buttons to control yaw. These are all relative to the airplane.
[Demo mod](https://github.com/minetest/minetest/files/4732404/eg_airplanes.zip)


SmallJoker:
Glad to see some rotation helper functions. 

I'd like to recommend renaming the following functions:
- `forward_from_rot` -> `rotation_to_horizontal`
- `up_from_rot` -> `rotation_to_vertical`

That would keep consistency with existing functions, such as
- `get_look_horizontal`
- `facedir_to_dir`
- `yaw_to_dir`

Moreover would it be possible to somehow combine the functions `*_from_rot` into a generic that converts a normal vector + rotation to a new vector? Like `vector.rotate` but for all axis.

NetherEran:
I renamed the functions and added a generic for rotating. Also updated the test mod. Currently vector.rotation_to_* are there but they're kind of trivial. So if requested I will remove them.

Desour:
Related: #8515

SmallJoker:
Tested the mod quickly and noticed that `get_look_vertical` (player) and `get_rotation().x` are inverted/inconsistent. Great..

The PR features look good so far. Please add a few unittests (for "busted") here: [vector_spec.lua](https://github.com/minetest/minetest/blob/master/builtin/common/tests/vector_spec.lua).

builtin/common/vector.lua
```diff
@@ -141,3 +141,72 @@ function vector.sort(a, b)
...
+	--flip the angle because the formula is for the right hand rule
+	angle = -angle
+	local cosangle = math.cos(angle)
+	local sinangle = math.sin(angle)
+	axis = vector.normalize(axis)
+	--https://en.wikipedia.org/wiki/Rodrigues%27_rotation_formula
+	return vector.add(vector.add(
+		vector.multiply(v, cosangle),
+		vector.multiply(vector.cross(axis, v), sinangle)),
+		vector.multiply(axis, vector.dot(axis, v) * (1 - cosangle)))
```
SmallJoker, line 155: `a * cosangle + b * (1 - cosangle)` can usually be shortened to
`cosangle * (a - b) + b` where `a` and `b` are vectors calculated by `vector.multiply` above.

Before: (ignoring `vector.dot`)
 - 7 multiplications (6 in vector, 1 simple)
 - 1 subtraction (simple)
 - 3 additions (vector)

After: (ignoring `vector.dot`)
 - 3 subtractions (vector)
 - 3 multiplications (vector)
 - 3 additions (vector)

```Lua
local dot_axis = vector.multiply(axis, vector.dot(axis, v))
return vector.add(
	vector.multiply(vector.cross(axis, v), sinangle)),
	vector.add(
		vector.multiply(cosangle, vector.subtract(v, dot_axis))
		dot_axis
	)
```
PS: Not tested, but shouldn't this work?

SmallJoker:
Excellent, thank you!

I'm in favour of this PR because it exactly provides generic functions that are wanted to work with rotations, whereas a full matrix API is for more math-savvy modders.
@Desour Do you have any recommendations regarding the function naming or organisation - also to ease interaction with your PR in the future?

Desour:
To be honest, I don't really understand most of those calculations (eg. what the heck is `forwup`?). But something you might want to consider is [handling the cases where sinus or cosinus gets 0 specially](https://github.com/minetest/minetest/pull/8515/files#diff-2a9edd3cd4ff818074ab695501413fb9R122).

Just using forward and up vectors ([which is also easily possible with matrices btw.](https://github.com/Desour/helicopter/blob/546d5fdb490f5a67f121d5dc7747dfeebc0d7e9a/init.lua#L107)) is probably indeed fine enough in many cases.

Function names should be mostly fine, I guess (at least there should be no horrible naming collisions). However, I suggest renaming `rotation_to_horizontal` eg. to `forward_at_rotation`, as it's confusing.

Btw. the rotation angles are applied in the same order as in `obj:set_rotation`, right? (It should be intrinsic first yaw, then pitch, then roll.)

You might also want to make the doc more clear, eg. what means forward? -> (0,0,1).

If all these functions do correctly what I think they are doing, it should be easy to change the implementation later, to simplify the source code.

NetherEran:
> To be honest, I don't really understand most of those calculations (eg. what the heck is forwup?). But something you might want to consider is handling the cases where sinus or cosinus gets 0 specially.

Did that and also added more comments so hopefully the code is easier to understand now.
> Function names should be mostly fine, I guess (at least there should be no horrible naming collisions). However, I suggest renaming rotation_to_horizontal eg. to forward_at_rotation, as it's confusing.

I wasn't really happy with the previous names of those functions, I think this is a better name.
> Btw. the rotation angles are applied in the same order as in obj:set_rotation, right? (It should be intrinsic first yaw, then pitch, then roll.)

As far as I understand the math behind my code yes.
> You might also want to make the doc more clear, eg. what means forward? -> (0,0,1).

Did that too.

The demo mod is also updated now.

builtin/common/vector.lua
Desour: You could maybe test this by choosing vectors with only a small angle between axis and vector. Then compare if the angle between before and after. If it doesn't stay the same, this might be less precise.
```lua
local axis = vector.new(1, 0, 0)
local v = vector.new(420, 1, 0)
local rotated = vector.rotate_around_axis(v, axis, 123)
local rotated_two_times = vector.rotate_around_axis(rotated, axis, 345)
local old_angle = vector.angle(axis, v)
assert.equal(old_angle, vector.angle(axis, rotated))
assert.equal(old_angle, vector.angle(axis, rotated_two_times))
```

Desour:
:thinking: I wonder if negating the angle is enough to use something for a right-handed system in a left-handed system.

NetherEran:
> 🤔 I wonder if negating the angle is enough to use something for a right-handed system in a left-handed system.

> Would you mind adding unit-tests that rotate with arbitrary angles (not multiples of math.pi/2; math.pi/4=45° should be fine enough though) around all three base directions (+x, +z, ...) and test if the rotation was in the correct angle and did not invert an axis (eg. cross vector_before and vector_after and look if the resulting vector points in the direction of the axis)?

Done. Actually I changed the way vector.rotate_around_axis is adjusted to being left handed. This way it's a little bit more efficient but mathematically exactly as negating the angle.

builtin/common/tests/vector_spec.lua
```diff
@@ -92,15 +92,67 @@ describe("vector", function()
 			rotated3 = vector.rotate_around_axis(rotated3, axis3, -math.pi / 2)
 			assert.True(almost_equal(rotate3, rotated3))
 		end)
+		it("is_left_handed", function()
+			local v_before1 = {x = 0, y = 1, z = -1}
+			local v_after1 = vector.rotate_around_axis(v_before1, {x = 1, y = 0, z = 0}, math.pi / 4)
+			assert.True(almost_equal(vector.normalize(vector.cross(v_after1, v_before1)), {x = 1, y = 0, z = 0}))
```
Desour, line 98: :thinking: `v1, v2, vector.cross(v1, v2)` are a left-handed system (eg. if you insert (1, 0, 0) for v1 and (0, 1, 0) for v1 into vector.cross, you'll get (0, 0, 1), which is left-handed in minetest).
Aren't `v_after1` and `v_before1` in .cross in the wrong order here? Trying it with my fingers, it shows me that it was a right-handed rotation.

NetherEran: > `v1, v2, vector.cross(v1, v2) are a left-handed system`

The cross product is right handed. By putting v_after1 and v_before1 in this order I however get a cross product for a left handed coordinate system. See [this](https://upload.wikimedia.org/wikipedia/commons/thumb/b/b0/Cross_product_vector.svg/800px-Cross_product_vector.svg.png) grafic.

The reason the cross product appears left handed in minetest is convention. I can't however write tests based on that, because left handed rotations would appear right handed and vice versa.

I might be getting something wrong or missing something though.

Desour: The graphic you linked uses a right-handed system. [Minetest's `vector.cross`](https://github.com/minetest/minetest/blob/27d611fe5561db20b380a16fdc6bcf1fefaf5d39/builtin/common/vector.lua#L84-L90) does the same calculations as the "normal" cross product. As minetest uses a left-handed system, the result will be left-handed in minetest (in the game, for players).
A left-handed system is mirrored to a right-handed one. A right-handed rotation will look like a left-handed rotation in the mirror.

You could also test this manually in game by making an entity that rotates itself pi/4 around some axis when you rightclick it.

Desour: I've made a small mod for testing and can confirm, that it's right-handed:
```lua
local particle_pos = vector.new(1, 1, 0)
local axis = vector.new(0, 1, 0)
local radians_per_sec = math.pi

	local player = minetest.get_connected_players()[1]
	if not player then
		return
	end

	minetest.add_particle({
		pos = vector.add(particle_pos, player:get_pos()),
		velocity = {x=0, y=0, z=0},
		acceleration = {x=0, y=0, z=0},
		expirationtime = dtime * 1.5,
		size = 1,
		collisiondetection = false,
		collision_removal = false,
		object_collision = false,
		vertical = false,
		texture = "bubble.png",
	})

	particle_pos = vector.rotate_around_axis(particle_pos, axis, radians_per_sec * dtime)
end)
```

NetherEran: Thanks, I'm pretty sure I get it now. I think I got confused because in the beginning I implemented vector.forward_at_rotation and vector.up_at_rotation using vector.rotate_around_axis. But since euler angles in MT are counterclockwise I got right looking results with the wrong code.

I guess it's time to fix this now.

builtin/common/vector.lua
```diff
@@ -141,3 +141,95 @@ function vector.sort(a, b)
...
 		{x = math.max(a.x, b.x), y = math.max(a.y, b.y), z = math.max(a.z, b.z)}
 end
+
+local function sin(x)
+	if x % math.pi == 0 then
+		return 0
+	else
+		return math.sin(x)
+	end
+end
```
sfan5, line 151: This feels like the wrong approach to ensure that results are exact

builtin/common/vector.lua
```diff
@@ -141,3 +141,95 @@ function vector.sort(a, b)
...
+	local sinangle = sin(angle)
+	axis = vector.normalize(axis)
+	--https://en.wikipedia.org/wiki/Rodrigues%27_rotation_formula
+	local dot_axis = vector.multiply(axis, vector.dot(axis, v))
+	return vector.add(
+		vector.multiply(vector.cross(axis, v), sinangle),
+		vector.add(
+		vector.multiply(vector.subtract(v, dot_axis), cosangle),
+		dot_axis
+	))
```
sfan5, line 172: This isn't very readable...
How would this look with calculations inlined?

builtin/common/vector.lua
```diff
@@ -141,3 +141,95 @@ function vector.sort(a, b)
...
+	}
+	local keys = {"z", "x", "y"}
+	--compute matrix multiplication: `matrix` * `v`
+	local ret = vector.new(0, 0, 0)
+	for i = 1, 3 do
+		local row = matrix[i]
+		for ii = 1, 3 do
+			ret[keys[i]] = ret[keys[i]] + row[ii] * v[keys[ii]]
+		end
+	end
```
sfan5, line 209: Please unroll and inline the calculations

builtin/common/vector.lua
```diff
@@ -141,3 +141,95 @@ function vector.sort(a, b)
...
+end
+
+
+function vector.forward_at_rotation(rot)
+	return vector.rotate({x = 0, y = 0, z = 1}, rot)
+end
+
+function vector.up_at_rotation(rot)
+	return vector.rotate({x = 0, y = 1, z = 0}, rot)
+end
```
sfan5, line 220: These are so trivial I don't think we need them in the API.
For `vector.forward_at_rotation` the implication that Z is "forward" can also be problematic.

doc/lua_api.txt
```diff
@@ -2896,6 +2896,25 @@ For the following functions `x` can be either a vector or a number:
...
+    * Returns `v1` rotated around axis `v2` by `a` radians according to
+      the left hand rule.
+* `vector.forward_at_rotation(r)`:
+    * Returns a unit vector pointing forward ({x = 0, y = 0, z = 1}) rotated by `r`.
+    * vector.forward_at_rotation(object:get_rotation()) points forward relative to an entity.
+* `vector.up_at_rotation(r)`:
+    * Returns a unit vector pointing up ({x = 0, y = 1, z = 0}) rotated by `r`.
+    * vector.up_at_rotation(object:get_rotation()) points up relative to an entity.
+* `vector.directions_to_rotation(v1, v2):
+    * Returns a rotation vector for `v1` pointing forward and `v2` pointing up.
```
sfan5, line 2915: I feel like this needs a better description, because I don't get it.

rubenwardy: so, a direction by itself cannot be resolved to a single rotation - there are many roll rotations that match the same direction. Think a plane moving in the same direction but spinning. The up vector is needed to resolve that.

I'd prefer the arguments to be named better: `forward` and `up`.

doc/lua_api.txt
```diff
@@ -2896,6 +2896,25 @@ For the following functions `x` can be either a vector or a number:
 * `vector.divide(v, x)`:
     * Returns a scaled vector or Schur quotient.
 
+For the following functions `a` is an angle in radians and `r` is a rotation
+vector ({x = <pitch>, y = <yaw>, z = <roll>}) where pitch, yaw and roll are
+in radians.
+
+* `vector.rotate(v, r):
```
rubenwardy, line 2903: you're missing the ending ` on all of these

SmallJoker: and `<pitch>` might be parsed as HTML

builtin/common/vector.lua
Desour: p_gimeno once said that special casing is better than rounding: http://irc.minetest.net/minetest-dev/2019-05-11#i_5542561

doc/lua_api.txt
NetherEran: Adding the missing ` s right now.

Things that can be parsed as HTML are all over lua_api.txt, see e.g. [formspec docs](https://github.com/minetest/minetest/blob/454dbf83a9bf292910c1495a2aa49fd8b960c28f/doc/lua_api.txt#L438). I don't think me removing mine  individually would be of any value without a general 'no html syntax' rule for the docs.

builtin/common/vector.lua
sfan5: A comment should be added here to explain the situation around rounding/exactness then.

doc/lua_api.txt
```diff
@@ -2896,6 +2896,23 @@ For the following functions `x` can be either a vector or a number:
...
+    * Applies the rotation `r` to `v` and returns the result.
+    * `vector.rotate({x = 0, y = 0, z = 1}, r)` and
+      `vector.rotate({x = 0, y = 1, z = 0}, r)` return vectors pointing
+      forward and up relative to an entity's rotation `r`.
+* `vector.rotate_around_axis(v1, v2, a)`:
+    * Returns `v1` rotated around axis `v2` by `a` radians according to
+      the left hand rule.
+* `vector.directions_to_rotation(direction, up)`:
+    * Returns a rotation vector for `direction` pointing forward using `up`
+      ad the up vector..
```
sfan5, line 2913:
```suggestion
      as the up vector..
```

builtin/common/vector.lua
```diff
@@ -141,3 +141,82 @@ function vector.sort(a, b)
...
+	return vector.new(
+		matrix[2][2] * v.x + matrix[2][3] * v.y + matrix[2][1] * v.z,
+		matrix[3][2] * v.x + matrix[3][3] * v.y + matrix[3][1] * v.z,
+		matrix[1][2] * v.x + matrix[1][3] * v.y + matrix[1][1] * v.z
+	)
+end
+
+function vector.directions_to_rotation(forward, up)
+	local rot = {x = math.asin(forward.y), y = -math.atan2(forward.x, forward.z), z = 0}
+	-- calculate vector pointing uwith roll = 0, just based on forward vector
```
sfan5, line 211:
```suggestion
	-- calculate vector pointing up with roll = 0, just based on forward vector
```

builtin/common/vector.lua
```diff
@@ -141,3 +141,82 @@ function vector.sort(a, b)
...
+	local forwup = vector.rotate({x = 0, y = 1, z = 0}, rot)
+	-- 'forwup' and 'up' are now in a plane with 'forward' as normal.
+	-- the angle between them is equal to math.abs(roll)
+	rot.z = vector.angle(forwup, up)
+	-- if 'up' rotated by rot.z is equal to forwup, then roll = -rot.z, else roll = rot.z
```
sfan5, line 216: This comment states the obvious, more helpful would be a short explanation <i>why</i> rot.z needs to be inverted in some cases

doc/lua_api.txt
```diff
@@ -2896,6 +2896,23 @@ For the following functions `x` can be either a vector or a number:
...
+    * `vector.rotate({x = 0, y = 0, z = 1}, r)` and
+      `vector.rotate({x = 0, y = 1, z = 0}, r)` return vectors pointing
+      forward and up relative to an entity's rotation `r`.
+* `vector.rotate_around_axis(v1, v2, a)`:
+    * Returns `v1` rotated around axis `v2` by `a` radians according to
+      the left hand rule.
+* `vector.directions_to_rotation(direction, up)`:
+    * Returns a rotation vector for `direction` pointing forward using `up`
+      as the up vector.
+    * `direction` and `up` need to be unit vectors in a 90 degree angle to each other.
```
rubenwardy, line 2914: This makes this function not very useful - you'd only be able to end up with a single axis of rotation. Unless the docs are wrong?

sfan5: I don't see an issue, `direction=(1, 0, 0) up=(0, 1, 0)`  is not the only way to make a 90° angle

rubenwardy: Let's say my direction is (1, 1, 0) and up is (0, 1, 0). They aren't at right angles, and this makes this function not very nice to use

rubenwardy: Maybe I misunderstand the point of this function. If up is (0, 1, 0), then this function should be equivalent to:

```lua
local function direction_to_rotation(direction)
	return {
		x = math.atan2(direction.y, math.sqrt(direction.z*direction.z+direction.x*direction.x)),
		y = -math.atan2(direction.x, direction.z),
		z = 0
	}
end
```

rubenwardy: I guess you'd need to rotate up along the plane formed by the original up and the direction, to reach a right angle

NetherEran: What do you think about just returning zero as roll if 'up' is omitted?

NetherEran:
Alright.
If the `up` argument is not provided to vector.directions_to_rotation it will now just assume the roll to be zero. The input vectors are also no longer required to be unit vectors and and if both are provided, an error is thrown if they're not in a right angle.

`vector.rotate_around_axis()` is also right handed again, but this time on purpose. Minetest uses a left handed coordinate system but right handed euler angles, so having one, and only one rotation helper be left handed would be messy.

Here's a screenshot of particles being spawned with a changing angle. One ring is calculated using `vector.rotate`, one using `minetest.yaw_to_dir` and one with `vector.rotate_around_axis`
screenshot_20200520_192252

This is the code that produces this. As you can see it's much nicer this way than if `vector.rotate_around_axis` was left handed.
```lua
local ppos = {x = 0, y = 12, z = 0}
local angle = 0
        local pos1 = vector.rotate({x = 0, y = 0, z = 3}, {x = 0, y = angle, z = 0})
        local pos2 = vector.multiply(minetest.yaw_to_dir(angle), 2)
        local pos3 = vector.rotate_around_axis({x = 0, y = 0, z = 1}, {x = 0, y = 1, z = 0}, angle)
        pos1 = vector.add(ppos, pos1)
        pos2 = vector.add(ppos, pos2)
        pos3 = vector.add(ppos, pos3)
        angle = angle + math.pi / 180
        minetest.add_particle(
                                {
                                        pos = pos1,
                                        texture = "bubble.png",
                                })
        minetest.add_particle(
                                {
                                        pos = pos2,
                                        texture = "bubble.png",
                                })
        minetest.add_particle(
                                {
                                        pos = pos3,
                                        texture = "bubble.png",
                                })
end)
```

paramat:
IRC comment http://irc.minetest.net/minetest-dev/2020-05-29#i_5695475

> 23:31 PGimeno       #9572 - "-- Compute matrix multiplication: `matrix` * `v`" That's not what the code that follows is doing. The matrix should be properly rearranged.

SmallJoker:
PR is kinda stalling, and I'm sorry to bring in some more points:
http://irc.minetest.net/minetest-dev/2020-06-04#i_5697421

> vector.directions_to_rotation calls very expensive functions. It's probably simpler to construct a matrix (two normalizations, **two cross **products, and then to convert that matrix to Euler.

(** = edited)

@NetherEran do you think this would improve the execution speed? It would be great to if this part could be optimized even further. Although I don't think it's a requirement for this PR, instead should be done in a follow-up PR.

> We have dir_to_facedir, dir_to_wallmounted, then directions_to_rotation?
> I'd call it dir_to_rotation; the second (upvector) parameter can be seen as an extra

That is indeed more consistent with the API. If the other devs do not have any objections, this is optimally changed before merge.

NetherEran:
> @NetherEran do you think this would improve the execution speed? 
Sounds like it would but I have no idea how to implement it and I was going to ask for help once I got time. But since I need help with it anyways, a follow-up PR to optimize it by someone with a better knowlege of linear algebra would probably be better.

One small, easy optimization that I'll probably put in is replacing the `vector.distance()` in vector.lua, line 321 with subtractions, additions and multiplications which should be faster than the two `math.sqrt()` it calls right now.

SmallJoker:
@sfan5 Does this PR still look good to you? The most important change was 7c4cfa6.

sfan5:
Didn't retest but yes still :+1: from me.

Andrey2470T:
Question: why is 'up' parameter necessary in `dir_to_rotation`?

NetherEran:
> Question: why is 'up' parameter necessary in `dir_to_rotation`?

it is optional actually

Andrey2470T:
I see it is optional, but I want to know what it does.

NetherEran:
It clarifies where up is relative to the object. For example if only a forward vector were supplied, the game would have no way of knowing the roll of this airplane:
screenshot_20201120_114524

If the up vector is left out it  basically assumed to be {x = 0, y = 1, z = 0}, so up relative to the rest of the world.

PR #9574, opened 2020-03-31
Add on_authplayer callback and a last_login parameter to on_joinplayer

sorcerykid:
This PR is a feature port from Minetest S3 that provides additional callback support for post-authentication processing.

  * Replace on_auth_fail callback with more versatile on_authplayer
  * Better clarify account login process in Lua API documentation
  * Change initial timestamp for newly registered accounts to -1

The `on_authplayer` callback makes it possible for scripts to record a successful login. It deprecates the existing `on_auth_fail` callback (which I moved to deprecated.lua). This is particularly vital for mods like Auth Redux that need to be able to [count failed logins](https://forum.minetest.net/viewtopic.php?p=326978#p326978), while deleting the counter afterward. Currently the only workaround is to wait until on_joinplayer to delete the counter. But on servers with long media loading times, the player may give up and disconnect, resulting in a stale counter sitting in memory.

The `on_joinplayer` callback now includes a `last_login` parameter for scripts to determine the _real_ last-login time of players when joining the game. This makes customized greetings possible like "Hi singleplayer! You last joined 5 days ago." Right now, the only way to accomplish this is by injecting a callback at the head of the registered_on_joinplayers table. That's far from ideal since it makes a lot of assumptions about when the last-login time is recalculated. I feel this is a cleaner approach.

I also set the last-login timestamp to -1 during create_auth, so that scripts can present a different welcome message in the case of new players, rather than showing the current time. The timestamp is calculated, as per usual by the authentication handler afterward.

## To do

This PR is Ready for Review.

## How to test

```
minetest.register_on_joinplayer( function ( player, last_login )
        if last_login then
                print( string.format( "on_joinplayer: player_name=%s, last_login=%d (%0.1f mins ago)",
                        player:get_player_name( ), last_login, ( os.time( ) - last_login ) / 60 ) )
        else
                print( string.format( "on_joinplayer: player_name=%s, last_login=nil",
                        player:get_player_name( ) ) )
        end
end )

minetest.register_on_authplayer( function ( player_name, ip, is_success )
        print( string.format( "on_authplayer: player_name=%s, ip=%s, is_success=%s",
                player_name, ip, is_success and "true" or "false" ) )
end )

minetest.register_on_auth_fail( function ( player_name, ip )
        print( string.format( "on_auth_fail (deprecated): player_name=%s, ip=%s",
                player_name, ip ) )
end )
```

### DIFF
--- a/builtin/game/auth.lua
+++ b/builtin/game/auth.lua
@@ -41,7 +41,6 @@ core.builtin_auth_handler = {
 		return {
 			password = auth_entry.password,
 			privileges = privileges,
-			-- Is set to nil if unknown
 			last_login = auth_entry.last_login,
 		}
 	end,
@@ -53,7 +52,7 @@ core.builtin_auth_handler = {
 			name = name,
 			password = password,
 			privileges = core.string_to_privs(core.settings:get("default_privs")),
-			last_login = os.time(),
+			last_login = -1,  -- Defer login time calculation until on_joinplayer
 		})
 	end,
 	delete_auth = function(name)
--- a/builtin/game/auth.lua
+++ b/builtin/game/auth.lua
@@ -52,7 +52,7 @@ core.builtin_auth_handler = {
 			name = name,
 			password = password,
 			privileges = core.string_to_privs(core.settings:get("default_privs")),
-			last_login = -1,  -- Defer login time calculation until on_joinplayer
+			last_login = -1,  -- Defer login time calculation until record_login (called by on_joinplayer)
 		})
 	end,
 	delete_auth = function(name)
--- a/builtin/game/chat.lua
+++ b/builtin/game/chat.lua
@@ -1056,7 +1056,7 @@ core.register_chatcommand("last-login", {
 			param = name
 		end
 		local pauth = core.get_auth_handler().get_auth(param)
-		if pauth and pauth.last_login ~= -1 then
+		if pauth and pauth.last_login and pauth.last_login ~= -1 then
 			-- Time in UTC, ISO 8601 format
 			return true, "Last login time was " ..
 				os.date("!%Y-%m-%dT%H:%M:%SZ", pauth.last_login)
--- a/builtin/game/chat.lua
+++ b/builtin/game/chat.lua
@@ -1056,7 +1056,7 @@ core.register_chatcommand("last-login", {
 			param = name
 		end
 		local pauth = core.get_auth_handler().get_auth(param)
-		if pauth and pauth.last_login then
+		if pauth and pauth.last_login ~= -1 then
 			-- Time in UTC, ISO 8601 format
 			return true, "Last login time was " ..
 				os.date("!%Y-%m-%dT%H:%M:%SZ", pauth.last_login)
--- a/builtin/game/deprecated.lua
+++ b/builtin/game/deprecated.lua
@@ -75,8 +75,10 @@ core.setting_save = setting_proxy("write")
 -- core.register_on_auth_fail
 --
 
-core.register_on_auth_fail = function (func)
-	core.log("deprecated", "core.register_on_auth_fail is obsolete and should be replaced by core.register_on_authplayer instead.")
+function core.register_on_auth_fail(func)
+	core.log("deprecated", "core.register_on_auth_fail " ..
+		"is obsolete and should be replaced by " ..
+		"core.register_on_authplayer instead.")
 
 	core.register_on_authplayer(function (player_name, ip, is_success)
 		if not is_success then
--- a/builtin/game/deprecated.lua
+++ b/builtin/game/deprecated.lua
@@ -70,3 +70,17 @@ core.setting_get = setting_proxy("get")
 core.setting_setbool = setting_proxy("set_bool")
 core.setting_getbool = setting_proxy("get_bool")
 core.setting_save = setting_proxy("write")
+
+--
+-- core.register_on_auth_fail
+--
+
+core.register_on_auth_fail = function (func)
+	core.log("deprecated", "core.register_on_auth_fail is obsolete and should be replaced by core.register_on_authplayer instead.")
+
+	core.register_on_authplayer(function (player_name, ip, is_success)
+		if not is_success then
+			func(player_name, ip)
+		end
+	end)
+end
--- a/src/network/serverpackethandler.cpp
+++ b/src/network/serverpackethandler.cpp
@@ -412,7 +412,7 @@ void Server::handleCommand_ClientReady(NetworkPacket* pkt)
 	m_clients.event(peer_id, CSE_SetClientReady);
 
 	s64 last_login;
-	m_script->getAuthLastLogin(playersao->getPlayer()->getName(), &last_login);
+	m_script->getAuth(playersao->getPlayer()->getName(), nullptr, nullptr, &last_login);
 	m_script->on_joinplayer(playersao, last_login);
 
 	// Send shutdown timer if shutdown has been scheduled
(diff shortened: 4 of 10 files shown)


builtin/game/deprecated.lua
```diff
@@ -70,3 +70,17 @@ core.setting_get = setting_proxy("get")
 core.setting_setbool = setting_proxy("set_bool")
 core.setting_getbool = setting_proxy("get_bool")
 core.setting_save = setting_proxy("write")
+
+--
+-- core.register_on_auth_fail
+--
+
+core.register_on_auth_fail = function (func)
```
SmallJoker, line 78: This notation is mostly used in builtin:
`function core.register_on_auth_fail(func)`

src/network/serverpackethandler.cpp
```diff
@@ -1717,25 +1721,25 @@ void Server::handleCommand_SrpBytesM(NetworkPacket* pkt)
 {
 	RemoteClient* client = getClient(pkt->getPeerId(), CS_Invalid);
 	ClientState cstate = client->getState();
+	std::string addr_s = getPeerAddress( pkt->getPeerId( ) ).serializeString( );
```
SmallJoker, line 1724: No spaces before or after `(` `)`

src/script/cpp_api/s_player.cpp
```diff
@@ -156,7 +156,9 @@ void ScriptApiPlayer::on_joinplayer(ServerActiveObject *player)
 	lua_getfield(L, -1, "registered_on_joinplayers");
 	// Call callbacks
 	objectrefGetOrCreate(L, player);
-	runCallbacks(1, RUN_CALLBACKS_MODE_FIRST);
+	if(last_login != -1)
```
SmallJoker, line 159: Space after `if`

src/script/cpp_api/s_player.cpp
```diff
@@ -156,7 +156,9 @@ void ScriptApiPlayer::on_joinplayer(ServerActiveObject *player)
...
+		lua_pushinteger(L, last_login);
+	runCallbacks(last_login != -1 ? 2 : 1, RUN_CALLBACKS_MODE_FIRST);
```
SmallJoker, line 161: Please push `nil` and pass two values, rather than leaving it dynamic. This way it would be possible in mods to check whether the server supports this feature (`{...}` generic argument capture).

src/script/cpp_api/s_player.cpp
```diff
@@ -216,16 +218,19 @@ void ScriptApiPlayer::on_playerReceiveFields(ServerActiveObject *player,
...
+	// Get core.registered_on_authplayers
 	lua_getglobal(L, "core");
-	lua_getfield(L, -1, "registered_on_auth_fail");
+	lua_getfield(L, -1, "registered_on_authplayers");
+
+	// Call callbacks
 	lua_pushstring(L, name.c_str());
 	lua_pushstring(L, ip.c_str());
-	runCallbacks(2, RUN_CALLBACKS_MODE_FIRST);
+	lua_pushboolean( L, is_success );
```
SmallJoker, line 232: Invalid space placement.

src/script/cpp_api/s_server.cpp
```diff
@@ -59,6 +59,32 @@ bool ScriptApiServer::getAuth(const std::string &playername,
 	return true;
 }
 
+bool ScriptApiServer::getAuthLastLogin(const std::string &playername,
```
SmallJoker, line 62: `ScriptApiServer::getAuth(playername, nullptr, nullptr)`
could be re-used instead of adding a new function. The auth handler must return the other fields any way and checking those doesn't take much time.

builtin/game/deprecated.lua
```diff
@@ -70,3 +70,17 @@ core.setting_get = setting_proxy("get")
...
+	core.log("deprecated", "core.register_on_auth_fail is obsolete and should be replaced by core.register_on_authplayer instead.")
```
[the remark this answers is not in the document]
sorcerykid: Thanks for the review. Is there any particular convention of where it should be split (aside from between words of course)? Longest line first, or evenly divided lines?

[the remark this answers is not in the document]
SmallJoker: There's no regulations where to split the text. I'd do whatever looks better - in this case: evenly divided lines.

src/script/cpp_api/s_server.cpp
sorcerykid: Thanks for the tip, Wouldn't adding another argument to the declaration of `getAuth` require that all other calls to that function elsewhere pass NULL? I was mainly trying to avoid changing any more files than necessary. But I can gladly do it this way if that's what you'd prefer.

sfan5: You can set the default argument value in the header file, then existing calls won't have be edited.

sorcerykid:
The requested corrections have been made, and the PR is ready for re-review. Please lmk if I should rebase.

SmallJoker:
@sorcerykid Thanks for the changes. Looks good. PR needs a rebase.

sorcerykid:
Rebased and retested. The PR is now ready for review.

builtin/game/chat.lua
```diff
@@ -1056,7 +1056,7 @@ core.register_chatcommand("last-login", {
 			param = name
 		end
 		local pauth = core.get_auth_handler().get_auth(param)
-		if pauth and pauth.last_login then
+		if pauth and pauth.last_login ~= -1 then
```
sfan5, line 1059: Isn't it possible that last_login is still nil for old auth entries?

builtin/game/auth.lua
```diff
@@ -53,7 +52,7 @@ core.builtin_auth_handler = {
 			name = name,
 			password = password,
 			privileges = core.string_to_privs(core.settings:get("default_privs")),
-			last_login = os.time(),
+			last_login = -1,  -- Defer login time calculation until on_joinplayer
```
sfan5, line 55: It's much more obvious to point to the place where `last_login` is actually set:
```suggestion
			last_login = -1,  -- Defer login time calculation until record_login (called by on_joinplayer)
```

builtin/game/chat.lua
sorcerykid: How old are we talking about? As far as I understand, create_auth has always set the last login to os.time(), at least as far back as 0.4.14. Correct me if I'm wrong, but I don't think it's possible for it to be nil unless the database is corrupted.

https://github.com/sorcerykid/minetest/blob/24e8b0ac1ea45719937948607259f13866c8bc64/builtin/game/auth.lua#L129

SmallJoker: Right. This field was not created for accounts in old 0.4.x releases, thus still does not exist for players that never logged in since then. Needs checking what value `last_login` has (nil?).

sfan5: Considering the old code had `pauth.last_login` I'd just change this to `pauth.last_login and pauth.last_login ~= -1`

sorcerykid: Yes, good point @sfan5 and @SmallJoker, I was just thinking that myself. Doesn't hurt to check anyway. Will amend accordingly :)

builtin/game/deprecated.lua
```diff
@@ -70,3 +70,17 @@ core.setting_get = setting_proxy("get")
...
+
+	core.register_on_authplayer(function (player_name, ip, is_success)
```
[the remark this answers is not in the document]
sorcerykid: Oops, I just noticed I overlooked this. I'll get this corrected as well in the next commit.

builtin/game/chat.lua
sfan5: Bump.

sorcerykid: May apologies, I've just been super busy the past few days, but I'm planning on getting this completed today.

builtin/game/deprecated.lua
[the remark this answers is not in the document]
sorcerykid: It looks like this actually was resolved since I adopted the favoured notation for functions.

sorcerykid:
It looks like everything is now resolved.

If either of you get a chance, please also take a look at #9569, as it's pretty straightforward. And it would to nice have these two PRs merged around the same time since they're somewhat related.

sorcerykid:
Thank you for all your help @sfan5 and @SmallJoker. It's much appreciated.

PR #9585, opened 2020-04-03
Add max_viewrange setting

appgurueu:
- Goal of the PR: adds the `max_viewrange` setting (default: `0`) which limits player viewrange regarding objects.
- How does the PR work? Simply clamps the object sending range in add & remove functions.
- If not a bug fix, why is this PR needed? What usecases does it solve? PvP servers have to deal with the problem that players see opponents by using a simple glitch: They set the viewrange very low, trigger mapblock unloading, and then quickly set it to infinite viewrange. During a short blink, they are able to spot all players on the map, no matter the distance.

## To do

This PR is Ready for Review.

- [ ] Consider applying the same to mapblocks

## How to test

Set `max_viewrange` to any value and have two players which are far away. They won't see each other.

### DIFF
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -1243,6 +1243,9 @@ server_unload_unused_data_timeout (Unload unused server data) int 29
 #    Maximum number of statically stored objects in a block.
 max_objects_per_block (Maximum objects per block) int 64
 
+#    Limits the maximum viewrange for objects
+max_viewrange (Maximum object view range) float 0
+
 #    See https://www.sqlite.org/pragma.html#pragma_synchronous
 sqlite_synchronous (Synchronous SQLite) enum 2 0,1,2
 
--- a/minetest.conf.example
+++ b/minetest.conf.example
@@ -1502,6 +1502,9 @@
 #    type: int
 # max_objects_per_block = 64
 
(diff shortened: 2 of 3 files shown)


sfan5:
Sounds like a duplicate of `player_transfer_distance` to me.

appgurueu:
Hmm. Possible. Although this applies to all objects.

SmallJoker:
Objects are controlled by `active_object_send_range_blocks`, and players by `player_transfer_distance`. This PR adds the same features again, which doesn't make sense.

Wuzzy2:
> They set the viewrange very low, trigger mapblock unloading, and then quickly set it to infinite viewrange. During a short blink, they are able to spot all players on the map, no matter the distance.

This sounds like a deeper underlying issue.
But it's true. You can see stuff that should be hidden simply because the area that would block the view is not loaded. That's because all unloaded areas are transparent.
I think it should be fixed for better gameplay. Players get an advantage simply because they either play with the rande setting or they lucked out due to slow/weird loading of mapchunks. I think this is a complex problem and there won't be a simple fix. It's a deeper design flaw IMO, and it needs a complete rethinking about how unloaded areas are treated.

Care to post a new issue for this?

appgurueu:
Closing this - those settings indeed already exist.

PR #9594, opened 2020-04-04
Builtin: Make item entities glow less

sfan5:
Dropped torches glowed so much that I thought it was a bug at first, but it's not.
To make the glow less annoying, halve the light level.

before:
grafik
after:
grafik
for comparison, an unlit item:
grafik

## To do

## How to test

Drop torch at night


sfan5:
Krock, rubenwardy and paramat have previously agreed that this is okay:
http://irc.minetest.net/minetest-dev/2020-03-07#i_5649019

PR #9598, opened 2020-04-05
Overall improvements to log messages

sfan5:
In detail:
* duplicate log messages removed (also if the same info is logged a bit later)
* attempted(!) media loads only with `--trace` (errors still show, obviously)
* itemdef, craftdef registering, client-side nodedef deserialization only with `--trace` 
* client-side packetcounter shows packet names; every 30s instead of 20s
* mod storage saving only logged if there were changes
* mod load times logged in milliseconds

## To do

This PR is Ready for Review.

## How to test

Look at a `--verbose` log file and notice that you do not have to scroll through 4000 (actual value) of irrelevant lines to see something of substance.

### DIFF
--- a/src/client/client.cpp
+++ b/src/client/client.cpp
@@ -61,6 +61,20 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 extern gui::IGUIEnvironment* guienv;
 
 /*
+	Utility classes
+*/
+
+void PacketCounter::print(std::ostream &o) const
+{
+	for (const auto &it : m_packets) {
+		auto name = it.first >= TOCLIENT_NUM_MSG_TYPES ? "?"
+			: toClientCommandTable[it.first].name;
+		o << "cmd " << it.first << " (" << name << ") count "
+			<< it.second << std::endl;
+	}
+}
+
+/*
 	Client
 */
 
@@ -336,12 +350,12 @@ void Client::step(float dtime)
 	{
 		float &counter = m_packetcounter_timer;
 		counter -= dtime;
-		if(counter <= 0.0)
+		if(counter <= 0.0f)
 		{
-			counter = 20.0;
+			counter = 30.0f;
 
 			infostream << "Client packetcounter (" << m_packetcounter_timer
-					<< "):"<<std::endl;
+					<< "s):"<<std::endl;
 			m_packetcounter.print(infostream);
 			m_packetcounter.clear();
 		}
@@ -621,14 +635,17 @@ void Client::step(float dtime)
 
 	m_mod_storage_save_timer -= dtime;
 	if (m_mod_storage_save_timer <= 0.0f) {
-		verbosestream << "Saving registered mod storages." << std::endl;
 		m_mod_storage_save_timer = g_settings->getFloat("server_map_save_interval");
+		int n = 0;
 		for (std::unordered_map<std::string, ModMetadata *>::const_iterator
 				it = m_mod_storages.begin(); it != m_mod_storages.end(); ++it) {
 			if (it->second->isModified()) {
 				it->second->save(getModStoragePath());
+				n++;
 			}
 		}
+		if (n > 0)
+			infostream << "Saved " << n << " modified mod storages." << std::endl;
 	}
 
 	// Write server map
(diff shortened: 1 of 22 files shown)


Wuzzy2:
OK, I have made some superficial tests, but I don't think I'm a good tester here (I'm just not familiar with the codebase).

But the main promise seem to be true, at least. The `verbose` logging has been drastically reduced. The mod loading times also seem to work corrercly.

But most of this PR is a bit over my head, to be honest …

```
-       // FIXME: setWindowIcon returns a bool result but it is unused.
-       // For now continue to return this result.
```

You removed this FIXME, without mentioning this in your post. Why? Is this already fixed, it was just forgotten to be removed? Please explain.

```
-       u32 dr = m_env.getDayNightRatio();
-       infostream << "Client: time_of_day=" << time_of_day
-                       << " time_speed=" << time_speed
-                       << " dr=" << dr << std::endl;
+       //u32 dr = m_env.getDayNightRatio();
+       //infostream << "Client: time_of_day=" << time_of_day
+       //              << " time_speed=" << time_speed
+       //              << " dr=" << dr << std::endl;
```

Why not remove it entirely instead of commenting it out? I mean, you have this awesome version control called “Git” which keeps track of it. You might have heard of it. ;-)

sfan5:
> Why? Is this already fixed, it was just forgotten to be removed?

The result is evidently not unused so someone must just have forgotten to remove the comment.

> Why not remove it entirely instead of commenting it out? I mean, you have this awesome version control called “Git” which keeps track of it.

I know, but in case someone needs this exact log message back, they can easily just uncomment it.
This is not the best justification but meh.

src/client/client.cpp
```diff
@@ -621,14 +635,16 @@ void Client::step(float dtime)
...
-		verbosestream << "Saving registered mod storages." << std::endl;
 		m_mod_storage_save_timer = g_settings->getFloat("server_map_save_interval");
+		int n = 0;
 		for (std::unordered_map<std::string, ModMetadata *>::const_iterator
 				it = m_mod_storages.begin(); it != m_mod_storages.end(); ++it) {
 			if (it->second->isModified()) {
 				it->second->save(getModStoragePath());
 			}
 		}
+		if (n > 0)
```
SmallJoker, line 646: `n` is always `0`. Missing increment in the loop.

src/client/game.cpp
```diff
@@ -2010,7 +2010,7 @@ void Game::processItemSelection(u16 *new_playeritem)
 	for (u16 i = 0; i <= max_item; i++) {
 		if (wasKeyDown((GameKeyType) (KeyType::SLOT_1 + i))) {
 			*new_playeritem = i;
-			infostream << "Selected item: " << new_playeritem << std::endl;
+			infostream << "Selected item: " << i << std::endl;
```
SmallJoker, line 2013: IMO this log could be removed. There's visual feedback for this action.

src/client/client.cpp
```diff
@@ -653,8 +669,8 @@ bool Client::loadMedia(const std::string &data, const std::string &filename)
 	};
 	name = removeStringEnd(filename, image_ext);
 	if (!name.empty()) {
-		verbosestream<<"Client: Attempting to load image "
-		<<"file \""<<filename<<"\""<<std::endl;
+		TRACESTREAM(<< "Client: Attempting to load image "
```
SmallJoker, line 672: Just my opinion: Wouldn't it look better to include the first `<<` operator into the `log.h` define?

src/map.cpp
```diff
@@ -1828,7 +1828,7 @@ void ServerMap::save(ModifiedState save_level)
 	if(save_level == MOD_STATE_CLEAN
 			|| block_count != 0) {
 		infostream<<"ServerMap: Written: "
-				<<block_count<<" block files"
+				<<block_count<<" blocks"
```
SmallJoker, line 1831: Re-styling would be nice here.

src/server.cpp
```diff
@@ -2938,10 +2940,9 @@ void Server::UpdateCrafting(RemotePlayer *player)
 	if (!clist || clist->getSize() == 0)
 		return;
 
-	if (!clist->checkModified()) {
-		verbosestream << "Skip Server::UpdateCrafting(): list unmodified" << std::endl;
+	if (!clist->checkModified())
 		return;
-	}
+	verbosestream << "Server: Updating crafting" << std::endl;
```
SmallJoker, line 2945: There's a visual feedback when this happens: updated craft preview slot. The previous log message is used to determine how many calls that can be skipped for speed optimization (incremental inventory).

src/client/client.cpp
sfan5: I'm sure it would, but that would be lots of line changes for small benefit.

src/server.cpp
sfan5: So you are saying this should be removed?

PR #9603, opened 2020-04-06
Add all src folders to doxygen

Desour:
- Goal of the PR:
Currently doxygen ignores files in `gui`, `server` and co.. This PR fixes that.
- How does the PR work?
Adds all the paths to `Doxyfile.in`.
- This is a bugfix.

## To do

## How to test

- Use cmake.
- Do `make doc`.
- Open `<your minetest folder>/doc/html/files.html` and see that there's everything.


SmallJoker:
Why not `RECURSIVE = YES` so that all `src` paths are checked?

Desour:
Done.

PR #9604, opened 2020-04-06
Allow to select en as language in settings menu

### LINKED ISSUE #9599: English language missing from language menu
##### Minetest version

```
5.2.0-win64
```

##### OS / Hardware

Operating system: Windows 7
CPU: Intel Core i5 2300 2.8 GHz

GPU model:
OpenGL version:

##### Summary

When trying to change language to english, there is no "en" option in Settings->All Settings->Client and Server->Language
(OS language is pt-BR, so the menu options are these but in portuguese)
Tested in version 5.1.1, the "en" option is there as expected
##### Steps to reproduce

Open minetest, go to "Settings" tab, click "All Settings", click Client and Server, click "language", click the "Edit" button, look for "en" language

Desour:
- Fixes #9599.
- https://github.com/minetest/minetest/commit/e80c0bdea5756089db1a7e11fa4510e19b3336a2 is the commit that removed en. ~~(But it added el, which seems to be another kind of english? (At least it makes everything english (and changes the font?).))~~

## To do

## How to test

Select en as language in advanced settings and restart.

### DIFF
--- a/builtin/settingtypes.txt
+++ b/builtin/settingtypes.txt
@@ -1373,7 +1373,7 @@ name (Player name) string
 
 #    Set the language. Leave empty to use the system language.
 #    A restart is required after changing this.
-language (Language) enum   ,ar,ca,cs,da,de,dv,el,eo,es,et,eu,fil,fr,hu,id,it,ja,ja_KS,jbo,kk,kn,lo,lt,ms,my,nb,nl,nn,pl,pt,pt_BR,ro,ru,sl,sr_Cyrl,sv,sw,th,tr,uk,vi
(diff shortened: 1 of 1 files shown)


SmallJoker:
Was broken since e80c0bdea575

sfan5:
> But it added el, which seems to be another kind of english?

`el` is Greek, which just happens to be a very incomplete translation.

SmallJoker:
I think there should be an explicit note to add English into the list, since it's not a directory inside `po`. Thus it gets easily lost on updates.

Desour:
Where would one add such a note?

SmallJoker:
Added to the wiki ("Releasing Minetest"). Else than that there's only the option to also include it in the settings description, which isn't user-friendly.

PR #9611, opened 2020-04-07
Miscellaneous networking improvements

### LINKED ISSUE #2862: Calling hud_change directly after hud_add on same HUD element does not change it
I probably have found a problem in the Lua API of Minetest 0.4.12.

If you call `hud_add`, then shortly after call `hud_change` on the same HUD element, it will not be changed. `hud_change` only seems to work if some time has elapsed between `hud_add` and `hud_change`.

Example input for `luacmd` mod:

```
/lua id = me:hud_add({hud_elem_type="text", text="Original text", number=0xFFFFFF, position={x=0.5, y=0.6}}); me:hud_change(id, "text", "I have changed!")
```

Alternatively, you can use this code in a mod (works only in singleplayer mode):

```
local id = player:hud_add({hud_elem_type="text", text="Original text", number=0xFFFFFF, position={x=0.5, y=0.6}})
player:hud_change(id, "text", "I have changed!")
```

This code will first add a text HUD element saying “Original text”, then immediately changing (at least it attempts to!) the text to “I have changed!”.

If you run the code, the expected result would be that your screen displays the text “I have changed!”. In my test, however, the text will reliably be “Original text”.

For comparison, here is another test for `luacmd`, this time, I have added a 2 second delay for `hud_change`:

```
/lua id = me:hud_add({hud_elem_type="text", text="Original text", number=0xFFFFFF, position={x=0.5, y=0.6}}); minetest.after(2, function() me:hud_change(id, "text", "I have changed!"))
```

This reliably first shows “Original text”, then it changes to “I have changed!” after 2 seconds.

sfan5:
fixes #2862
This happens to also help with #9592, but that issue is most likely user error anyway.

## To do

## How to test

Play Minetest in both single- and multiplayer and make sure things still work.

### DIFF
--- a/src/network/connection.cpp
+++ b/src/network/connection.cpp
@@ -924,7 +924,7 @@ UDPPeer::UDPPeer(u16 a_id, Address a_address, Connection* connection) :
 	Peer(a_address,a_id,connection)
 {
 	for (Channel &channel : channels)
-		channel.setWindowSize(g_settings->getU16("max_packets_per_iteration"));
+		channel.setWindowSize(START_RELIABLE_WINDOW_SIZE);
 }
 
 bool UDPPeer::getAddress(MTProtocols type,Address& toset)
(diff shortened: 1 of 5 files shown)


src/network/serveropcodes.cpp
```diff
@@ -188,9 +198,9 @@ const ClientCommandFactory clientCommandFactoryTable[TOCLIENT_NUM_MSG_TYPES] =
 	null_command_factory, // 0x48
 	{ "TOCLIENT_HUDADD",                   1, true }, // 0x49
 	{ "TOCLIENT_HUDRM",                    1, true }, // 0x4a
-	{ "TOCLIENT_HUDCHANGE",                0, true }, // 0x4b
-	{ "TOCLIENT_HUD_SET_FLAGS",            0, true }, // 0x4c
-	{ "TOCLIENT_HUD_SET_PARAM",            0, true }, // 0x4d
+	{ "TOCLIENT_HUDCHANGE",                1, true }, // 0x4b
```
sfan5, line 201: This is the change that fixes the first linked issue.

src/network/connection.cpp
```diff
@@ -924,7 +924,7 @@ UDPPeer::UDPPeer(u16 a_id, Address a_address, Connection* connection) :
 	Peer(a_address,a_id,connection)
 {
 	for (Channel &channel : channels)
-		channel.setWindowSize(g_settings->getU16("max_packets_per_iteration"));
+		channel.setWindowSize(START_RELIABLE_WINDOW_SIZE);
```
sfan5, line 927: It makes no sense for this to be configurable and `max_packets_per_iteration` is used with an entirely different meaning in another place already.

src/network/connection.h
```diff
@@ -555,15 +555,15 @@ class Peer {
...
+		unsigned int m_increment_packets_remaining = 0;
 
 		virtual u16 getNextSplitSequenceNumber(u8 channel) { return 0; };
 		virtual void setNextSplitSequenceNumber(u8 channel, u16 seqnum) {};
 		virtual SharedBuffer<u8> addSplitPacket(u8 channel, const BufferedPacket &toadd,
 				bool reliable)
 		{
-			fprintf(stderr,"Peer: addSplitPacket called, this is supposed to be never called!\n");
+			errorstream << "Peer::addSplitPacket called,"
+					<< " this is supposed to be never called!" << std::endl;
```
ClobberXD, line 566: The following sounds better, IMO:

` this is never supposed to be called!`

SmallJoker: If it's supposed to be never called, it's kinda pointless to change it.

ClobberXD: I commented before I realised that the PR has already been merged. Should've noted this before, really. Sorry. :)

PR #9617, opened 2020-04-08
Improve protocol-level receiving code

sfan5:
- stop recreating buffer each time
- remove duplicate `WaitData()` call
- fix some theoretical latency issue with buffered packets
- higher timeout -> less busy-looping

## To do

### DIFF
--- a/src/network/connection.cpp
+++ b/src/network/connection.cpp
@@ -1173,7 +1173,9 @@ Connection::Connection(u32 protocol_id, u32 max_packet_size, float timeout,
 	m_bc_peerhandler(peerhandler)
 
 {
-	m_udpSocket.setTimeoutMs(5);
+	/* Amount of time Receive() will wait for data, this is entirely different
+	 * from the connection timeout */
+	m_udpSocket.setTimeoutMs(500);
 
 	m_sendThread->setParent(this);
 	m_receiveThread->setParent(this);
--- a/src/network/connectionthreads.cpp
+++ b/src/network/connectionthreads.cpp
@@ -812,6 +812,14 @@ void *ConnectionReceiveThread::run()
 	ThreadIdentifier);
 	PROFILE(ThreadIdentifier << "ConnectionReceive: [" << m_connection->getDesc() << "]");
 
+	// use IPv6 minimum allowed MTU as receive buffer size as this is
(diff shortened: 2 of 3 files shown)


src/network/connection.cpp
```diff
@@ -1173,7 +1173,9 @@ Connection::Connection(u32 protocol_id, u32 max_packet_size, float timeout,
 	m_bc_peerhandler(peerhandler)
 
 {
-	m_udpSocket.setTimeoutMs(5);
+	/* Amount of time Receive() will wait for data, this is entirely different
+	 * from the connection timeout */
+	m_udpSocket.setTimeoutMs(500);
```
nerzhul, line 1178: that can cause the server to hang i think. We may distinct the client acceptable version and the server. I think 10 ms is acceptacle on server but not 500, this will hang the entire network layer

sfan5: This runs on the `ConnectionReceiveThread` which has no job to do other than receiving.

src/network/connectionthreads.cpp
```diff
@@ -892,157 +900,139 @@ void *ConnectionReceiveThread::run()
...
+				" Ignoring." << std::endl);
+			return;
+		}
 
-			peer->ResetTimeout();
+		peer->ResetTimeout();
 
-			Channel *channel = 0;
+		Channel *channel = nullptr;
+		if (dynamic_cast<UDPPeer *>(&peer))
```
SmallJoker, line 998: else?
This check is either superfluous, or the else case needs a `FATAL_ERROR_IF` sanity check if `!channel`

sfan5: Did you mean to comment on the `if (channel)` below?

SmallJoker: I mean if this line's check returns `false`, `channel` would stay a nullptr and cause "random" segfaults in `processPacket` L1024.

sfan5: `handlePacketType_Reliable` and `handlePacketType_Control` already have an `assert(channel != NULL);`
I'll add a check nonetheless.

SmallJoker: Indeed. Although `assert` does not catch such errors in Release builds - the situation where such a problem would appear most likely.

PR #9618, opened 2020-04-08
Install the `locale` directory in a standard location without subfolder

Calinou:
Other established FOSS projects don't seem to be doing this (try `ls /usr/share/locale`).

This PR was inspired by the patch currently used in the Flatpak distribution: https://github.com/flathub/net.minetest.Minetest/blob/5844f665b2686278f9e288e1d8112f0668458021/minetest-i18n.patch

## To do

This PR is Ready for Review.

## How to test

Build Minetest and install it using `make install`.


sfan5:
Couldn't find it in Debian's list of patches from a quick look, but the file list reveals they also do this.
:+1:

nerzhul:
okay to me too

nerzhul:
seems travis github status is buggy. It's green on all steps

PR #9622, opened 2020-04-09
Update API site build

pauloue:
- markdown fixed the bug that caused installation with newer pip versions to fail
- removed part of the markdown patch that is now in upstream markdown

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -325,9 +325,8 @@ package:docker:
 
 pages:
   stage: deploy
-  image: python:3.7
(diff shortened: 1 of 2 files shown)


nerzhul:
thanks, is this possible to have python 3.8 or it's not supported yet ?

pauloue:
I changed it to python:latest which is currently 3.8

nerzhul:
i prefer to not use latest, when python will publish next version that may break the build

pauloue:
ok, it's 3.8 now

PR #9629, opened 2020-04-10
Drop genericobject.{cpp,h}

nerzhul:
This file is not for generic object but for ActiveObject message passing.
Put ownership of the various commands to the right objects and cleanup the related code.

That should remove some memory overhead due to some extra memory copy on some function, visible on servers with many active objects

Protect m_messages_out from outside modifications

## To do

This PR is Ready for Review.

## How to test

Build and test various active objects.

No function change here


src/content_sao.cpp
```diff
@@ -289,6 +288,120 @@ void UnitSAO::notifyObjectPropertiesModified()
...
+	// parameters
+	writeS16(os, m_attachment_parent_id);
+	os << serializeString(m_attachment_bone);
+	writeV3F32(os, m_attachment_position);
+	writeV3F32(os, m_attachment_rotation);
+	return os.str();
+}
+
+std::string UnitSAO::generateUpdateBonePositionCommand(const std::string &bone,
+	const v3f &position, const v3f &rotation)
```
sfan5, line 305: is there a reason this isn't a const method?

src/content_sao.cpp
```diff
@@ -289,6 +288,120 @@ void UnitSAO::notifyObjectPropertiesModified()
...
+		os<<serializeString(armor_group.first);
+		writeS16(os, armor_group.second);
+	}
+	return os.str();
+}
+
+
+std::string UnitSAO::generateUpdatePositionCommand(const v3f &position, const v3f &velocity,
+	const v3f &acceleration, const v3f &rotation, bool do_interpolate, bool is_movement_end,
+	f32 update_interval)
```
sfan5, line 358: same here

src/content_sao.cpp
nerzhul: it's static, it cannot be const because there is no membership

src/content_sao.cpp
nerzhul: same here :)

PR #9633, opened 2020-04-10
Move SAO to dedicated files & drop the old now useless factory

nerzhul:
* Move UnitSAO to dedicated files
* add vscode cmake plugin files to .gitignore
* move ServerActiveObject files to server folder & proper file naming
* move PlayerSAO to dedicated file
* move LuaEntitySAO to dedicated file
* delete TestSAO
* drop the old useless factory from the Minetest 0.3 era, as now we have only (for now) a single useful type. This change will improve the server performance on massive LuaEntitySAO creation/load
* cleanup headers, mapgen doesn't require content_sao or any other SAO reference

**Don't squash this PR on merge, we need to keep changes separated**

## To do

This PR is Ready for Review

## How to test

Build & test with some SAO types with 2 clients, throw SAO on the ground see them work properly, either the PlayerSAO and the LuaEntitySAO


sfan5:
Travis logs currently seem to be broken, but I think clang-format is failing on the new files (`luaentity_sao.cpp`, `player_sao.cpp`, `unit_sao.cpp`).
Cleanup looks good to me other than that.

nerzhul:
nice to hear. I will try to see if we can "unbreak" travis :D

sfan5:
<pre>
.git/rebase-apply/patch:873: trailing whitespace.
		ServerActiveObject *obj = createSAO((ActiveObjectType) s_obj.type, s_obj.pos, 
</pre>

sorcerykid:
I just have to say this is a huge improvement, so much cleaner and more organized than it was before.

PR #9641, opened 2020-04-11
Replace travis with github actions

nerzhul:
Travis is quite old, slow and unfriendly to maintain. Use the new github
actions for some of our checks instead of travis.

This speeds up the build, github infra is faster, and also maintained, unlike travis old runners :(

This should permit in the future to restore the Mac OSX build and maybe windows native builds

## To do

This PR is a Ready for Review but there is a missing build

* [x] Windows binary builds

## How to test

Just look at the ci ?


nerzhul:
travis is totally broken currently... Nice haha

sfan5:
I don't think there's a pressing need to move to Github actions, having all checks on a single platform is quite nice. On the other hand travis is quite slow sometimes and this can speed it up.

I'm neutral to this. Would be nice to hear other coredevs opinion on this.

.github/workflows/clang-tidy.yml
```diff
@@ -0,0 +1,23 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+    - uses: actions/checkout@v2
+
+    - name: Install clang-tidy
+      run: |
+        sudo apt-get update -qyy
+        sudo apt-get install clang-tidy-8 -qyy
```
sfan5, line 19: Doesn't github have a way to pre-install packages without doing it manually?

nerzhul: i'm searching a way but it doesn't seems, looked for this project while searching, but github is less flexible  than gitlab which permits to use any docker image
https://github.com/onqtam/doctest/pull/285/files

nerzhul: but i take the point that i should convert everything in a single MR :)

nerzhul: it seems no, it's as raw as gitlab

BuckarooBanzay:
@nerzhul Interesting, what about integration-tests to verify that the final build actually runs?
I'm using this pattern for quite a while in my mods and tools now:

* Example Action: https://github.com/blockexchange/blockexchange/blob/master/.github/workflows/integration-test.yml
* Test script: https://github.com/blockexchange/blockexchange/blob/master/test/integration-test-standalone.sh

Also here for my custom minetest build:
* Action: https://github.com/pandorabox-io/minetest_docker/blob/master/.github/workflows/docker.yml
* Integration-Test script: https://github.com/pandorabox-io/minetest_docker/blob/master/test.sh

This spins up the built binary and does some (currently stupid) sanity checks on that..
You could even test mapgen related functions (https://github.com/mt-mods/jumpdrive/blob/master/integration_test.lua)

nerzhul:
It will be the same as before currently, i just want to have iso builds, but after we can have PR to improve that.  Don't forget that gitlab does the artifact build

nerzhul:
@sfan5 , @rubenwardy , @SmallJoker pr is now ready, all builds have been implemented in a very similar way. Improvements on the pipeline can be done later. Here we are iso with what travis gave us before

nerzhul:
PR can be merged when you want

.github/workflows/validate.yml
```diff
@@ -0,0 +1,239 @@
...
+      - name: Install compiler
+        run: |
+          sudo apt-get update -qyy
+          sudo apt-get install wget
+          wget http://minetest.kitsunemimi.pw/mingw-w64-x86_64_9.2.0_ubuntu18.04.tar.xz -O mingw.tar.xz
+          sed -e "s|%PREFIX%|x86_64-w64-mingw32|" -e "s|%ROOTPATH%|/usr/x86_64-w64-mingw32|" < ./util/github/toolchain_mingw.cmake.in > ./util/buildbot/toolchain_mingw64.cmake
+          sudo tar -xaf mingw.tar.xz -C /usr
+      - name: Build
+        run: |
+          ./util/buildbot/buildwin64.sh cmakebuild
```
sfan5, line 239: what about `EXISTING_MINETEST_DIR` and `NO_MINETEST_GAME`?

If you let the script download fresh code from github.com/minetest/minetest it won't actually test the changes.
This is also incorrect in the `.gitlab-ci.yml` by the way.

.github/workflows/validate.yml
```diff
@@ -0,0 +1,239 @@
...
+          install_linux_deps
+
+      - name: Build
+        run: |
+          mkdir cmakebuild
+          cd cmakebuild
+          cmake -DCMAKE_BUILD_TYPE=Debug -DRUN_IN_PLACE=TRUE -DENABLE_GETTEXT=TRUE -DBUILD_SERVER=TRUE ${CMAKE_FLAGS} ..
+          make -j2
+        env:
+          CMAKE_FLAGS: "-DCMAKE_C_COMPILER=clang-9 -DCMAKE_CXX_COMPILER=clang++"
```
sfan5, line 206: same issue in other places
```suggestion
          CMAKE_FLAGS: "-DCMAKE_C_COMPILER=clang-9 -DCMAKE_CXX_COMPILER=clang++-9"
```

sfan5:
The "build passing" badge in the README should also be replaced.

.github/workflows/validate.yml
```diff
@@ -0,0 +1,239 @@
...
+          source ./util/github/lint.sh
+          perform_lint
+
+  clang_tidy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install clang-tidy
+      run: |
+        sudo add-apt-repository "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-9 main"
```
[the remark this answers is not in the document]
nerzhul: okay perfect i just copied our travis friend which seems to have some junk :)

.github/workflows/validate.yml
```diff
@@ -0,0 +1,239 @@
...
+        run: |
+          sudo apt-get install g++-6 gcc-6 -qyy
+          source ./util/github/common.sh
+          install_linux_deps
+
+      - name: Build
+        run: |
+          mkdir cmakebuild
+          cd cmakebuild
+          cmake -DCMAKE_BUILD_TYPE=Debug -DRUN_IN_PLACE=TRUE -DENABLE_GETTEXT=TRUE -DBUILD_SERVER=TRUE ${CMAKE_FLAGS} ..
```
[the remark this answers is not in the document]
nerzhul: i think we can factorize directly , and shell script error management is just error prone. Here it failed on the first error. Will try how to fix that

.github/workflows/validate.yml
```diff
@@ -0,0 +1,239 @@
...
+          valgrind --leak-check=full --leak-check-heuristics=all --undef-value-errors=no --error-exitcode=9 ./bin/minetest --run-unittests
+
+  win32:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install compiler
+        run: |
+          sudo apt-get update -qyy
+          sudo apt-get install wget
```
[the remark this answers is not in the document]
nerzhul: i installed it to be sure, i can try to not install it

util/github/common.sh
```diff
@@ -44,7 +44,7 @@ install_macosx_deps() {
 }
 
 # Relative to git-repository root:
-TRIGGER_COMPILE_PATHS="src/.*\.(c|cpp|h)|CMakeLists.txt|cmake/Modules/|util/travis/|util/buildbot/"
+TRIGGER_COMPILE_PATHS="src/.*\.(c|cpp|h)|CMakeLists.txt|cmake/Modules/|util/github/|util/buildbot/"
 
 needs_compile() {
 	RANGE="$TRAVIS_COMMIT_RANGE"
```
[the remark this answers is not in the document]
nerzhul: nowhere i think, it's just a sed

.gitlab-ci.yml
```diff
@@ -219,7 +219,7 @@ build:fedora-24:
     - apt-get update -y
     - apt-get install -y wget xz-utils unzip git cmake gettext
     - wget -q http://minetest.kitsunemimi.pw/mingw-w64-${WIN_ARCH}_9.2.0_ubuntu18.04.tar.xz -O mingw.tar.xz
-    - sed -e "s|%PREFIX%|${WIN_ARCH}-w64-mingw32|" -e "s|%ROOTPATH%|/usr/${WIN_ARCH}-w64-mingw32|" < util/travis/toolchain_mingw.cmake.in > ${TOOLCHAIN_OUTPUT}
+    - sed -e "s|%PREFIX%|${WIN_ARCH}-w64-mingw32|" -e "s|%ROOTPATH%|/usr/${WIN_ARCH}-w64-mingw32|" < util/github/toolchain_mingw.cmake.in > ${TOOLCHAIN_OUTPUT}
```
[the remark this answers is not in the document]
nerzhul: okay fine to me

nerzhul:
Thanks for the review, i will take time to fix changes

.github/workflows/validate.yml
nerzhul: i understand for the NO_MINETEST_GAME flag. I just read the EXISTING_MINETEST_DIR old behaviour but i don't understand the point here, the build runs just perfectly on the CI

nerzhul:
@sfan5 your turn

.github/workflows/validate.yml
sfan5: If I create a PR that introduces a compiler error on Windows the following will happen:
* the CI runs
* it clones the source from `github.com/minetest/minetest`
* the source is master is bug-free so the build completes successfully
* the CI completes successfully

This misses the entire point.

util/github/common.sh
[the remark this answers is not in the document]
sfan5: This check has a point though, it should not just be removed.

Before starting the CI process this function would be called to determine if compiling was necessary, so that C++ compiling would be skipped if you e.g. change something in `builtin/`.
Is it possible to do the same thing with Github actions?

.gitlab-ci.yml
```diff
@@ -275,6 +276,7 @@ build:win64:
     NO_PACKAGE: "1"
     WIN_ARCH: "x86_64"
     TOOLCHAIN_OUTPUT: "util/buildbot/toolchain_mingw64.cmake"
+    NO_MINETEST_GAME: 1
```
sfan5, line 279: Since some users use these packages I think MTG should stay here.
(The only thing I meant to suggest in my other comment was `EXISTING_MINETEST_DIR`)

util/ci/build.sh
```diff
@@ -0,0 +1,9 @@
+#! /bin/bash
+
+set -e
+set -u
```
sfan5, line 4: this can also be put into the first line:
`#!/bin/bash -eu`

util/ci/common.sh
```diff
@@ -21,9 +21,10 @@ install_linux_deps() {
 	local pkgs=(libirrlicht-dev cmake libbz2-dev libpng-dev \
 		libjpeg-dev libxxf86vm-dev libgl1-mesa-dev libsqlite3-dev \
 		libhiredis-dev libogg-dev libgmp-dev libvorbis-dev libopenal-dev \
-		gettext libpq-dev postgresql-server-dev-all libleveldb-dev)
+		gettext libpq-dev postgresql-server-dev-all libleveldb-dev \
+		libcurl4-openssl-dev)
 	# for better coverage, build some jobs with luajit
-	if [[ "$CC" == "clang"* && -z "$VALGRIND$FREETYPE" ]]; then
+	if [[ "x$WITH_LUAJIT" != "x" ]]; then
```
sfan5, line 27:
```suggestion
	if [ -n "$WITH_LUAJIT" ]; then
```

util/ci/clang-tidy.sh
```diff
@@ -0,0 +1,18 @@
...
+	-DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
+	-DRUN_IN_PLACE=TRUE \
+	-DENABLE_GETTEXT=TRUE \
+	-DENABLE_SOUND=FALSE \
+	-DBUILD_SERVER=TRUE ..
+make GenerateVersion
+
+cd ..
+
+./util/ci/run-clang-tidy.py -clang-tidy-binary=clang-tidy-9 -p cmakebuild -quiet -config="$(cat .clang-tidy)" 'src/.*'
```
sfan5, line 18: for readability:
```suggestion
./util/ci/run-clang-tidy.py \
	-clang-tidy-binary=clang-tidy-9 -p cmakebuild \
	-quiet -config="$(cat .clang-tidy)" \
	'src/.*'
```

util/github/common.sh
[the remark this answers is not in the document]
nerzhul: yep we can filter actions depending on what we modified, like in gitlab

.github/workflows/validate.yml
nerzhul: ok i see, will take a look

util/github/common.sh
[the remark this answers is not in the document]
sfan5: Does it do that yet?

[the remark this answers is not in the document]
nerzhul: not yet but we can add this later it's minor as the builds are far more faster than travis. In 5 minutes you have your result with actions whereas before it tooks 30 mins...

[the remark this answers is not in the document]
nerzhul: it seems we can filter by having multiple workflows, i will try to split one of the linter to see if we can trigger it properly

PR #9642, opened 2020-04-11
Reduce ServerEnvironment propagation (helps pathfinder to become client side friendly)

nerzhul:
ServerEnvironment is a huge class with many accessors. In various places it's not needed

Remove it to reduce the ServerEnvironment usage everywhere.

Idea here is to reduce size of some of our objects to transport lightweight managers and permit easier testing

## To do

## How to test
You can easily test it with any mod with pathfinding or nodetimers. It's just pointer types replacement no functional changes.

### DIFF
--- a/src/script/lua_api/l_nodetimer.cpp
+++ b/src/script/lua_api/l_nodetimer.cpp
@@ -41,11 +41,9 @@ int NodeTimerRef::l_set(lua_State *L)
 {
 	MAP_LOCK_REQUIRED;
 	NodeTimerRef *o = checkobject(L, 1);
-	ServerEnvironment *env = o->m_env;
-	if(env == NULL) return 0;
 	f32 t = readParam<float>(L,2);
 	f32 e = readParam<float>(L,3);
-	env->getMap().setNodeTimer(NodeTimer(t, e, o->m_p));
+	o->m_map->setNodeTimer(NodeTimer(t, e, o->m_p));
 	return 0;
 }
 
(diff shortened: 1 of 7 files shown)


SmallJoker:
`ServerEnvironment` will be needed as soon the pathfinder takes entity collisions into account (probably never).

nerzhul:
i looked a little bit at the collision code and i think it won't be needed with some refactor. Collision has nothing to do with Environment, just the Map and content defs logically.
Is this fine to you @SmallJoker

src/script/lua_api/l_nodetimer.cpp
```diff
@@ -98,37 +86,21 @@ int NodeTimerRef::l_get_elapsed(lua_State *L)
...
+void NodeTimerRef::create(lua_State *L, v3s16 p, ServerMap *map)
 {
-	NodeTimerRef *o = new NodeTimerRef(p, env);
+	NodeTimerRef *o = new NodeTimerRef(p, map);
 	*(void **)(lua_newuserdata(L, sizeof(void *))) = o;
 	luaL_getmetatable(L, className);
 	lua_setmetatable(L, -2);
 }
 
-void NodeTimerRef::set_null(lua_State *L)
```
sfan5, line 126: This was unused, right?

nerzhul: if not the compilation failed no ? :D 
yes, and i don't see the idea to drop the m_env pointer from the nodetimerref, do you have a usecase for that, which will trigger the way that each thing you will do on nodetimer will return nil ?
As you may see there is not line removal on the lua api registration

sfan5: The usecase would be to invalidate the reference if the `ServerEnvironment`/`Map` disappears, but that obviously doesn't happen.
So this is okay.

PR #9665, opened 2020-04-14
Verify database connection on interval

nerzhul:
This PR permits to add periodic DB verification (each 5 seconds)

This permits to ensure database is up & running, solving some connection issues when there is a firewall between and the connection was inactive for a long time.

Also fix the pg verification command which was totally wrong (just check the old connection state instead of verifying it)

The verification is no noop on all backends except on pg currently. But that can change if we add mysql support or something like this later :)

## How to test

Run this on postgresql remote server with a firewall between server & pg. Play the game 2 mins. Disconnect, reconnect on the server after 10 mins. Server works.
Before this PR: you reconnect, server crashes due to invalid pg connection

### DIFF
--- a/src/database/database-postgresql.cpp
+++ b/src/database/database-postgresql.cpp
@@ -90,13 +90,19 @@ void Database_PostgreSQL::connectToDatabase()
 	initStatements();
 }
 
-void Database_PostgreSQL::verifyDatabase()
+void Database_PostgreSQL::pingDatabase()
 {
-	if (PQstatus(m_conn) == CONNECTION_OK)
-		return;
+	// Verify DB connection with ping
+	try {
+		ping();
+	} catch (const DatabaseException &e) {
+		// If ping failed, show the error and try reconnect
+		PQreset(m_conn);
 
-	PQreset(m_conn);
-	ping();
+		errorstream << e.what() << std::endl
+			<< "Reconnecting to database " << m_connect_string << std::endl;
+		connectToDatabase();
+	}
 }
 
 void Database_PostgreSQL::ping()
@@ -151,7 +157,7 @@ void Database_PostgreSQL::createTableIfNotExists(const std::string &table_name,
 
 void Database_PostgreSQL::beginSave()
 {
-	verifyDatabase();
+	pingDatabase();
 	checkResults(PQexec(m_conn, "BEGIN;"));
 }
 
@@ -227,7 +233,7 @@ bool MapDatabasePostgreSQL::saveBlock(const v3s16 &pos, const std::string &data)
 		return false;
 	}
 
-	verifyDatabase();
+	pingDatabase();
 
 	s32 x, y, z;
 	x = htonl(pos.X);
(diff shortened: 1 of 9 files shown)


appgurueu:
Each 5 seconds seems a bit frequent to me?

nerzhul:
@appgurueu no in fact, because a keepalive is more frequent in many protocols. If you look at various connection pools (like java or golang) implementation they run it each second on each query too

nerzhul:
linter failure, will fix if team is okay with the correction

src/database/database.h
```diff
@@ -32,6 +32,7 @@ class Database
 	virtual void beginSave() = 0;
 	virtual void endSave() = 0;
 	virtual bool initialized() const { return true; }
+	virtual void verifyDatabase() = 0;
```
sfan5, line 35: I think `pingConnection` would be a better name.
Also an empty default implementation would be nice.

src/serverenvironment.cpp
```diff
@@ -1219,6 +1219,11 @@ void ServerEnvironment::step(float dtime)
 		}
 	}
 
+	if (m_database_check_interval.step(dtime, 5.0f)) {
```
sfan5, line 1222: I know this is nitpicking but wouldn't 10 work just as well?

src/database/database-postgresql.cpp
```diff
@@ -92,11 +92,18 @@ void Database_PostgreSQL::connectToDatabase()
...
+		ping();
+	} catch (const DatabaseException &e) {
+		// If ping failed, show the error and try reconnect
+
+		errorstream << e.what() << std::endl;
+		PQreset(m_conn);
 
-	PQreset(m_conn);
-	ping();
+		actionstream << "Reconnecting to database " << m_connect_string << std::endl;
```
sfan5, line 104: I think this is better merged into a single error message.

src/database/database.h
nerzhul: the problem is that the ping is just the pg name, in other backend it's not the same name but i can rename if you prefer
i can try an empty default, it was on a previous first try, maybe if it works i can remove upper classes empty body but not sure du to compiler ambigous

sfan5: The issue I have with the name is that "verify the database" sounds a lot like you would be checking the consistency of correctness of the database structures or contents.

src/serverenvironment.cpp
nerzhul: i think 10 can suffice yes, but more longer can lead to timeouts in some configurations

src/database/database-postgresql.cpp
nerzhul: fine

src/database/database.h
nerzhul: okay then i move to ping

sfan5:
<pre>
Applying: Verify database connection on interval
.git/rebase-apply/patch:36: trailing whitespace.
		errorstream << e.what() << std::endl 
.git/rebase-apply/patch:383: trailing whitespace.
{ 
.git/rebase-apply/patch:384: trailing whitespace.
	dbase->pingDatabase(); 
warning: 3 lines add whitespace errors.
</pre>

nerzhul:
rofl, i should see if there is a parameter in vscode :) i fix that and merge

PR #6803, opened 2017-12-18
Optimize wieldmesh

numberZero:
Fixes #3069, fixes #6720.

This PR changes handling of extruded images. Instead of using single mesh for all textures (of a single size), separate optimized mesh is used for each texture. Optimization is simple and fast, working in *O(texture_pixel_count)* time, and resulting mesh uses *O(sqrt(texture_pixel_count))* memory. `createExtrusionMesh` caches and returns that mesh, but actual work is done by the `Extruder` class.

As the whole extrusion code is new, it is put in separate files; existing ones are updated minimally to support or to use it.

Note: soft overlays are not extruded at all in this PR. This may cause problems with e.g. colored books *(from some mod that I don’t remember),* but extrusion would cause Z-fighting unless done in a special, more complicated way.
Object with an overlay

PR #9671, opened 2020-04-14
Optimize get_objects_inside_radius calls

nerzhul:
our previous implementation calls the ActiveObjectMgr to return ids and then lookup those ids in the same map and test each object
Instead now we call the global map to return the pointers directly and we ask filtering when building the list using lamba.

This drop double looping over ranges of active objects (and then filtered one) and drop x lookups on the map regarding the first call results

## To do

## How to test

Have item_drops and a mob framework with mobs and just see it works (a little bit better)

### DIFF
--- a/src/collision.cpp
+++ b/src/collision.cpp
@@ -360,17 +360,19 @@ collisionMoveResult collisionMoveSimple(Environment *env, IGameDef *gamedef,
 				// Calculate distance by speed, add own extent and 1.5m of tolerance
 				f32 distance = speed_f->getLength() * dtime +
 					box_0.getExtent().getLength() + 1.5f * BS;
-				std::vector<u16> s_objects;
-				s_env->getObjectsInsideRadius(s_objects, *pos_f, distance);
 
-				for (u16 obj_id : s_objects) {
-					ServerActiveObject *current = s_env->getActiveObject(obj_id);
-
-					if (!self || (self != current &&
-							self != current->getParent())) {
-						objects.push_back((ActiveObject*)current);
+				// search for objects which are not us, or we are not its parent
+				// we directly use the callback to populate the result to prevent
+				// a useless result loop here
+				auto include_obj_cb = [self, &objects] (ServerActiveObject *obj) {
+					if (!self || (self != obj && self != obj->getParent())) {
+						objects.push_back((ActiveObject *)obj);
 					}
-				}
+					return false;
+				};
+
+				std::vector<ServerActiveObject *> s_objects;
+				s_env->getObjectsInsideRadius(s_objects, *pos_f, distance, include_obj_cb);
 			}
 		}
 
--- a/src/server/activeobjectmgr.cpp
+++ b/src/server/activeobjectmgr.cpp
@@ -111,17 +111,19 @@ void ActiveObjectMgr::removeObject(u16 id)
 }
 
 // clang-format on
-void ActiveObjectMgr::getObjectsInsideRadius(
-		const v3f &pos, float radius, std::vector<u16> &result)
+void ActiveObjectMgr::getObjectsInsideRadius(const v3f &pos, float radius,
+		std::vector<ServerActiveObject *> &result,
+		std::function<bool(ServerActiveObject *obj)> include_obj_cb)
 {
 	float r2 = radius * radius;
 	for (auto &activeObject : m_active_objects) {
 		ServerActiveObject *obj = activeObject.second;
-		u16 id = activeObject.first;
 		const v3f &objectpos = obj->getBasePosition();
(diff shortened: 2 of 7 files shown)


src/collision.cpp
```diff
@@ -360,17 +360,19 @@ collisionMoveResult collisionMoveSimple(Environment *env, IGameDef *gamedef,
...
 
-				for (u16 obj_id : s_objects) {
-					ServerActiveObject *current = s_env->getActiveObject(obj_id);
-
-					if (!self || (self != current &&
-							self != current->getParent())) {
-						objects.push_back((ActiveObject*)current);
+				// search for objects which are not us, or we are not its parent
+				// we directly use the callback to populate the result to prevent
+				// a useless result loop here
```
sfan5, line 366: Wouldn't passing `objects` to `getObjectsInsideRadius` work?

src/collision.cpp
```diff
@@ -360,17 +360,19 @@ collisionMoveResult collisionMoveSimple(Environment *env, IGameDef *gamedef,
...
+				auto includeObjCallback = [self, &objects](ServerActiveObject *obj) {
```
sfan5, line 367: variables are generally snake_case, also spacing between [] and ()

src/server/activeobjectmgr.cpp
```diff
@@ -111,17 +111,19 @@ void ActiveObjectMgr::removeObject(u16 id)
 }
 
 // clang-format on
-void ActiveObjectMgr::getObjectsInsideRadius(
-		const v3f &pos, float radius, std::vector<u16> &result)
+void ActiveObjectMgr::getObjectsInsideRadius(const v3f &pos, float radius,
+		std::vector<ServerActiveObject *> &result,
+		std::function<bool(ServerActiveObject *obj)> includeObjCallback)
```
sfan5, line 116: snake_case here too

src/serverenvironment.cpp
```diff
@@ -1603,14 +1603,12 @@ void ServerEnvironment::getSelectedActiveObjects(
 	const core::line3d<f32> &shootline_on_map,
 	std::vector<PointedThing> &objects)
 {
-	std::vector<u16> objectIds;
+	std::vector<ServerActiveObject *> objectIds;
```
sfan5, line 1606: should be renamed to `objects` too

src/collision.cpp
nerzhul: unfortunately not because it's a parent class, compiler refuses it :(

src/serverenvironment.cpp
nerzhul: renamed to objs, we have objects as a parameter already

SmallJoker:
Found fitting test mod (N° 02/65): https://github.com/minetest/minetest/pull/6803#issuecomment-353639173
Addition in collision.cpp L360: `ScopeProfiler sp(g_profiler, "Objects in radius", SPT_AVG);`
Addition in test mod: `
minetest.registered_entities["__builtin:item"].initial_properties.collide_with_objects = true`
Test case: `/a 3`

PR + profiler: 82us (8.2e-5)
master + profiler: 142us (0.000142)
Each ~3000 calls/second
Regarding the collision code (only), I measured a ~40% speed improvement.

src/serverenvironment.h
```diff
@@ -323,9 +323,10 @@ class ServerEnvironment : public Environment
 	bool swapNode(v3s16 p, const MapNode &n);
 
 	// Find all active objects inside a radius around a point
-	void getObjectsInsideRadius(std::vector<u16> &objects, const v3f &pos, float radius)
+	void getObjectsInsideRadius(std::vector<ServerActiveObject *> &objects, const v3f &pos, float radius,
+			std::function<bool(ServerActiveObject *obj)> includeObjCallback)
```
sfan5, line 327: snake_case

nerzhul: ah i missed that one :)

nerzhul:
> Found fitting test mod (N° 02/65): [#6803 (comment)](https://github.com/minetest/minetest/pull/6803#issuecomment-353639173)
> Addition in collision.cpp L360:
> 
>     * `ScopeProfiler sp(g_profiler, "Objects in radius", SPT_AVG);`
> 
> 
> Addition in the linked test mod:
> 
>     * ` minetest.registered_entities["__builtin:item"].initial_properties.collide_with_objects = true`
> 
> 
> Test command: `/a 3`
> 
> PR + profiler: 82us (8.2e-5)
> master + profiler: 142us (0.000142)
> Each ~3000 calls/second
> Regarding the collision code (only), I measured a ~40% speed improvement.

thanks for the tests @SmallJoker

PR #9677, opened 2020-04-15
Add an option to disable unittest build, & disable them on Docker build

nerzhul:
All is in the title.
It's not always useful to build unittests, especially on the docker build and on release builds (maybe).
Add a switch for that

## To do

This PR is Ready for Review.

## How to test
build the project with -DBUILD_UNITTESTS=0 and see that the unittest flag disappear from our command line


SmallJoker:
Opting out unittests for faster compiling and lighter binaries. I get the idea, but IMO those tests can also be used as basic sanity checks which should be possible to run at all times when in doubt.

sfan5:
Well there are certain platforms where you can't even run unittests, Android is one of them and also currently builds without them.

nerzhul:
docker build has no interest to contain the unittest, it's a base image you just run. I don't see the point to add a command line option saying hey compile with unittests if you want to run it, unittests are for devel purposes

nerzhul:
are we fine ?

nerzhul:
PR has been updated to:
* show an error message and keep the flag
* disable client unittests too (forgotten in first version)

PR #9682, opened 2020-04-15
Swap out -ffast-math for a safe subset of optimization flags

### LINKED ISSUE #3943: testStreamRead and testBufReader failures
Is this a real bug or just because the values were set to 53.53467f and tested as 53.534f?

```
Test assertion failed: readF1000(is) == 53.534f
    at test_serialization.cpp:305
[FAIL] testStreamRead - 0ms
Test assertion failed: buf.getF1000() == 53.534f
    at test_serialization.cpp:472
[FAIL] testBufReader - 0ms

```

Minetest 0.4.13-dev-f9a9038-dirty
gentoo linux
gcc (Gentoo 4.9.3 p1.3, pie-0.6.3) 4.9.3

### LINKED ISSUE #5330: Touching a wall while falling, sometimes gets fall damage without landing on it, in version 0.4.15
It's difficult to reproduce this bug:
https://www.youtube.com/watch?v=9G1zKrO7NgA

This needs to be fixed as soon as possible.

sfan5:
fixes #3943, fixes #5330 
prevents very visible item collision issues in #9668

All flags used are supported since (at least) gcc 4.7: https://gcc.gnu.org/onlinedocs/gcc-4.7.4/gcc.pdf
clang keeps flag-compatibility with gcc

## Can't this break anything?

Not at all. The BSDs build without this flag, Debian/Ubuntu **always** built without it (until recently).
The debug builds that were running our unittests (CI) build without this.
The only thing this can "break" is performance, but from a quick test with the collision code there is no observable difference.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -713,6 +713,11 @@ else()
 		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DWIN32_LEAN_AND_MEAN")
 	endif()
 
+	# Use a safe subset of flags to speed up math calculations:
+	# - we don't need errno or math exceptions
+	# - we don't deal with Inf/NaN or signed zero
+	set(MATH_FLAGS "-fno-math-errno -fno-trapping-math -ffinite-math-only -fno-signed-zeros")
+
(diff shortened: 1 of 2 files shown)


oilboi:
👍

PR #9683, opened 2020-04-16
Fix typo in Windows build instructions

oilboi:
Fix - to / so it shows that it's a subdirectory


sfan5:
I think that's correct the way it is, you can create a `minetest-build` directory for CMake to put its files in.

SmallJoker:
The currently documented path is saner, because there's no way that build files will find their way into the repository - even if the most common paths are `.gitignore`'d.

PR #9689, opened 2020-04-16
serverpackethandler: Reduce pkt->getPeerId() invocations and more

HybridDog:
This is a code tidy PR. I have replaced the `pkt->getPeerId()` so that the method is called only once instead of many times; additionally I've done some whitespace fixes/changes next to the affected code.

While doing the whitespace changes I noticed two missing spaces and one double space in the message strings, and fixed those. [Here is an example for a missing space](https://github.com/minetest/minetest/blob/45999b74e610b13e8cda20c0c420e152b9837ea6/src/network/serverpackethandler.cpp#L1626-L1627)


src/network/serverpackethandler.cpp
```diff
@@ -761,12 +763,13 @@ void Server::handleCommand_ChatMessage(NetworkPacket* pkt)
...
-	RemotePlayer *player = m_env->getPlayer(pkt->getPeerId());
+	session_t peer_id = pkt->getPeerId();
+	RemotePlayer *player = m_env->getPlayer(peer_id);
 	if (player == NULL) {
-		errorstream << "Server::ProcessData(): Canceling: "
-				"No player for peer_id=" << pkt->getPeerId()
-				<< " disconnecting peer!" << std::endl;
-		DisconnectPeer(pkt->getPeerId());
+		errorstream <<
+			"Server::ProcessData(): Canceling: No player for peer_id=" <<
```
nerzhul, line 770: i don't think this way to call errorstream is compliant with our guidelines no ?

src/network/serverpackethandler.cpp
```diff
@@ -370,18 +372,17 @@ void Server::handleCommand_ClientReady(NetworkPacket* pkt)
 	PlayerSAO* playersao = StageTwoClientInit(peer_id);
 
 	if (playersao == NULL) {
-		actionstream
-			<< "TOSERVER_CLIENT_READY stage 2 client init failed for peer_id: "
-			<< peer_id << std::endl;
+		actionstream <<
+			"TOSERVER_CLIENT_READY stage 2 client init failed for peer_id: " <<
+			peer_id << std::endl;
```
sfan5, line 377: better like this for consistency, also `actionstream` was wrong
```suggestion
		errorstream << "TOSERVER_CLIENT_READY stage 2 client init failed "
			"peer_id=" << peer_id << std::endl;
```

src/network/serverpackethandler.cpp
sfan5: The preferred style is
```c++
    errorstream
        << "foobar" << ...
```
but I don't think this style is disallowed by the guidelines.

HybridDog: I don't find a [guideline](https://dev.minetest.net/Code_style_guidelines#Spaces) which disallows this.
I've moved the message to a separate line so that the constant string is not separated, which should make it easier to read.

HybridDog: Those messages are lots of duplicate code, but fixing that is not part of this issue.

HybridDog: > The preferred style is […]

In Lua I always do the string concatenation the other way round, so I've did it here, too.

src/network/serverpackethandler.cpp
```diff
@@ -109,11 +110,11 @@ void Server::handleCommand_Init(NetworkPacket* pkt)
...
 	if (depl_serial_v == SER_FMT_VER_INVALID) {
-		actionstream << "Server: A mismatched client tried to connect from "
-				<< addr_s << std::endl;
-		infostream<<"Server: Cannot negotiate serialization version with "
-				<< addr_s << std::endl;
-		DenyAccess(pkt->getPeerId(), SERVER_ACCESSDENIED_WRONG_VERSION);
+		actionstream << "Server: A mismatched client tried to connect from " <<
+			addr_s << std::endl;
+		infostream << "Server: Cannot negotiate serialization version with " <<
+			addr_s << std::endl;
```
sfan5, line 116: should be made more useful or removed, my suggestion:
```suggestion
			addr_s << " client_max=" << (int)client_max << std::endl;
```

src/network/serverpackethandler.cpp
```diff
@@ -159,56 +160,56 @@ void Server::handleCommand_Init(NetworkPacket* pkt)
...
 	// Don't enforce for users that have some admin right or mod permits it.
 	if (m_clients.isUserLimitReached() &&
 			playername != g_settings->get("name") &&
 			!m_script->can_bypass_userlimit(playername, addr_s)) {
-		actionstream << "Server: " << playername << " tried to join from "
-				<< addr_s << ", but there" << " are already max_users="
-				<< g_settings->getU16("max_users") << " players." << std::endl;
-		DenyAccess(pkt->getPeerId(), SERVER_ACCESSDENIED_TOO_MANY_USERS);
+		actionstream << "Server: " << playername << " tried to join from " <<
+			addr_s << ", but there" << " are already max_users=" <<
```
sfan5, line 210:
```suggestion
			addr_s << ", but there are already max_users=" <<
```

src/network/serverpackethandler.cpp
```diff
@@ -834,11 +836,12 @@ void Server::handleCommand_Password(NetworkPacket* pkt)
...
-	RemoteClient* client = getClient(pkt->getPeerId(), CS_Created);
+	session_t peer_id = pkt->getPeerId();
+	RemoteClient* client = getClient(peer_id, CS_Created);
 	if (client->net_proto_version >= 25) {
-		infostream << "Server::handleCommand_Password(): Denying change: "
-			<< " Client protocol version for peer_id=" << pkt->getPeerId()
-			<< " too new!" << std::endl;
+		infostream << "Server::handleCommand_Password(): Denying change: " <<
+			"Client protocol version for peer_id=" << peer_id << " too new!" <<
+			std::endl;
```
SmallJoker, line 844: Dead code. Can be removed.

SmallJoker, line 844: This function is dead at least since 0.5.0-dev. Feel free to purge it.

src/network/serverpackethandler.cpp
```diff
@@ -52,11 +52,12 @@ void Server::handleCommand_Init(NetworkPacket* pkt)
 	if(pkt->getSize() < 1)
 		return;
 
-	RemoteClient* client = getClient(pkt->getPeerId(), CS_Created);
+	session_t peer_id = pkt->getPeerId();
+	RemoteClient* client = getClient(peer_id, CS_Created);
```
SmallJoker, line 56: Please correct the pointer placement for all touched lines. Most common one: `RemoteClient* `

HybridDog: I've changed the asterisk position for the four RemoteClient pointers.

src/network/serverpackethandler.cpp
HybridDog: done

PR #9697, opened 2020-04-17
Fix configuration caching in log_deprecated

HybridDog:
I've also fixed a wrong comment.

### DIFF
--- a/src/script/common/c_internal.cpp
+++ b/src/script/common/c_internal.cpp
@@ -157,9 +157,9 @@ static void script_log(lua_State *L, const std::string &message,
 
 void log_deprecated(lua_State *L, const std::string &message, int stack_depth)
 {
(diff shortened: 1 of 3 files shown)


src/script/common/c_internal.cpp
```diff
@@ -157,19 +157,20 @@ static void script_log(lua_State *L, const std::string &message,
 
 void log_deprecated(lua_State *L, const std::string &message, int stack_depth)
 {
-	static bool configured = false;
-	static bool do_log     = false;
-	static bool do_error   = false;
+	static thread_local bool configured = false;
```
sfan5, line 160: either all of these or none should be thread local

HybridDog: fixed

PR #9699, opened 2020-04-17
Apply a scrollbar's default value to scroll containers

### LINKED ISSUE #9691: scroll_container content is reset to the top everytime a formspec is reloaded
Try out the `redesign` branch of `craftguide`: https://github.com/minetest-mods/craftguide/tree/redesign

1. Open the craftguide
2. Move the scrollbar down
3. Click an item

The scrollbar thumb is still correctly positioned but the `scroll_container` content is reset to the top. You have to move the thumb of the scrollbar to show the appropriate content.

@Desour @rubenwardy

Desour:
- Fixes #9691.
- I've also updated the test formspec in minimal.
- `GUIScrollContainer::setScrollBar` now updates the scrolling, which makes sense, as there's a new scrollbar with an arbitrary scroll position.
(I thought about replacing the pointer parameter with a reference because the pointer must not be a nullptr. But the code style guidelines forbid this, it's too C++y.)

## To do

## How to test

- Go to minimal and do `/formspec`. Then, after clicking the scroll container tab, try dragging the scrollbar and don't notice sudden jumps.
- Alternative: Do the steps in #9691.

### DIFF
--- a/games/minimal/mods/test/formspec.lua
+++ b/games/minimal/mods/test/formspec.lua
@@ -251,7 +251,7 @@ Number]
 			"box[1,22.5;4,1;#a00a]"..
 		"scroll_container_end[]"..
 		"scrollbaroptions[max=170]".. -- lowest seen pos is: 0.1*170+6=23 (factor*max+height)
-		"scrollbar[7.5,0;0.3,4;vertical;scrbar;0]"..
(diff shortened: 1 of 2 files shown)


An0n3m0us:
Works! Great job!

PR #9712, opened 2020-04-19
Fix alias handling of get_content_id

### LINKED ISSUE #9632: register_alias_force() breaks get_content_id()
##### Minetest version

5.2.0 Release, but PR #9458 brought the problem to light

##### Summary
`minetest.register_alias_force(a, b)` can cause `get_content_id(a)` to fail.

before 5.2.0 it failed silently, now it causes the game to exit, but it shouldn't be failing at all - the lookup should succeed.

Mods that use get_content_id() won't know whether it will fail/exit before they invoke it because *another mod* may have invoked register_alias_force(). Having mods always perform an alias lookup before every get_content_id() call works around this, but doesn't feel like the right solution.

An example is the nether mod, which has a dependency on stairs. If another mod (moreblocks) invokes register_alias_force("stairs:stair_cobble", "moreblocks:stair_cobble") then nether will crash when it tries to get the id of a node it has a dependency for.

I've discussed it further in PR #9458

##### Steps to reproduce

It doesn't always seem to reproduce, so here's a mod that does it
[aliastest.zip](https://github.com/minetest/minetest/files/4464137/aliastest.zip)

Create a new MTG world with just that mod and run it.

The mod runs the following code with a dependency on default
```
-- in the beginning, get_content_id() works
local it_works  = minetest.get_content_id("default:cobble")
local it_works2 = minetest.get_content_id("default:obsidian")
local it_works3 = minetest.get_content_id("default:dirt_with_grass")

-- register_alias() doesn't stop get_content_id() from working
minetest.register_alias("default:obsidian", "default:dirt_with_grass")
local it_still_works = minetest.get_content_id("default:obsidian")

-- register_alias_force() breaks get_content_id()
minetest.register_alias_force("default:cobble", "default:dirt_with_grass")
local it_fails = minetest.get_content_id("default:cobble")


minetest.log("before alias, id=" .. it_works)
minetest.log("after alias, id=" .. it_still_works)
minetest.log("after alias-force, id=" .. it_fails)
```

sfan5:
...also some miscellaneous cleanups in itemdef.cpp 
fixes #9632

## To do

## How to test

<b>Put the following code in the top level of a mod, not inside a callback or chatcommand.</b>

Test that the following code succeeds:
```lua
minetest.register_alias("aliastest1", "air")
assert(minetest.get_content_id("aliastest1") == minetest.get_content_id("air"))
```

Test that the following code throws an error:
```lua
minetest.register_alias("aliastest2", "asdf:asdf")
minetest.get_content_id("aliastest2")
```

### DIFF
--- a/src/itemdef.cpp
+++ b/src/itemdef.cpp
@@ -270,17 +270,16 @@ public:
 		// Convert name according to possible alias
 		std::string name = getAlias(name_);
 		// Get the definition
-		std::map<std::string, ItemDefinition*>::const_iterator i;
-		i = m_item_definitions.find(name);
-		if(i == m_item_definitions.end())
+		auto i = m_item_definitions.find(name);
+		if (i == m_item_definitions.cend())
 			i = m_item_definitions.find("unknown");
-		assert(i != m_item_definitions.end());
+		assert(i != m_item_definitions.cend());
 		return *(i->second);
 	}
 	virtual const std::string &getAlias(const std::string &name) const
 	{
-		StringMap::const_iterator it = m_aliases.find(name);
-		if (it != m_aliases.end())
+		auto it = m_aliases.find(name);
+		if (it != m_aliases.cend())
 			return it->second;
 		return name;
 	}
@@ -300,8 +299,7 @@ public:
 		// Convert name according to possible alias
 		std::string name = getAlias(name_);
 		// Get the definition
-		std::map<std::string, ItemDefinition*>::const_iterator i;
-		return m_item_definitions.find(name) != m_item_definitions.end();
+		return m_item_definitions.find(name) != m_item_definitions.cend();
 	}
 #ifndef SERVER
 public:
@@ -443,11 +441,9 @@ public:
 	}
 	void clear()
 	{
-		for(std::map<std::string, ItemDefinition*>::const_iterator
-				i = m_item_definitions.begin();
-				i != m_item_definitions.end(); ++i)
+		for (auto &i : m_item_definitions)
 		{
-			delete i->second;
+			delete i.second;
 		}
 		m_item_definitions.clear();
 		m_aliases.clear();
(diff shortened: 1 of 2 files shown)


src/itemdef.cpp
```diff
@@ -270,17 +270,16 @@ class CItemDefManager: public IWritableItemDefManager
 		// Convert name according to possible alias
 		std::string name = getAlias(name_);
 		// Get the definition
-		std::map<std::string, ItemDefinition*>::const_iterator i;
-		i = m_item_definitions.find(name);
-		if(i == m_item_definitions.end())
+		auto i = m_item_definitions.find(name);
+		if(i == m_item_definitions.cend())
```
nerzhul, line 274: maybe fix the space ?

src/itemdef.cpp
```diff
@@ -443,8 +441,7 @@ class CItemDefManager: public IWritableItemDefManager
 	}
 	void clear()
 	{
-		for(std::map<std::string, ItemDefinition*>::const_iterator
-				i = m_item_definitions.begin();
+		for(auto i = m_item_definitions.begin();
```
nerzhul, line 444: you can convert to for (auto &idef: m_item_definitions) { delete idef; } no ?

Treer:
Thanks for fixing that!

HeyITGuyFixIt:
If we get an error from proper usage of `minetest.get_content_id`, it should now be because of a node that doesn't exist?

Treer:
@HeyITGuyFixIt The PR that introduced that change to `get_content_id` is #9458, and they make the argument for the behaviour change there.

My cloudlands mod was also using it to check for a node's existence, so I had to update to code like this:
`if minetest.registered_nodes[name] ~= nil then result = minetest.get_content_id(name) end`

Obviously I had found the old behaviour useful, but I must admit that if you ignore the alias handling bug it uncovered (i.e. what this PR fixes) very few mods seem to have been broken by the new `get_content_id` behaviour. Checking for the existence of nodes seems to be a surprisingly uncommon practice in Minetest mods.

PR #9713, opened 2020-04-19
mapblock_mesh: Optimize a few things

sfan5:
Doubt there's a noticeable difference, but this code is more efficient.

## To do

### DIFF
--- a/src/client/mapblock_mesh.cpp
+++ b/src/client/mapblock_mesh.cpp
@@ -225,7 +225,7 @@ static u16 getSmoothLightCombined(const v3s16 &p,
 		return f.light_propagates;
 	};
 
-	std::array<bool, 4> obstructed = {{ 1, 1, 1, 1 }};
+	bool obstructed[4] = { true, true, true, true };
 	add_node(0);
 	bool opaque1 = !add_node(1);
 	bool opaque2 = !add_node(2);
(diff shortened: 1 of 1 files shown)


src/client/mapblock_mesh.cpp
```diff
@@ -963,8 +961,7 @@ static void updateFastFaceRow(
 		face_dir_corrected = next_face_dir_corrected;
 		std::memcpy(lights, next_lights, ARRLEN(lights) * sizeof(u16));
```
pauloue, line 962:
```suggestion
		std::memcpy(lights, next_lights, sizeof(lights));
```
And the same on line 924

src/client/mapblock_mesh.cpp
```diff
@@ -384,44 +410,16 @@ static void getNodeVertexDirs(const v3s16 &dir, v3s16 *vertex_dirs)
...
-		// faces towards Y- (assume Z+ as "down" in texture)
-		vertex_dirs[0] = v3s16( 1,-1, 1);
-		vertex_dirs[1] = v3s16(-1,-1, 1);
-		vertex_dirs[2] = v3s16(-1,-1,-1);
-		vertex_dirs[3] = v3s16( 1,-1,-1);
-	}
+
+	// Direction must be (1,0,0), (-1,0,0), (0,1,0), (0,-1,0),
+	// (0,0,1), (0,0,-1)
+	assert(dir.X * dir.X + dir.Y * dir.Y + dir.Z * dir.Z == 1);
```
nerzhul, line 416: i don't think this make a very important difference, it seems we have more calculations to do

sfan5: The assert only runs in debug mode.

PR #9718, opened 2020-04-21
FormSpec: Add universal style selector `*`

v-rob:
This PR adds the universal selector `*` to `style` and `style_type`.  It can be combined with state selectors.  This makes it easier to style everything, e.g. set `noclip` or `textcolor` on everything.

## To do

This PR is Ready for Review.

## How to test

```
formspec_version[3]
size[1,2.5]
position[0.35,0.5]
style_type[*;noclip=true]
style[*;textcolor=green]
button[2,0;3,1;;Button]
label[2,2;Label]
style[*:hovered;textcolor=yellow]
style_type[*:pressed;bgcolor=purple]
button[6,0;3,1;;Button]
image_button[6,1.5;3,1;;;Image Button]
```

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -2553,7 +2553,7 @@ Elements
 
 * Set the style for the element(s) matching `selector` by name.
 * `selector` can be one of:
-    * `<name>` - An element name.
+    * `<name>` - An element name. Includes `*`, which represents every element.
     * `<name>:<state>` - An element name, a colon, and one or more states.
 * `state` is a list of states separated by the `+` character.
     * If a state is provided, the style will only take effect when the element is in that state.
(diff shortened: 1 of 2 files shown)


Df458:
Looks good. Could you also add support for `style[*]` as well? I think it would be better for consistency, and I have some ideas for further extensions of style selectors that could make good use of this feature.

v-rob:
Done.  Updated check too.

SmallJoker:
`style[mybutton_*;textcolor=green]` is treated like
`style[*;textcolor=green]`

Bug or feature?

Df458:
I missed this the first time through, but that sounds like a bug to me. IMO,the wildcard character should either have to be alone to work -or- Allow prefix/suffix for partial matches.

I think either of those behaviors would be fine (the former is probably the easiest), but the current behavior should be changed.

v-rob:
Have you actually tested and found this to happen?  If so, it's definitely a bug, but a strange one, because `theme_by_type` and `theme_by_name` are `unordered_map`s, and `find` should find if the entire string is a key.

SmallJoker:
@v-rob I read the code wrong. Somehow I assumed `theme_by_type.find` would search all asterisks within the string of each key. Sorry for the trouble.

PR #9726, opened 2020-04-22
Simplify how parseElement splits element string

pyrollo:
Image buttons are not rendered in formspec if their texture uses modifiers.

This PR simply fix this by adding "image_button" to the compatibility workaround in `parseElement`.

By the way, I'm not sure about this workaround being still needed, maybe a better fix would be to remove it completely.

EDIT: Now this PR simplifies how parseElement splits element string in type/description parts.

## To do

## How to test
My test code:
```
minetest.register_chatcommand("t", {
	params = "",
	description = "Test",
	func = function(name, param)
		minetest.show_formspec(name, "test", [[
			size[1,1]
			image_button[0,0;1,1;default_stone.png^[colorize:#F008;test;]
		]]);
	end
})
```
Without the fix --> Empty formspec
With the fix --> Expected image button displayed

### DIFF
--- a/src/gui/guiFormSpecMenu.cpp
+++ b/src/gui/guiFormSpecMenu.cpp
@@ -2668,24 +2668,12 @@ void GUIFormSpecMenu::parseElement(parserData* data, const std::string &element)
 	if (parseVersionDirect(element))
 		return;
 
-	std::vector<std::string> parts = split(element,'[');
-
-	// ugly workaround to keep compatibility
-	if (parts.size() > 2) {
-		if (trim(parts[0]) == "image") {
-			for (unsigned int i=2;i< parts.size(); i++) {
-				parts[1] += "[" + parts[i];
-			}
-		}
-		else { return; }
-	}
-
-	if (parts.size() < 2) {
+	size_t pos = element.find('[');
+	if (pos == std::string::npos)
 		return;
-	}
 
-	std::string type = trim(parts[0]);
-	std::string description = trim(parts[1]);
+	std::string type = trim(element.substr(0, pos));
+	std::string description = element.substr(pos+1);
 
 	if (type == "container") {
 		parseContainer(data, description);


sfan5:
If you use formspec escaping on the texture name (as it is intended), this workaround should not be necessary.

pyrollo:
Oh yes, thanks! That works great :)

But I still don't understand what this `parseElement` code is for: 

```c++
	std::vector<std::string> parts = split(element,'[');

	// ugly workaround to keep compatibility
	if (parts.size() > 2) {
		if (trim(parts[0]) == "image" || trim(parts[0]) == "image_button") {
			for (unsigned int i=2;i< parts.size(); i++) {
				parts[1] += "[" + parts[i];
			}
		}
		else { return; }
	}

	if (parts.size() < 2) {
		return;
	}

	std::string type = trim(parts[0]);
	std::string description = trim(parts[1]);
```

If I understand correctly this ensures that there are no other "[" char in the formspec element, except for `image` element.

The whole thing could be replaced by something like (untested) :

```c++
	size_t pos = element.find('[');
	if (pos == string::npos)
		return;

	std::string type = trim(element.substr(0, pos));
	std::string description = trim(element.substr(pos+1));
```

`type` would hold the left part of the first [ char and all the rest (`description`) is passed to subsequent pasre* functions.

src/gui/guiFormSpecMenu.cpp
```diff
@@ -2668,24 +2668,12 @@ void GUIFormSpecMenu::parseElement(parserData* data, const std::string &element)
...
-	if (parts.size() < 2) {
+	size_t pos = element.find('[');
+	if (pos == std::string::npos)
 		return;
-	}
 
-	std::string type = trim(parts[0]);
-	std::string description = trim(parts[1]);
+	std::string type = trim(element.substr(0, pos));
+	std::string description = trim(element.substr(pos+1));
```
Desour, line 2676: Why is there a `trim` for `description` btw.? This makes it impossible to eg. set a space as default value for a `field[]`.

v-rob: Ah, that's why I couldn't have newlines at the end of my vertlabels.  I think this should be removed.

pyrollo: You are right, this should be the responsibility of element specific parse function to handle `description` part.

PR #9736, opened 2020-04-23
Add chat_font_size setting

### LINKED ISSUE #9616: Client side minetest.get_node_light(pos, 0.5) crashes the game
minetest.get_node_light(pos, 0.5) crashes the game when called in a client side mod

### LINKED ISSUE #9660: Add Separate chat_font_size Setting To Improve In Game Accessibly
## Problem
I requires a separate chat size to the UI's settings since chat still feels/looks small to me even on font size 30 and 28 for failback. 

## Solutions

Separate optional chat font size setting to be able to set custom font size JUST for chat. To improve in game accessibly in this regard.  

## Alternatives

No real alternatives since beyond ~30 font size UI suffers greatly on my setup.

SmallJoker:
Closes #9660 (EDIT: autocomplete failed)
Default font sizes are used when the setting value is `0` or below (clamped by Settings).

## To do

This PR is Ready for Review.

## How to test

1) No settings change: test the chat prompt and the recent chat - must use the default sizes.
2) `chat_font_size = 35`: Huge fonts for the chat prompt and recent chat

### DIFF
--- a/src/client/gameui.cpp
+++ b/src/client/gameui.cpp
@@ -76,6 +76,11 @@ void GameUI::init()
 	m_guitext_chat = gui::StaticText::add(guienv, L"", core::rect<s32>(0, 0, 0, 0),
 		//false, false); // Disable word wrap as of now
 		false, true, guiroot);
+	u16 chat_font_size = g_settings->getU16("chat_font_size");
+	if (chat_font_size != 0) {
+		m_guitext_chat->setOverrideFont(g_fontengine->getFont(
+			chat_font_size, FM_Unspecified));
+	}
 
 	// Profiler text (size is updated when text is updated)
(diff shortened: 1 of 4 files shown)


src/client/gameui.cpp
```diff
@@ -76,6 +76,11 @@ void GameUI::init()
 	m_guitext_chat = gui::StaticText::add(guienv, L"", core::rect<s32>(0, 0, 0, 0),
 		//false, false); // Disable word wrap as of now
 		false, true, guiroot);
+	u16 chat_font_size = g_settings->getU16("chat_font_size");
```
sfan5, line 79: wouldn't it be better if the default value was empty instead of 0?

SmallJoker: This is an integer setting, hence setting this value to an empty string will not be possible using the `Settings` formspec.

PR #9740, opened 2020-04-24
Add MSVC build

adrido:
This adds MSVC builds to GitHub actions including x86, x64, portable and installer

## To do

This PR is Ready for Review.

## How to test

Wait until build completes, go to Actions download artifact and try it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -171,7 +171,7 @@ jobs:
           NO_PACKAGE: 1
   msvc:
     name: VS 2019 ${{ matrix.config.arch }}-${{ matrix.type }}
-    runs-on: windows-latest 
+    runs-on: windows-2019 
     env:
       VCPKG_VERSION: c7ab9d3110813979a873b2dbac630a9ab79850dc
 #                    2020.04
@@ -235,4 +235,3 @@ jobs:
         with:
           name: msvc-${{ matrix.config.arch }}-${{ matrix.type }}
           path: .\Package\
-          
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -169,3 +169,70 @@ jobs:
         env:
           NO_MINETEST_GAME: 1
           NO_PACKAGE: 1
+  msvc:
+    name: VS 2019 ${{ matrix.config.arch }}-${{ matrix.type }}
+    runs-on: windows-latest 
+    env:
+      VCPKG_VERSION: c7ab9d3110813979a873b2dbac630a9ab79850dc
+#                    2020.04
+      vcpkg_packages: irrlicht zlib curl[winssl] openal-soft libvorbis libogg sqlite3 freetype luajit
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          - {
+              arch: x86,
+              generator: "-G'Visual Studio 16 2019' -A Win32",
+              vcpkg_triplet: x86-windows
+            }
+          - {
+              arch: x64,
+              generator: "-G'Visual Studio 16 2019' -A x64",
+              vcpkg_triplet: x64-windows
+            }
+        type: [portable, installer]
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Restore from cache and run vcpkg
+        uses: lukka/run-vcpkg@v2
+        with:
+          vcpkgArguments: ${{env.vcpkg_packages}}
+          vcpkgDirectory: '${{ github.workspace }}\vcpkg'
+          appendedCacheKey: ${{ matrix.config.vcpkg_triplet }}
+          vcpkgGitCommitId: ${{ env.VCPKG_VERSION }}
+          vcpkgTriplet: ${{ matrix.config.vcpkg_triplet }}
+
+      - name: CMake
+        run: |
+          cmake ${{matrix.config.generator}}  `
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -169,3 +169,70 @@ jobs:
         env:
           NO_MINETEST_GAME: 1
           NO_PACKAGE: 1
+  msvc:
+    name: VS 2019 ${{ matrix.config.arch }}-${{ matrix.type }}
+    runs-on: windows-latest
```
nerzhul, line 174: don't use latest please

.github/workflows/build.yml
```diff
@@ -169,3 +169,70 @@ jobs:
...
+          {
+            cpack -G ZIP -B "$env:GITHUB_WORKSPACE\Package"
+          }
+        env:
+          TYPE: ${{matrix.type}}
+
+      - name: Package Clean
+        run: rm -r $env:GITHUB_WORKSPACE\Package\_CPack_Packages
+
+      - uses: actions/upload-artifact@v1
```
nerzhul, line 234: we don't use artifacts uploads currently, you can drop it

adrido: That's IMO the most important step after the build, as nobody else is doing windows builds, and there is no way to get the installer.

.github/workflows/build.yml
```diff
@@ -169,3 +169,69 @@ jobs:
...
+              arch: x86,
+              generator: "-G'Visual Studio 16 2019' -A Win32",
+              vcpkg_triplet: x86-windows
+            }
+          - {
+              arch: x64,
+              generator: "-G'Visual Studio 16 2019' -A x64",
+              vcpkg_triplet: x64-windows
+            }
+        type: [portable, installer]
```
sfan5, line 193: `portable` is pretty much guaranteed to work, it's enough if these jobs test only the installer

adrido: Someone still have to do it. If there is a change, that cant be tested by simply compiling it, users have to either compile the changes by themself or they can now just download the package and and try it without the need of compiling. 

It runs parallel, its already implemented, why not use it?

sfan5: > Someone still have to do it.

The Gitlab pipeline already tests ZIP package generation.

> If there is a change, that cant be tested by simply compiling it, users have to either compile the changes by themself or they can now just download the package and and try it without the need of compiling.

I am not saying artifacts should be removed.

> It runs parallel, its already implemented, why not use it?

I don't really like have the CI waste time even if it's "free".

nerzhul: remember that we can run 10 jobs in parallel in the free version here, but we may just run this on master, it's less runs that on each pull request, and is more reasonable no ? :)

sfan5: There's only 10 jobs in parallel? Then this definitely does make a difference.

adrido: There are 20 concurrent jobs: https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits

Gitlab only compiles the master branch, not the PR.

As long as there is a working artifact it is not "waste time" imo.
----
If you really want to save some jobs on the PR, I would suggest to skip the installer builds on PR. Potential testers can then have the stable release version installed, but can still test the portable version without breaking one of their worlds. Would that be ok?

nerzhul: oh yeah it's 20, we are fine then :) sorry i misread the doc at the beginning :)

sfan5: > Potential testers can then have the stable release version installed, but can still test the portable version without breaking one of their worlds. Would that be ok?

Yeah skip the `installer` variant then.

adrido: > oh yeah it's 20, we are fine then

> Yeah skip the `installer` variant then.

😂 Keep it as its currently is, or skip the installer?

sfan5: I still propose having just `portable` builds since that fulfills both the "test compiling with MSVC" and "produce usable artifact".

rubenwardy: Can you produce installer builds on tagged commits?

adrido: > Can you produce installer builds on tagged commits?
could be possible https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif But it depends on how does the if work in combination with the matrix

sfan5:
I don't know if Github's documentation is wrong or whatever, but here you can see 10 jobs running and one waiting to be ran. This implies there are not 20 but 10 concurrent jobs.
Bildschirmfoto_2020-05-02_12-55-40

.github/workflows/build.yml
adrido: > 
> 
> > Can you produce installer builds on tagged commits?
> 
> could be possible https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idif But it depends on how does the if work in combination with the matrix

Sadly it isn't possible yet https://github.community/t5/GitHub-Actions/Matrix-cannot-be-used-in-jobs-level-if/td-p/41048

I would suggest to create a build-release.yml that is doing the full job (including minetest game) if a tag is pushed. But I think that will be better done in a separate PR

I currently disabled the installer builds. If someone is working on the installer he can simply enable them.

adrido:
grafik
11 Jobs running, a few seconds later there were 13 running 1 one finished.
So I couldn't reproduce it, did you report it to GitHub?

nerzhul:
merged, thanks for your time. Maybe someone can provide a macosx one after that ? (don't know if we have mac osx users yet)

PR #9746, opened 2020-04-25
Settings: Unwrap overcomplicated flags calculation

SmallJoker:
Requires #9737 to work.

## To do

This PR is Ready for Review.

## How to test

1) Settings as follows: https://github.com/minetest/minetest/pull/9737#issuecomment-619174680
2) Generate a new world, join and leave
3) Verify that the flags functions are still working: https://github.com/minetest/minetest/pull/9284#issue-359210913

### DIFF
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -484,30 +484,18 @@ v3f Settings::getV3F(const std::string &name) const
 u32 Settings::getFlagStr(const std::string &name, const FlagDesc *flagdesc,
 	u32 *flagmask) const
 {
-	u32 flags = 0;
-	u32 mask_default = 0;
-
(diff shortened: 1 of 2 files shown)


src/settings.cpp
```diff
@@ -1057,14 +1065,27 @@ void Settings::overrideDefaults(Settings *other)
 		}
 		const FlagDesc *flagdesc = getFlagDescFallback(setting.first);
 		if (flagdesc) {
-			// Flags cannot be copied directly.
```
sfan5, line 1060: while you're editing this method, is it possible to make the `Settings *other` argunment `const`?

SmallJoker:
Superseded by #10661

Issue #6461, opened 2017-09-25
minetest --version build date update


HybridDog:
I've found a problem: When compiling minetest after git pulling, version.cpp is only compiled if there're changes. So the date shown by `minetest --version` isn't updated.
To fix it, version.cpp needs to be compiled every time make is executed.

See also #6331

sfan5:
who wanted a date in there anyway, this is stupid

SmallJoker:
His and my fault: #6331 
Moving it into version.cpp allowed a nice build date stamp in the version info. However, it wasn't tested or analysed enough to notice the problem of incremental building.

PR #9749, opened 2020-04-25
 Fix detection of in-place path_locale when RUN_IN_PLACE=0 

sfan5:
<b>don't squash</b>

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -28,13 +28,3 @@
 		#define BUILD_TYPE "Debug"
 	#endif
 #endif
-
-#define BUILD_INFO \
-	"BUILD_TYPE=" BUILD_TYPE "\n"          \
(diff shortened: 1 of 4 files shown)


src/config.h
```diff
@@ -28,13 +28,3 @@
 		#define BUILD_TYPE "Debug"
 	#endif
 #endif
-
-#define BUILD_INFO \
```
SmallJoker, line 32: 50b2185ced9dc64e054c
#6461

sfan5: As far as I've understood, this only affects putting a build date there.

If the version changes, the `cmake_config_githash.h` is updated and then `version.cpp` will also be recompiled. Am I missing something?

Wuzzy2:
Works for me. ^^

src/config.h
SmallJoker: Makes sense. We can investigate this if #defines other than the build date are affected.

PR #9756, opened 2020-04-26
Add PostgreSQL authentication backend

nerzhul:
Add support for PostgreSQL authentication Database.
This is certainly the most important database backend, as it permits to share auth between servers for a server galaxy.

## To do

This PR is Ready for Review.

### DIFF
--- a/src/database/database-postgresql.cpp
+++ b/src/database/database-postgresql.cpp
@@ -166,6 +166,11 @@ void Database_PostgreSQL::endSave()
 	checkResults(PQexec(m_conn, "COMMIT;"));
 }
 
+void Database_PostgreSQL::rollback()
+{
+	checkResults(PQexec(m_conn, "ROLLBACK;"));
+}
+
 MapDatabasePostgreSQL::MapDatabasePostgreSQL(const std::string &connect_string):
 	Database_PostgreSQL(connect_string),
 	MapDatabase()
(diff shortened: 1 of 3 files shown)


src/database/database-postgresql.cpp
```diff
@@ -637,4 +642,174 @@ void PlayerDatabasePostgreSQL::listPlayers(std::vector<std::string> &res)
...
+		lastLoginStr.c_str()
+	};
+
+	PGresult *result = execPrepared("auth_create", 3, values, false, false);
+
+	int numrows = PQntuples(result);
+	if (numrows == 0) {
+		errorstream << "Strange behaviour on auth creation, no ID returned." << std::endl;
+		PQclear(result);
+		rollback();
```
sfan5, line 756: doesn't this require `beginSave` to be called first?

src/database/database-postgresql.cpp
```diff
@@ -637,4 +642,174 @@ void PlayerDatabasePostgreSQL::listPlayers(std::vector<std::string> &res)
...
+{
+	connectToDatabase();
+}
+
+void AuthDatabasePostgreSQL::createDatabase()
+{
+	createTableIfNotExists("auth",
+		"CREATE TABLE auth ("
+			"id SERIAL,"
+			"name VARCHAR(32) UNIQUE,"
```
sfan5, line 656: 32 seems a bit short
In fact the [postgres documentation](https://www.postgresql.org/docs/current/datatype-character.html) suggests to just use `TEXT` in all cases.

src/database/database-postgresql.cpp
nerzhul: yep it's misplaced, my bad

src/database/database-postgresql.cpp
nerzhul: yep i prefer to use TEXT too, it was just to keep compat with sqlite3 format, but it's not a good pg practice, removing

nerzhul:
@sfan5 it's fixed and ready for a merge

src/database/database-postgresql.cpp
```diff
@@ -637,4 +642,174 @@ void PlayerDatabasePostgreSQL::listPlayers(std::vector<std::string> &res)
...
+			"privilege TEXT,"
+			"PRIMARY KEY (id, privilege),"
+			"CONSTRAINT fk_id FOREIGN KEY (id) REFERENCES auth (id) ON DELETE CASCADE"
+		");");
+}
+
+void AuthDatabasePostgreSQL::initStatements()
+{
+	prepareStatement("auth_read", "SELECT id, name, password, last_login FROM auth WHERE name = $1");
+	prepareStatement("auth_write", "UPDATE auth set name = $1, password = $2, last_login = $3 WHERE id = $4");
```
sfan5, line 674: just for consistency
```suggestion
	prepareStatement("auth_write", "UPDATE auth SET name = $1, password = $2, last_login = $3 WHERE id = $4");
```

Issue #9652, opened 2020-04-12
Node digging particles only spawn on top of node


oilboi:
Node digging particles only spawn on top of node but the collision detection seems to work a lot better

Wuzzy2:
Confirmed. Issue in 0ac999ded725f8efcd26db284161683e37efeecf (5.3.0-dev), but not 5.2.0, so it must have been introduced very recently.

It looks very odd when the node is mined, as all the particles will fall down at once. In 5.2.0, it looked much better.


How to reproduce:
Disable Creative Mode, dig dirt in MTG with your hand.

sfan5:
When you noclip inside the node you can see that particles are spawned inside but pushed upwards.
Reverting 3ad5388c6d3894e2f4aa7158cc2b62b626f0f967 fixes this, so #9343 is the cause.

@TheTermos any idea about the preferred way to fix this?

TheTermos:
Looking into this.

TheTermos:
Particles are spawned inside a node, so they are in fact in collision.
I'm working on a quick and dirty solution.

Slow and proper solution would be for particles to be emitted outside a node, in the middle of the dug face.

TheTermos:
The quickest and the dirtiest: change particles to non physical.

Little eye candy is lost, because particles will not collide with the ground, but should be ok as a temporary solution.

client/particles.cpp line 654:
`true,`
becomes
`false,`
that's all

oilboi:
> The quickest and the dirtiest: change particles to non physical.
> 
> Little eye candy is lost, because particles will not collide with the ground, but should be ok as a temporary solution.
> 
> client/particles.cpp line 654:
> `true,`
> becomes
> `false,`
> that's all

Is there a simple way to find nodes that are nodeboxes, ignore, or air then spawn the particles near there with a velocity outwards of the center of the node?

TheTermos:
There is a simple way of finding exactly the spot just outside the particular face being dug, I'd go with that first.

KaylebJay:
I actually like the new particles look... I had no idea it was a 'bug' :D

KaylebJay:
![screenshot_20200423_071844](https://user-images.githubusercontent.com/44441970/80106391-07df6d80-8537-11ea-8786-3f55a65266b6.png)
The particles do look nicer but I can see the effect after looking closer. Its definitely a bug.

TheTermos:
[...]

PR #9764, opened 2020-04-27
Collision detection - #9343 follow-up

TheTermos:
This PR is a follow-up to #9343. It addresses the following issues found during 5.3 testing

- Oddly slow velocity passed from localplayer.cpp (e.g. 1e-43,0,0)
If velocity is used in calculations as a divisor this can cause the result to be infinity. 
Passed velocity components are now being truncated to 4 decimal places.

- #9652
Clipping recovery has been modified so the outcome is more similar to how it used to be in the old system.

- Improved collision detection by truncating calculated distance of movement to 2 decimal places
This helps distribute floating point error on one side of the correct value (to make sure an object rather falls short that overshoot)

- minor performance optimization
If there's a collision, it's along Y 99% of the time. Y is now tested first.

## To do

This PR is Ready for Review.

## How to test
- compare how object-object and player-object collisions work in this PR and 5.2.0
example entity on_step: 
```
self.object:set_acceleration({x=0,y=-10,z=0})		
self.object:set_velocity({z=1.0,y=0,x=0.0})
```
Jump on the entity and try to stay on top
Go in front of it and push against it

- dig some nodes and observe the digging particles.

### DIFF
--- a/src/collision.cpp
+++ b/src/collision.cpp
@@ -54,6 +54,22 @@ struct NearbyCollisionInfo {
 	aabb3f box;
 };
 
+// Helper functions:
+// Truncate floating point numbers to specified number of decimal places
+// in order to move all the floating point error to one side of the correct value
+static inline f32 truncate(const f32 val, const u16 exp)
+{
+	return std::trunc(val * 10.0f * exp) / (10.0f * exp);
+}
+
+static inline v3f truncate(const v3f& vec, const u16 exp)
+{
+	return v3f(
+		truncate(vec.X, exp),
+		truncate(vec.Y, exp),
+		truncate(vec.Z, exp)
+	);
+}
 
 // Helper function:
 // Checks for collision of a moving aabbox with a static aabbox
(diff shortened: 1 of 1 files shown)


src/collision.cpp
```diff
@@ -54,6 +54,22 @@ struct NearbyCollisionInfo {
 	aabb3f box;
 };
 
+// Helper functions:
+// Truncate floating point numbers to specified number of decimal places
+// in order to move all the floating point error to one side of the correct value
+static inline f32 truncate(const f32 val, const u16 exp)
+{
+	return std::trunc(val * 10.0f * exp) / (10.0f * exp);
```
sfan5, line 62:
```suggestion
	return truncf(val * 10.0f * exp) / (10.0f * exp);
```

src/collision.cpp
```diff
@@ -489,8 +506,8 @@ collisionMoveResult collisionMoveSimple(Environment *env, IGameDef *gamedef,
 						pos_f->Z += speed_f->Z * nearest_dtime;
 				}
 			} else {
-				*pos_f += *speed_f * nearest_dtime;
-				dtime -= nearest_dtime;
+				*pos_f += truncate(*speed_f * nearest_dtime, 2);
+				dtime -= nearest_dtime;
```
sfan5, line 510: trailing space

src/collision.cpp
TheTermos: I hear math.h is supposed to be deprecated. How about `std::truncf`?

sfan5: Is it? I thought `std::foo` vs `foo` is just a style choice (assuming no function overloads).

TheTermos: yes, looks like in this case they both point to the same place.
Done

src/collision.cpp
```diff
@@ -54,6 +54,22 @@ struct NearbyCollisionInfo {
 	aabb3f box;
 };
 
+// Helper functions:
+// Truncate floating point numbers to specified number of decimal places
+// in order to move all the floating point error to one side of the correct value
+static inline f32 truncate(const f32 val, const u16 exp)
+{
+	return truncf(val * 10.0f * exp) / (10.0f * exp);
```
SmallJoker, line 62: `exp` is misleading. `factor` would fit better to express the linearity.
If you'd instead like to specify the decimal places directly: `pow(10.0f, exp)`
In terms of performance I'd prefer keeping this as a simple factor, and passing `100.0f` as argument below.

TheTermos: Not a fan of that solution, you'd be able to pass just any float there, but I'm going to change it if you insist.

src/collision.cpp
```diff
@@ -66,95 +82,94 @@ CollisionAxis axisAlignedCollision(
...
-						(speed.X < 0 && staticbox.MinEdge.X < movingbox.MinEdge.X)) {
-					if (
-						(std::max(movingbox.MaxEdge.Y + speed.Y * time, staticbox.MaxEdge.Y)
-							- std::min(movingbox.MinEdge.Y + speed.Y * time, staticbox.MinEdge.Y)
-							- relbox.MinEdge.Y < 0) &&
+		if (*dtime <= dtime_max) {
+			inner_margin = std::max(-0.5f * (staticbox.MaxEdge.Y - staticbox.MinEdge.Y), -2.0f);
+
+			if ((speed.Y > 0 && staticbox.MinEdge.Y - movingbox.MaxEdge.Y > inner_margin) ||
+				(speed.Y < 0 && movingbox.MinEdge.Y - staticbox.MaxEdge.Y > inner_margin)) {
```
SmallJoker, line 108: Would it be possible to simplify these checks a bit?

```C++
aabb3f moved_box = movingbox;
moved_box.MinEdge *= speed * time;
moved_box.MaxEdge *= speed * time;

f32 direction = (speed.Y == 0.0f) ? 0.0f : speed.Y / abs(speed.Y);

if (staticbox.MinEdge.Y - movingbox.MaxEdge.Y > inner_margin * direction) {
	if (std::max(moved_box.MaxEdge.X, staticbox.MaxEdge.X)
			- std::max(moved_box.MinEdge.X, staticbox.MinEdge.X) < relbox.MinEdge.X)
		// if ( .. Z

```

To avoid code duplication, it would also be possible to make use of a generic function like 
```C++
template <float v3f::*A, float v3f::*B, float v3f::*B>
static bool checkCollision(const v3f &speed, const aabb3f &staticbox, const aabb3f &movingbox)
{
	if (staticbox.MinEdge.*A - movingbox.MaxEdge.*A > inner_margin * direction) {
}
```

src/collision.cpp
SmallJoker: I don't mind. Regarding performance, a factor would be a better choice. Feel free to use `exp` if you prefer to keep the decimal point argument.

src/collision.cpp
TheTermos: - simplify: 
I wish but unfortunately the suggested code isn't equivalent.
- generic:
Yes, that probably wouldn't improve readability that much and I'm not terribly familiar with pointers to members so I really prefer not to do that.

src/collision.cpp
TheTermos: Shouldn't matter that much performance wise, the operation should on average be done 3 * #movingboxes / step, but if precision is increased to 64 then this function should be used more often, so
changed as requested.

src/collision.cpp
SmallJoker: Which part of the suggested code is not equivalent? Is it the msising `direction != 0.0f` in the first condition? Just wondering what I missed there.

PR #9769, opened 2020-04-27
Server class code cleanups

nerzhul:
This PR is composed of 2 tiny cleanups & 1 responbitility split
* drop a useless bool return on overrideDayNightRatio
* make Server::init() private 
* Move inventory manager responsibility outside of Server, to a ServerInventoryMgr object

## To do

This PR is Ready for Review.

## How to test

Launch a server with bunch of detached inventories, it should work as intended

Note: during my refacto i saw that some code parts are not perfect, but i didn't refactored them, like the sendDetachedInventories() which will recheck by sendDetachedInventory call the peer_id.
This part is interesting because if i connect to a server i will send detached inventories to some attached to other players, whereas they doesn't care

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -887,7 +887,6 @@ static bool run_dedicated_server(const GameParams &game_params, const Settings &
 			// Create server
 			Server server(game_params.world_path, game_params.game_spec,
 					false, bind_addr, true, &iface);
-			server.init();
 
 			g_term_console.setup(&iface, &kill, admin_nick);
 
@@ -922,7 +921,6 @@ static bool run_dedicated_server(const GameParams &game_params, const Settings &
 			// Create server
 			Server server(game_params.world_path, game_params.game_spec, false,
 				bind_addr, true);
-			server.init();
 			server.start();
 
 			// Run server
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -3442,15 +3442,12 @@ void Server::setClouds(RemotePlayer *player, const CloudParams &params)
 	SendCloudParams(player->getPeerId(), params);
 }
 
-bool Server::overrideDayNightRatio(RemotePlayer *player, bool do_override,
+void Server::overrideDayNightRatio(RemotePlayer *player, bool do_override,
 	float ratio)
 {
-	if (!player)
-		return false;
-
+	sanity_check(player);
 	player->overrideDayNightRatio(do_override, ratio);
 	SendOverrideDayNightRatio(player->getPeerId(), do_override, ratio);
-	return true;
 }
 
 void Server::notifyPlayers(const std::wstring &msg)
--- a/src/server.cpp
+++ b/src/server.cpp
@@ -64,6 +64,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include "chat_interface.h"
 #include "remoteplayer.h"
 #include "server/player_sao.h"
+#include "server/serverinventorymgr.h"
 #include "translation.h"
 
 class ClientNotFoundException : public BaseException
@@ -218,10 +219,8 @@ Server::Server(
 	m_path_world(path_world),
 	m_gamespec(gamespec),
 	m_simple_singleplayer_mode(simple_singleplayer_mode),
-	m_max_chatmessage_length(g_settings->getU16("chat_message_max_size")),
 	m_dedicated(dedicated),
 	m_async_fatal_error(""),
-	m_liquid_transform_every(g_settings->getFloat("liquid_update")),
 	m_con(std::make_shared<con::Connection>(PROTOCOL_ID,
 			512,
 			CONNECTION_TIMEOUT,
@@ -230,12 +229,9 @@ Server::Server(
 	m_itemdef(createItemDefManager()),
 	m_nodedef(createNodeDefManager()),
 	m_craftdef(createCraftDefManager()),
-	m_lag(g_settings->getFloat("dedicated_server_step")),
 	m_thread(new ServerThread(this)),
 	m_clients(m_con),
 	m_admin_chat(iface),
-	m_csm_restriction_flags(g_settings->getU64("csm_restriction_flags")),
-	m_csm_restriction_noderange(g_settings->getU32("csm_restriction_noderange")),
 	m_modchannel_mgr(new ModChannelMgr())
 {
 	if (m_path_world.empty())
@@ -343,11 +339,6 @@ Server::~Server()
 	infostream << "Server: Deinitializing scripting" << std::endl;
 	delete m_script;
 
-	// Delete detached inventories
-	for (auto &detached_inventory : m_detached_inventories) {
-		delete detached_inventory.second;
-	}
-
 	while (!m_unsent_map_edit_queue.empty()) {
 		delete m_unsent_map_edit_queue.front();
 		m_unsent_map_edit_queue.pop();
@@ -393,6 +384,9 @@ void Server::init()
 
 	m_script = new ServerScripting(this);
 
+	// Must be created before mod loading because we have some inventory creation
+	m_inventory_mgr = std::unique_ptr<ServerInventoryManager>(new ServerInventoryManager());
+
 	m_script->loadMod(getBuiltinLuaPath() + DIR_DELIM "init.lua", BUILTIN_MOD_NAME);
 
 	m_modmgr->loadMods(m_script);
@@ -427,6 +421,7 @@ void Server::init()
 	// Initialize Environment
 	m_env = new ServerEnvironment(servermap, m_script, this, m_path_world);
 
+	m_inventory_mgr->setEnv(m_env);
 	m_clients.setEnv(m_env);
 
 	if (!servermap->settings_mgr.makeMapgenParams())
@@ -447,6 +442,13 @@ void Server::init()
 	servermap->addEventReceiver(this);
 
 	m_env->loadMeta();
+
+	// Those settings can be overwritten in world.mt, they are
+	// intended to be cached after environment loading.
+	m_liquid_transform_every = g_settings->getFloat("liquid_update");
+	m_max_chatmessage_length = g_settings->getU16("chat_message_max_size");
+	m_csm_restriction_flags = g_settings->getU64("csm_restriction_flags");
+	m_csm_restriction_noderange = g_settings->getU32("csm_restriction_noderange");
 }
 
 void Server::start()
@@ -1180,82 +1182,6 @@ void Server::onMapEditEvent(const MapEditEvent &event)
 		return;
 
 	m_unsent_map_edit_queue.push(new MapEditEvent(event));
-}
-
-Inventory* Server::getInventory(const InventoryLocation &loc)
-{
-	switch (loc.type) {
-	case InventoryLocation::UNDEFINED:
-	case InventoryLocation::CURRENT_PLAYER:
-		break;
-	case InventoryLocation::PLAYER:
-	{
-		RemotePlayer *player = m_env->getPlayer(loc.name.c_str());
-		if(!player)
-			return NULL;
-		PlayerSAO *playersao = player->getPlayerSAO();
-		if(!playersao)
-			return NULL;
-		return playersao->getInventory();
-	}
-		break;
-	case InventoryLocation::NODEMETA:
-	{
-		NodeMetadata *meta = m_env->getMap().getNodeMetadata(loc.p);
-		if(!meta)
-			return NULL;
-		return meta->getInventory();
-	}
-		break;
-	case InventoryLocation::DETACHED:
-	{
-		if(m_detached_inventories.count(loc.name) == 0)
-			return NULL;
-		return m_detached_inventories[loc.name];
-	}
-		break;
-	default:
-		sanity_check(false); // abort
-		break;
-	}
-	return NULL;
-}
-
-void Server::setInventoryModified(const InventoryLocation &loc)
-{
-	switch(loc.type){
-	case InventoryLocation::UNDEFINED:
-		break;
-	case InventoryLocation::PLAYER:
-	{
-
-		RemotePlayer *player = m_env->getPlayer(loc.name.c_str());
-
-		if (!player)
-			return;
-
-		player->setModified(true);
-		player->inventory.setModified(true);
-		// Updates are sent in ServerEnvironment::step()
-	}
-		break;
-	case InventoryLocation::NODEMETA:
-	{
-		MapEditEvent event;
-		event.type = MEET_BLOCK_NODE_METADATA_CHANGED;
-		event.p = loc.p;
-		m_env->getMap().dispatchEvent(event);
-	}
-		break;
-	case InventoryLocation::DETACHED:
-	{
-		// Updates are sent in ServerEnvironment::step()
-	}
-		break;
-	default:
-		sanity_check(false); // abort
-		break;
-	}
 }
 
 void Server::SetBlocksNotSent(std::map<v3s16, MapBlock *>& block)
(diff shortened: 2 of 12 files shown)


src/server/serverinventorymgr.cpp
```diff
@@ -0,0 +1,180 @@
...
+bool ServerInventoryMgr::removeDetachedInventory(const std::string &name)
+{
+	const auto &inv_it = m_detached_inventories.find(name);
+	if (inv_it == m_detached_inventories.end())
+		return false;
+
+	delete inv_it->second;
+	m_detached_inventories.erase(inv_it);
+
+	if (!m_env) // Mods are not done loading
```
nerzhul, line 137: note: this is a historical hack, i really don't like it, but it's keep as is

src/server.cpp
```diff
@@ -442,15 +445,12 @@ void Server::init()
 	servermap->addEventReceiver(this);
 
 	m_env->loadMeta();
-
-	m_liquid_transform_every = g_settings->getFloat("liquid_update");
-	m_max_chatmessage_length = g_settings->getU16("chat_message_max_size");
-	m_csm_restriction_flags = g_settings->getU64("csm_restriction_flags");
-	m_csm_restriction_noderange = g_settings->getU32("csm_restriction_noderange");
```
SmallJoker, line 449: I'd prefer to keep the settings getters nicely grouped here.

src/server.cpp
```diff
@@ -2712,19 +2636,17 @@ void Server::sendRequestedMedia(session_t peer_id,
 	}
 }
 
-void Server::sendDetachedInventory(const std::string &name, session_t peer_id)
+void Server::sendDetachedInventory(Inventory* inventory, const std::string &name, session_t peer_id)
```
SmallJoker, line 2639: Misplaced pointer, in header too.

src/server.cpp
```diff
@@ -2712,19 +2636,17 @@ void Server::sendRequestedMedia(session_t peer_id,
...
 {
-	const auto &inv_it = m_detached_inventories.find(name);
-	const auto &player_it = m_detached_inventories_player.find(name);
+	std::string attachedPlayerName;
```
SmallJoker, line 2641: snake_case_for_variable_names. Example: `detached_name`, `attached_inv_player`

src/server.cpp
```diff
@@ -2712,19 +2636,17 @@ void Server::sendRequestedMedia(session_t peer_id,
...
+	m_inventory_mgr->getAttachedPlayer(name, attachedPlayerName);
```
SmallJoker, line 2642: I think it would be saner to use the inventory name as key, and call a function within `m_inventory_mgr` to retrieve `Inventory *` and the player name both at once. This practice is confusing.

src/server.cpp
```diff
@@ -2760,15 +2682,10 @@ void Server::sendDetachedInventory(const std::string &name, session_t peer_id)
...
-		const std::string &name = detached_inventory.first;
-		if (incremental) {
-			Inventory *inv = detached_inventory.second;
-			if (!inv || !inv->checkModified())
-				continue;
-		}
-
-		sendDetachedInventory(name, peer_id);
+	std::unordered_map<std::string, Inventory *> inventories;
+	m_inventory_mgr->getDetachedInventories(inventories, incremental);
```
SmallJoker, line 2686: Rather than copying the map, I'd prefer to use a lambda expression as callback here.

src/server.cpp
nerzhul: the problem i wanted to avoid is the double lookup on Inventory objects. I may try to refactor a little bit more to find a better solution

src/server.cpp
nerzhul: fine, i will do that tomorrow

src/server.cpp
nerzhul: okay, as you like

nerzhul: i understand, but they are inited outside of the constructor. I will add them grouped in the constructor

src/server.cpp
nerzhul: when you have many detached inventories, and a player connects, you will lookup many times on the inventories to verify if they are attached. Maybe we should refactor detached inventories to have a ownership directly on it, then ?

nerzhul: okay i found a way to solve it smoothly and drop the old lookup hack :)

nerzhul:
@SmallJoker all your comments are addressed and i found a way to remove the ugly lookup code in the sending packet function :)

nerzhul:
@SmallJoker is this fine ? anything more to do, it's mergeable now

src/server.cpp
```diff
@@ -2760,16 +2665,35 @@ void Server::sendDetachedInventory(const std::string &name, session_t peer_id)
...
+				return;
 		}
 
-		sendDetachedInventory(name, peer_id);
-	}
+		// if we are pushing inventories to a specific player
+		// we should filter to send only the right inventories
+		if (!peer_name.empty()) {
+			std::string attached_player;
+			imgr->getAttachedPlayer(name, attached_player);
```
SmallJoker, line 2688: Following would be more common and easier to understand:
`std::string attached_player = imgr->getAttachedPlayer(name);`

src/server/serverinventorymgr.cpp
```diff
@@ -0,0 +1,181 @@
...
+	}
+	return true;
+}
+
+void ServerInventoryMgr::applyOnInventories(std::function<void(ServerInventoryMgr *imgr,
+				const std::string &, Inventory *)>
+				apply_cb)
+{
+	for (const auto &detached_inventory : m_detached_inventories) {
+		apply_cb(this, detached_inventory.first, detached_inventory.second);
```
SmallJoker, line 169: I thought you wanted to include the `incremental` checks there, hence the suggestion for a callback.
Now it's no better than a simple map getter function like `const auto &inventories = m_inventory_mgr.getInventories();`.
If you would like to use callbacks anway (despite being kinda useless right now), it might not be needed to pass `this`. That parameter is already known by the caller (i.e. is superfluous).

src/server/CMakeLists.txt
```diff
@@ -4,5 +4,6 @@ set(server_SRCS
 	${CMAKE_CURRENT_SOURCE_DIR}/mods.cpp
 	${CMAKE_CURRENT_SOURCE_DIR}/player_sao.cpp
 	${CMAKE_CURRENT_SOURCE_DIR}/serveractiveobject.cpp
+	${CMAKE_CURRENT_SOURCE_DIR}/serverinventorymgr.cpp
```
SmallJoker, line 7: What about Android?

nerzhul: android is not using those cmakelists ? i should update :D

src/server/serverinventorymgr.cpp
nerzhul: at the beginning i wanted to performing this sending from the manager, then i found a more generic way to apply things without taking the ownership of the map while searching for a good function name. I can restore the incremental behaviour if you prefer, yes

src/server.cpp
nerzhul: obviously yes

src/server/CMakeLists.txt
nerzhul: `$(wildcard ../../../src/server/*.cpp)           \`

The new android build is <3 :)

nerzhul:
@SmallJoker you should be happy now :)

src/server/serverinventorymgr.h
```diff
@@ -0,0 +1,54 @@
...
+
+	void sendDetachedInventories(const std::string &peer_name, bool incremental,
+			std::function<void(const std::string &, Inventory *)> apply_cb);
+	const std::string &getAttachedPlayer(const std::string &name) const;
+
+private:
+	ServerEnvironment *m_env = nullptr;
+
+	// key = name
+	std::map<std::string, Inventory *> m_detached_inventories;
```
sfan5, line 51: This should rather be a `std::unordered_map<std::string, DetachedInventory>`
with 
```cpp
struct DetachedInventory {
  Inventory *inv;
  std::string players; // "" means visible to all players
};
```

This avoids doing a second lookup in `sendDetachedInventories` and also allows `getAttachedPlayer` to be removed.

src/server/serverinventorymgr.h
```diff
@@ -0,0 +1,54 @@
...
+
+class ServerEnvironment;
+
+class ServerInventoryMgr : public InventoryManager
+{
+public:
+	ServerInventoryMgr(ServerEnvironment *env);
+	virtual ~ServerInventoryMgr();
+	/*
+		Shall be called with the environment and the connection locked.
```
sfan5, line 33: What is "the connection" here?

src/server/serverinventorymgr.h
```diff
@@ -0,0 +1,54 @@
...
+*/
+
+#pragma once
+
+#include "inventorymanager.h"
+#include <functional>
+
+class ServerEnvironment;
+
+class ServerInventoryMgr : public InventoryManager
```
sfan5, line 27: should rather be `ServerInventoryManager`
the class name is used only a few times, makes no sense to abbreviate it

src/server/serverinventorymgr.cpp
```diff
@@ -0,0 +1,198 @@
...
+bool ServerInventoryMgr::removeDetachedInventory(const std::string &name)
+{
+	const auto &inv_it = m_detached_inventories.find(name);
+	if (inv_it == m_detached_inventories.end())
+		return false;
+
+	delete inv_it->second;
+	m_detached_inventories.erase(inv_it);
+
+	if (!m_env) // Mods are not done loading
```
sfan5, line 144: Since `m_env` is set in the class constructor, this check never does anything.

src/server.cpp
```diff
@@ -2760,16 +2665,20 @@ void Server::sendDetachedInventory(const std::string &name, session_t peer_id)
...
-		const std::string &name = detached_inventory.first;
-		if (incremental) {
-			Inventory *inv = detached_inventory.second;
-			if (!inv || !inv->checkModified())
-				continue;
+	// Lookup player name, to filter detached inventories just after
+	std::string peer_name;
+	if (peer_id != PEER_ID_INEXISTENT) {
+		RemotePlayer *player = m_env->getPlayer(peer_id);
+		if (player) {
```
sfan5, line 2672: should return if player does not exist

src/server.cpp
```diff
@@ -388,6 +389,12 @@ void Server::init()
 
 	m_script = new ServerScripting(this);
 
+	// Initialize Environment
+	m_env = new ServerEnvironment(servermap, m_script, this, m_path_world);
```
sfan5, line 393: This is a pretty significant change, I don't think the environment should be initialized before mods are loaded.
If you need `m_inventory_mgr` to be created at this point, it should rather have a method to set the `m_env` when its actually created.
( see also my comment about the `if (!m_env)`)

src/main.cpp
```diff
@@ -887,7 +887,6 @@ static bool run_dedicated_server(const GameParams &game_params, const Settings &
 			// Create server
 			Server server(game_params.world_path, game_params.game_spec,
 					false, bind_addr, true, &iface);
-			server.init();
 
 			g_term_console.setup(&iface, &kill, admin_nick);
```
sfan5, line 891: Are you sure g_term_console does not rely on `init()` being called before `setup()`?

src/server/serverinventorymgr.h
nerzhul: i don't know, it's just a code move, it's right that there is no interest to keep it there

src/server/serverinventorymgr.h
nerzhul: i will take a look to see how we can optimize that, but your comment sounds to increase more again the quality on this code part (and it's why spliting big objects is good !!)

nerzhul: i don't remember why this map passed outside of my radar on the c++11 move, anyway let's go

src/server/serverinventorymgr.cpp
nerzhul: yep exact, it was just a code move, maybe a historical ugly hack :)

src/server.cpp
nerzhul: yep exact, i'll add a warning too in case ot

src/main.cpp
nerzhul: i will recheck, but i don't see any link between server and console setup

src/server.cpp
nerzhul: environment is just a constructor and doesn't run or register anything, it just open the world.mt configuration currently. Also mods rely on environment, they must load after, i can put a setter but it's more error prone in terms of init that anythings, adding more spaghettis in our init code.

src/main.cpp
nerzhul: the console uses only a run queue waiting events from other components, it's pretty isolated, no problem here

src/server.cpp
sfan5: > Also mods rely on environment, they must load after

No they do not, mods have always been loaded before the environment was created.

The initialization order should not just be changed because it's convenient. Also, if this is done warnings like this will not work anymore:
https://github.com/minetest/minetest/blob/cb159f8d8af4556391db8a6875657625733eeb11/src/script/lua_api/l_env.cpp#L643-L648

src/server.cpp
```diff
@@ -2760,16 +2665,23 @@ void Server::sendDetachedInventory(const std::string &name, session_t peer_id)
...
-	for (const auto &detached_inventory : m_detached_inventories) {
-		const std::string &name = detached_inventory.first;
-		if (incremental) {
-			Inventory *inv = detached_inventory.second;
-			if (!inv || !inv->checkModified())
-				continue;
+	// Lookup player name, to filter detached inventories just after
+	std::string peer_name;
+	if (peer_id != PEER_ID_INEXISTENT) {
+		if (RemotePlayer *player = m_env->getPlayer(peer_id))
```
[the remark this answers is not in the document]
nerzhul: it's just proper lifetime management. Fine  to add braces, if i remembered we don't need braces for a single line, no ?

src/server.cpp
```diff
@@ -442,15 +446,12 @@ void Server::init()
 	servermap->addEventReceiver(this);
 
 	m_env->loadMeta();
-
-	m_liquid_transform_every = g_settings->getFloat("liquid_update");
-	m_max_chatmessage_length = g_settings->getU16("chat_message_max_size");
```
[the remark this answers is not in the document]
nerzhul: i see, i really don't like our settings interface where everything is in spaghetti, i will rollback and add a comment for the future, as it appears totally unclear here, it's why i moved them

src/server/serverinventorymgr.cpp
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	Inventory *inv = new Inventory(idef);
+	sanity_check(inv);
+	m_detached_inventories[name].inventory = inv;
+	if (!player.empty()) {
+		m_detached_inventories[name].owner = player;
+
+		// if player is connected, send him the inventory
+		if (RemotePlayer *p = m_env->getPlayer(name.c_str()))
```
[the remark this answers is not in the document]
nerzhul: exact, it's original code issue, how many issues we found with this refactor on the original code :)

src/server.cpp
nerzhul: understood. I restored the old behaviour on detached inventory creation.

nerzhul:
@sfan5 your turn, hope this time we can see this nice improvement merged

src/server.h
```diff
@@ -253,10 +247,8 @@ class Server : public con::PeerHandler, public MapEventReceiver,
 
 	void deleteParticleSpawner(const std::string &playername, u32 id);
 
-	// Creates or resets inventory
-	Inventory *createDetachedInventory(const std::string &name,
-			const std::string &player = "");
-	bool removeDetachedInventory(const std::string &name);
+	ServerInventoryManager *getInventoryMgr() { return m_inventory_mgr.get(); }
```
sfan5, line 250: `ServerInventoryManager *getInventoryMgr() const { ... }` ?

src/server/serverinventorymgr.cpp
```diff
@@ -0,0 +1,191 @@
...
+	case InventoryLocation::NODEMETA: {
+		NodeMetadata *meta = m_env->getMap().getNodeMetadata(loc.p);
+		if (!meta)
+			return NULL;
+		return meta->getInventory();
+	} break;
+	case InventoryLocation::DETACHED: {
+		if (m_detached_inventories.count(loc.name) == 0)
+			return NULL;
+		return m_detached_inventories[loc.name].inventory;
```
sfan5, line 64:
```suggestion
		auto it = m_detached_inventories.find(loc.name);
		if (it == m_detached_inventories.end())
			return nullptr;
		return it.second.inventory;
```

src/server.h
nerzhul: not sure we can as sometimes we create detached inventories, will verify that

src/server/serverinventorymgr.cpp
nerzhul: oh yeah obviously a good change, the original code was crazy and didn't reviewed it :)

Issue #9771, opened 2020-04-29
Translations broke


Wuzzy2:
##### Minetest version
```
a36c9c3e930608386b983ea76158793fe9d622f6
```

Translations broke again. Minetest refuses to compile with Gettext.

Reverting 9364cc7e5f51b60f1bc72477cafa8a6dabf0d4d8 fixes this.

Wuzzy2:
Oops. I meant to say:

“Reverting a368e7e79303fa02ef97580a2c387dcee57bc3c4 fixes this.”

juozaspo:
Minetes with gettext enabled compiles fine on my system (Ubuntu 20.04) but it does not generate locale files so translations does not work at all. No error in building whatsoever. 
Compilation log after cleaning git source tree with command `git clean -d -x -i`, keeping only mt configuration files such minetest.conf, worlds, etc:
```
juozas@xubuntu:~/minetest_dev$ cmake . -DRUN_IN_PLACE=TRUE -DCMAKE_BUILD_TYPE=Debug -DENABLE_GETTEXT=TRUE
-- The C compiler identification is GNU 9.3.0
-- The CXX compiler identification is GNU 9.3.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- *** Will build version 5.3.0-dev-debug ***
-- Found Irrlicht: /usr/lib/x86_64-linux-gnu/libIrrlicht.so  
-- Using GMP provided by system.
-- Found GMP: /usr/lib/x86_64-linux-gnu/libgmp.so  
-- Using bundled JSONCPP library.
-- Found LuaJIT: /usr/lib/x86_64-linux-gnu/libluajit-5.1.so (found version "") 
-- Using LuaJIT provided by system.
-- Found CURL: /usr/lib/x86_64-linux-gnu/libcurl.so  
-- cURL support enabled.
-- Found GettextLib: /usr/include  
-- Found OpenAL: /usr/lib/x86_64-linux-gnu/libopenal.so  
[...]

PR #9772, opened 2020-04-29
Fix gettext detection and locale building (broken since a368e7e)

juozaspo:
This PR fixes cmake gettext detection and locale generation broken since a368e7e, also might solve #9771

- Fix enabling gettext broken since a368e7e by replacing wrong variable (GETTEXT_FOUND) with correct (GETTEXTLIB_FOUND) in `src/CMakeLists.txt`, to make sure variable name matches the new variable name provided in `cmake/Modules/FindGettextLib.cmake`
- Result fix is correct setting of USE_GETTEXT variable and correct locale file generation and building with gettext.

## To do

## How to test

- (optional) Clean git source tree by running command `git clean -d -x -i` (You can exclude files by filtering listed files by pattern as prompted while using this command)
- Rerun cmake and observe its' output (should include line beginning with `GetText enabled; locales found`), and after running make minetest should be translated like it was before mentioned commit was merged.


sfan5:
Whoops, looks like I missed that when testing.

orbea:
I missed it too, sorry.

PR #9776, opened 2020-04-29
Show relative screenshot filename on chat when applicable

juozaspo:
This PR allows printing the relative filename of screenshot to client instead of absolute when screenshot_path is relative. When screenshot path is absolute it prints absolute path like before.

- Fix the issue where absolute filename with dots (e.g. `/path/to/minetest/bin/../screenshots/filename.ext`) is displayed in chat when screenshot path is relative (e.g. `screenshots`)
- The files listed below have all the necessary changes and modifications implementing such functions are listed near each file:
  - `src/client/client.cpp`, `Client::makeScreenshot()`:
    - Upon saving screenshot relative path is generated using following method
      - Initially, output text is set to filename, with relative components removed by using function `fs::AbsolutePath`
      - Screenshot dir then is checked if it is root dir, value is compared to output of `fs::GetRootDir`, If it is not the same, the operations listed below are performed to get the relative file name
        - Common branching path of minetest share path and absolute screenshot path is generated using new function `fs::GetCommonPath`, dfference of levels is stored in the variable pointing to third argument of that function
        - If screenshot path is not absolute and level difference is lover than set threshold, the relative path generated using a new function `fs::BuildRelativePath` is used inplace, otherwise no changes are done.
      - If screenshot dir is root, no changes are done to output filename
  - `src/filesys.h` & `src/filesys.cpp`
    - `RemoveRelativePathComponents`
      - Upon return, the resulting path checked if it is not same as root dir, trailing character is not removed if path is the root dir.
    - `AbsolutePath`
      - unix logic is changed:
        - If path is absolute, `RemoveRelativePathComponents` is used to return clean path
        - Otherwise current working dir is get using `getCurrentWorkingDir`, if it fails the function returns same value as above, otherwise it returns working dir + path
      - Suggestions welcome, this function might not be currently finalized atm.
    - New function `SplitComponentFromPath`, splits last component from path into variable supplied to second argument, also updating variable supplied to first argument.
      - Returns true if path contains delimiter, component is top file/directory in the path.
      - Otherwise returns false, component is same as the path.
    - New function `GetRootDir`, returns root directory, on windows returning notation like `C:\`, etc.
    - New function `GetCommonPath`, compares two paths, returns a path to the branching point of first and second argument, third argument contains returned level difference between them. 
  - `src/porting.h`
    - added missing definition of `porting::getCurrentWorkingDir`, used by some functions in `filesys.cpp`
- Those changes make sure path is shown w/o upper level markers and single dots anywhere in the middle of the path except for upper level markers on beginning of the relative path where needed. Also it makes sure no extra slash is printed when path is empty too or is show as root (`/`)
- Any suggestions or bug reports welcome, most features work good so far.

## To do

This PR is a Work in Progress ~Ready for Review~.

  - [x] Make sure absolute screenshot dir is derived properly
  - [x] If absolute value of screenshot dir is on drive or system root, displayed (relative) path should change to absolute
  - [x] When screenshot dir is not on system root
    - [x] Absolute path for minetest directory should be provided by `fs::AbsolutePath`
    - [x] Common path & level count from minetest share dir to it should be provided by new function `fs::GetCommonPath`, level difference should be stored in the variable supplied to the third argument of this function
   - [x] If screenshot path config variable were absolute or level difference were too high screenshot dir should be set to be absolute 
      - [x] Otherwise relative path should be built using function `fs::BuildRelativePath`
  - [x] Basename should have no information prepended
  - [x] Relative and absolute filenames should consist of basename and path to screenshots, joined using new function `fs::JoinPaths`
  - [x] Absolute path should be used to create screenshot dir and for screenshot file saving
  - [ ] Make sure the new functions are efficient, no bugs, etc.
  - [ ] Make sure everything else works, might need some testing (messages in `infostream` would show if anything would be broken)
  - [ ] Remove debug messages when done

## How to test

- Build minetest from master, take screenshot, notice both relative and absolute screenshot paths print as absolute, parent directory markers (`../`) and dots (`./`) not removed or substituted with alternative path
- Build minetest from PR, do tests by setting screenshot path as shown below, files should save in paths relative to share or minetest dir, depending on build
  - default screenshot path - displayed file name should have a default sub-directory preceding it, no markers shown before it
  - path with multiple ascending and descending levels should resolve to shorter and direct path, e.g `../something/../minetest`  should resolve to `minetest` and `../something/..` should resolve to `..`, w/o delimiter after it
  - empty screenshot path - only file name should be displayed, no markers before it
  - path pointing to parent directory - displayed file name should include correct count of parent level markers depending on parent level depth
  - path descending with more parent levels then set limit (currently 3) - displayed file name should always be absolute
  - any absolute path - displayed file name should be absolute, any kind of unneeded markers and extra delimiters should be removed


ExeVirus:
For the record, the core devs are aware of your tireless maintaining of this PR, they're just overwhelmed. I might take a look at this soon, but I'm just a relatively new mod dev with a lot of C++ background. Thank you!

SmallJoker:
FYI: You don't need to rebase this branch all the time. It's only required if there are conflicts stated in the build check section below.

This is another nice-to-have feature PR, and not on my priority list. However, I'll have a look at it sometime.

juozaspo:
@SmallJoker I regularly rebase this branch each time I update my local copy of my minetest fork saved on my pc. I could reduce frequency of updates if necessary but as how I like to keep my forks up to date most of the times I can't guarantee that.

src/client/client.cpp
```diff
@@ -1792,18 +1792,78 @@ void Client::makeScreenshot()
...
+			screenshot_dir = "";
+			if (level_diff != 0)
+				for (int i = 0; i < level_diff; i++)
+					screenshot_dir += ".." + ((i < level_diff-1) ? std::string(DIR_DELIM) : "");
+			std::size_t pos=common_path.length()
+				+ ((screenshot_dir_abs.length() > common_path.length())?1:0);
+			screenshot_dir += screenshot_dir_abs.substr(pos,screenshot_dir_abs.length());
+		}
+	} else
+		screenshot_dir = screenshot_dir_abs;
```
v-rob, line 1855:
```C++
} else {
	screenshot_dir = screenshot_dir_abs;
}
```

src/client/client.cpp
```diff
@@ -1792,18 +1792,78 @@ void Client::makeScreenshot()
...
+		std::size_t mtpath_pos, screenpath_pos;
+		std::string mtpath_str = fs::AbsolutePath(porting::path_user), screenpath_str = screenshot_dir_abs;
+		std::string mtpath_name, screenpath_name, common_path;
+
+		do {
+			if (!mtpath_done) mtpath_levels++;
+			mtpath_pos = mtpath_str.find(std::string(DIR_DELIM));
+			if (mtpath_pos != std::string::npos) {
+				mtpath_name = mtpath_str.substr(0, mtpath_pos);
+				mtpath_str.erase(0, mtpath_pos + std::string(DIR_DELIM).length());
```
v-rob, line 1820: DIR_DELIM is always 1 character, so no need to use `length()`

src/client/client.cpp
```diff
@@ -1792,18 +1792,78 @@ void Client::makeScreenshot()
...
+				screenpath_name = screenpath_str;
+				screenpath_done = true;
+			}
+			if (!levels_diverged && mtpath_name == screenpath_name) {
+				common_path += mtpath_name
+					+ ((!mtpath_done && !screenpath_done) ? std::string(DIR_DELIM) : "");
+				common_levels++;
+			}
+			else if (!levels_diverged) levels_diverged = true;
+		} while ((mtpath_pos != std::string::npos) || (screenpath_pos != std::string::npos));
```
v-rob, line 1839: The parentheses around the `!=` are redundant.

src/client/client.cpp
```diff
@@ -1792,18 +1792,78 @@ void Client::makeScreenshot()
...
+		int level_diff = mtpath_levels - common_levels;
+
+		if (fs::IsPathAbsolute(g_settings->get("screenshot_path")) || level_diff >= 4) {
+			screenshot_dir = screenshot_dir_abs;
+		} else {
+			screenshot_dir = "";
+			if (level_diff != 0)
+				for (int i = 0; i < level_diff; i++)
+					screenshot_dir += ".." + ((i < level_diff-1) ? std::string(DIR_DELIM) : "");
+			std::size_t pos=common_path.length()
```
v-rob, line 1850: Spaces around `=`

src/client/client.cpp
```diff
@@ -1792,18 +1792,78 @@ void Client::makeScreenshot()
...
+				+ ((screenshot_dir_abs.length() > common_path.length())?1:0);
+			screenshot_dir += screenshot_dir_abs.substr(pos,screenshot_dir_abs.length());
```
v-rob, line 1852: Space after `,`

v-rob (commented):
I did a quick run-through on code style, but nothing else since my testing shows that it doesn't seem to work with relative paths for `screenshot_path`.  I get errors like

image

I'm testing on Windows 10, so it might be that it only works on Linux (I assume that's what you're using).

juozaspo:
> I did a quick run-through on code style, but nothing else since my testing shows that it doesn't seem to work with relative paths for `screenshot_path`. I get errors like
> 
> image
> 
> I'm testing on Windows 10, so it might be that it only works on Linux (I assume that's what you're using).

In your error I see both unix and windows delimiters in the path shown on your error, unix delimiters should be changed to windows's style (`\`) and preceding colon symbol should be removed. What is the contents of screenshot_path in your conf file, it should not have unix delimiters in it. I'll update the code and see what I can do.

v-rob:
My setting is just set to "screenshots", which is the default setting.  I don't know how those delimiters are getting in either since I really don't see them in the code.  Perhaps it has something to do with the recent update to Irrlicht 1.9?

juozaspo:
I'm working on it. Will update later.

juozaspo:
The program tries to write on absolute path, filename_abs value needs to be checked, also the path might not be writable for some reason as in Win10 restrictions.

juozaspo:
Added some debug messages. Try now with --info from command line, the output right after triggering screenshot key should be helpful.

sfan5:
More importantly this should be put into a generic function in `filesys.cpp` instead of residing in the client code.

juozaspo:
Currently I'm unable to build this pr on wine as cmake fails to detect correct toolchain and compiler, as described in #11236.

src/client/client.cpp
```diff
@@ -1792,18 +1792,78 @@ void Client::makeScreenshot()
...
+	// path values such as the root or path with non-existing folders
+	io::IFileSystem *irrfs = RenderingEngine::get_filesystem();
+	io::path path = irrfs->getAbsolutePath(screenshot_dir.c_str());
+	std::string screenshot_dir_abs = std::string(path.c_str());
+
+	std::size_t pos = screenshot_dir_abs.rfind(std::string(DIR_DELIM));
+	if (screenshot_dir_abs.length() > 1 && pos == screenshot_dir_abs.length()-1)
+		screenshot_dir_abs = screenshot_dir_abs.substr(0, pos);
+
+	if (screenshot_dir_abs != "/") {
```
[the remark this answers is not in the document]
juozaspo: Code changed. Now it checks if path is directory root instead of just delimiter, also on windows, wrong delimiter returned by irrlicht is replaced with correct so it might work.

src/client/client.cpp
```diff
@@ -1792,18 +1792,78 @@ void Client::makeScreenshot()
...
+		bool mtpath_done = false, screenpath_done = false, levels_diverged=false;
+		int mtpath_levels = 0, common_levels = 0;
+		std::size_t mtpath_pos, screenpath_pos;
+		std::string mtpath_str = fs::AbsolutePath(porting::path_user), screenpath_str = screenshot_dir_abs;
+		std::string mtpath_name, screenpath_name, common_path;
+
+		do {
+			if (!mtpath_done) mtpath_levels++;
```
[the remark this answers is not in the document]
juozaspo: done.

src/client/client.cpp
juozaspo: done.

src/client/client.cpp
```diff
@@ -1792,18 +1792,78 @@ void Client::makeScreenshot()
...
+			} else if (!screenpath_done) {
+				screenpath_name = screenpath_str;
+				screenpath_done = true;
+			}
+			if (!levels_diverged && mtpath_name == screenpath_name) {
+				common_path += mtpath_name
+					+ ((!mtpath_done && !screenpath_done) ? std::string(DIR_DELIM) : "");
+				common_levels++;
+			}
+			else if (!levels_diverged) levels_diverged = true;
```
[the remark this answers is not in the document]
juozaspo: done. also after if expression there should be a matching bracket if there's one shown below last comand

src/client/client.cpp
juozaspo: done.

src/client/client.cpp
juozaspo: done.

src/client/client.cpp
juozaspo: fixed.

src/client/client.cpp
juozaspo: done.

v-rob:
Is this ready for testing again, or is it still a draft?

juozaspo:
I want to test for Windows as I don't have a way to do this myself. Also the debug output right after screenshot trigger should be useful (run minetest --info parameter or check debug.txt file), it would show how stuff is parsed and confirm if everything works right or not. If everything works, all that's needed to do is remove debug messages and convert the PR to ready state.

juozaspo:
I've tried to build both win32 and win64 using buildbot scripts in util subdirectory, 32 bit does not produce working build, 64 bit build fails, while on github action it works. Probably incompatible mingw version, I don't know.

juozaspo:
Finaly were able to cross-compile win64 build. Needed to copy win32 posix toolchain, change some files, modify 64 bit script to use that toolchain, copy needed posix dll files, game works. To get info in debug.txt `debug_log_level = info` needs to be added to minetest.conf, saving screenshots works on my end on wine.

Some debug info upon saving a screenshot:
```
2021-06-17 18:54:17: INFO[Main]: Client::makeScreenshot(): screenshot_dir_abs: Z:/home/juozas/Darbastalis/minetest-5.5.0-c9447d20b-win64/screenshots
2021-06-17 18:54:17: INFO[Main]: Client::makeScreenshot(): patched screenshot_dir_abs: Z:\home\juozas\Darbastalis\minetest-5.5.0-c9447d20b-win64\screenshots
2021-06-17 18:54:17: INFO[Main]: Client::makeScreenshot(): dir_root: Z:\
2021-06-17 18:54:17: INFO[Main]: Client::makeScreenshot(): level_diff: 0
2021-06-17 18:54:17: INFO[Main]: Client::makeScreenshot(): common_path: Z:\home\juozas\Darbastalis\minetest-5.5.0-c9447d20b-win64
2021-06-17 18:54:17: INFO[Main]: Client::makeScreenshot(): screenshot_dir: screenshots
2021-06-17 18:54:17: INFO[Main]: Client::makeScreenshot(): filename_base: screenshot_20210617_185417
2021-06-17 18:54:17: INFO[Main]: Client::makeScreenshot(): basename: screenshot_20210617_185417.png
2021-06-17 18:54:17: INFO[Main]: Client::makeScreenshot(): filename: screenshots\screenshot_20210617_185417.png
2021-06-17 18:54:17: INFO[Main]: Client::makeScreenshot(): filename_abs: Z:\home\juozas\Darbastalis\minetest-5.5.0-c9447d20b-win64\screenshots\screenshot_20210617_185417.png
2021-06-17 18:54:18: INFO[Main]: Saved screenshot to 'screenshots\screenshot_20210617_185417.png'
```

juozaspo:
Win32 build works too. It failed before as I've copied wrong (non-posix) files from mingw. After copying correct it does not crash any more, saving screenshots works too.
Debug info build made from previous commit.
```
2021-06-17 19:21:01: INFO[Main]: Client::makeScreenshot(): screenshot_dir_abs: Z:/home/juozas/Darbastalis/minetest-5.5.0-c9447d20b-win32/screenshots
2021-06-17 19:21:01: INFO[Main]: Client::makeScreenshot(): patched screenshot_dir_abs: Z:\home\juozas\Darbastalis\minetest-5.5.0-c9447d20b-win32\screenshots
2021-06-17 19:21:01: INFO[Main]: Client::makeScreenshot(): dir_root: Z:\
2021-06-17 19:21:01: INFO[Main]: Client::makeScreenshot(): level_diff: 0
2021-06-17 19:21:01: INFO[Main]: Client::makeScreenshot(): common_path: Z:\home\juozas\Darbastalis\minetest-5.5.0-c9447d20b-win32
2021-06-17 19:21:01: INFO[Main]: Client::makeScreenshot(): screenshot_dir: screenshots
2021-06-17 19:21:01: INFO[Main]: Client::makeScreenshot(): filename_base: screenshot_20210617_192101
2021-06-17 19:21:01: INFO[Main]: Client::makeScreenshot(): basename: screenshot_20210617_192101.png
2021-06-17 19:21:01: INFO[Main]: Client::makeScreenshot(): filename: screenshots\screenshot_20210617_192101.png
2021-06-17 19:21:01: INFO[Main]: Client::makeScreenshot(): filename_abs: Z:\home\juozas\Darbastalis\minetest-5.5.0-c9447d20b-win32\screenshots\screenshot_20210617_192101.png
2021-06-17 19:21:01: INFO[Main]: Saved screenshot to 'screenshots\screenshot_20210617_192101.png'
```

If it doesn't work with latest commit, revert to one with debug then try again. Post debug info from console or debug.txt (add debug_log_level = info to minetest.conf first) in that case

src/client/client.cpp
```diff
@@ -1787,18 +1787,89 @@ void Client::makeScreenshot()
 	if (fs::IsPathAbsolute(g_settings->get("screenshot_path")))
 		screenshot_dir = g_settings->get("screenshot_path");
 	else
-		screenshot_dir = porting::path_user + DIR_DELIM + g_settings->get("screenshot_path");
+		screenshot_dir = porting::path_user + std::string(DIR_DELIM) + g_settings->get("screenshot_path");
+
+	// Using Irrlicht methods since those defined in filesys.cpp are
```
rubenwardy, line 1792: This is gross, and should probably be moved to filesys.cpp or at least another function

rubenwardy: I'm expecting at least two functions here: `getAbsolutePath(path)` and `joinPath(base, filename)`

juozaspo: No longer using irrlicht. Going to test already defined function in filesys.cpp and see if it works instead. If it does not i'll try to fix it.

juozaspo: Also the unix function realpath (used by a function in filesys.cpp) does not return correct path when source path included non-existing files and/or directories, I'm trying to make a workaround for it.

rubenwardy: Irrlicht has a compatible license, so you could look to see how it implemented it

juozaspo: Implemented a working workaround in filesys.cpp, also fixed a bug in RemoveRelativePathComponents where it would not work correctly with path that is drive root, absolute path should now be generated correctly.

rubenwardy:
Is this ready for review? It's still marked as a drag

juozaspo:
I'm a quite busy right now, I'll update this later. This needs a bit of rewrite and some essential functions has to be written first

rubenwardy:
I've implemented a `fs::RelativePath` function here: https://github.com/minetest/minetest/commit/3c532f1fa746b2fab34573e18101180cf4fbe466

juozaspo:
The above commit is nowhere in the master. To successfully use it this needs to be added to the master first. The master I rebased onto so far doesn't have such commit. You might need to post a PR and include a way to test it before merging.

rubenwardy:
It's on the relative_path branch of my fork of the Minetest repo. I'm not sure why it linked to the upstream repo, GitHub is weird

rubenwardy:
Note: you don't need to use that commit. I posted it here as you might find it useful

juozaspo:
Did a major rewrite and squashed useless commits. Should work properly now. Debug messages not removed yet.

sfan5 (changes requested):

sfan5:
Generally the implementation here could be simpler. You also don't need more than one new function.
Consider this code:
```cpp
std::string MakeRelative(std::string base, std::string path, int allow_parent)
{
	auto StartsWith = [] (const std::string &p, const std::string &p2) -> bool {
		if (!strncmp(p.c_str(), p2.c_str(), p2.size()))
			return p.size() == p2.size() || IsDirDelimiter(p[p2.size()]);
		return false;
	};
	base = RemoveRelativePathComponents(base);
	path = RemoveRelativePathComponents(path);

	// Match `base` to the beginning of `path`, removing path components until it does
	size_t pos = std::string::npos;
	size_t dotdot_count = 0;
	while (dotdot_count <= allow_parent) {
		if (StartsWith(path, base)) {
			pos = base.size();
			break;
		}
		if (base.empty())
			break;
		base = RemoveLastPathComponent(base);
		dotdot_count++;
	}
	if (pos == std::string::npos)
		return path; // Couldn't make relative
	// Skip extra delimiters
	while (pos != path.size() && IsDirDelimiter(path[pos]))
		pos++;

	std::string ret;
	for (size_t i = 0; i < dotdot_count; i++)
		ret.append(".." DIR_DELIM);
	ret += path.substr(pos);
	return ret;
}
```
which behaves as follows:
```
MakeRelative("/home/user/minetest", "/home/user/minetest/screenshots/1.png", 0) = "screenshots/1.png"
MakeRelative("/home/user/minetest", "/home/user/minetest/1.png", 0) = "1.png"
MakeRelative("/home/user/minetest", "/home/user/minetest", 0) = ""
MakeRelative("/home/user/minetest", "/home/user/minetest/", 0) = ""
MakeRelative("/home/user/minetest", "/home/user/1.png", 0) = "/home/user/1.png"
MakeRelative("/home/user/minetest", "/home/user/1.png", 1) = "../1.png"
MakeRelative("/home/user/minetest", "/home/1.png", 1) = "/home/1.png"
MakeRelative("/home/user/minetest", "/home/1.png", 2) = "../../1.png"
```

src/filesys.cpp
```diff
@@ -715,20 +715,107 @@ std::string RemoveRelativePathComponents(std::string path)
 	pos = path.size();
 	while (pos != 0 && IsDirDelimiter(path[pos-1]))
 		pos--;
-	return path.substr(0, pos);
+	if (path != GetRootDir(path))
+		return path.substr(0, pos);
+	else
+		return path;
```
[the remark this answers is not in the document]
juozaspo: If the path is root dir (/) w/o the change it would become empty and not working as expected as substr would remove the one and only directory separator.

src/filesys.cpp
```diff
@@ -715,20 +715,107 @@ std::string RemoveRelativePathComponents(std::string path)
...
+		char buf[BUFSIZ];
+		if(!porting::getCurrentWorkingDir(buf, sizeof(buf))) {
+			errorstream << "fs::AbsolutePath(): Unable to get the current working dir"
+				<< std::endl;
+			return RemoveRelativePathComponents(path);
+		}
+		return RemoveRelativePathComponents(std::string(buf)
+				+ std::string(DIR_DELIM) + path);
+	}
+#endif
```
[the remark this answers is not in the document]
juozaspo: realpath does not work with non existing directories, it returns NULL and errno is set on error. Consider this example and it's output:
```
#include <iostream>
#include <cstdlib>
#include <cstring>
#include <cerrno>
using namespace std;
void func (const char *path);
int main(int argc, char* argv[]) {
	func("/");
	func("/home/juozas/minetest");
	func("/home/juozas/minetest_dev");
	func("/home/juozas/minetest/screenshots");
	func("/home/juozas/minetest/../Paveikslėliai");
	func("/home");
	func("/home/juozas/minetest");
	func("/foo/bar/baz/../bzz");
	return 0;
}
void func (const char *path) {
	char* retval = realpath(path,NULL);
	if (retval == NULL) { 
		retval = new char[sizeof strerror(errno)];
		strcpy(retval,strerror(errno));
	}
	cout << "realpath(\"" << path << "\") = \"" << retval << "\";" << endl;
}
```

```
realpath("/") = "/";
realpath("/home/juozas/minetest") = "/home/juozas/minetest";
realpath("/home/juozas/minetest_dev") = "/home/juozas/minetest_dev";
realpath("/home/juozas/minetest/screenshots") = "/home/juozas/minetest/screenshots";
realpath("/home/juozas/minetest/../Paveikslėliai") = "/home/juozas/Paveikslėliai";
realpath("/home") = "/home";
realpath("/home/juozas/minetest") = "/home/juozas/minetest";
realpath("/foo/bar/baz/../bzz") = "No such file or directory";
```

juozaspo:
Rebuilt algorithm, relative filename is now genarated from absolute filename, will remove debug after some testing.

src/client/client.cpp
```diff
@@ -1823,18 +1823,41 @@ void Client::makeScreenshot()
...
 
-	std::string filename_base = screenshot_dir
-			+ DIR_DELIM
-			+ std::string("screenshot_")
-			+ std::string(timetstamp_c);
+	std::string filename_base = std::string("screenshot_")
+		+ std::string(timetstamp_c);
 	std::string filename_ext = "." + g_settings->get("screenshot_format");
-	std::string filename;
+	std::string basename, filename, filename_abs;
```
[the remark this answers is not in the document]
juozaspo: no longer the issue, I've changed the location where the relative filename is generated as you suggested, it is right before file name is displayed on chat

SmallJoker:
https://irc.minetest.net/minetest-dev/2022-09-25#i_6018920

My two cents:

Users might have Minetest installed system-wide or run them from the Downloads directory without paying much attention due to to desktop or start menu links. I find the full path more helpful. It does exactly tell where the screenshot is, without leaving any doubts. The screenshot message is printed anyway and the path is usually not too long to be of a concern.

juozaspo:
I do not wish to contribute to this any more since this pr was closed w/i prior warning or explaining why, and I'd like to have this deleted and related commits deleted as long as old behavior is preferred.

PR #9777, opened 2020-04-29
fix models turn black when inside solid node

TheTermos:
This is a quick and easy fix for #2467

How it works:
Cao's light isn't updated anymore if calculated position is inside a node with `light_propagates ` being false.

There still may be some weird lightning e.g. if player is noclipping through cave floor in a specific way or objects are left inside nodes for a long time, but come on, this is basically a two liner, why not merge and worry about complex solutions later.

## To do

This PR is Ready for Review.

## How to test

Clip player and objects into nodes and see what happens.

### DIFF
--- a/src/client/clientenvironment.cpp
+++ b/src/client/clientenvironment.cpp
@@ -330,7 +330,7 @@ void ClientEnvironment::step(float dtime)
 			MapNode n = this->m_map->getNode(p, &pos_ok);
 			if (pos_ok) {
 				const NodeDefManager *nodemgr = m_client->ndef();
-				//don't update if clipping with a solid node.
+				// Don't update if clipping with a solid node so object doesn't turn black
 				if (nodemgr->get(n).light_propagates) {
 					light = n.getLightBlend(day_night_ratio, nodemgr);
 					cao->updateLight(light);
--- a/src/client/clientenvironment.cpp
+++ b/src/client/clientenvironment.cpp
@@ -328,12 +328,18 @@ void ClientEnvironment::step(float dtime)
 			// Get node at head
 			v3s16 p = cao->getLightPosition();
 			MapNode n = this->m_map->getNode(p, &pos_ok);
-			if (pos_ok)
-				light = n.getLightBlend(day_night_ratio, m_client->ndef());
-			else
+			if (pos_ok) {
+				const NodeDefManager *nodemgr = m_client->ndef();
+				//don't update if clipping with a solid node.
+				if (nodemgr->get(n).light_propagates) {
+					light = n.getLightBlend(day_night_ratio, nodemgr);
+					cao->updateLight(light);
+				}
+			}
+			else {
 				light = blend_light(day_night_ratio, LIGHT_SUN, 0);
-
-			cao->updateLight(light);
+				cao->updateLight(light);
+			}
 		}
 	};
 
--- a/src/client/clientenvironment.cpp
+++ b/src/client/clientenvironment.cpp
@@ -335,8 +335,7 @@ void ClientEnvironment::step(float dtime)
 					light = n.getLightBlend(day_night_ratio, nodemgr);
 					cao->updateLight(light);
 				}
-			}
-			else {
+			} else {
 				light = blend_light(day_night_ratio, LIGHT_SUN, 0);
 				cao->updateLight(light);
 			}


src/client/clientenvironment.cpp
```diff
@@ -328,12 +328,18 @@ void ClientEnvironment::step(float dtime)
 			// Get node at head
 			v3s16 p = cao->getLightPosition();
 			MapNode n = this->m_map->getNode(p, &pos_ok);
-			if (pos_ok)
-				light = n.getLightBlend(day_night_ratio, m_client->ndef());
-			else
+			if (pos_ok) {
+				const NodeDefManager *nodemgr = m_client->ndef();
+				//don't update if clipping with a solid node.
```
sfan5, line 333:
```suggestion
				// Don't update if clipping with a solid node so object doesn't turn black
```

SmallJoker (dismissed):

sfan5:
Testing with spears as of [this commit](https://gitlab.com/echoes91/spears/-/commit/b5ce10d1574f885b835c20663303cb4bb7749eee), I can still manage to get spears to appear as black if thrown into a ceiling. The fix does work for the player model however.

I wonder how big of a problem it is that lightning never gets updated if you e.g. stick a spear into a wall at day and it stays glowing at night.

TheTermos:
> I can still manage to get spears to appear as black if thrown into a ceiling.

It definitely works with entities, I bet it's something about spears implementation, probably entities are being replaced after impact?

> I wonder how big of a problem it is that lightning never gets updated if you e.g. stick a spear into a wall at day and it stays glowing at night.

I figure it's less of a problem than if they turn black instantly.
I mean I think it's a better fallback in case no viable node is found, should a way of looking for the right node be agreed upon in the future.

paramat:
> There still may be some weird lightning e.g. [...] objects are left inside nodes for a long time, but come on, this is basically a two liner, why not merge and worry about complex solutions later.

I really like the cleverness and simplicity of this PR, but disagree with this attiude.
It is known the implementation has problems, so those problems should be considered now. Dealing with it later will probably mean significantly changing the initial implementation, might as well get it right first time and save everyone the extra work.
> I wonder how big of a problem it is that lightning never gets updated if you e.g. stick a spear into a wall at day and it stays glowing at night.

I assume lighting would not update if artificial lighting is changed too.
I actually consider these new bugs as much of a problem as the one fixed. An object being black when mostly inside a node is actually somewhat reasonable: The object is mostly inside a node, the node is dark inside, so the object is dark.
Objects being bright at night or the brightness not reacting to changing artificial light is a more 'broken' behaviour.

So i think the implementation needs a rethink, this will have to happen later anyway, might as well do it now.
I think making an object no longer update its light level is a clever, simple idea, but too crude and not acceptable, sorry :-1:

TheTermos:
> An object being black when mostly inside a node is actually slightly reasonable: The object is mostly inside a node, the node is dark inside, so the object is dark.

To be honest, I'm a bit tired of dealing with your reasoning of this quality. You make taking you seriously quite difficult.

Everyone, I still remember my first impression, that was little over a year ago. Turning black gives just terrible quality impression. Meanwhile there are PRs meaning to solve this issue sitting open for years.

Anyway I don't think that 100% effective method of determining model lighting is likely to happen, so whatever the solution is going to be, some kind of fallback is a necessity, so what is it going to be, the one that always and immediately gives unintended results, or the one that does it only sometimes and over long time.

paramat:
Well, i mean 'reasonable' in terms of thinking about a super-simplified approach to lighting, i do not mean it is 'fine for MT'. That sentence is absolutely meant to be taken in context together with the following sentence, as in: What your PR does, with no light updates, would look even more broken in my opinion. Sorry if i was unclear, i can see how that can be taken the wrong way.

TheTermos:
> That sentence is absolutely meant to be taken in context together with the following sentence, as in:  What your PR does, with no light updates, would look even more broken in my opinion.

Opinion is a matter of opinion, if you prefer stuff turning black instantly over retaining last known valid value then who am I to argue.
But your rationalizing reasoning above is irredeemable regardless of context.

jastevenson303:
>your PR does, with no light updates, would look even more broken in my
opinion. Sorry if i was unclear

Classic

paramat:
Possible close? Due to merge of https://github.com/minetest/minetest/commit/3f0cbbc372d5f359af18f5cbad37a1b165d3df4e

SmallJoker:
I am happy with the solution from #10006. It works well and I don't think it's really necessary to freeze the light level when diving into the ground.

TheTermos:
> I am happy with the solution from #10006. It works well

A bit disappointing.

#10006 is quite horrible. Yes it's going to work most of the time for the trivial scenario - player vs nodes, because player model happens to be almost entirely contained within its collision box, but I'm assuming you guys probably intend this thing to be capable of much more, don't you?

#10006 is flawed in the same way the old code was, and so it needs this solution too.
But imo the best course of action would be to do it properly based on the bounding box.

PR #9818, opened 2020-05-05
Make bgcolor tint button background images

Df458:
This PR addresses one of the tasks in #9044 by making the `bgcolor_*` style properties tint button background images. This is very useful, because it lets users potentially set custom background images in a formspec prepend while still benefiting from the ability to make colored buttons.

This PR, like most of my recent button PRs, is relatively simple on the code side. The only notable change is exposing color tint to `draw2DImage9Slice`, but that function already *had* a color tint that it was setting be default so it's not a big change either.

## How to test
I'm still recovering from my health issues, so I haven't gone to the trouble of building a big test mod showcasing the possibilities. However, I've tinted one of the buttons in the `/formspec` styling tab, and further tests should be pretty straightforward: Just use `bgcolor` and `bgimg` on the same button elements.

### DIFF
--- a/src/gui/guiButton.cpp
+++ b/src/gui/guiButton.cpp
@@ -313,11 +313,12 @@ void GUIButton::draw()
 
 		// PATCH
 		video::ITexture* texture = ButtonImages[(u32)imageState].Texture;
+		video::SColor image_colors[] = { BgColor, BgColor, BgColor, BgColor };
 		if (BgMiddle.getArea() == 0) {
 			driver->draw2DImage(texture,
 					ScaleImage? AbsoluteRect : core::rect<s32>(pos, sourceRect.getSize()),
 					sourceRect, &AbsoluteClippingRect,
-					0, UseAlphaChannel);
+					image_colors, UseAlphaChannel);
 		} else {
 			core::rect<s32> middle = BgMiddle;
 			// `-x` is interpreted as `w - x`
@@ -327,7 +328,7 @@ void GUIButton::draw()
 				middle.LowerRightCorner.Y += texture->getOriginalSize().Height;
 			draw2DImage9Slice(driver, texture,
 					ScaleImage ? AbsoluteRect : core::rect<s32>(pos, sourceRect.getSize()),
-					middle, &AbsoluteClippingRect);
+					middle, &AbsoluteClippingRect, image_colors);
 		}
 		// END PATCH
 	}
@@ -722,6 +723,8 @@ GUIButton* GUIButton::addButton(IGUIEnvironment *environment,
 
 void GUIButton::setColor(video::SColor color)
 {
+	BgColor = color;
+
 	float d = 0.65f;
 	for (size_t i = 0; i < 4; i++) {
 		video::SColor base = Environment->getSkin()->getColor((gui::EGUI_DEFAULT_COLOR)i);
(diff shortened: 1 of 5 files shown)


SmallJoker:
It works.
What about texture packs? The mod-defined tint color might not match the used texture.

Df458:
> What about texture packs? The mod-defined tint color might not match the used texture.

We can address that by adding a new style property for defining non-tintable regions in the background (like the overlay property on item definitions). I'm not sure if it's worth going to such a length though, what do you think?

SmallJoker:
@Df458 Non-tintable regions wouldn't look that great - the entire button should be tinted the same.
How about downscaling the image to one pixel, and take that as the background color? Though that might not yield the desired color either..

Df458:
I think I'm having some trouble following your argument, my assumption was that you were talking about situations like buttons with different 'border' and 'background' colors in their textures but your response seems to indicate that you're thinking of some other issue.

Could you give an example?

SmallJoker:
@Df458 Imagine a game where all buttons are styled in a (for example) grey tone. Assume someone wants to use their own textures - let's say - green buttons. Using 9-slice textures this is possible:


grafik

However, by hardcoding a color value in the formspec style, it is no longer possible to change regular button textures to any other color.

Df458:
I see now. In this case, not much can be done. Even if we can tell what the 'original color' of a button is, recoloring it based on that information is a seriously non-trivial issue. Rather than multiplying colors like we do elsewhere, we'd have to adjust the HSV values based on the color difference I think? I'm not even sure where you'd start with an operation like that, and I'm not convinced the result would actually look good either.

In the end, I think it's down to either allowing modders to easily recolor buttons and accepting the potential cost, or not merging this at all.

Personally though, [despite being a texture pack maker](https://content.minetest.net/packages/Hugues%20Ross/rpg16/) I think the downside of not allowing this is far greater. Modders can already force a button texture to a particular color with texture modifiers, and this case has the exact same downsides described above. However, to do so they need to know the exact background texture being used. This forces them to repeat the texture name in multiple places, and in the worst-case scenario (a mod supporting 2 games with different default button textures) could create a no-win situation where the modder can't easily make a colored button that fits.

Df458:
@SmallJoker how do we want to proceed here?

If you feel the issue you've pointed out is a blocker, then it's best to close this and move on. Otherwise, I'll rebase and port the example to devtest right away so the review process can continue.

SmallJoker:
@Df458 What I like most about this PR is that less textures need to be created but still get a very similar look. Though hardcoded modifiers limit the theming freedoms of texture packs, artists will face less work due to automatic colouring.
The benefits overweight the downsides in this case. Sorry for not replying earlier - I did not know how to proceed either. I'm keen approving if you'd like to rebase this PR.

Df458:
@SmallJoker I've rebased the PR and re-added the example to devtest (Sorry, forgot to notify you)

KaylebJay:
This is a really great feature. One approval, could we get another core dev to look at this, perhaps? (for 5.4.0, of course :slightly_smiling_face: )

Df458:
Rebased on 5.4.0. Let's give this the final push it needs!

SmallJoker:
Code still looks good. Approval is up to date.

src/gui/guiButton.cpp
```diff
@@ -750,22 +753,26 @@ void GUIButton::setFromStyle(const StyleSpec& style)
...
+						Colors[i] = multiplyColorValue(Colors[i], COLOR_HOVERED_MOD);
 				}
-			}
 		}
 
 	} else {
+		BgColor = video::SColor(255, 255, 255, 255);
 		for (size_t i = 0; i < 4; i++) {
 			video::SColor base =
 					Environment->getSkin()->getColor((gui::EGUI_DEFAULT_COLOR)i);
```
v-rob, line 778: `BgColor` should be set to what `base` is set to instead of white.  It doesn't make a difference now, but if `BgColor` is used in the future somewhere else, this could cause a strange bug.

rubenwardy (approved):

PR #9833, opened 2020-05-07
Damage texture modifier

### LINKED ISSUE #9607: Allow damage texture modifiers other than ^[brighten
## Problem

Sometimes you don't want the entity to flash white when you punch it. e.g. When you punch a mob you may want it to flash red instead. I propose allowing this to happen without hacky lua workarounds.

## Solutions

I propose to simply allow the entities to specify what texture modifier to use within their definition table. With a key similar to damage_texture_overlay  and "^[brighten" as the default value you would have the same functionality as before by default but if you wanted to you could set the value to be something like "^[invert:rgb" to get some really weird punches.

appgurueu:
- Implements damage texture modifiers
- Adds a new object property `damage_texture_modifier`
- Closes #9607

## To do

This PR is Ready for Review.

## How to test

```lua
    player:set_properties{damage_texture_modifier = "^[invert:rgb"}
end)
```

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -6546,7 +6546,7 @@ Player properties need to be saved manually.
         -- Defaults to 'true'.
 
         damage_texture_modifier = "^[brighten",
-        -- By default "^[brighten", texture modifier to be applied when object is hit
+        -- Texture modifier to be applied for a short duration when object is hit
     }
 
 Entity definition
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -6544,6 +6544,9 @@ Player properties need to be saved manually.
         -- deleted when the block gets unloaded.
         -- The get_staticdata() callback is never called then.
         -- Defaults to 'true'.
+
+        damage_texture_modifier = "^[brighten",
+        -- By default "^[brighten", texture modifier to be applied when object is hit
     }
 
 Entity definition
--- a/src/client/content_cao.cpp
+++ b/src/client/content_cao.cpp
@@ -1691,7 +1691,7 @@ bool GenericCAO::directReportPunch(v3f dir, const ItemStack *punchitem,
 					v2f(m_prop.visual_size.X, m_prop.visual_size.Y) * BS);
 			m_env->addSimpleObject(simple);
 		}
-		if (m_reset_textures_timer < 0) {
+		if (m_reset_textures_timer < 0 && !m_prop.damage_texture_modifier.empty()) {
 			m_reset_textures_timer = 0.05;
 			if (result.damage >= 2)
 				m_reset_textures_timer += 0.05 * result.damage;
--- a/src/client/content_cao.cpp
+++ b/src/client/content_cao.cpp
@@ -1502,7 +1502,7 @@ void GenericCAO::processMessage(const std::string &data)
 	} else if (cmd == AO_CMD_SET_TEXTURE_MOD) {
 		std::string mod = deSerializeString(is);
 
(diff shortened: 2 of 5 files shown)


src/client/content_cao.cpp
```diff
@@ -1622,7 +1622,7 @@ void GenericCAO::processMessage(const std::string &data)
 				m_reset_textures_timer = 0.05;
```
sfan5, line 1622: the 2 comment lines above here should be removed too

src/object_properties.cpp
```diff
@@ -166,4 +167,5 @@ void ObjectProperties::deSerialize(std::istream &is)
 	eye_height = readF32(is);
 	zoom_fov = readF32(is);
 	use_texture_alpha = readU8(is);
+	damage_texture_modifier = deSerializeString(is);
```
sfan5, line 170: needs backwards compatibility

src/client/content_cao.cpp
appgurueu: just removed it

src/object_properties.cpp
appgurueu: sfan5: how should I implement that? checking for `is_open` ?

src/client/content_cao.cpp
sfan5: Still there

appgurueu: oops, removed it in the other file but not there - now done

src/client/content_cao.cpp
```diff
@@ -1617,12 +1617,10 @@ void GenericCAO::processMessage(const std::string &data)
 						v2f(m_prop.visual_size.X, m_prop.visual_size.Y) * BS);
 				m_env->addSimpleObject(simple);
 			} else if (m_reset_textures_timer < 0) {
```
sfan5, line 1619: same for the other
```suggestion
			} else if (m_reset_textures_timer < 0 && !m_prop.damage_texture_modifier.empty()) {
```

doc/lua_api.txt
```diff
@@ -6544,6 +6544,9 @@ Player properties need to be saved manually.
         -- deleted when the block gets unloaded.
         -- The get_staticdata() callback is never called then.
         -- Defaults to 'true'.
+
+        damage_texture_modifier = "^[brighten",
+        -- By default "^[brighten", texture modifier to be applied when object is hit
```
sfan5, line 6549:
```suggestion
        -- Texture modifier to be applied when object is hit
```

src/client/content_cao.cpp
ClobberXD: Is this to account for mods explicitly overriding the default texture modifier with `""`?

sfan5: Yes, it works fine without this but there is no need to set the texture timers and whatnot when the mod wants the damage texture disabled.

ClobberXD: Ahh, understood. Thanks.

src/object_properties.cpp
```diff
@@ -166,4 +168,7 @@ void ObjectProperties::deSerialize(std::istream &is)
 	eye_height = readF32(is);
 	zoom_fov = readF32(is);
 	use_texture_alpha = readU8(is);
+	try {
+		damage_texture_modifier = deSerializeString(is);
+	} catch (SerializationError &e) {}
```
SmallJoker, line 173:
```suggestion
	} catch (SerializationError &e) {
	} catch (PacketError &e) {}
```

PR #9858, opened 2020-05-12
CSM: Bugfixes to camera:get_pos() and camera:get_fov()

### LINKED ISSUE #9857: Minetest.camera:get_pos() returns pos with wrong decimal place
For example: if player position is:
x= -1285.9
y=3.5
z=1576.5

then minetest.camera reports the camera adjusted position as:
x = -12858.57421875
y = 49.700000762939
z = 15765.104492188

sfan5:
closes #9857

## To do

### DIFF
--- a/clientmods/preview/init.lua
+++ b/clientmods/preview/init.lua
@@ -79,7 +79,7 @@ core.register_on_item_use(function(itemstack, pointed_thing)
 		return false
 	end
 
-	local pos = vector.add(core.localplayer:get_pos(), core.camera:get_offset())
+	local pos = core.camera:get_pos()
 	local pos2 = vector.add(pos, vector.multiply(core.camera:get_look_dir(), 100))
 
 	local rc = core.raycast(pos, pos2)
--- a/src/script/lua_api/l_camera.cpp
+++ b/src/script/lua_api/l_camera.cpp
@@ -51,6 +51,7 @@ void LuaCamera::create(lua_State *L, Camera *m)
 	lua_setfield(L, objectstable, "camera");
 }
 
+// set_camera_mode(self, mode)
 int LuaCamera::l_set_camera_mode(lua_State *L)
 {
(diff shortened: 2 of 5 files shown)


clientmods/preview/init.lua
```diff
@@ -79,7 +79,7 @@ core.register_on_item_use(function(itemstack, pointed_thing)
 		return false
 	end
 
-	local pos = vector.add(core.localplayer:get_pos(), core.camera:get_offset())
+	local pos = core.camera:get_pos()
```
SmallJoker, line 82: These are not equal in the two 3rd perspective modes. This code is not so important, so it might suffice to note that this shootline starts from the camera, rather than the head position.

src/script/lua_api/l_camera.cpp
```diff
@@ -115,38 +120,40 @@ int LuaCamera::l_get_offset(lua_State *L)
 	return 1;
 }
 
+// get_look_dir(self)
```
SmallJoker, line 123: `// FIXME: wouldn't localplayer be a better place for this?`

for the sake of completeness

src/script/lua_api/l_camera.cpp
```diff
@@ -115,38 +120,40 @@ int LuaCamera::l_get_offset(lua_State *L)
...
+	if (!camera)
+		return 0;
 
-	push_v3f(L, v);
+	push_v3f(L, camera->getDirection());
 	return 1;
 }
 
+// get_look_horizontal(self)
+// FIXME: wouldn't localplayer be a better place for this?
```
SmallJoker, line 135: Would be consistent with the server-side `ObjectRef`, but since these values are calculated locally I think they should be kept in the camera object.

clientmods/preview/init.lua
sfan5: This is just debug and testing code.

Issue #8390, opened 2019-03-17
minetest.register_on_player_inventory_action ignores many inventory changes


Wuzzy2:
##### Issue type
 - Bug report
 
##### Minetest version
```
5.0.0
```

##### Summary
The function `minetest.register_on_player_inventory_action` ignores many inventory changes. There are many ways the inventory of a player can change without triggering this callback:

* Pick up an item by punching it
* Obtain an item by mining
* Clear inventory with `/clearinv` or `/pulverize`
* Any inventory change by a mod
* Probably other events (?)

However, in contrast, throwing away an item DOES trigger this callback. o_O

This behaviour does not make sense at all, there doesn't seem to be any logic behind this.

This behaviour makes the uses for this function rather limited.

Expecteed behaviour:
All of the above events trigger this function, as they all change the inventory in some way.

##### Steps to reproduce
Use this test mod in singleplayer:
```
local i = 0
minetest.register_on_player_inventory_action(function(player, action, inventory, inventory_info)
	minetest.log("error", "Inventory action triggered! sequence number: "..i)
	i = i + 1
end)
```

Then do anything of the actions listed above and try to do other things with your inventory. For events that get detected, a message appears. Otherwise, it doesn't.

SmallJoker:
Déjà vu. Is this a duplicate?

The behaviour is easy to explain:
A) The inventory is modified in C++: Callback is run
B) The inventory is modified in Lua: Callback is not run

Would you want a callback flood for a filled inventory which is then cleared? Oh god, that would perform horribly if all Lua actions caused this callback to be executed.

ClobberXD:
@SmallJoker: Your explanation makes perfect sense, but I still think this behaviour is useless, if only a certain few methods of moving items between inventories can trigger the callbacks. What's the point of having a callback if it isn't called for all actions? If there's the issue of this callback being invoked all the time, we could simply add a warning like we've done for `register_on_dignode` and the like, that the callback will be invoked way too often, and that it isn't recommended to overuse this callback for the same reason.

Wuzzy2:
> The behaviour is easy to explain:
> A) The inventory is modified in C++: Callback is run
> B) The inventory is modified in Lua: Callback is not run

Not documented. ;-)

Linuxdirk:
So in current state the callback is pretty much useless for _real-life_ usage? What are inventory changes by C++ that happen in everyday gameplay?

SmallJoker:
[...]

PR #9862, opened 2020-05-13
Items: on_wield and on_unwield callbacks implemented

CrazyminerMC:
- Goal of the PR
Add two new callbacks
It adds two callbacks when an item is wielded and unwielded

## To do

 Ready for Review.

## How to test

Register a new node and assign the on_wield and the on_unwield callbacks 
`on_wield = function(itemstack, user)`
`on_unwield = function(itemstack, user)`

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -6887,7 +6887,7 @@ Used by `minetest.register_node`, `minetest.register_craftitem`, and
         -- Shall drop item and return the leftover itemstack.
         -- The dropper may be any ObjectRef or nil.
         -- default: minetest.item_drop
-        
+
         on_wield = function(itemstack, user),
         -- default: nil
         -- Called when the player switches to this item in the hotbar.
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -6890,7 +6890,7 @@ Used by `minetest.register_node`, `minetest.register_craftitem`, and
         
         on_wield = function(itemstack, user),
         -- default: nil
-        -- It defines the behaviour of an item when it's wielded by a player.
+        -- Called when the player switches to this item in the hotbar.
         -- The user may be any ObjectRef or nil.
 
         on_use = function(itemstack, user, pointed_thing),
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -6893,6 +6893,11 @@ Used by `minetest.register_node`, `minetest.register_craftitem`, and
         -- Called when the player switches to this item in the hotbar.
         -- The user may be any ObjectRef or nil.
 
+        on_unwield = function(itemstack, user),
+         -- default: nil
+         -- Called when the player switches from this item to another slot in the hotbar.
+         -- The user may be any ObjectRef or nil.
+
         on_use = function(itemstack, user, pointed_thing),
         -- default: nil
         -- Function must return either nil if no item shall be removed from
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -6887,6 +6887,11 @@ Used by `minetest.register_node`, `minetest.register_craftitem`, and
         -- Shall drop item and return the leftover itemstack.
         -- The dropper may be any ObjectRef or nil.
         -- default: minetest.item_drop
+        
+        on_wield = function(itemstack, user),
+        -- default: nil
+        -- It defines the behaviour of an item when it's wielded by a player.
+        -- The user may be any ObjectRef or nil.
 
         on_use = function(itemstack, user, pointed_thing),
         -- default: nil
--- a/src/inventorymanager.cpp
+++ b/src/inventorymanager.cpp
@@ -531,6 +531,17 @@ void IMoveAction::apply(InventoryManager *mgr, ServerActiveObject *player, IGame
 		}
 	}
 
+	if (from_inv.type == InventoryLocation::PLAYER || to_inv.type == InventoryLocation::PLAYER){
+		// if it's moved in the selected slot and there are no items like such already
+		if (to_i == player->getWieldIndex() &&
+			to_stack_was.name != src_item.name)
+			PLAYER_TO_SA(player)->item_OnWield(src_item, player);
+		// if the whole stack is moved away from the selected slot
+		else if (from_i == player->getWieldIndex() &&
+				 count == from_stack_was.count)
+			PLAYER_TO_SA(player)->item_OnUnwield(src_item, player);
+	}
+
 	mgr->setInventoryModified(from_inv);
 	if (inv_from != inv_to)
 		mgr->setInventoryModified(to_inv);
@@ -679,6 +690,11 @@ void IDropAction::apply(InventoryManager *mgr, ServerActiveObject *player, IGame
 
 			if (item2.count != actually_dropped_count)
 				errorstream<<"Could not take dropped count of items"<<std::endl;
+
+			// If players drop the last item in the stack they had in hand
+			if (from_i == player->getWieldIndex() &&
+					(src_item.count - actually_dropped_count == 0))
+				PLAYER_TO_SA(player)->item_OnUnwield(item2, player);
 		}
 
 		src_item.count = actually_dropped_count;
--- a/src/network/serverpackethandler.cpp
+++ b/src/network/serverpackethandler.cpp
@@ -858,6 +858,10 @@ void Server::handleCommand_PlayerItem(NetworkPacket* pkt)
 
 	*pkt >> item;
 
+	ItemStack previous_item;
+	playersao->getWieldedItem(&previous_item, nullptr);
+	m_script->item_OnUnWield(previous_item, playersao);
+
 	playersao->getPlayer()->setWieldIndex(item);
 
 	ItemStack selected_item;
--- a/src/network/serverpackethandler.cpp
+++ b/src/network/serverpackethandler.cpp
@@ -860,7 +860,7 @@ void Server::handleCommand_PlayerItem(NetworkPacket* pkt)
 
 	ItemStack previous_item;
 	playersao->getWieldedItem(&previous_item, nullptr);
-	m_script->item_OnUnWield(previous_item, playersao);
+	m_script->item_OnUnwield(previous_item, playersao);
 
 	playersao->getPlayer()->setWieldIndex(item);
 
--- a/src/network/serverpackethandler.cpp
+++ b/src/network/serverpackethandler.cpp
@@ -1262,13 +1262,23 @@ void Server::handleCommand_Interact(NetworkPacket *pkt)
 			}
 
 			pointed_object->rightClick(playersao);
-		} else if (m_script->item_OnPlace(
-				selected_item, playersao, pointed)) {
-			// Placement was handled in lua
+		} else {
+			// Need to save the selected_item because item_OnPlace
+			// changes the referenced selected_item
+			ItemStack old_selected_item = selected_item;
+			// Handle placement in lua
+			if (m_script->item_OnPlace(
+					selected_item, playersao, pointed)) {
+				// If the last item was placed, launch item_OnUnwield()
+				if (selected_item.count == 0) {
+					m_script->item_OnUnwield(
+							old_selected_item, playersao);
+				}
 
-			// Apply returned ItemStack
-			if (playersao->setWieldedItem(selected_item)) {
-				SendInventory(playersao, true);
+				// Apply returned ItemStack
+				if (playersao->setWieldedItem(selected_item)) {
+					SendInventory(playersao, true);
+				}
 			}
 		}
 
--- a/src/network/serverpackethandler.cpp
+++ b/src/network/serverpackethandler.cpp
@@ -858,15 +858,16 @@ void Server::handleCommand_PlayerItem(NetworkPacket* pkt)
 
 	*pkt >> item;
 
-	ItemStack previous_item;
-	playersao->getWieldedItem(&previous_item, nullptr);
-	m_script->item_OnUnwield(previous_item, playersao);
-
-	playersao->getPlayer()->setWieldIndex(item);
-
 	ItemStack selected_item;
-	playersao->getWieldedItem(&selected_item, nullptr);
-	m_script->item_OnWield(selected_item, playersao);
+	player->getWieldedItem(&selected_item, nullptr);
+	if (!selected_item.empty())
+		m_script->item_OnUnwield(selected_item, playersao);
+
+	player->setWieldIndex(item);
+
+	player->getWieldedItem(&selected_item, nullptr);
+	if (!selected_item.empty())
+		m_script->item_OnWield(selected_item, playersao);
 }
 
 void Server::handleCommand_Respawn(NetworkPacket* pkt)
--- a/src/network/serverpackethandler.cpp
+++ b/src/network/serverpackethandler.cpp
@@ -859,6 +859,10 @@ void Server::handleCommand_PlayerItem(NetworkPacket* pkt)
 	*pkt >> item;
 
 	playersao->getPlayer()->setWieldIndex(item);
+
+    ItemStack selected_item;
+    playersao->getWieldedItem(&selected_item, nullptr);
+    m_script->item_OnWield(selected_item, playersao);
 }
 
 void Server::handleCommand_Respawn(NetworkPacket* pkt)
--- a/src/network/serverpackethandler.cpp
+++ b/src/network/serverpackethandler.cpp
@@ -860,9 +860,9 @@ void Server::handleCommand_PlayerItem(NetworkPacket* pkt)
 
 	playersao->getPlayer()->setWieldIndex(item);
 
-    ItemStack selected_item;
-    playersao->getWieldedItem(&selected_item, nullptr);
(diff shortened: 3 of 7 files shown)


ClobberXD:
Nice work. I haven't seen the code yet, but I'd like the callback to also have access to the previously wielded item. i.e.

```lua
on_wield = function(itemstack, old_itemstack, user)
```

This would allow mods to run different code based on what item the player previously wielded. Also, would it be a good idea to pass the current and old wield indices as well? I can't immediately think of a use-case, but it would certainly come in handy for some purposes.

Zughy:
(as I worked on the PR too) I can't really think of a situation where a player wants to chain two items to make something happen, can you provide an example @ClobberXD ?

ClobberXD:
> I can't really think of a situation where a player wants to chain two items to make something happen

I can't think of any use-cases that *require the player* to cycle from one item to another. However, think of it this way: a player is always wielding something - on_wield runs when a player wields a new item and "stops" wielding the previous item. A wield action consists of the player also "stopping" to wield the old item, IMO. While the use-cases for knowing the old item aren't immediately obvious, that isn't a justification for passing only half of the information. :)

> also, the index can be retrieved already by using the current parameters (user:get_wield_index()) . I don't think it's necessary passing it by default

This is partially for the sake of convenience, but again, it's not possible for a mod to retrieve the previously wielded item or its wield index (unless it keeps track of previous and current using globalsteps). As aforementioned, I can't immediately think of use-cases, but I'm pretty sure a mod would put this feature to good use *if this feature exists*.

Zughy:
I see your point Clobber, how about implementing another separate callback, such as `on_switch`, to be called when an item gets changed?

ClobberXD:
> I see your point Clobber, how about implementing another separate callback, such as on_switch or on_unwield, to be called when an item gets changed?

Would a separate callback for an couple of extra params be justified? I'm not quite sure about that...

Btw, there's also going to be a `minetest.register_on_wielditem_change` global callback (see #7587). If `on_wield` isn't suitable for a mod's use-case, `on_wielditem_change` should fit the bill in most cases.

doc/lua_api.txt
```diff
@@ -6887,6 +6887,11 @@ Used by `minetest.register_node`, `minetest.register_craftitem`, and
         -- Shall drop item and return the leftover itemstack.
         -- The dropper may be any ObjectRef or nil.
         -- default: minetest.item_drop
+        
+        on_wield = function(itemstack, user),
+        -- default: nil
+        -- It defines the behaviour of an item when it's wielded by a player.
```
sfan5, line 6893:
```suggestion
        -- Called when the player switches to this item in the hotbar.
```

src/network/serverpackethandler.cpp
```diff
@@ -859,6 +859,10 @@ void Server::handleCommand_PlayerItem(NetworkPacket* pkt)
 	*pkt >> item;
 
 	playersao->getPlayer()->setWieldIndex(item);
+
+    ItemStack selected_item;
+    playersao->getWieldedItem(&selected_item, nullptr);
+    m_script->item_OnWield(selected_item, playersao);
```
sfan5, line 865: spaces -> tabs

src/script/cpp_api/s_item.cpp
```diff
@@ -88,6 +88,25 @@ bool ScriptApiItem::item_OnPlace(ItemStack &item,
...
+	int error_handler = PUSH_ERROR_HANDLER(L);
+	
+	// Push callback function on stack
+	if (!getItemCallback(item.name.c_str(), "on_wield"))
+	  return false;
+		
+	// Call function
+	LuaItemStack::create(L, item);
+	objectrefGetOrCreate(L, user);
+	PCALL_RES(lua_pcall(L, 2, 1, error_handler));
```
sfan5, line 105: the third argument to lua_pcall here should be `0` since you're not expecting a return value
then the pop below also needs to be changed to `1`

doc/lua_api.txt
CrazyminerMC: Done

src/script/cpp_api/s_item.cpp
CrazyminerMC: Done

src/network/serverpackethandler.cpp
CrazyminerMC: Done

pandaro:
> Btw, there's also going to be a minetest.register_on_wielditem_change global callback (see #7587). If on_wield isn't suitable for a mod's use-case, on_wielditem_change should fit the bill in most cases.

i will suggest to implement an "on_unwild" callback. It look like something lost, because if something happen when a tool is wielded and unhappen when it is unwielded we need to check with a generic minetest.register_on_wielditem_change. It's evident. For example: suppose to have a pick that give you immortality unwielded...Why i must introduce a global callback just to check if i wielded this pick and unwielded this pick, when i can define this behaviour in the pick definition.

ClobberXD:
@pandaro A global callback makes sense when the logic is de-coupled from specific items - that's the objective of "global" callbacks, really. :)

If you want to, say, keep track of *all* wielditem changes, adding (or possibly even overriding) `on_wield` and `on_unwield` for all items is probably the worst way to do it. As you can see, the usecases for per-item and global callbacks are very much different from each other.

****

But thinking about it again, an `on_unwield` callback to complement `on_wield` makes total sense to me now. It's rather silly to expect mods to use the global callback for this. But IMO, this still doesn't justify not passing the previous wield item to `on_wield`, @Zughy :)

KaylebJay:
This is great :)) So many uses for this new function. :+1: on_unwield does not make quite as much sense, but it would be nice to have in the back pocket too.

sfan5:
> But IMO, this still doesn't justify not passing the previous wield item to `on_wield`

Do you have an example usecase for that?

ClobberXD:
> Do you have an example usecase for that?

I can't come up with a use-case for this at the moment, sorry. But as I noted in a previous comment:

> A wield action consists of the player also "stopping" to wield the old item, IMO. While the use-cases for knowing the old item aren't immediately obvious, that isn't a justification for passing only half of the information.

My opinion is that an on_wield is an action that comprises of "stopping" to wield the old item, and beginning to wield the new item. There's the argument that one could use the global `on_wielditem_change` that passes the old item too, but this means that mods would have to use the global callback if they want to check the previous wielditem, even if it's for only one specific item.

pandaro:
I have just compiled a minetest build with this PR and seems to work as well.
This PR is really nice for me so if i can do something in order to help, maybe(maybe) i can do it.

sfan5:
Tested it and it works, but here's some suggested code changes:
```diff
diff --git a/src/network/serverpackethandler.cpp b/src/network/serverpackethandler.cpp
index 66bd5a028..519f387f0 100644
--- a/src/network/serverpackethandler.cpp
+++ b/src/network/serverpackethandler.cpp
@@ -861,15 +861,16 @@ void Server::handleCommand_PlayerItem(NetworkPacket* pkt)
 
        *pkt >> item;
 
-       ItemStack previous_item;
-       playersao->getWieldedItem(&previous_item, nullptr);
-       m_script->item_OnUnwield(previous_item, playersao);
+       ItemStack selected_item;
+       player->getWieldedItem(&selected_item, nullptr);
+       if (!selected_item.empty())
+               m_script->item_OnUnwield(selected_item, playersao);
 
-       playersao->getPlayer()->setWieldIndex(item);
+       player->setWieldIndex(item);
 
-       ItemStack selected_item;
-       playersao->getWieldedItem(&selected_item, nullptr);
-       m_script->item_OnWield(selected_item, playersao);
+       player->getWieldedItem(&selected_item, nullptr);
+       if (!selected_item.empty())
+               m_script->item_OnWield(selected_item, playersao);
 }
 
 void Server::handleCommand_Respawn(NetworkPacket* pkt)
```

An alternative to `!selected_item.empty()` would be to implement the full hand list handling (second parameter of `getWieldedItem` + return value).

pandaro:
I'm in hard testing this PR but i have some trouble in implementing my things so i decide to post it there cause my suspect is that there is a problem with `set_tool_capabilities`
here a example code:
```
minetest.register_tool("rpgtool:1", {
	description = "rpgtool 1",
	inventory_image = "default_tool_steelaxe.png",
	_rpg = true,
	on_unwield = function(itemstack,user)
		local meta = itemstack:get_meta()
		meta:set_tool_capabilities({
			max_drop_level=0,
			range=10,
			groupcaps={
				cracky={times={[2]=0.1, [3]=0.1}, uses=100, maxlevel=1}
			},
			damage_groups = {fleshy=0},
		})
	return itemstack
end,
	on_wield = function(itemstack,user)
		local meta = itemstack:get_meta()
		meta:set_tool_capabilities({
			max_drop_level=0,
			range=10,
			groupcaps={
				cracky={times={[2]=0.1, [3]=0.1}, uses=100, maxlevel=1}
			},
			damage_groups = {fleshy=10},
		})
		return itemstack
	end,
	on_place = function(itemstack, placer, pointed_thing)
		local meta = itemstack:get_meta()
		meta:set_tool_capabilities({
			max_drop_level=0,
			range=10,
			groupcaps={
				cracky={times={[2]=0.1, [3]=0.1}, uses=100, maxlevel=1}
			},
			damage_groups = {fleshy=30},
		})
		return itemstack
	end,
	on_secondary_use = function(itemstack, user, pointed_thing)
		local meta = itemstack:get_meta()
		meta:set_tool_capabilities({
			max_drop_level=0,
			range=10,
			groupcaps={
				cracky={times={[2]=0.1, [3]=0.1}, uses=100, maxlevel=1}
			},
			damage_groups = {fleshy=20},
		})
		return itemstack
	end,

	tool_capabilities = {
		max_drop_level=0,
		range = 4,
		groupcaps={
			cracky={times={[2]=2.00, [3]=2.00}, uses=10, maxlevel=1}
		},
		damage_groups = {fleshy=2},
	},
})
```

if anyone have time to try it out, can see that with `on_place` and `on_secondary_use` the tool capability will change, but using `on_wield` or `on_unwield` the tool capability do not change or i'm loosing something.
Thanks for your time

ClobberXD:
@pandaro `on_wield` and `on_unwield` don't return anything, unlike `on_use` and co.

pandaro:
Thanks @ClobberXD, there Is a way ti implement a return? The same as on_place and on_use? Can be usefull!

pandaro:
Sorry if I intrude again even if I'm not a core dev.
I would like to understand if it is actually possible to `return` the value of `Itemstack` to this function to make this call interoperable with `set_tool_capabilities`.
I don't know if it's a good idea or not, and I don't know if I can get the same result in another way.
thank you

Zughy:
> Thanks @ClobberXD. @CrazyminerMC There Is a way to implement a return? The same as on_place and on_use? Can be usefull!

Quoting sfan5 a few messages before: "Do you have an example usecase for that?"

If someone comes up with something in the future, it could be implemented then. As for now, 5.3.0 is going to be released on June 12th, so I'd avoid adding something not even a week before (considering the slowness of PRs)

pandaro:
> Quoting sfan5 a few messages before: "Do you have an example usecase for that?"

yes, i have post my code that is my personal usecase.
But if this will block your project do not worry, leave this PR as it is.
PS i'm interessed in your project!

sfan5:
Discussion in -dev today revealed a pretty big thing that this (or the other) PR do not handle:
http://irc.minetest.net/minetest-dev/2020-06-12#i_5701175

CrazyminerMC:
Now, I need some help to resolve one other problem I discovered, I need to launch on_wield when an item, which I pick up from the ground, is put in my selected slot. It looks like the item is added to the inventory from a lua script, am I right ? 
https://github.com/minetest/minetest/blob/0a1181f7637adab6fe1ba5d3c09038912a06b37f/src/server/luaentity_sao.cpp#L321

ClobberXD:
The builtin item entity uses `InvRef::add_item` to add the entity's item to the hitter's inventory:

https://github.com/minetest/minetest/blob/0a1181f7637adab6fe1ba5d3c09038912a06b37f/builtin/game/item_entity.lua#L315-L326

`add_item` adds the new item to the inventory list, and marks the list as dirty (L631). The dirty list (along with all other dirty/modified lists) is sent to the player in the next run of `ServerEnvironment::step`.

https://github.com/minetest/minetest/blob/0a1181f7637adab6fe1ba5d3c09038912a06b37f/src/inventory.cpp#L585-L633

CrazyminerMC:
Technically the best way to call on_wield and on_unwield is from `InventoryList::removeItem` and `InventoryList::addItem` but inside `inventory.cpp` I can't find the reference to the player

ClobberXD:
@CrazyminerMC Inventory objects don't contain references to Players; they're supposed to be independent of SAOs, and all Inventory objects are handled similarly, except for inventory sending of player inventories (only sent to the owner).

pandaro:
my opinion is that:
This PR should be merged as it is unless other technical issues, which concern code cleanliness and quality, do not prevent this.
Managing other situations requires other codes placed in other places: the problem of managing an object taken from the ground requires a dedicated function that has been missing for a long time. It has already been requested in other [issues:](https: //github.com/minetest/minetest/pull/7712).
An "on_item_pickup" function is needed beyond this PR and should not keep it locked.
Thank you for listening.

src/inventorymanager.cpp
```diff
@@ -531,6 +531,17 @@ void IMoveAction::apply(InventoryManager *mgr, ServerActiveObject *player, IGame
 		}
 	}
 
+	if (from_inv.type == InventoryLocation::PLAYER || to_inv.type == InventoryLocation::PLAYER){
```
sfan5, line 534: This needs to be split into two ifs, checking `to_i == player->getWieldIndex()` when to_inv isn't the player inventory causes bugs.

src/inventorymanager.cpp
```diff
@@ -531,6 +531,17 @@ void IMoveAction::apply(InventoryManager *mgr, ServerActiveObject *player, IGame
...
+		// if it's moved in the selected slot and there are no items like such already
+		if (to_i == player->getWieldIndex() &&
+			to_stack_was.name != src_item.name)
```
sfan5, line 537: This should compare metadata too (I think `to_stack_was != src_item`) works.

src/inventorymanager.cpp
```diff
@@ -531,6 +531,17 @@ void IMoveAction::apply(InventoryManager *mgr, ServerActiveObject *player, IGame
...
+			PLAYER_TO_SA(player)->item_OnWield(src_item, player);
+		// if the whole stack is moved away from the selected slot
+		else if (from_i == player->getWieldIndex() &&
+				 count == from_stack_was.count)
+			PLAYER_TO_SA(player)->item_OnUnwield(src_item, player);
+	}
```
sfan5, line 543: I also don't see any handling for swaps here, it needs to call <i>both</i> unwield on the old item and wield on the new one

src/server/luaentity_sao.cpp
```diff
@@ -360,9 +360,17 @@ u16 LuaEntitySAO::punch(v3f dir,
...
+	ItemStack old_selected_item;
+	puncher->getWieldedItem(&old_selected_item, nullptr);
+
 	bool damage_handled = m_env->getScriptIface()->luaentity_Punch(m_id, puncher,
 			time_from_last_punch, toolcap, dir, result.did_punch ? result.damage : 0);
 
+	puncher->getWieldedItem(&selected_item, nullptr);
+	// If LUA script added something to the selected slot launch item_OnWield
+	if (old_selected_item != selected_item)
+		puncher->getEnv()->getScriptIface()->item_OnWield(selected_item, puncher);
```
sfan5, line 372: Even if not very useful in practice, the convention is that Lua modifications to inventories does <i>not</i> run callbacks.

(This check also isn't very reliable if it only works in on_punch)

sfan5:
Looking at the required changes to get the callbacks to work when the item changes inside the inventory makes me wonder if this is worth it. The engine is entirely missing tracking for Lua inventory changes (#8390) and this PR can't be expected to hack that in.
On the other hand how useful is `on_wield` really when the item can vanish at any time without any notification?

Zughy:
@sfan5 so if I got this right, the engine is nowhere near ready to feature such callback (will it be eventually tho?).
About the utility of the callback, it's the only way we can implement different crosshairs (via HUD) in our game when changing the weapon in the player's hand (please don't suggest us to use `on_globalstep`). I could name at least a few other use-cases, so I definitely don't think the utility of it should be questioned

sfan5:
> I definitely don't think the utility of it should be questioned

I am not questioning the utility of `on_wield` in general.
I am questioning how useful it is when it cannot be relied upon (`on_wield` might not be called when there is an item, `on_unwield` might not be called when the item is gone, ...)

> please don't suggest us to use `on_globalstep`

But that works doesn't it?

Zughy:
> I am questioning how useful it is when it cannot be relied upon
Oh, I'm sorry sfan then. Still, I don't know the engine that much so I can't think of a proper solution off the top of my head...

> But that works doesn't it?
Yes, but it's quite a refrain when something doesn't work by default; according to this, I should have put a lot of things on that function, so I always try to avoid it like hell

CrazyminerMC:
As there are evident steps needed before even considering implementing this, it's pointless leaving this open for now. We hope to see those steps eventually

PR #9888, opened 2020-05-17
Implement grouped mode for find_nodes_in_area

sfan5:
usecase: minetest/minetest_game#2683

## To do

This PR is Ready for Review.

### DIFF
--- a/doc/client_lua_api.txt
+++ b/doc/client_lua_api.txt
@@ -768,12 +768,15 @@ Call these functions only at load time!
     * `nodenames`: e.g. `{"ignore", "group:tree"}` or `"default:dirt"`
     * `search_center` is an optional boolean (default: `false`)
       If true `pos` is also checked for the nodes
-* `minetest.find_nodes_in_area(pos1, pos2, nodenames)`: returns a list of
-  positions.
+* `minetest.find_nodes_in_area(pos1, pos2, nodenames, [grouped])`
+    * `pos1` and `pos2` are the min and max positions of the area to search.
     * `nodenames`: e.g. `{"ignore", "group:tree"}` or `"default:dirt"`
-    * First return value: Table with all node positions
-    * Second return value: Table with the count of each node with the node name
-      as index.
+    * If `grouped` is true the return value is a table indexed by node name
+      which contains lists of positions.
+    * If `grouped` is false or absent the return values are as follows:
+      first value: Table with all node positions
+      second value: Table with the count of each node with the node name
+      as index
     * Area volume is limited to 4,096,000 nodes
 * `minetest.find_nodes_in_area_under_air(pos1, pos2, nodenames)`: returns a
   list of positions.
--- a/src/script/lua_api/l_env.cpp
+++ b/src/script/lua_api/l_env.cpp
@@ -798,8 +798,7 @@ int ModApiEnvMod::l_find_node_near(lua_State *L)
 	return 0;
 }
 
-// find_nodes_in_area(minp, maxp, nodenames) -> list of positions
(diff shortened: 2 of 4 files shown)


KaylebJay:
Good feature, thanks sfan5.

sfan5:
If anyone feels like testing: it would be interesting to know if this is faster than using a vmanip to read all nodes and then iterating them in Lua.

src/script/lua_api/l_env.cpp
```diff
@@ -826,45 +833,73 @@ int ModApiEnvMod::l_find_nodes_in_area(lua_State *L)
...
-		ndef->getIds(readParam<std::string>(L, 3), filter);
-	}
 
-	std::vector<u32> individual_count;
-	individual_count.resize(filter.size());
+		// last filter table is at top of stack
+		u32 i = filter.size() - 1;
+		do {
+			if (idx[i] == 0)
+				lua_pop(L, 1);
```
SmallJoker, line 868: What would you think about a new comment here for clarification?
`// No such node found -> drop the empty table`

src/script/lua_api/l_env.cpp
```diff
@@ -826,45 +833,73 @@ int ModApiEnvMod::l_find_nodes_in_area(lua_State *L)
...
+
+		v3s16 p;
+		for (p.X = minp.X; p.X <= maxp.X; p.X++)
+		for (p.Y = minp.Y; p.Y <= maxp.Y; p.Y++)
+		for (p.Z = minp.Z; p.Z <= maxp.Z; p.Z++) {
+			content_t c = map.getNode(p).getContent();
+
+			auto it = std::find(filter.begin(), filter.end(), c);
+			if (it != filter.end()) {
+				u32 filt_index = it - filter.begin();
```
SmallJoker, line 858:
```suggestion
				// Calculate index of the filter table and append this position to its end
				u32 filt_index = it - filter.begin();
```

src/script/lua_api/l_env.cpp
```diff
@@ -826,45 +833,73 @@ int ModApiEnvMod::l_find_nodes_in_area(lua_State *L)
...
 
-	std::vector<u32> individual_count;
-	individual_count.resize(filter.size());
+		// last filter table is at top of stack
+		u32 i = filter.size() - 1;
+		do {
+			if (idx[i] == 0)
+				lua_pop(L, 1);
+			else
+				lua_setfield(L, base, ndef->get(filter[i]).name.c_str());
```
SmallJoker, line 870: `// Found >= 1 node -> push tables to the return value`

doc/client_lua_api.txt
```diff
@@ -768,12 +768,15 @@ Call these functions only at load time!
...
       If true `pos` is also checked for the nodes
-* `minetest.find_nodes_in_area(pos1, pos2, nodenames)`: returns a list of
-  positions.
+* `minetest.find_nodes_in_area(pos1, pos2, nodenames, [grouped])`
+    * `pos1` and `pos2` are the min and max positions of the area to search.
     * `nodenames`: e.g. `{"ignore", "group:tree"}` or `"default:dirt"`
-    * First return value: Table with all node positions
-    * Second return value: Table with the count of each node with the node name
-      as index.
+    * If `grouped` is true the return value is a table indexed by node name
```
LoneWolfHT, line 774: Can the node name be something like 'group:lava' in this case?

sfan5: No, the names returned are always nodes even if you put in a group.

LoneWolfHT: Hmmm, I guess a modder could use `minetest.get_item_group()` if they wanted to apply to all nodes in a certain group.
I have no problems with this then

PR #9893, opened 2020-05-19
Cleanup of particle & particlespawner structures and code

sfan5:
## To do

This PR is Ready for Review.

## How to test

Test that particles & particlespawners including all parameters still work

### DIFF
--- a/src/client/particles.cpp
+++ b/src/client/particles.cpp
@@ -37,32 +37,31 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 	Utility
 */
 
-v3f random_v3f(v3f min, v3f max)
+static f32 random_f32(f32 min, f32 max)
+{
+	return rand() / (float)RAND_MAX * (max - min) + min;
+}
+
+static v3f random_v3f(v3f min, v3f max)
 {
 	return v3f(
-		rand() / (float)RAND_MAX * (max.X - min.X) + min.X,
-		rand() / (float)RAND_MAX * (max.Y - min.Y) + min.Y,
-		rand() / (float)RAND_MAX * (max.Z - min.Z) + min.Z);
+		random_f32(min.X, max.X),
+		random_f32(min.Y, max.Y),
+		random_f32(min.Z, max.Z));
 }
+
+/*
+	Particle
+*/
 
 Particle::Particle(
 	IGameDef *gamedef,
 	LocalPlayer *player,
 	ClientEnvironment *env,
-	v3f pos,
-	v3f velocity,
-	v3f acceleration,
-	float expirationtime,
-	float size,
-	bool collisiondetection,
-	bool collision_removal,
-	bool object_collision,
-	bool vertical,
+	const ParticleParameters &p,
 	video::ITexture *texture,
 	v2f texpos,
 	v2f texsize,
-	const struct TileAnimationParams &anim,
-	u8 glow,
 	video::SColor color
 ):
 	scene::ISceneNode(RenderingEngine::get_scene_manager()->getRootSceneNode(),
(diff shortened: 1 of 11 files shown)


src/network/clientpackethandler.cpp
```diff
@@ -958,114 +958,68 @@ void Client::handleCommand_SpawnParticle(NetworkPacket* pkt)
...
+	std::string datastring(pkt->getString(0), pkt->getSize());
+	std::istringstream is(datastring, std::ios_base::binary);
 
-		// This is horrible but required (why are there two ways to deserialize pkts?)
-		std::string datastring(pkt->getRemainingString(), pkt->getRemainingBytes());
-		std::istringstream is(datastring, std::ios_base::binary);
-		animation.deSerialize(is, m_proto_ver);
-		glow = readU8(is);
-		object_collision = readU8(is);
-	} catch (...) {}
```
sfan5, line 1044: Note that the compatibility code removed here was for 766fb7b46edab47f96b8b940e390daf5319261b7, which is pre-5.0.

src/network/clientpackethandler.cpp
```diff
@@ -958,114 +958,68 @@ void Client::handleCommand_SpawnParticle(NetworkPacket* pkt)
...
-	try {
-		vertical = readU8(is);
-		collision_removal = readU8(is);
-		animation.deSerialize(is, m_proto_ver);
-		glow = readU8(is);
-		object_collision = readU8(is);
-	} catch (...) {}
+	ParticleParameters p;
+
+	p.pos                = readV3F32(is);
```
SmallJoker, line 963: How about moving this to `ParticleParameters::deSerialize()`?

src/client/particles.cpp
```diff
@@ -81,33 +80,28 @@ Particle::Particle(
 	m_material.setTexture(0, texture);
 	m_texpos = texpos;
 	m_texsize = texsize;
-	m_animation = anim;
+	m_animation = p.animation;
```
SmallJoker, line 83: What would be the disadvantages of keeping `ParticleParameters` inside the class, as replacement for those variables? The diff might be larger, but I think that would be fine.

src/network/clientpackethandler.cpp
```diff
@@ -958,114 +958,68 @@ void Client::handleCommand_SpawnParticle(NetworkPacket* pkt)
...
-		std::istringstream is(datastring, std::ios_base::binary);
-		animation.deSerialize(is, m_proto_ver);
-		glow = readU8(is);
-		object_collision = readU8(is);
-	} catch (...) {}
+	ParticleSpawnerParameters p;
+	u32 server_id;
+	u16 attached_id = 0;
+
+	p.amount             = readU16(is);
```
SmallJoker, line 992: Respectively `ParticleSpawnerParameters::serialize()` here.

src/network/clientpackethandler.cpp
sfan5: Kinda wanted to avoid a second `particles.cpp`, but sure can do.

src/client/particles.cpp
sfan5: Larger diff was my original reason not to. The code uses and modifies lots of variables that would be inside the structure and I think that doesn't really make sense to keep them together for that.

src/network/clientpackethandler.cpp
sfan5: Hm actually, for the particlespawner the code mixes between reading to `p` and reading stuff like `server_id` or `attached_id`. It's a little messy to pull that apart.
Should I do this for `ParticleParameters` only?

SmallJoker: Yes, I've seen the messy part, thus only suggested serialize functions for it for `ParticleParameters`. However, that could also be done at a later time.

src/client/particles.cpp
SmallJoker: I see. Let's keep it as-is for now.

src/network/clientpackethandler.cpp
sfan5: Really not sure if this is any better than before: https://github.com/minetest/minetest/pull/9893/commits/7806c2dd03e7999469740ac0f9afe0491baaeb26

SmallJoker: Thank you. It looks better to me because it properly separates the struct serialization from other variables.

PR #9896, opened 2020-05-19
Potential fix for GUI scaling filter clipping animated images and 9slice backgrounds

### LINKED ISSUE #9892: animated_image doesn't work when gui_scaling_filter is enabled
When `gui_scaling_filter` is enabled, `animated_image` shows only one frame, the others stay invisible. `background` are broken too.

Comparison with `craftguide`:

###  `gui_scaling_filter` disabled:

screenshot_20200519_001456

###  `gui_scaling_filter` enabled:

screenshot_20200519_001242

@Df458 @v-rob @rubenwardy @sfan5

Df458:
Probably fixes #9892

I don't know too much about the GUI scaling filter's operation, so I can't guarantee that this is the correct move. It definitely fixes animated images, though 9-slice textures still look odd (although they now display the full texture like they're supposed to).

I don't know enough about this feature to guarantee that this is correct behavior, but going by the comments in the code and the math being done, I'm pretty sure this is the right fix. Ideally, someone with more knowledge of the feature should thoroughly test this to ensure that it works correctly.

### DIFF
--- a/src/client/imagefilters.cpp
+++ b/src/client/imagefilters.cpp
@@ -91,7 +91,7 @@ void imageScaleNNAA(video::IImage *src, const core::rect<s32> &srcrect, video::I
 	u32 dy, dx;
 	video::SColor pxl;
 
-	// Cache rectsngle boundaries.
+	// Cache rectangle boundaries.
 	double sox = srcrect.UpperLeftCorner.X * 1.0;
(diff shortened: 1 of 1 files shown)


EvidenceBKidscode:
Tested and this fix works. :+1:

PR #9898, opened 2020-05-20
Modernize include guards and add missing ones

Awkor:
This PR converts the remaining `#ifndef` header guards to `#pragma once`.

I used `guard2once` from [guardonce](https://github.com/cgmb/guardonce) to do the conversion.

### DIFF
--- a/src/client/hud.h
+++ b/src/client/hud.h
@@ -18,8 +18,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
 
-#ifndef CLIENT_HUD_HEADER
-#define CLIENT_HUD_HEADER
+#pragma once
 
 #include <vector>
 #include <IGUIFont.h>
@@ -147,4 +146,3 @@ void drawItemStack(
 		const v3s16 &angle,
 		const v3s16 &rotation_speed);
 
-#endif
--- a/src/gui/guiEditBoxWithScrollbar.h
+++ b/src/gui/guiEditBoxWithScrollbar.h
@@ -2,7 +2,8 @@
 // This file is part of the "Irrlicht Engine".
 // For conditions of distribution and use, see copyright notice in irrlicht.h
 
-#pragma once
+#ifndef GUIEDITBOXWITHSCROLLBAR_HEADER
+#define GUIEDITBOXWITHSCROLLBAR_HEADER
 
 #include "IGUIEditBox.h"
 #include "IOSOperator.h"
@@ -194,4 +195,5 @@ protected:
 };
 
 
+#endif // GUIEDITBOXWITHSCROLLBAR_HEADER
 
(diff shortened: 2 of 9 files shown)


src/client/hud.h
```diff
@@ -18,8 +18,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
 */
 
-#ifndef CLIENT_HUD_HEADER
-#define CLIENT_HUD_HEADER
```
ClobberXD, line 22: I'm pretty sure the `#ifndef` was left untouched on purpose, because there are two header files with the same name:

- `src/hud.h`
- `src/client/hud.h`

Would `#pragma once` take this quirk into account?

Awkor: From what I've gathered from [here](https://gcc.gnu.org/onlinedocs/cpp/Pragmas.html), [here](https://en.wikipedia.org/wiki/Pragma_once) and [here](https://docs.microsoft.com/en-us/cpp/preprocessor/once?view=vs-2019), I think this quirk should be handled correctly.

ClobberXD: That does indeed seem to be the case, thanks for clarifying.

orbea:
`#pragma once` is not standardized, ifdef guards are. This seems much more likely to introduce problems than solve them.

src/gui/guiEditBoxWithScrollbar.h
```diff
@@ -2,8 +2,7 @@
 // This file is part of the "Irrlicht Engine".
 // For conditions of distribution and use, see copyright notice in irrlicht.h
 
-#ifndef GUIEDITBOXWITHSCROLLBAR_HEADER
-#define GUIEDITBOXWITHSCROLLBAR_HEADER
+#pragma once
```
SmallJoker, line 5: Modified Irrlicht source. Please keep as-is.

SmallJoker:
@orbea A while back `#pragma once` was decided to be widely supported, and thus accepted as header guard for Minetest. Somewhere there's a PR about this.

@Awkor While you're at it - would you please be so nice and include the missing guards as mentioned in https://github.com/minetest/minetest/pull/9899#issuecomment-631614798 ?

ClobberXD:
`#pragma once` is supported on all mainstream compilers, even if it isn't a part of the C++ specification itself. See https://en.wikipedia.org/wiki/Pragma_once#Portability

Awkor:
@SmallJoker Done.

src/gui/guiSkin.h
```diff
@@ -2,8 +2,7 @@
 // This file is part of the "Irrlicht Engine".
 // For conditions of distribution and use, see copyright notice in irrlicht.h
 
-#ifndef __GUI_SKIN_H_INCLUDED__
-#define __GUI_SKIN_H_INCLUDED__
+#pragma once
```
sfan5, line 5: This is also a modified file from Irrlicht, so please keep the include guard.

Awkor: Should be fixed now.

PR #9903, opened 2020-05-20
Properly implement exposing the zoom key

appgurueu:
- Same as & inspired by #9901, but *works* (also with get_player_control_bits) and is documented
- Properly sets the bitflags
- Does it resolve any reported issue? -
- This PR is Ready for Review

##### How to test

```lua
	for _, player in pairs(minetest.get_connected_players()) do
		assert(controls.zoom == (player:get_player_control_bits() > 256))
		if controls.zoom then
			minetest.chat_send_all("Player "..player:get_player_name().." is pressing the zoom key!")
		end
	end
end)
```


SmallJoker:
It were a nice thing to assist the original author by highlighting the issues and proposing solutions rather than taking and re-doing the PR. There's absolutely no hurry, so please give them some time to get the PR ready.
This PR also lacks of a notice that the `zoom` key will only be available for 5.3.0-dev player and newer.

appgurueu:
> It were a nice thing to assist the original author by highlighting the issues and proposing solutions rather than taking and re-doing the PR. There's absolutely no hurry, so please give them some time to get the PR ready.
> This PR also lacks of a notice that the `zoom` key will only be available for 5.3.0-dev player and newer.

I will include a notice. Regarding "assisting the author": I would've done so if the first PR contained more effort than this one. But it was completely broken and I figured out it was not worth the hassle to basically dictate him more than 50% of his PR (especially as I'm not a maintainer and can't make edits which he could just merge with one click) for 2 trivial lines of code.

appgurueu:
@raistlin I have merged your branch

ClobberXD:
Both `get_player_control` and `get_player_control_bits` work as expected. :+1:

Here's the code I used for testing:

```lua
local timer = 0
local bits
	timer = timer - dtime
	if timer > 0 then
		return
	end
	timer = 1

	if not player then
		return
	end

	local new_bits = player:get_player_control_bits()
	if not bits then
		bits = new_bits
	else
		local diff = new_bits - bits
		if math.abs(diff) == 512 then
			minetest.chat_send_all("Bits changed! | zoom " .. (diff < 0 and "unset" or "set"))
			bits = new_bits
		end
	end

	if player:get_player_control().zoom then
		minetest.chat_send_all("zoom key pressed!")
	end
end)
```

This code should print `zoom key pressed!` to chat, when it detects the zoom key to be pressed. If the control bits change, `Bits changed! | zoom {set/unset}` will be printed to chat.

appgurueu:
My code does pretty much the same, except it isn't limited to once per second so it'll spam your chat ;)

ClobberXD:
I didn't notice that the first post already had testing code. Apologies. :)

KaylebJay:
Thanks @appgurueu, looks good now. Should open up a lot of new modding possibilities! :thumbsup:

PR #9904, opened 2020-05-20
Check for valid base64 before decoding

Lejo1:
Returns nil instead of some supper-strange bytes if the string is not valid base64

### DIFF
--- a/src/script/lua_api/l_util.cpp
+++ b/src/script/lua_api/l_util.cpp
@@ -323,8 +323,7 @@ int ModApiUtil::l_decode_base64(lua_State *L)
 	const std::string data = std::string(d, size);
 
 	if (!base64_is_valid(data)) {
-		lua_pushnil(L);
-		return 1;
+		return 0;
 	}
 
 	std::string out = base64_decode(data);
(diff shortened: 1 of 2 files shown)


src/script/lua_api/l_util.cpp
```diff
@@ -318,9 +318,16 @@ int ModApiUtil::l_decode_base64(lua_State *L)
...
 
 	size_t size;
-	const char *data = luaL_checklstring(L, 1, &size);
+	const char *d = luaL_checklstring(L, 1, &size);
+
+	const std::string data = std::string(d, size);
+
+	if (!base64_is_valid(data)) {
+		lua_pushnil(L);
+		return 1;
```
sfan5, line 327: You don't need to explicitly return nil by pushing it, you can instead just `return 0`

Lejo1: Thanks, for letting me that know!
Changed it.

src/script/lua_api/l_util.cpp
```diff
@@ -318,9 +318,15 @@ int ModApiUtil::l_decode_base64(lua_State *L)
 	NO_MAP_LOCK_REQUIRED;
 
 	size_t size;
-	const char *data = luaL_checklstring(L, 1, &size);
+	const char *d = luaL_checklstring(L, 1, &size);
+
+	const std::string data = std::string(d, size);
+
+	if (!base64_is_valid(data)) {
```
appgurueu, line 325: Code style, braces can be omitted

PR #9912, opened 2020-05-22
[NO SQUASH] Node texture appearance for particles+particlespawners

### LINKED ISSUE #9840: Spawn block breaking particles
It would create a lot of aesthetic opportunities in mods to be able to create block breaking particles as if the player broke the block.

sfan5:
closes #9840

second commit is so that short-lived particlespawners are not sent to clients 1000s of nodes away

## To do

This PR is Ready for Review.

## How to test

```lua
core.register_on_punchnode(function(pos, node, puncher, pointed_thing)
	pos.y = pos.y + 0.55
	local g = vector.new(0, -9.81, 0)
	core.add_particlespawner({
		minpos = pos, maxpos = pos,
		minvel = vector.new(-1, 2.5, -1),
		maxvel = vector.new(1, 5, 1),
		minacc = g, maxacc = g,
		minsize = 0, maxsize = 0,
		time = 3.2,
		amount = 400,
		node = node,
		node_tile = 3,
	})
end)
```
* Punch a node and observe a fountain of dig particles that have the texture of a side(!) face

### DIFF
--- a/doc/lua_api.txt
+++ b/doc/lua_api.txt
@@ -7831,6 +7831,8 @@ Used by `minetest.add_particle`.
 
         size = 1,
         -- Scales the visual size of the particle texture.
+        -- If `node` is set, size can be set to 0 to spawn a randomly-sized
+        -- particle (just like actual node dig particles).
 
         collisiondetection = false,
         -- If true collides with `walkable` nodes and, depending on the
@@ -7849,6 +7851,7 @@ Used by `minetest.add_particle`.
         -- If true faces player using y axis only
 
         texture = "image.png",
+        -- The texture of the particle
 
         playername = "singleplayer",
         -- Optional, if specified spawns particle only on the player's client
@@ -7859,6 +7862,17 @@ Used by `minetest.add_particle`.
         glow = 0
         -- Optional, specify particle self-luminescence in darkness.
         -- Values 0-14.
+
+        node = {name = "ignore", param2 = 0},
+        -- Optional, if specified the particle will have the same appearance as
+        -- node dig particles for the given node.
+        -- `texture` and `animation` will be ignored if this is set.
+
+        node_tile = 0,
+        -- Optional, only valid in combination with `node`
+        -- If set to a valid number 1-6, specifies the tile from which the
+        -- particle texture is picked.
+        -- Otherwise, the default behavior is used. (currently: any random tile)
     }
 
 
@@ -7888,7 +7902,9 @@ Used by `minetest.add_particlespawner`.
         maxsize = 1,
         -- The particles' properties are random values between the min and max
         -- values.
-        -- pos, velocity, acceleration, expirationtime, size
+        -- applies to: pos, velocity, acceleration, expirationtime, size
+        -- If `node` is set, min and maxsize can be set to 0 to spawn
+        -- randomly-sized particles (just like actual node dig particles).
 
         collisiondetection = false,
         -- If true collide with `walkable` nodes and, depending on the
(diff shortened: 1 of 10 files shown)


doc/lua_api.txt
```diff
@@ -7859,6 +7862,16 @@ Used by `minetest.add_particle`.
...
+
+        node = {name = "ignore", param1 = 0, param2 = 0},
+        -- Optional, if specified the particle will have the same appearance as
+        -- node dig particles for the given node.
+        -- `texture` and `animation` will be ignored if this is set.
+
+        node_tile = 0,
+        -- Optional, only valid in combination with `node`
+        -- If set to a valid number 1-6, specifies the tile from which the
+        -- particle texture is picked. Otherwise, the default behavior is used.
```
ClobberXD, line 7874: This last sentence sounds ambiguous to me - what's the default behaviour here?

sfan5: The default behaviour is to pick a random tile.
I intentionally meant to keep this unspecified because default behavior might change in the future and we don't want to guarantee it.

ClobberXD: Ahh, that's what I thought.

> I intentionally meant to keep this unspecified because default behavior might change in the future and we don't want to guarantee it.

Understood.

sfan5: I agree that the documentation is a bit confusing, maybe you have a suggestion how to word this better?

ClobberXD: The following would suffice, IMO:

`... Otherwise, a random tile is selected.`

(Btw, is the random selection per-particle or per-particlespawner for particlespawners?)

Maybe also add a note about this behaviour changing in the future?

`Note: This behaviour isn't final, and might change in the future.`

... or just 

`Note: This behaviour might change in the future.`

sfan5: > Btw, is the random selection per-particle or per-particlespawner for particlespawners?

per particle

ClobberXD: I see, thanks

doc/lua_api.txt
```diff
@@ -7859,6 +7862,17 @@ Used by `minetest.add_particle`.
         glow = 0
         -- Optional, specify particle self-luminescence in darkness.
         -- Values 0-14.
+
+        node = {name = "ignore", param1 = 0, param2 = 0},
+        -- Optional, if specified the particle will have the same appearance as
+        -- node dig particles for the given node.
```
SmallJoker, line 7868:
```
-- `param1` is ignored
```

src/particles.cpp
```diff
@@ -50,4 +53,11 @@ void ParticleParameters::deSerialize(std::istream &is, u16 protocol_ver)
 	animation.deSerialize(is, 6); /* NOT the protocol ver */
 	glow               = readU8(is);
 	object_collision   = readU8(is);
+	// This is kinda awful
```
SmallJoker, line 56: How about adding a version number on top that can be used reliably beginning from protocol version 40?

sfan5: The awful thing here is explicitly needing to check for EOF, other than that it's not different from other `try { ... } catch(PacketError) {}` code.
So I don't think it's good to bump the protocol version when it isn't strictly necessary.

doc/lua_api.txt
sfan5: Removed it from the shown table instead.

src/particles.cpp
SmallJoker: The difference to `PacketError` is that the variables aren't overwritten when EOF is reached - here it requires this workaround.
I suggested particle-versioning to get around protocol bumps in the future and to solve these checks without the "awful" part. The protocol version can be bumped a lot; it's a `u16`.

sfan5: Right. When `readU8`/`readU16` are replaced with an API that properly throws exceptions in the future then the awful code vanishes and the extra version number would be wasted work.
So IMO this shouldn't be done.

PR #9914, opened 2020-05-22
POC: Introduce SerializeStream helper class

SmallJoker:
### Goals

Solve cases like these:
https://github.com/minetest/minetest/blob/e79bc40c0a5312baf4e8c3e33048d50b41b4a2ff/src/network/clientpackethandler.cpp#L333-L334
https://github.com/minetest/minetest/blob/e79bc40c0a5312baf4e8c3e33048d50b41b4a2ff/src/server.cpp#L1400-L1404

This implementation can be used wherever `?stream` or `NetworkProtocol` is currently used, which is most of the code. If another stream backend is wanted, say - write to `vector<u8>` for better performance - this is possible too, but requires a new class to implement all stream functions.

### Other design ideas

**Option 1:** New class that inherits from `iostream`
  * Would basically implement `NetworkPacket` as stream
  * Requires too many functions to do it properly. Most stream functions won't ever be used
  * `stringstream` is sometimes slower than `vector` but if copies are avoided (such as `::str()`) these two are comparable in terms of speed.

**Option 2:** No `istream / ostream` dependencies or conversions
  * Single class that does all (de)serializing for network and disk
  * Touches way too many lines to be adopted within a single PR - if even, the PR would be a rebase mightmare

**Option 3:** Keep everything as-is. _(not an option)_

**Option 4:** Your idea goes here

## To do

This PR is a Proof of Concept. Basically everything is missing.

## How to test

1) Judge code
2) Compile Minetest
3) Check that the minimalistic unittest works on startup

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -110,6 +110,7 @@ FileLogOutput file_log_output;
 
 static OptionList allowed_options;
 
+#include "util/serializestream.h"
 int main(int argc, char *argv[])
 {
 	int retval;
@@ -184,6 +185,8 @@ int main(int argc, char *argv[])
 	if (g_settings->getBool("enable_console"))
 		porting::attachOrCreateConsole();
 
+	testit();
+	return 0;
 #ifndef __ANDROID__
 	// Run unit tests
 	if (cmd_args.getFlag("run-unittests")) {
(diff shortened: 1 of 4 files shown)


MoNTE48:
@SmallJoker https://github.com/minetest/minetest/pull/9927

SmallJoker:
Ideas have been collected. Will open a new PR once it's in an usable state.

